PR #384, opened 2020-06-23
Update button

pomfrida:


libraries/core-react/src/Button/Button.jsx
```diff
@@ -23,8 +17,8 @@ const Base = ({ base, baseDisabled: disabled }) => {
     color: ${base.color};
     fill: ${base.color};
     svg {
-      height: 16px;
-      width: 16px;
+      ${({ height }) => css({ height })}
+      ${({ width }) => css({ width })}
```
mimarz, line 21: Suggest changing this to one line

libraries/core-react/src/Button/Button.jsx
```diff
@@ -105,36 +98,63 @@ const ButtonBase = styled.button.attrs(({ type = 'button' }) => ({
...
+    width: button.icon_size[iconType].width,
+    height: button.icon_size[iconType].height,
+  }
+
   return (
     <ButtonBase
+      as={component}
       base={base}
+      type={href ? null : 'button'}
+      href={href ? href : null}
```
mimarz, line 146: Think this has to be `undefined` instead of `null` for React to properly drop adding the attributes?

libraries/core-react/src/Button/Button.jsx
```diff
@@ -105,36 +98,63 @@ const ButtonBase = styled.button.attrs(({ type = 'button' }) => ({
...
+  ref,
+) {
   const colorBase = colors[color] || {}
   const base = colorBase[variant] || {}
   const baseDisabled = colors.disabled[variant] || {}
 
+  const iconRight = rightIcon && <RightIcon>{rightIcon}</RightIcon>
+  const iconLeft = leftIcon && <LeftIcon>{leftIcon}</LeftIcon>
+  const iconType = variant === 'ghost_icon' ? variant : 'button_icon'
+  const component = href ? 'a' : 'button'
```
mimarz, line 132: Rename to `as` ?

apps/storybook-react/stories/Button.stories.jsx
```diff
@@ -67,6 +78,29 @@ export const contained = () => (
     <Button color="secondary">Secondary</Button>
     <Button color="danger">Danger</Button>
     <Button disabled>Disabled</Button>
+    <Button leftIcon={<Icon name="save" title="save"></Icon>}>Primary</Button>
```
vnys, line 81: I know this is a common pattern, but we have deliberately avoided it so far in favour of composition. I would do something like this instead: 

```jsx
<Button>
  <Icon name="save" title="save" />
  Primary
</Button>
```

apps/storybook-react/stories/Button.stories.jsx
```diff
@@ -86,6 +120,60 @@ export const outlined = () => (
...
+    <Button
+      variant="outlined"
+      rightIcon={<Icon name="save" title="save"></Icon>}
+    >
+      Primary
+    </Button>
+    <Button
+      variant="outlined"
+      color="secondary"
+      rightIcon={<Icon name="save" title="save"></Icon>}
```
vnys, line 159: Same here, just do 

```jsx
<Button>
  Primary
  <Icon name="save" title="save" />
</Button>
```

apps/storybook-react/stories/Button.stories.jsx
```diff
@@ -146,6 +282,28 @@ export const form = () => {
...
+  return (
+    <Wrapper>
+      <input
+        type="file"
+        id="file-upload"
+        style={{ display: 'none' }}
+        multiple
+      />
+      <label htmlFor="file-upload">
+        <Button as="span">Upload</Button>
```
vnys, line 295: This results in invalid html: 
image

Also not possible to use keyboard to focus on the button.

libraries/core-react/src/Button/Button.tokens.js
```diff
@@ -36,4 +36,14 @@ const colors = {
...
 export const button = {
   colors,
+  icon_size: {
+    ghost_icon: {
+      width: '24px',
+      height: '24px',
+    },
+    button_icon: {
+      width: '16px',
+      height: '16px',
```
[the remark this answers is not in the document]
pomfrida: Screenshot 2020-06-25 at 11 41 17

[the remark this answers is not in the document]
mimarz: Like this @pomfrida 
image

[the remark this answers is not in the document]
pomfrida: Lol I'm such a noob 🤓

libraries/core-react/src/Button/Button.jsx
```diff
@@ -105,36 +100,37 @@ const ButtonBase = styled.button.attrs(({ type = 'button' }) => ({
...
-}) => {
+export const Button = forwardRef(function Button(
+  { variant, children, disabled, className, color, tabIndex, href, ...other },
+  ref,
+) {
   const colorBase = colors[color] || {}
   const base = colorBase[variant] || {}
   const baseDisabled = colors.disabled[variant] || {}
 
+  const as = href ? 'a' : other.as ? other.as : 'button'
```
mimarz, line 111: Using the value of `as` for `type` on button can lead to invalid html.

libraries/core-react/src/Button/Button.jsx
```diff
@@ -105,36 +100,37 @@ const ButtonBase = styled.button.attrs(({ type = 'button' }) => ({
...
+  const baseProps = {
+    ...other,
+    ref,
+    as,
+  }
+
   return (
     <ButtonBase
       base={base}
+      type={href || other.as ? undefined : 'button'}
```
mimarz, line 122: We should maybe not ignore `type` if it comes from props either. Force it if href is put, else use provided type?

libraries/core-react/src/Button/Button.jsx
```diff
@@ -105,36 +100,37 @@ const ButtonBase = styled.button.attrs(({ type = 'button' }) => ({
...
+      href={href ? href : undefined}
```
mimarz, line 123: this can just use href from props, set the defaultProps to be `undefined`

libraries/core-react/src/Button/Button.jsx
```diff
@@ -84,17 +79,17 @@ const Base = ({ base, baseDisabled: disabled }) => {
   `
 }
 
-const ButtonBase = styled.button.attrs(({ type = 'button' }) => ({
-  type,
-}))`
+const ButtonBase = styled.button`
   margin: 0;
```
mimarz, line 83: Suggest we remove this margin as we have gotten feedback about margins not working on button

pomfrida: Added custom margin test, was success 🏆

mimarz: 👏

Issue #373, opened 2020-06-18
Add Slider onCommitedChange callback


cbuv:
**Is your feature request related to a problem? Please describe.**
Performance issue with slider values connected to other parts of a page.  
With a slider range of several years,  a drag could trigger thousands of state changes to components that use these values. The slider solution will trigger one change.   

**Describe the solution you'd like**
Add a onChangeCommited method like material ui provide for the slider component. 
https://material-ui.com/api/slider/
As with the onChange, the values array is passed back on a mouseUp event.

**Describe alternatives you've considered**
Implemented onMouseUp handler on the exported Slider component. But there's no way to know if the value is the start or end. Must have access to internal state of the component (the array with both values in the range).

PR #386, opened 2020-06-24
Add committed change for sliders

wenche:
Solve #373

### DIFF
--- a/apps/storybook-react/stories/Slider.stories.jsx
+++ b/apps/storybook-react/stories/Slider.stories.jsx
@@ -59,11 +59,10 @@ export const Examples = () => {
           onChange={(event, value) => {
             updateValueTwo(value)
           }}
-          min="0"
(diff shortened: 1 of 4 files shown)


wenche:
Also solves #387

apps/storybook-react/stories/Slider.stories.jsx
```diff
@@ -43,10 +46,32 @@ export const Examples = () => {
           onChange={onChange}
           ariaLabelledby="range-slider-label"
         />
-        <p>
+        <p style={{ marginTop: '1.5rem' }}>
```
vnys, line 49: We should have a couple of utility classes available for these things 🤔  Or Styled System props maybe…

wenche: Indeed... *kremt* tailwind

PR #400, opened 2020-06-30
Storefront with docz

### LINKED ISSUE #474: Use Docz
As part of the effort to fill in developer documentation in the Storefront, we need a component playground with examples and proptables. We’ll use Docz for that, which is built on top of Gatsby, which is what we’re already using in the storefront. To make it as easy as possible to upgrade to new versions of Docz in the future we should avoid custom solutions where possible and do things the Docz way. That means for example using Theme UI for styling where we don’t use EDS components, and adding route fields to frontmatter. 

- [x] Move pages and add frontmatter (@vnys) 
- [x] Style the sidebar like the navigation drawer (@vnys)
- [x] Use [Page component ](https://www.docz.site/docs/creating-your-themes#using-documents-settings)for table of content, tabs  (@wenche)
- [x] Add component status overview component
- [x]  Add remaining styles from[ layout.css](https://github.com/equinor/design-system/blob/storefront/apps/storefront/src/components/layout.css)

wenche:
resolves #474


wenche:
I'm in the middle of applying all the stuffs in /tempates/Page from the original storefront. My initial idea was to use the approach from [docz](https://www.docz.site/docs/creating-your-themes#using-documents-settings) and [the <ComponentProvider](https://www.docz.site/docs/components-api#componentsprovider) but I can't get it to work. See commit 4075eef . According to the [API](https://www.docz.site/docs/components-api#usedocs), docz also has a useDocs hook, but that returns all the docs.

wenche:
[https://www.docz.site/docs/gatsby-theme#customizing-components](Docz gatsby-theme) documentation shows the approach for document shadowing. I've made several components in the gatsby-theme-docz folder that overrides those from the theme. The `wrapper.js` file, see [docs](https://www.docz.site/docs/gatsby-theme#wrapping-the-entire-app) adds Helmet configuration. `index.js` adds [a theme](https://www.docz.site/docs/gatsby-theme#creating-your-own-docz-theme), but still using the base components from Gatsby theme. I used [this example](https://github.com/nejcm/docz-theme-extended/tree/master/theme/src/gatsby-theme-docz) from a SO thread to make things work.

wenche:
Useful links:
[Gatsby's docz documentation](https://www.gatsbyjs.org/tutorial/writing-documentation-with-docz/)
[The original theme we are shawing](https://github.com/doczjs/docz/tree/master/core/gatsby-theme-docz/src)
[Docz](https://www.docz.site/)

wenche:
Things to do:
 - There are some issues in console. I'm not sure about what to do with the InitialColorMode-warning. As [discussed on github](https://github.com/doczjs/docz/issues/1440) this is what the docs say you I had a be kind to future self vision about making it easier to apply dark mode. 
- There is still some css-files, some styled components, some ./styles.js from the theme and some settings in index.js. Quite messy indeed... But I would guess we should try to use the settings from the index.js-file where we can?
- We need to recreate the url structure from the /content folder as we still want the same url structure from the storefront (?). I guess we'll need to do some custom server woo as in the original storefront.
- As briefly mentioned, the build in useMenu-hooks doesn't support nested menus, that's why I just copy'n pasted from the original storefront.

vnys:
/azp run storefront-docz

baby steps 🙄

Issue #395, opened 2020-06-25
Add status badges to readme


vnys:
We should add status badges to the README. Something like this: 

| Package | Status | Version |
|--|--|--|
| [Core React](https://github.com/equinor/design-system/tree/develop/libraries/core-react) | [![Build Status](https://dev.azure.com/equinor-design-system/Equinor%20Design%20System/_apis/build/status/react-tests?branchName=develop)](https://dev.azure.com/equinor-design-system/Equinor%20Design%20System/_build/latest?definitionId=15&branchName=develop) | [![Version](https://img.shields.io/npm/v/@equinor/eds-core-react)](https://www.npmjs.com/package/@equinor/eds-core-react) |
| [Tokens](https://github.com/equinor/design-system/tree/develop/libraries/tokens) | [![Build Status](https://dev.azure.com/equinor-design-system/Equinor%20Design%20System/_apis/build/status/tokens?branchName=develop)](https://dev.azure.com/equinor-design-system/Equinor%20Design%20System/_build/latest?definitionId=17&branchName=develop) | [![Version](https://img.shields.io/npm/v/@equinor/eds-tokens)](https://www.npmjs.com/package/@equinor/eds-tokens)|
| [Icons](https://github.com/equinor/design-system/tree/develop/libraries/icons) | [![Build Status](https://dev.azure.com/equinor-design-system/Equinor%20Design%20System/_apis/build/status/icons?branchName=develop)](https://dev.azure.com/equinor-design-system/Equinor%20Design%20System/_build/latest?definitionId=16&branchName=develop) | [![Version](https://img.shields.io/npm/v/@equinor/eds-icons)](https://www.npmjs.com/package/@equinor/eds-icons)|
| [Figma broker](https://github.com/equinor/design-system/tree/develop/apps/figma-broker) | [![Build Status](https://dev.azure.com/equinor-design-system/Equinor%20Design%20System/_apis/build/status/figma-broker?branchName=develop)](https://dev.azure.com/equinor-design-system/Equinor%20Design%20System/_build/latest?definitionId=18&branchName=develop) | [![Version](https://img.shields.io/badge/dynamic/json?url=https://raw.githubusercontent.com/equinor/design-system/develop/apps/figma-broker/package.json&label=version&query=version&color=orange&prefix=v)](https://github.com/equinor/design-system/apps/figma-broker) |
[...]

PR #402, opened 2020-07-01
Add and update pipelines

vnys:
In this pr I’ve added pipelines for
- tokens
- icons
- figma-broker

I’ve also updated existing pipelines with eslint scripts.

The result of the pipelines will be used in badges in #395, but it also forces us to write cleaner code – the stories in particular have lots of errors.

In figma-broker there are no build scripts, so the pipeline only runs eslint. 
We had some discussions about Docker, but I don’t think it makes sense to build a Docker image at this point that’s not going to be used for anything – so I think we should revisit that once we decide to deploy the figma-broker somewhere.

There’s going to be some cleanup after this pr is merged into develop. In `storybook-react` alone there are now 101 errors in the stories. 🙄

~I’ll make~ I made that a separate issue. (#403)


apps/storybook-react/azure-pipelines.yml
```diff
@@ -36,6 +36,24 @@ variables:
     value: /apps/storybook-react
 
 steps:
+  - task: NodeTool@0
+    inputs:
+      versionSpec: '14.x'
+    displayName: 'Install Node.js'
+
+  - script: npm install -g pnpm@next
```
mimarz, line 44: Why `pnpm@next`?

PR #406, opened 2020-07-02
Pagination

pomfrida:


libraries/core-react/src/Pagination/Pagination.jsx
```diff
@@ -0,0 +1,208 @@
...
+const Navigation = styled.nav`
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  flex-wrap: nowrap;
+  margin-left: ${({ withItemIndicator }) =>
+    withItemIndicator ? tokens.spacingSmall : 0};
+`
+
+const UnorderedList = styled.ul`
```
wenche, line 33: I think a ol would be the best choice here

libraries/core-react/src/Pagination/Pagination.jsx
```diff
@@ -0,0 +1,208 @@
...
+            ),
+          )}
+        <Button
+          variant="ghost_icon"
+          onClick={moveRight}
+          aria-label="Go to next page"
+          disabled={activePage === pages}
+        >
+          <Icon name="chevron_right" title="next" />
+        </Button>
```
wenche, line 158: A list can't have buttons and icons (ellipsis) as direct children :) Should wrap them in li

wenche:
How easy would it be to have an onChange function with the chosen page? I miss the opportunity to hook into my router, e.g. myapp.com/resources?page=3  Could be something like
```
if (props.onChange) {
  return onChange(currentSelectedPage)
}
```

wenche (approved):
LGTM 👍

Issue #408, opened 2020-07-02
Wrapping table cell content in Tooltip alters alignment


lindsve:
**Describe the bug**

The cell alignment is changed from right aligned to center aligned when wrapping cell content with a `Tooltip` component.
<img width="441" alt="Skjermbilde 2020-07-02 kl  14 20 47" src="https://user-images.githubusercontent.com/159861/86358179-56179800-bc6f-11ea-8be2-1f6c6f1123d8.png">

**Steps to reproduce the bug**
```
    <Table>
      <Head>
        <Row>
          <Cell as="th" scope="col" width={"240px"}>
            Wellbore
          </Cell>
          <Cell as="th" scope="col" width={"120px"}>
            DtectLoss
          </Cell>
        </Row>
      </Head>
      <Body>
        <Row>
          <Cell>NO 12/34-C-B1</Cell>
          <Cell>OK</Cell>
        </Row>
        <Row>
          <Cell>NO 12/34-C-B2</Cell>
          <Cell>OK</Cell>
        </Row>
        <Row>
          <Cell>NO 34/12-A</Cell>
          <Cell><Tooltip placement={"top"} title={"nice title"}>Alarm</Tooltip></Cell>
        </Row>
      </Body>
    </Table>
```

**Expected behavior**

Not sure if this is expected behavior or by design, but I initially did not expect alignment to be changed

**Specifications**

- Version: @equinor/eds-core-react: 0.4.0,
- Browser: Microsoft Edge 83.0.478.50
- OS: MacOS 10.15.5

PR #409, opened 2020-07-02
Tooltip wrap bug

pomfrida:
Tooltip no longer move content if content is smaller than tooltip, as well as tooltip content doesn't wrap

### DIFF
--- a/apps/storybook-react/stories/Table.stories.jsx
+++ b/apps/storybook-react/stories/Table.stories.jsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { Table, Typography, Tooltip } from '@equinor/eds-core-react'
+import { Table, Typography } from '@equinor/eds-core-react'
 import './../style.css'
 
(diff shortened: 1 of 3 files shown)


apps/storybook-react/stories/Table.stories.jsx
```diff
@@ -39,9 +39,7 @@ export const simpleTable = () => (
             <Row>
               <Cell>Darth Vader</Cell>
               <Cell>Sith</Cell>
-              <Cell variant="numeric">
-                <Tooltip title="Darth Vader sucks">59</Tooltip>
-              </Cell>
+              <Cell variant="numeric">59</Cell>
```
vnys: …but it doesn’t remove the problem… The bug is still «The cell alignment is changed from right aligned to center aligned when wrapping cell content with a Tooltip component.» as reported in #408

pomfrida: 🤣  Fixed now

PR #454, opened 2020-07-07
Style th properly when parent is tbody

vnys:
- Left-aligns text in ths
- Uses borders and backgrounds based on parent selector

### DIFF
--- a/apps/figma-broker/transformers/typography.js
+++ b/apps/figma-broker/transformers/typography.js
@@ -1,6 +1,6 @@
 import R from 'ramda'
 import { withType, withName, removeNilAndEmpty } from '@utils'
-import { rootFontSize, rem, em, px } from '@units'
+import { rootFontSize, rem, em } from '@units'
 import { fillToRgba } from './colors'
 
 export { fillToHex, fillToHsla, fillToRgba } from './colors'
@@ -47,10 +47,10 @@ export const toTypography = (figmaNode, name) => {
     fontWeight,
     letterSpacing: letterSpacing ? em(letterSpacingEm) : null,
     lineHeight: em(lineHeightEm),
-    textDecoration,
-    textTransform: textCase ? `${textCase}CASE` : null,
+    textDecoration: textDecoration ? R.toLower(textDecoration) : null,
+    textTransform: textCase ? R.toLower(`${textCase}CASE`) : null,
     fontStyle: italic ? 'italic' : null,
-    textAlign,
+    textAlign: R.toLower(textAlign),
   }
 
   return removeNilAndEmpty(typography)
--- a/apps/figma-broker/transformers/typography.js
+++ b/apps/figma-broker/transformers/typography.js
@@ -14,7 +14,7 @@ export const toTypography = (figmaNode, name) => {
     fontSize,
     fontWeight,
     letterSpacing,
-    textAlignHorizontal = 'center',
+    textAlignHorizontal: textAlign,
     lineHeightPercentFontSize,
     textDecoration,
     textCase,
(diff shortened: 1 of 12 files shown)


libraries/core-react/src/Table/Cell.jsx
```diff
@@ -4,11 +4,28 @@ import styled from 'styled-components'
...
+    text: {
+      ...headerTextLevel1,
+      text: {
+        ...headerTextLevel2,
+        typography: {
+          ...headerTypography,
+          textAlign: 'left',
+        },
+      },
+    },
```
mimarz, line 28: We can simplify this further. We are already handling `textAlign` when exporting typography in `figma-broker`. 

Looks like we decided earlier to opt out `textAlign` prop if its value was left... 🤔

Simpler fix would be to update `figma-broker` transformer for typography [here](https://github.com/equinor/design-system/blob/4eba1e02c31d650a6052eb27da2f3ce7d4bf4458/apps/figma-broker/transformers/typography.js#L57) and then re-run `/create-component-tokens`

libraries/core-react/src/Table/Cell.jsx
```diff
@@ -34,18 +61,33 @@ const borderTemplate = (borders) =>
...
+        }
+
+        tbody & {
+          ${borderTemplate(borders.tbody)}
+          background: ${background.tbody};
+        }`
+      : `
+        ${borderTemplate(borders)}
+        background: ${background}`
+
```
mimarz, line 82: Add `css` function from `styled-components` for better readability?

vnys: That wouldn’t work in this case

mimarz:
```jsx
css`
thead & {
    ${borderTemplate(borders.thead)}
    background: ${background.thead};
}

tbody & {
    ${borderTemplate(borders.tbody)}
    background: ${background.tbody};
}`
```

css has support for string templates no?

vnys:
> We could also move the token logic to a `Cell.tokens.js` file?

Yes I guess, but I think that could be in a different issue – if the point is to remove the component tokens.

libraries/core-react/src/Table/Cell.jsx
vnys: The output from that is 
```css
thead &{, border-bottom: 2px solid rgba(220, 220, 220, 1); 
, background:,rgba(247, 247, 247, 1),;}tbody &{, border-bottom: 1px solid rgba(220, 220, 220, 1); 
, background:,rgba(255, 255, 255, 1),;}
```

vnys: Don’t think the css helper is appropriate to use here

apps/figma-broker/transformers/typography.js
```diff
@@ -52,10 +50,7 @@ export const toTypography = (figmaNode, name) => {
     textDecoration,
     textTransform: textCase ? `${textCase}CASE` : null,
     fontStyle: italic ? 'italic' : null,
-  }
-
-  if (textAlign !== 'left') {
-    typography = R.set(textAlignLens, textAlign, typography)
+    textAlign,
```
mimarz, line 53: Add back the lowercasing? `R.toLower(textAlignHorizontal)` a lot of unecessary updates, and its all uppercase in tokens files now 🙈

vnys: Yes, I was wondering about that 🤔 Text-transform is uppercase too…

mimarz: Hmm, yeah, it should be lowercase to... Don't you think?

vnys: Yes, changed that too.

PR #467, opened 2020-07-10
Button content alignments fix for Webkit browsers

### LINKED ISSUE #465: Centering ghost button icon in Safari/iOS
On ghost buttons the icon doesn't get centered in Safari/iOS. Reproduced in the EDS storybook. Fixed by adding margin: 0 auto to the svg. But might be more correct ways to fix. 

1. Open a ghost button in Safari
2. Get surprised by the left aligned icon

- Version: 
- Browser: Safari
- OS: Mac OSX and iOS

**Additional context**

image

mimarz:
Fixes #465

* Workaround for missaligment inside button using `css-grid` on Webkit browsers
* Fixed  colorfull 🦄  border on buttons on Webkit browsers

### DIFF
--- a/libraries/core-react/src/Button/Button.jsx
+++ b/libraries/core-react/src/Button/Button.jsx
@@ -8,7 +8,7 @@ import { typographyTemplate } from '../_common/templates'
 const { colors } = button
 
 // display:grid; does not work on Webkit browser engine, so we have to wrap content in element where css-grid works
-const WebkitFix = styled.span`
+const ButtonInner = styled.span`
   display: grid;
   grid-gap: 8px;
   grid-auto-flow: column;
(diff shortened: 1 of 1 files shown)


libraries/core-react/src/Button/Button.jsx
```diff
@@ -7,6 +7,14 @@ import { typographyTemplate } from '../_common/templates'
 
 const { colors } = button
 
+// display:grid; does not work on Webkit browser engine, so we have to wrap content in element where css-grid works
+const WebkitFix = styled.span`
```
vnys, line 11: Would call it something more generic – `ButtonInner` for example. Apart from that LGTM 👍

PR #472, opened 2020-07-15
Prepare release of EDS Core@0.5.0

### LINKED ISSUE #457: Increase legibillity
We’ve received feedback on poor legibility with the Equinor type in small font-sizes, tables in particular. This could be alleviated by replacing regular with medium in all cases where we’re under 16px in font-size. Let’s do a quick test to see how that works out.

vnys:
resolves #457

### DIFF
--- a/apps/storybook-react/package.json
+++ b/apps/storybook-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@equinor/eds-storybook-react",
-  "version": "0.4.0",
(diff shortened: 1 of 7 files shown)


vnys:
> You could also a line about reduced lib size due to optimising dependencies :D

Yeah, maybe 😉 But I think that’s something the users should take for granted 😛 If they care at all, we haven’t had any complaints about file size at least. But 1.2M was way too over the top.

PR #479, opened 2020-08-10
Trigger Tooltip on disabled elements

pomfrida:
Mouse Events on disabled elements are buggy.

[This is a known issue on Safari and Chrome browsers.](https://github.com/facebook/react/issues/19419) Testing in Firefox gives us the desired outcome.

Trying to find a workaround that goes for all browsers...

### DIFF
--- a/apps/storybook-react/stories/Tooltip.stories.jsx
+++ b/apps/storybook-react/stories/Tooltip.stories.jsx
@@ -28,7 +28,9 @@ const Wrapper = styled.div`
 
 const TextWrapper = styled.div`
   margin-bottom: 32px;
+  width: 800px;
 `
+
 export default {
   title: 'Components|Tooltip',
   component: Tooltip,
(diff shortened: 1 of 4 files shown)


pomfrida:
Could solve it with CSS: pointer-events: none;
But this will override that beautiful forbidden cursor... Hmm...

apps/storybook-react/stories/Tooltip.stories.jsx
```diff
@@ -151,3 +153,33 @@ export const WithKnobs = () => {
...
+        <Typography variant="h3">Tooltip with disabled elements</Typography>
+        <Typography variant="body_long">
+          Firefox, Edge and Chrome supports tooltip on disabled elements.
+        </Typography>
+        <Typography variant="body_long">
+          If you have Safari users, you will need to add inline style to your
+          disabled element, shown in the example below. This will help trigger
+          the mouse events correctly. Unfortunately, this workaround overwrites
+          the &apos;not-allowed&apos; cursor.
+        </Typography>
```
pomfrida, line 170: Updated the Storybook with a "Disabled element" example, with workaround for Safari browser for the developers to decide what to use.

libraries/core-react/src/Tooltip/Tooltip.jsx
```diff
@@ -95,7 +95,12 @@ export const Tooltip = forwardRef(function Tooltip(
   return (
     <Wrapper {...props}>
       <Anchor
+        tabIndex={0}
         onMouseOver={handleOpen}
+        onMouseEnter={handleOpen}
+        onPointerEnter={handleOpen}
+        onPointerLeave={handleClose}
```
pomfrida, line 102: onPointerEnter and onPointerLeave fixed the issue in Chrome.

PR #480, opened 2020-08-13
Replace the old storefront with new one

vnys:


vnys:
There are some inconsistencies in core-react. Will start again to rule out human error.

PR #483, opened 2020-08-13
Fix file upload and link buttons

### LINKED ISSUE #482: File upload and link buttons don’t look right
**Describe the bug**

After introducing the span-element to fix the Safari bug in #465 – the upload and link buttons don’t look right

1. See the [link button in the storybook](https://eds-storybook-react.azurewebsites.net/iframe.html?id=components-button--link&viewMode=story)
2. See the [upload button in the storybook](https://eds-storybook-react-development.azurewebsites.net/iframe.html?id=components-button--file-upload&viewMode=story)
3. Shake your head in disbelief

**Expected behavior**

The buttons should look like they used to before the bug fix

**Specifications**

- Version: 0.5.1
- Browser: Edge
- OS: MacOS

**Additional context**

**File upload**

image

**Link**

image

vnys:
resolves #482

### DIFF
--- a/libraries/core-react/src/Button/Button.jsx
+++ b/libraries/core-react/src/Button/Button.jsx
@@ -13,6 +13,7 @@ const ButtonInner = styled.span`
   grid-gap: 8px;
   grid-auto-flow: column;
   align-items: center;
+  height: 100%;
 `
 
 const Base = ({ base, baseDisabled: disabled }) => {
@@ -94,6 +95,7 @@ const ButtonBase = styled.button`
   text-decoration: none;
   position: relative;
   cursor: pointer;
+  display: inline-block;
 
   &::before {
     position: absolute;


libraries/core-react/src/Button/Button.jsx
```diff
@@ -13,6 +13,7 @@ const ButtonInner = styled.span`
   grid-gap: 8px;
   grid-auto-flow: column;
   align-items: center;
+  height: 100%;
```
vnys, line 16: fixes vertical alignment bug in link button

libraries/core-react/src/Button/Button.jsx
```diff
@@ -94,6 +95,7 @@ const ButtonBase = styled.button`
   text-decoration: none;
   position: relative;
   cursor: pointer;
+  display: inline-block;
```
vnys, line 98: overrides default display type when sending in an inline tag in the as-prop.

Issue #486, opened 2020-08-14
The storefront does not hydrate in Safari


vnys:
**Describe the bug**

The storefront does not hydrate in Safari, causing full page loads instead of client-side navigation upon clicking on links in the menu.

**Steps to reproduce the bug**

1. Go to https://eds.equinor.com
2. Click a link in the menu
3. Observe how the progress bar underneath the address bar indicates a full page load

The same goes for https://eds-storefront-development.equinor.com, which uses Docz.

**Expected behavior**

After the initial page load, React should take over and clicking links in the menu should not cause a full page load

**Specifications**

- Version: n/a
- Browser: Safari
- OS: macOS Mojave

**Additional context**

The fact that this has been going on for quite some time without us noticing actually means Gatsby as a concept works quite well – as long as you have a fallback that works without React. The old menu uses checkboxes and css to toggle submenus, but the new one i the Docz version of the Storefront is pure React – and hence does _not_ work without it. 

@wenche found out the reason Safari doesn’t render React could be related to lookbehind in RegEx which is not supported in Safari. So that would be the first place to look. In the Docz version of the storefront, there’s also an error in the console about an unhandled promise rejection – but that could be related to the first error. 

One way to deal with this could be to bring back some of the logic in the old menu so you could at least navigate without React – but the main rationale for moving to Docz was the playground component which is useless without React.

PR #487, opened 2020-08-17
Fix missing lookbehind support in Safari

wenche:
solve #486


apps/storefront-docz/src/components/figmaImage.jsx
```diff
@@ -41,11 +41,15 @@ const Title = styled(Link)`
...
-    '_',
-  )
+  // Can't use lookbehind because of Safari's lack of support
+  //const fileId = (/(?<=file\/).*(?=\/)/.exec(url) || [])[0] || ''
+  //const nodeId = ((/(?<=node-id=).*/.exec(url) || [])[0] || '').replace(
+  //  '%3A',
+  //  '_',
+  //)
+  const splitOnFile = url.split(/(file\/.*?\/)/)
+  const fileId = splitOnFile[1].split(/(\/)/)[2] || ''
```
mimarz, line 51: This could potentially give a `undefined` exception of `splitOnFile` if the split doesn't not split

PR #488, opened 2020-08-17
Popover and tooltip consistensy

pomfrida:
`Margin-right: auto` on both anchors needed for them to not ruin table cell structure. See the 'Custom Table' the Table examples in the Storybook.


wenche (changes requested):
But as far as I can see, the table example highlights the issue in the storefront with the arrow marker being calculated to the wrong place
Skjermbilde 2020-08-19 kl  10 36 45

pomfrida:
@wenche Omg 🤣  🔫  Simple magic fixes everything: display: inline-flex instead of flex. I'll remove the custom table demo before merging, it's just for you guys to see

PR #491, opened 2020-08-18
Move updated files from old documentation branch

vnys:
Two files got lost between the old and the new storefront. Let’s merge them into develop and get them out into the world.


BeckyBrekke:
@vnys @wenche 
I can't find this page: https://eds.equinor.com/patterns/data-visualisation/ and found this PR that looks like it was supposed to fix that. Is it lost in space some place?

PR #494, opened 2020-08-19
Add Wenche and Frida to codeowners

vnys:
Adding Wenche and Frida to CODEOWNERS means every time we make a pull request we’re all added as reviewers by default for everything but the documentation – which is owned by @BeckyBrekke.


pomfrida (approved):
Nice 👍

PR #497, opened 2020-08-19
Hide drafts in prod

### LINKED ISSUE #495: Draft mode pages showing in storefront
Storefront pages that are in draft mode used to be hidden from users accessing the storefront. Now they are visible. 

Go to storefront and see that all pages are visible regardless of their mode. 
Example:   
Faq Storefront: https://eds.equinor.com/help/faq/
Github: https://github.com/equinor/design-system/blob/documentation/apps/storefront/docs/help/faq/faq.mdx

**Expected behavior**
The content of the FAQ page should be not shown. I can't remember if the page was still shown but empty or if it was not in the navigation at all. We discussed what was best a while back. Victor or Mickey will remember hopefully. 

**Specifications**

- Version: 
- Browser: 
- OS: 

**Additional context**

vnys:
Preventing drafts from being built is a pain in the 🍑 which requires us to tap into Gatsby’s `createPages` api without getting in the way of Docz – so I’ve chosen a path of less resistance – which is preventing drafts from showing in the menu and replacing the content of the drafts pages with the word “placeholder” and also adding `<meta name="robots" content="noindex" />` to the drafts pages to prevent them from being indexed. All this is happening in prod, in dev the content of the drafts will be visible. 

I’ve also removed the public folder, as this was committed before public was added to .gitignore – so just scroll past that part when reviewing the code.

Update: I’ve built the storefront for prod and deployed it to https://eds-storefront-redesign.azurewebsites.net. If you open the menu on “Help”, you’ll see that “FAQ” is missing – compared to https://eds-storefront-development.azurewebsites.net. The FAQ-page is [still there](https://eds-storefront-redesign.azurewebsites.net/help/faq/), but does not show the contents compared to the [development stage](https://eds-storefront-development.azurewebsites.net/help/faq/). Run `document.getElementsByTagName('meta').robots` in the console to print out the meta-tag.

resolves #495


apps/storefront/src/gatsby-theme-docz/components/MainContainer/index.js
```diff
@@ -121,75 +129,91 @@ export const MainContainer = ({ children, doc, ...rest }) => {
...
-                    const routeSegment = tab.toLowerCase().replace(/\s/g, '-')
-                    const firstTwoRouteSegments = /^\/([a-z-]+\/?){2}/
-                    const categoryRoute = route.match(firstTwoRouteSegments)[0]
+        isPrivate ? (
+          <>
+            <Helmet>
+              <meta name="robots" content="noindex" />
+            </Helmet>
+            <ContentHeader>
+              <H1>Placeholder</H1>
```
vnys, line 137: Show the text “placeholder” instead of actual header to prevent showing up in search results

apps/storefront/src/gatsby-theme-docz/components/MainContainer/index.js
```diff
@@ -121,75 +129,91 @@ export const MainContainer = ({ children, doc, ...rest }) => {
...
-              <nav aria-label="tabbed content naviagtion">
-                <Tabs>
-                  {tabs.map((tab, index) => {
-                    const routeSegment = tab.toLowerCase().replace(/\s/g, '-')
-                    const firstTwoRouteSegments = /^\/([a-z-]+\/?){2}/
-                    const categoryRoute = route.match(firstTwoRouteSegments)[0]
+        isPrivate ? (
+          <>
+            <Helmet>
+              <meta name="robots" content="noindex" />
```
vnys, line 134: Prevent indexing of drafts by search engines

apps/storefront/package.json
```diff
@@ -42,6 +42,7 @@
     "polished": "^3.6.5",
     "prop-types": "^15.7.2",
     "query-string": "^6.13.1",
+    "ramda": "^0.27.1",
```
vnys, line 45: Thought I’d try something fancy 🙄

apps/storefront/src/gatsby-theme-docz/components/Sidebar/index.js
```diff
@@ -123,15 +124,18 @@ export const Sidebar = ({ className, open, onClick }) => {
...
+
+  const subMenus = useMemo(
+    () =>
+      R.pipe(
+        R.filter(notRootRoute),
+        R.filter(routesWithTwoSegments),
+        R.filter(handleDrafts),
+        R.reduce((acc, { title, route }) => [...acc, { title, route }], []),
+      )(docs),
+    [docs],
```
vnys, line 138: Not sure there’s any real benefit to using Ramda here instead of native filter and reduce, but one thing you _can_ do which is quite nice is to intercept a piped function with R.tap(console.log) to see what’s happening at that particular point. Also since we’re using Ramda extensively in Figma Broker, we might as well get into the habbit of using it in other places as well – it’s a powerful library.

mimarz: Since you are using Ramda, maybe the `R.reduce` for picking props can be replaced with [R.pick](https://ramdajs.com/docs/#pick) ?

apps/storefront/src/gatsby-theme-docz/components/Sidebar/index.js
```diff
@@ -123,15 +124,18 @@ export const Sidebar = ({ className, open, onClick }) => {
...
     doc.route.replace(/^\/|\/$/g, '').split('/').length < 3
 
-  const subMenus = useMemo(() =>
-    docs
-      .filter(notRootRoute)
-      .filter(routesWithTwoSegments)
-      .map((doc) => ({
-        title: doc.title,
-        route: doc.route,
-        current: current.route.includes(doc.route),
```
mimarz, line 133: current is removed, is this intended?

mimarz (approved):
Looks good 👍  Just a tip on Ramda and question on one prop no longer being picked

apps/storefront/src/gatsby-theme-docz/components/Sidebar/index.js
vnys: Yes, the `<Link />` component handles this on it’s own

BeckyBrekke:
Proposed solution sounds fine. I guess the most important thing if that the pages don't show up in the menu.

Issue #249, opened 2020-03-02
Add TypeScript type definitions


vnys:
- [ ] **Add types for eds-tokens**
- [x] **Add types for eds-icons**
- [ ] **Add types for eds-core-react:**
  - [ ] Accordion
  - [x] Avatar
  - [ ] Banner 
  - [ ] Breadcrumbs
  - [ ] Button
  - [ ] Card
  - [ ] Chip
  - [ ] Dialog
  - [x] Divider
  - [x] Icon
  - [ ] List
  - [ ] Pagination
  - [ ] Popover
  - [ ] Progress
  - [ ] Scrim
  - [ ] Search
  - [ ] Selection controls
  - [ ] SideSheet
  - [ ] Slider
  - [ ] Snackbar
  - [ ] Table
  - [ ] TOC
  - [ ] Tabs
  - [x] Textfield
  - [ ] Tooltip
  - [x] Top bar
  - [x] Typography

PR #507, opened 2020-08-21
Types for tokens library

wenche:
One step for #249


wenche:
I think that we perhaps can rely on type inference for this library, it's just some plain objects and ts should figure the types by itself.

PR #511, opened 2020-08-24
Hide drafts tabs in the storefront

vnys:
In [production ](https://eds.equinor.com) pages that are in drafts mode are hidden, while they're visible in [development](https://eds-storefront-development.azurewebsites.net). This pr makes sure tabs that link to pages in drafts mode behave in the same way.

### DIFF
--- a/apps/storefront/docs/updates/release-information/known-issues.mdx
+++ b/apps/storefront/docs/updates/release-information/known-issues.mdx
@@ -8,3 +8,5 @@ tabs:
 mode: draft
 route: /updates/release-information/known-issues/
 ---
+
+Here be dragons…
--- a/apps/storefront/src/gatsby-theme-docz/components/MainContainer/index.js
+++ b/apps/storefront/src/gatsby-theme-docz/components/MainContainer/index.js
@@ -132,24 +132,18 @@ export const MainContainer = ({ children, doc, ...rest }) => {
 
   const isDraft = findByRoute(drafts)
 
-  const reduceIndexed = R.addIndex(R.reduce)
+  const mapIndexed = R.addIndex(R.map)
 
   const toLowerRemoveSpaces = R.pipe(R.toLower, R.replace(/\s/g, '-'))
 
   const publishedTabs = (route) => (tabs) =>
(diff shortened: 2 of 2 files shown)


apps/storefront/src/gatsby-theme-docz/components/MainContainer/index.js
```diff
@@ -110,15 +111,62 @@ const LandingPage = styled.div``
...
+            route:
+              route.match(/^\/([a-z-]+\/?){2}/)[0] +
+              (i === 0
+                ? ''
+                : R.pipe(R.toLower, R.curry(R.replace)(/\s/g, '-'))(curr) +
+                  '/'),
+          },
+        ],
+        [],
+      ),
```
vnys, line 158: 🤦

apps/storefront/docs/updates/release-information/known-issues.mdx
```diff
@@ -8,3 +8,5 @@ tabs:
 mode: draft
 route: /updates/release-information/known-issues/
 ---
+
+Here be dragons…
```
vnys, line 12: `<MainContainer>` needs children

apps/storefront/src/gatsby-theme-docz/components/MainContainer/index.js
vnys: Not particularly pleased with how that publishedTabs function looks.

mimarz: @vnys It looks like you are using the index to ignore the first entry in the list. Maybe you can use the [tail](https://ramdajs.com/docs/#tail) function instead? Would be something like this instead, `pipe(tail, map, filter)`

vnys: Actually I don’t ignore the first entry in the list, I just do something different to the route. Not sure how I would go about doing something like that with Ramda. But i replaced reduce with map, makes it a bit easier to read.

wenche (approved):
Looks good 👍

PR #519, opened 2020-08-26
Removed size prop in Button Icon Story

pomfrida:
The size prop isn't supported inside buttons yet so this code was removed to avoid confusion


mimarz (approved):
LGTM 👍

PR #524, opened 2020-08-28
Avatar types

wenche:
Partially solves #249


wenche:
Had to rewrite the tests slightly because of an error. 🤷‍♀️

PR #527, opened 2020-08-31
Typescript eslint

### LINKED ISSUE #522: Configure Eslint for typescript
- https://github.com/typescript-eslint/typescript-eslint
- https://github.com/typescript-eslint/typescript-eslint/blob/master/docs/getting-started/linting/README.md

vnys:
resolves #522 

Eslint complains about a lot of stuff that has to do with `any` types. We need to discuss what do do about those. We _can_ set the error level to warn for now, that would at least give us constant reminders that we are doing something wrong 😁 

## One important thing!

In package.yaml there are two scripts, one called `lint` and the other called `lint:all` which calls `lint` with a specified path of `.`. That means that if you want to run eslint in _only_ the libraries folder for example, you can do that with `pnpm run lint libraries`


.eslintrc.js
```diff
@@ -0,0 +1,122 @@
...
+        labelAttributes: [],
+        controlComponents: [],
+        assert: 'either',
+        depth: 25,
+      },
+    ],
+  },
+  overrides: [
+    {
+      files: ['**/*.jsx'],
```
vnys, line 95: default is js, this adds jsx files too without having to specify it as a cli argument

tsconfig.base.json
```diff
@@ -0,0 +1,17 @@
+{
+  "compilerOptions": {
+    /* some comment */
```
vnys, line 3: 🙈

wenche (approved):
Let's try it out 👍

mimarz (approved):
Looks good, I guess 🤪👍

PR #531, opened 2020-08-31
[WIP]: Scrim types

wenche:
Part of #249


wenche:
With this I'm stuck with an empty props in docgen. I don't know why this i happening. From es build file
```
try {
  Scrim.displayName = "Scrim";
  Scrim.__docgenInfo = {
    description: "",
    displayName: "Scrim",
    props: {}
  };
} catch (e) {}
```

wenche:
My impression is that the code and the tests itself have some flaws... I think we are struggling both with TS and suboptimal code here. E.g.
```
 /** Stupid hack to avoid crash on ...rest*/
  [x: string]: any
```
Quick fix from mighty Internet to avoid crash in the tests with `<Scrim **onKeyDown**={handleClose}>`

Also it's hard to wrap my head around the differents between TS KeyboardEvent and React.KeyboardEvent. It's a mess right now :/

wenche:
Closing this pull request since it is far behind the rest of the code base. (Aka long lived PR's are not a good idea :) )

wenche:
See PR #654

PR #592, opened 2020-09-10
Bugfix onClose for Menu

pomfrida:


libraries/core-react/src/Menu/Menu.test.jsx
```diff
@@ -26,8 +26,9 @@ describe('Menu', () => {
     expect(menuContainer).toHaveStyleRule('background', 'red')
   })
   it('is visible when open is true & anchorEl is set', () => {
+    const handleOnClose = jest.fn()
     render(
-      <Menu open>
+      <Menu open onClose={handleOnClose}>
```
mimarz, line 31: remove the new `onClose` functions in tests. onClose is optional so should be tested accordingly

mimarz (approved):
Looks like some testing code is still present in the `Menu.stories` that needs to be removed before merging, the `onClick` and `onClose={onClick}` handlers. 

Apart from that it looks might sweet! 🍬👍

apps/storybook-react/stories/Menu.stories.jsx
```diff
@@ -68,9 +68,9 @@ export default {
 
 const simpleMenuTemplate = (
   <>
-    <MenuItem>Item 1</MenuItem>
-    <MenuItem>Item 2</MenuItem>
-    <MenuItem>Item 3</MenuItem>
+    <MenuItem onClick={onClick}>Item 1</MenuItem>
+    <MenuItem onClick={onClick}>Item 2</MenuItem>
+    <MenuItem onClick={onClick}>Item 3</MenuItem>
```
mimarz, line 73: dis

apps/storybook-react/stories/Menu.stories.jsx
```diff
@@ -223,7 +223,12 @@ export const EdgeDetection = () => {
       >
         Top left
       </FloatingAnchor>
-      <Menu id="menu-topleft" open={Boolean(topLeft)} anchorEl={topLeft}>
+      <Menu
+        id="menu-topleft"
+        open={Boolean(topLeft)}
+        anchorEl={topLeft}
+        onClose={onClick}
```
mimarz, line 230: & dis

apps/storybook-react/stories/Menu.stories.jsx
vnys: What?

apps/storybook-react/stories/Menu.stories.jsx
vnys: What dis @mimarz? 😁

vnys: If it’s the name of the event handler you’re referring to, I would call that something else than `onClick` since it’s being called by other events as well.

vnys: …preferably something that starts with `handle` since that’s the convention we agreed on way back some time :wink: `handleClick`, `handleChange` etc.

vnys: This syntax may come in handy 🤷 https://github.com/equinor/design-system/blob/develop/apps/storybook-react/stories/Accordion/Accordion.stories.jsx#L119-L122

vnys (approved):
Minor stuff, apart from that LGTM 👍

PR #593, opened 2020-09-10
Support for directly setting icon on Icon component

### LINKED ISSUE #584: Provide an alternative to Icon.add

Normally, when we use icons in our components, we will import the specific icon and use it directly, e.g.

```jsx
import AssignmentIcon from '@material-ui/icons/Assignment';

<AssignmentIcon />
```

EDS requires pre-registering the icons via `Icon.add`, which means we have to maintain this list in our project separately from the component that uses it, and doesn't feel particularly developer-friendly.

I would like to avoid using `Icon.add`. The simplest option that occurs to me is to have a data property to Icon where we can pass the object imported from eds-icons, e.g.

```jsx
import { save } from '@equinor/eds-icons';

<Icon data={save} />
```

Alternatively. provide pre-defined components for each icon, like material-ui. The latter option seems like a lot of work unless it was automatically generated.

Currently, we have just written a wrapper for Icon that automatically invokes Icon.add for us via hooks.

mimarz:
* Use `data` property to directly set icon data being rendered
* Provided data object must follow [icon object structure](https://github.com/equinor/design-system/tree/develop/libraries/icons#example-of-javascript-object-data)
* Code housekeeping
* Fixed malformed proptypes description (even though we are changing it soon 🤪)

resolves #584

### DIFF
--- a/apps/storybook-react/stories/Icon.stories.jsx
+++ b/apps/storybook-react/stories/Icon.stories.jsx
@@ -70,6 +70,24 @@ export const IconExamples = () => (
     <Wrapper>
       <Icon name="grid_off" title="Save me!" />
     </Wrapper>
+    <h2>Using data prop with eds-icon</h2>
+    <Wrapper>
+      <Icon data={save} />
+    </Wrapper>
+    <h2>Using data prop with custom icon</h2>
+    <Wrapper>
+      <Icon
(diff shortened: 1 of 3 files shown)


vnys (approved):
LGTM 👍

Issue #608, opened 2020-09-15
Divider is not displayed


sudoole:
**Describe the bug**

The Divider component is not visible. I have to manually set a width for it to be displayed.

**Steps to reproduce the bug**

1. Import the Divider component: `import { Dialog, Divider } from "@equinor/eds-core-react"`;
2. Use it in a React component:
```
function DividerTest(props: CreateOrderComponentInterface) {
  return (
    <Dialog>
      <Title>Title</Title>
      Before
      <Divider />
      After
    </Dialog>
  );
}
```
3. The result does not show a divider where i placed it, but the divider used within the Dialog component is displayed:
![divider-original](https://user-images.githubusercontent.com/254171/93210807-8cd65800-f760-11ea-8322-ac74863cdac5.png)
4. Inspecting the components shows that the Divider component misses a width property when comparing to the one that appear under the Dialog title:
![Selection_012](https://user-images.githubusercontent.com/254171/93211333-5e0cb180-f761-11ea-8a07-11d40f6f51b1.png)

5. Manually setting width will display the divider:
```
const Divider = styled(EdsDivider)`
  width: 100%;
`;
```
results in:
![divider-custom](https://user-images.githubusercontent.com/254171/93211456-901e1380-f761-11ea-9966-d433e752de37.png)

**Expected behavior**

I expect a visible divider.

**Specifications**

- Version: 0.6.1
- Browser: Firefox
- OS: Linux

**Additional context**

pomfrida:
Thanks for reporting this issue. Added a PR fix so you won't have to manually set the width any more in grid elements.

PR #609, opened 2020-09-16
Divider width 100%

pomfrida:
Solves #608


mimarz (changes requested):
What bug issue does this fix?

mimarz (approved):
LGTM 👍

PR #613, opened 2020-09-16
Move application of header background to thead instead of Cell

### LINKED ISSUE #610: Table header does not fill width when Table is set to 100% width
**Describe the bug**

When setting the table witdh to `100%` the header does not fill width as cells do.

image

1. `width: 100%;` on `Table`

**Expected behavior**

Expect the styling to match that of the cell filling up the width of the table

image

**Additional context**

We apply styling the the `th` element only. Maybe applying the the header background and border to `thead` instead would make this look better

mimarz:
* Header background is applied to `Head` instead of `Cell` when as is `th`
  * Will now feel more native when `thead` is stretched to fit table width
* Created `Table.tokens` file and moved relevant code there
* Simplified token extraction from`table` token
* Added example with fixed header in stories

resolves #610

### DIFF
--- a/libraries/core-react/src/Table/Table.tokens.js
+++ b/libraries/core-react/src/Table/Table.tokens.js
@@ -1,0 +1,33 @@
+import tableTokens from '@equinor/eds-tokens/components/table/table.json'
+
+const { header, cell } = tableTokens
+
+const variants = {
+  header: { text: header.text },
+  cell: {
+    text: cell.text,
+    numeric: cell.monospaced_numeric,
+    icon: cell.icon,
+    input: cell.input,
+  },
+}
+
+export const getTokens = (as, variant) => {
+  switch (as) {
+    case 'th':
+      return {
+        ...variants.header[variant],
+        // borders: {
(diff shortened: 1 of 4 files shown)


mimarz:
Example with fixed header can be omitted. Most important was to show `thead`being streched.

libraries/core-react/src/Table/Table.tokens.js
```diff
@@ -0,0 +1,33 @@
...
+      return {
+        ...variants.header[variant],
+        // borders: {
+        //   thead: variants.header[variant].borders,
+        //   tbody: variants.cell[variant].borders,
+        // },
+        // background: {
+        //   thead: variants.header[variant].background,
+        //   tbody: variants.cell[variant].background,
+        // },
```
vnys, line 27: Remove comments

vnys (approved):
LGTM 🌮 Just some comments that should be removed…

libraries/core-react/src/Table/Table.tokens.js
mimarz: Fixed

PR #625, opened 2020-09-23
Added typescript declarations to core-react

### LINKED ISSUE #624: Add typescript declarations for eds-core-react

mimarz:
* Turns out the official `@rollup/plugin-typescript` plugin still has some [issues...](https://github.com/rollup/plugins/issues?q=is%3Aissue+is%3Aopen+typescript).
* Switching to `rollup-plugin-typescript2` works 👍
* Added missing `ts-no-checks` for now since we had to change `index.js` file to `.ts` for declarations to work
* Declarations are saved under `dist/types` as they are not bundled into one file and thus follow the same folder structure as is under `/src` but as type files

image

resolves #624

### DIFF
--- a/libraries/core-react/package.json
+++ b/libraries/core-react/package.json
@@ -5,6 +5,7 @@
   "main": "dist/core-react.cjs.js",
   "module": "dist/core-react.es.js",
   "browser": "dist/core-react.umd.js",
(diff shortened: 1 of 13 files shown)


vnys (approved):
LGTM 👍

PR #641, opened 2020-10-01
🐛 fix missing props spread on BannerActions

### LINKED ISSUE #478: BannerAction doesn't applies CSS with the style prop
**Describe the bug**

The BannerAction component is missing the `{...props} ` on the rendered element. Inline styles with prop style={{}} will not have affect (and other props as well ofc)

Solution: Add props spread

SvSven:
Fixes #478


vnys (approved):
LGTM 👍

PR #657, opened 2020-10-08
Fix pagination bug

### LINKED ISSUE #647: Pagination component generates react warning on first and last component
Get a react warning when Pagination reaches first or last page.

The bug in console from react:
"Expected `onClick` listener to be a function, instead got `false`."
"If you used to conditionally omit it with onClick={condition && value}, pass onClick={condition ? value : undefined} instead."

1. Use the Pagination component and look in console. 
2. Click on first or last page.
3. Will happen each time activePage < 1 or activePage > pages

**Expected behavior**

Expect not to get any warnings from component.

**Specifications**

- Version: 0.6.0
- Browser: Chrome
- OS: Windows

**Additional context**

image

Fix will probably be changing from conditionally omit to ternary in these two places:
https://github.com/equinor/design-system/blob/develop/libraries/core-react/src/Pagination/Pagination.jsx#L117-L120
https://github.com/equinor/design-system/blob/develop/libraries/core-react/src/Pagination/Pagination.jsx#L158-L161

Before:
```javascript
activePage > 1 && ((event) => { onPageChange(event, activePage - 1) })
```

To:
```javascript
activePage > 1 ? ((event) => { onPageChange(event, activePage - 1) }) : undefined
```

vnys:
resolves #647

### DIFF
--- a/libraries/core-react/src/Pagination/Pagination.jsx
+++ b/libraries/core-react/src/Pagination/Pagination.jsx
@@ -114,10 +114,11 @@ export const Pagination = forwardRef(function Pagination(
           <Button
             variant="ghost_icon"
             onClick={
-              activePage > 1 &&
-              ((event) => {
-                onPageChange(event, activePage - 1)
-              })
+              activePage > 1
+                ? (event) => {
+                    onPageChange(event, activePage - 1)
+                  }
+                : undefined
             }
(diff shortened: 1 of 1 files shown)


mimarz (approved):
LGTM 👍

wenche (approved):
👍

PR #658, opened 2020-10-08
✨ Types for List

### LINKED ISSUE #575: Refactor Lists to Typescript
Create typings for [Lists](https://eds.equinor.com/components/lists/)

wenche:
Resolve #575


wenche:
It seems that this way of reexporting in `./index` don't create a good experience in the editor when someone hovers the component.
Scrim

vs List


We probably need to solve this by using another technique.

vnys:
So this happens if we `import` and then `export` instead of just `export` directly?

```jsx
import { List } from './List'
import { ListItem } from './ListItem'

List.ListItem = ListItem

export { List }
```

vs. 

```jsx
export { Scrim } from './Scrim'
```

We could move `List.ListItem = ListItem` to `List`? Would that help?

mimarz:
> So this happens if we `import` and then `export` instead of just `export` directly?
> 
> ```js-jsx
> import { List } from './List'
> import { ListItem } from './ListItem'
> 
> List.ListItem = ListItem
> 
> export { List }
> ```
> 
> vs.
> 
> ```js-jsx
> export { Scrim } from './Scrim'
> ```
> 
> We could move `List.ListItem = ListItem` to `List`? Would that help?

Its because of the first example _in combination with typings_. Look at [TopBar](https://github.com/equinor/design-system/blob/typescript/libraries/core-react/src/TopBar/index.tsx) as an example

We still have to test and see how this looks like when consumed. Best way would be to update Storybook stories to `.tsx`

wenche:
When I was fiddling with the `Banner` component, I did it like this:

In Banner.tsx
```
import { BannerActions, BannerActionsProps } from './BannerActions'
import { BannerIcon, BannerIconProps } from './BannerIcon'
import { BannerMessage, BannerMessageProps } from './BannerMessage'
...

type Props = {
  BannerActions?: BannerActionsProps
  BannerIcon?: BannerIconProps
  BannerMessage: BannerMessageProps
}
...
Banner.BannerActions = BannerActions
Banner.BannerIcon = BannerIcon
Banner.BannerMessage = BannerMessage

Banner.displayName = 'Banner'

export { Banner }
```

In index.ts
```
import { Banner } from './Banner'
import { BannerIcon } from './BannerIcon'
import { BannerMessage } from './BannerMessage'
import { BannerActions } from './BannerActions'

Banner.BannerIcon = BannerIcon
Banner.BannerMessage = BannerMessage
Banner.BannerActions = BannerActions

export { Banner }
```
I can't remember if that worked 100% percent but I think so? 🤔

wenche:
Skjermbilde 2020-10-09 kl  09 50 31

Auto suggestions in the code works

vnys:
But why is that `import` done twice? If those child components are imported into `Banner`, it shouldn’t be necessary to do it again in `index`?

wenche:
> But why is that `import` done twice? If those child components are imported into `Banner`, it shouldn’t be necessary to do it again in `index`?

Don't forget rule no 1: We focus on making things works without warnings and errors, we have no idea if things actually makes sense 😂  I can't remember if it didn't worked or if I just didn't got that far, but the Banner wasn't finished at that point :)

vnys (approved):
LGTM 👍

PR #667, opened 2020-10-12
Add types for table component

### LINKED ISSUE #573: Refactor Table to Typescript
Create typings for [Table](https://eds.equinor.com/components/table/)

wenche:
Resolve #573


wenche:
Finally a component without refs 😓 

We have a eslint rule of [explicit return type.](https://github.com/typescript-eslint/typescript-eslint/blob/v3.10.1/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.md) If I just added prop types, it was not happy.

Skjermbilde 2020-10-12 kl  13 16 25
I added React.FunctionComponent as proposed by [Fettblog](https://fettblog.eu/typescript-react/components/) and a lot of other places on the mighty Internet

```
type Props = React.HTMLAttributes<HTMLTableElement>

export const Table: FunctionComponent<Props> = (props) => {
  const { children } = props
  return <TableBase {...props}>{children}</TableBase>
}
```
🤷‍♀️

libraries/core-react/src/Table/Table.tsx
```diff
@@ -8,7 +7,9 @@ const TableBase = styled.table`
   border-collapse: collapse;
 `
 
-export const Table = (props) => {
+type Props = React.HTMLAttributes<HTMLTableElement>
+
+export const Table: FunctionComponent<Props> = (props: Props) => {
```
vnys, line 12: Should we use `React.FC` instead of `FunctionComponent` since we’re using `React.HTMLAttributes`? 🤔

vnys:
> I added React.FunctionComponent as proposed by [Fettblog](https://fettblog.eu/typescript-react/components/) and a lot of other places on the mighty Internet

Great guide, haven’t seen it before 😊 And didn’t know he was the author of the Smashing Magazine book, maybe I should buy that 🤔 Or get a Smashing Magazine membership and get access to all their books.

libraries/core-react/src/Table/Table.tsx
vnys: …just a thought. But if the fettblog author don’t recommend using shorthand I’m not going to argue against that 😛

wenche: Might be? I'm actually not sure about what I think about `React.HTMLAttributes`... You would after all write 
```
import React, {useState} from 'react
...
const t = useState(...)
```
instead of React.useState right? So I kind of feel it's quite verbose to repeat React ?

vnys: True. So then maybe we should change that instead, and `import { HTMLAttributes }`.

mimarz: What is you just sett the return type to `JSX:Element` ? 

Repeating "Props" in generic and parameters seems messy

wenche: Maybe? I think the current code is more the result of copy'n paste from different sources than consistency

wenche: I don't actually feel we are repeating more than usual? But could write it like this instead
```
type Props = React.HTMLAttributes<HTMLTableElement>

export const Table: FunctionComponent<Props> = ({children) => {
  return <TableBase {...props}>{children}</TableBase>
}
```

mimarz: On second thought, I dont think you need to set the type on the parameter when you are setting the type on variable declaration @wenche

wenche: And yeah, you can do it like that @mimarz 
```
type Props = React.HTMLAttributes<HTMLTableElement>

export const Table = (props: Props): JSX.Element => {
  const { children } = props
  return <TableBase {...props}>{children}</TableBase>
}
```

wenche: Maybe just tune that Eslint rule? 
https://www.reddit.com/r/reactjs/comments/i4jx85/reactfc_vs_jsxelement/

PR #681, opened 2020-10-15
Bring back the docgen plugin

### LINKED ISSUE #622: Find alternative to docgen for props table generation
`babel-plugin-react-docgen-typescript` does not seemed to be [maintained](https://github.com/strothj/babel-plugin-react-docgen-typescript/issues/18) anymore.

* Find alternative? [TSdoc](https://github.com/microsoft/tsdoc)?
* Fork `babel-plugin-react-docgen-typescript` ?

vnys:
The babel-plugin-react-docgen has been updated and works as expected as far as I can see. It doesn’t have a maintainer at the moment, so long term we should either take on maintainer responsibility or find a better solution. I expect there will be more suitable tools available once we teak the leap to typescript for the storybook as well as the components.

resolves #622

### DIFF
--- a/libraries/core-react/rollup.config.js
+++ b/libraries/core-react/rollup.config.js
@@ -18,6 +18,9 @@ const globals = {
 }
 
 const extensions = ['.jsx', '.js', '.tsx', '.ts']
+
+const buildForStorybook = process.env.STORYBOOK
+
 export default [
   {
     input: './src/index.ts',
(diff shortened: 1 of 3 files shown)


libraries/core-react/rollup.config.js
```diff
@@ -49,10 +55,7 @@ export default [
         sourcemap: true,
         globals,
       },
-      {
-        file: pkg.main,
-        format: 'cjs',
-      },
+      ...(buildForStorybook ? [] : [{ file: pkg.main, format: 'cjs' }]),
```
mimarz, line 58: Why do we not build the `commonjs` dist for storybook?

vnys: Because the storybook don’t use that file so it’s a waste of time

mimarz (approved):
LGTM 👍

PR #686, opened 2020-10-16
Accordian typescript

### LINKED ISSUE #564: Refactor Accordion to Typescript
Create typings for [Accordion](https://eds.equinor.com/components/accordion/)

mimarz:
resolves #564


libraries/core-react/src/Accordion/Accordion.tokens.ts
```diff
@@ -23,11 +23,24 @@ const {
   },
 } = tokens
 
-export const accordion = {
+type AccordianToken = {
```
wenche, line 26: You mix Accordion and Accordian :)

libraries/core-react/src/Accordion/AccordionItem.tsx
```diff
@@ -16,7 +24,7 @@ const AccordionItem = forwardRef(function AccordionItem(
   },
   ref,
 ) {
-  const [expanded, setExpanded] = useState(isExpanded)
+  const [expanded, setExpanded] = useState<boolean>(isExpanded)
```
wenche, line 27: I wasn't sure if we should add <boolean> explicit or not since Typescript used type inference correct and eslint was  happy 🤷‍♀️  But maybe it is a good idea to write it explicit anyway?

wenche (changes requested):

libraries/core-react/src/Accordion/Accordion.tokens.ts
mimarz: Fixed!

libraries/core-react/src/Accordion/AccordionItem.tsx
mimarz: Probably something that snuck in when I was refactoring. It was inferred correctly when I removed it 👍

wenche (approved):
LGTM 👍

PR #689, opened 2020-10-16
Toc types

### LINKED ISSUE #558: Refactor Navigation Table of contents to Typescript
Create typings for [Navigation Table of contents](https://eds.equinor.com/components/navigation-table-of-contents/)

wenche:
Resolve #558


wenche:
Started to write Type for the token object, but since it on a completely different format than the other tokens, it didn't make sense beyond TS automatic detection 🤷‍♀️  Should probably be aligned with a more common token object format

vnys:
I’m scratching my head a bit too with the Dialog tokens… not sure if it gives us something or if it’s just a waste of time

wenche:
It gives us something if we use a common format, but quite a few of the components does not

libraries/core-react/src/TableOfContents/TableOfContents.tsx
```diff
@@ -1,13 +1,20 @@
-// @ts-nocheck
-import React, { forwardRef } from 'react'
-import PropTypes from 'prop-types'
+import React, { forwardRef, HTMLAttributes } from 'react'
 import styled, { css } from 'styled-components'
 import { List, Typography } from '..'
```
mimarz, line 3: This needs to be updated to direct imports instead of .. imports #708

libraries/core-react/src/TableOfContents/TableOfContents.tsx
```diff
@@ -19,7 +26,7 @@ const StyledTableOfContents = styled.nav`
     `}
 `
 
-const TocList = styled((props) => <List {...props} />)`
+const TocList = styled((props: ListType) => <List {...props} />)`
```
mimarz, line 29: This has nothing to do with the typescript rewrite, more clean code. But these should be normal "extends". 

For examples; 
`const TocList = styled((props: ListType) => <List {...props} />)` 
is equivalent 
 `const TocList = styled(List)`

mimarz (changes requested):
Looks good from a typescript point of view. Need to fix some imports as discovered in #708 for tests to run. We have also updated the failing test, so rebasing with latest from typescript to test could help.

There is also a clean code mention I think you can do. This would eliminate the need to export `ListType` from `<List>`

wenche:
Moved to #714 to avoid lot's of changes outside the TOC changes 🙈

PR #690, opened 2020-10-16
Typescript conversion of Chips

### LINKED ISSUE #581: Refactor Chips to Typescript
Create typings for [Chips](https://eds.equinor.com/components/chips/)

Make sure you fork the repository and branch from `typescript` before starting on this issue

joelmheim:
Resolves #581 

Pair programmed with @nicholasdalhaug .

Tests for Button and Chips fails due to problems with the same tokens file. This was also the case before our typescript conversion of Chips.


libraries/core-react/src/Chip/Chip.tsx
```diff
@@ -36,6 +35,8 @@ const StyledChips = styled.div.attrs(({ clickable, deletable }) => ({
   grid-auto-columns: max-content;
   align-items: center;
   z-index: 999;
+  tabIndex: ${({ clickable, deletable }) => (clickable || deletable ? 0 : null)},
```
wenche, line 38: You can't rewrite it like this, as role and tabIndex is not CSS and it has to be in `attrs` function of styled components as it used to be. See e.g. MenuItem.tsx:
```
const ListItem = styled.li.attrs<StyleAttrsProps>(({ isFocused }) => ({
  role: 'menuitem',
  tabIndex: isFocused ? -1 : 0,
}))<StyleProps>`
  width: auto;
  position: relative;
  z-index: 2;

  ${typographyTemplate(typography)}
  ${spacingsTemplate(spacings)};
...
`
```

libraries/core-react/src/Chip/Chip.tsx
```diff
@@ -124,27 +124,38 @@ const StyledChips = styled.div.attrs(({ clickable, deletable }) => ({
...
     css`
       padding-left: 8px;
     `}
-
 `
-
-export const Chip = forwardRef(function EdsChips(
-  { children, onDelete, disabled, onClick, variant, ...rest },
+type Props = {
+  disabled?: boolean,
```
wenche, line 134: Would you please add a `/** Comment describing what this prop is */` above each prop? When we render this in a PropTable-component in the storybook and in eds.equinor.com it adds the comment in the documentation

libraries/core-react/src/Chip/Chip.tsx
```diff
@@ -159,14 +170,14 @@ export const Chip = forwardRef(function EdsChips(
...
-      // Delete takes presidens, else click action is activated
+      // Delete takes precedence, else click action is activated
       if (clickable && !deletable) {
         handleClick(event)
       }
     }
   }
 
-  const resizedChildren = React.Children.map(children, (child) => {
+  const resizedChildren = React.Children.map(children, (child: React.DetailedReactHTMLElement<any, any>) => {
```
wenche, line 180: This <any> gives quite a few warnings in eslint :)

wenche (commented):
Thanks for contributing! 👏 

Some minor comments.

It doesn't look like prettier and eslint is up and running in your IDE? We don't actually use` ; `for instance :)

libraries/core-react/src/Chip/Chip.tsx
joelmheim: Yes, I see that. Have fixed it so the properties now are assigned to attrs.

libraries/core-react/src/Chip/Chip.tsx
joelmheim: Fixed this by changing to type React.ReactElement

libraries/core-react/src/Chip/Chip.tsx
joelmheim: Have added comments from the old PropTypes for the ones that we still specify. Some of them have been removed as they are standard Div properties.

joelmheim:
Also added some specific typing to the Chip.tests.tsx file. This could be of interest to other implementations, as it involves casting the target property of Event (or MouseEvent) to HTMLElement in order to get to the id property, or parentElement property.

wenche (approved):
Great 👍

libraries/core-react/src/Chip/Chip.tsx
wenche: 👍

PR #691, opened 2020-10-19
Convert component tokens to typescript

mimarz:
Suggestion on simplifying component tokens usage. 

* Json import and "parsing" is done in tokens package
* Exposed on `tokens` property under new property `components`
Example 

```jsx

 const  {
    components: {
        table,
        button.
  }
} = tokens
```

* All tests are green expect for Accordian which is dependant in Tabs which is still written in js...

Update:
* Changed all component token files from `json` to `ts`. This will lead up to better tokens later on when we apply types to component token files.


vnys (approved):
Smashing pumpkins 🎃

PR #700, opened 2020-10-19
Breadcrumbs Types

### LINKED ISSUE #553: Refactor Navigation Breadcrumbs to Typescript
Create typings for [Navigation Breadcrumbs](https://eds.equinor.com/components/navigation-breadcrumbs/)

pomfrida:
Resolves #553


libraries/core-react/src/Breadcrumbs/Breadcrumb.tsx
```diff
@@ -1,12 +1,12 @@
-// @ts-nocheck
 import React, { forwardRef } from 'react'
-import PropTypes from 'prop-types'
 import styled, { css } from 'styled-components'
 import { Typography } from '../Typography'
 import { Tooltip } from '../Tooltip'
 import { breadcrumbs as tokens } from './Breadcrumbs.tokens'
 
-const StyleTypography = styled(Typography)`
+type StyledProps = Pick<Props, 'maxWidth'>
```
vnys, line 7: `Pick` is one of those things I didn’t know about, that we should probably use in more components…

libraries/core-react/src/Breadcrumbs/Breadcrumb.tsx
```diff
@@ -27,13 +27,25 @@ const StyleTypography = styled(Typography)`
   ${({ maxWidth }) => css({ maxWidth })}
 `
 
-export const Breadcrumb = forwardRef(function Breadcrumb(
-  { className, children, maxWidth, ...rest },
+type Props = {
+  /*
```
vnys, line 31: why a blank line here?

libraries/core-react/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -38,13 +36,20 @@ const Collapsed = styled(Typography)`
...
 
-export const Breadcrumbs = forwardRef(function Breadcrumbs(
-  { className, children, collapse, ...rest },
+type Props = {
+  /*
+   * Collapses the list of breadcrumbs so that only the first
+   * and last breadcrumb will be shown, with an ellipsis in between.
+   */
+  collapse?: boolean
+} & JSX.IntrinsicElements['nav']
```
vnys, line 45: This is also something we do differently from component to component… `React.HTMLAttributes<HTMLNavElement> `vs. `JSX.IntrinsicElements['nav']`

vnys (approved):
LGTM 👍 Some extra lines in comments that could be removed, but nothing serious.

libraries/core-react/src/Breadcrumbs/Breadcrumbs.tsx
pomfrida: @mimarz and I discussed this yesterday, if I remember correctly the difference was that `JSX.IntrinsicElements['nav']` included more props?

libraries/core-react/src/Breadcrumbs/Breadcrumb.tsx
mimarz: Recommend having a look at the Typescript [utility types](https://www.typescriptlang.org/docs/handbook/utility-types.html)  :)

libraries/core-react/src/Breadcrumbs/Breadcrumbs.tsx
mimarz: @vnys Funfact, [nav](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav) only uses the `HTMLElement` dom interface

vnys: > @vnys Funfact, [nav](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav) only uses the `HTMLElement` dom interface

Noted 😉

libraries/core-react/src/Breadcrumbs/Breadcrumb.tsx
pomfrida: It's pretty neat! 😄  @mimarz also taught me `Omit<Type, Keys>` from [Typescript Utility Types](https://www.typescriptlang.org/docs/handbook/utility-types.html) where you can remove `Keys` from `Type` and re-use it

PR #701, opened 2020-10-19
Omit tests for Button and Accordion

vnys:
Tests fail for Button and Accordion, which makes it hard for external contributors to know if the tests fail because of something in _their_ code. In this pr the test-files have been renamed `*.tst.tsx` so they will be skipped.


mimarz:
This has been fixed now by #702 without omiting said tests :)

PR #705, opened 2020-10-19
Fix Chip tests

joelmheim:
Discovered problems with Chip tests while refactoring other component. Not sure what is wrong, but changing to importing from ./Chip instead of ../ fixed it (iteratively for all imports...).


vnys:
Thanks, I wasn’t aware that this was a problem, it should import from the **index.ts** file when you just use `..` to go up one level. I think we do that in other components too 🤔 I’ll look into that and see if something has changed… Can you change the branch you want to merge this into from `develop` to `typescript`?

joelmheim:
Have updated the branch. Sorry about that, was a bit quick. See you are getting some extra commits in the PR, but the changed files are correct. You could fix that by squashing the commits on merge.

mimarz (approved):
LGTM 👍

PR #712, opened 2020-10-20
Slider Typescript

### LINKED ISSUE #565: Refactor Slider to Typescript
Create typings for [Slider](https://eds.equinor.com/components/slider/)

pomfrida:
Resolves #565


libraries/core-react/src/Slider/Slider.tokens.ts
```diff
@@ -23,7 +23,71 @@ const {
   typography: { paragraph },
 } = tokens
 
-export const slider = {
+type Slider = {
```
wenche, line 26: Add on the list for normalizing of tokens

libraries/core-react/src/Slider/Slider.tsx
```diff
@@ -144,8 +149,40 @@ const SrOnlyLabel = styled.label`
...
+  /** Id for the elements that labels this slider */
+  ariaLabelledby: string
+  /** Components value, range of numbers */
+  value: SliderValueType
+  /** Function to be called when value change */
+  onChange?: (
+    event: MouseEvent | KeyboardEvent,
+    newValue: SliderValueType,
+  ) => void
+  /* Function to be called when value is committed by mouseup event */
```
wenche, line 164: It seems like only two ** works for enabling the `description` feature in props tables. Should probably be fixed in a lot of components...

wenche (approved):
A pretty complicated component 🙈 

I can see you changed the logic slightly for sliderValue, but I guess it still works with both range and slider 😓 

Looks good! 👍

PR #723, opened 2020-10-21
Use component src in the Storybook

### LINKED ISSUE #721: Docgen info not generated for typescript components
**Describe the bug**

1. Rewrite a component in typescript
2. Run build:for-storybook

**Expected behavior**

The bundle should have a docgen object for each component

vnys:
part of solving #721 

I’ve disabled the `displayName` in most of the components, but _some_ components explisitly require it according to Eslint – I have not looked into why at this point. I suggest merging this in, and then as a next step start debugging why some components are missing props. 

There’s a new way of documenting arguments with controls similar to addon-knobs, see the <Divider /> for an example of how this works, or read the section on [args in the docs](https://storybook.js.org/docs/react/writing-stories/args). 

To write a story in typescript, just change the suffix to `tsx`. Again see the Divider story for an example. 

Note that dockerizing the Storybook does not work at the moment and will have to be fixed before merging the typescript branch into develop


apps/storybook-react/.storybook/main.js
```diff
@@ -1,5 +1,11 @@
 module.exports = {
-  stories: ['../docs/**/*.story.mdx', '../stories/**/*.stor(y|ies).(jsx|mdx)'],
+  // typescript: {
+  //   reactDocgen: 'react-docgen',
+  // },
```
vnys, line 4: Leaving this as a reminder. Storybook will move to `react-docgen` in v7.0

apps/storybook-react/.storybook/main.js
```diff
@@ -1,5 +1,11 @@
...
+  stories: [
+    '../docs/**/*.story.mdx',
+    '../stories/**/*.stor(y|ies).(jsx|mdx|tsx)',
```
vnys, line 7: Adding `tsx`, assuming we won’t need `ts` too.

libraries/core-react/package.json
```diff
@@ -81,6 +84,7 @@
     "@equinor/eds-tokens": "workspace:*",
     "@testing-library/react-hooks": "^3.3.0",
     "@types/lodash": "^4.14.162",
+    "babel-plugin-react-docgen-typescript": "^1.5.1",
```
vnys, line 87: this should be removed

libraries/core-react/src/Popover/PopoverAnchor.tsx
```diff
@@ -12,4 +12,4 @@ export const PopoverAnchor = forwardRef<HTMLDivElement, Props>(
   },
 )
 
-PopoverAnchor.displayName = 'eds-popover-anchor'
+// PopoverAnchor.displayName = 'eds-popover-anchor'
```
pomfrida, line 15: Remove comment, this `displayName` is needed in `Popover.tsx`

pomfrida (changes requested):
See comments

libraries/core-react/src/Divider/Divider.tsx
```diff
@@ -17,25 +17,25 @@ const StyledDivider = styled.hr<StyleProps>`
   height: ${(props) => props.dividerHeight};
 `
 
-type Props = {
-  /** Color variants
- @default medium */
+export type DividerProps = {
```
mimarz, line 20: Is there a specific reason you renamed from `Props` to `DividerProps`? We have been using `Props` naming on all components for the "main" component properties.

mimarz (changes requested):
Suggest updating your branch with changes from #729 to avoid errors with removal of `displayName` and simplification.

Can we not use `Eds` prefix for our components also? Feels wrong to just use `Button` or `Divider` as it might clash with other stuff, or harder to debug.

libraries/core-react/src/Divider/Divider.tsx
```diff
@@ -17,25 +17,25 @@ const StyledDivider = styled.hr<StyleProps>`
...
-  const styleProps = {
-    backgroundColor: tokens.color[color],
-    marginTop: tokens[variant].spacings.top,
-    marginBottom: tokens[variant].spacings.bottom,
-    dividerHeight: tokens.height,
-  }
-  return <StyledDivider {...styleProps} className={className} ref={ref} />
-})
+export const Divider = forwardRef<HTMLHRElement, DividerProps>(
+  ({ color = 'medium', variant = 'medium', className = '' }, ref) => {
```
mimarz, line 30: This is an anonymous function, maybe thats why you eslint complains when you remove manual definition of `displayName`?

libraries/core-react/src/Popover/PopoverAnchor.tsx
vnys: The generated displayName will be `PopoverAnchor` so I changed the checking to `popoverChildren.type.displayName === 'PopoverAnchor'` in `Popover.tsx`.

apps/storybook-react/stories/Divider.stories.tsx
```diff
@@ -0,0 +1,48 @@
+import React, { FunctionComponent } from 'react'
```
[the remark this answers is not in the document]
vnys: Fixed

libraries/core-react/src/Divider/Divider.tsx
vnys: We have to export the types to use them in the stories. I forgot that yesterday, so it wasn’t that clear, but take a look at [the story](https://github.com/equinor/design-system/blob/020ce6cacb4366333843d787a1256ef352fc2526/apps/storybook-react/stories/Divider.stories.tsx#L32) where the props are being used, and also index.ts where [the props are exported](https://github.com/equinor/design-system/blob/020ce6cacb4366333843d787a1256ef352fc2526/libraries/core-react/src/index.ts#L5). If we have several components in the same story it’s better to have individual names. Exporting the props like that is removed from the bundle, but something we need to do when we use the typescript files directly in the Storybook.

libraries/core-react/src/Divider/Divider.tsx
vnys: Could be, I’ll look into that

vnys:
> Can we not use `Eds` prefix for our components also? Feels wrong to just use `Button` or `Divider` as it might clash with other stuff, or harder to debug.

That’s a bigger debate which I don’t think should hold back this pr. But I have deliberately commented out instead of removing the `displayNames` for a reason, in case we find other ways to fix this.

vnys:
> Suggest updating your branch with changes from #729 to avoid errors with removal of `displayName` and simplification.

Will do

libraries/core-react/src/Divider/Divider.tsx
vnys: It is indeed because it’s an arrow function. Normally javascript infers the function name from the variable that holds the arrow function, but in this case that does not work. So then we either have to refactor the components that use arrow functions, or just accept that we have to explicitly declare the `displayName` in these cases.

pomfrida (approved):
LGTM 👍

wenche (approved):
Looks good as a starting point for the rest of the work 👍

libraries/core-react/src/Divider/Divider.tsx
mimarz: Should be no problem to change it from arrow function to named function in most cases tho. Suggest we conform on on way of doing this and change our components using arrow functions to named functions then?

libraries/core-react/src/Divider/Divider.tsx
mimarz: So either we call the explicitly, which means we need to change all existing components `Props` or we can just rename it when importing in the stories. Which do we favour? 🤔

libraries/core-react/src/Divider/Divider.tsx
vnys: Yes, I agree. I made an issue (#732)

mimarz (approved):
Looks good, but there are some actions points we need to figure out for code uniformity from this 😅  "DividerProps" is already diverging from all our other components

* What do we call our component "props" type?
* How do we name our functions for `displayName`? Prefix or no prefix etc.
* Change arrow function components to named functions?

libraries/core-react/src/Divider/Divider.tsx
wenche: I can live with both approaches 🤷‍♀️  Initially I thought that it will be OK to rename imports as most of the components (?) only have one single component, but when I browsed the mighty Internet to see what other people do they often write <Component>Props, including the Storybook docs. So it might be a defacto standard?

vnys: Yes, I’ve seen that too…

PR #729, opened 2020-10-22
Children type checking

### LINKED ISSUE #726: Change logic for checking children types
We are currently using the `displayName` property for checking if children are of an specific type, for handling context based logic. 

This approach has been know to be fragile as its dependant on a string, often written by hand. Also it feels weird 🤪

@nicholasdalhaug Showed us a good way to improve this logic in [Banner](https://github.com/equinor/design-system/blob/72ff0768d01fe5c6bf8daee676355e740b05b9ac/libraries/core-react/src/Banner/Banner.tsx#L36)

Identified components which needs this logic changed:
* AccordianHeader
* Popover
* Menulist

Bonus: 
This will free us from the need to set `displayName` and let it be generated by functions names instead, as mentioned by @vnys in #721

mimarz:
resolves #726

* Removed usages of checking using `displayName === "some-components-displayName`
* Simplified `Popover` children splitting from its anchor element

### DIFF
--- a/libraries/core-react/src/Popover/Popover.tsx
+++ b/libraries/core-react/src/Popover/Popover.tsx
@@ -1,6 +1,13 @@
-import React, { forwardRef, useRef, HTMLAttributes, ReactNode } from 'react'
+import React, {
+  forwardRef,
+  useRef,
+  HTMLAttributes,
+  ReactNode,
+  isValidElement,
+} from 'react'
 import styled from 'styled-components'
 import { PopoverItem } from './PopoverItem'
+import { PopoverAnchor } from './PopoverAnchor'
 
 const Container = styled.div`
   position: relative;
@@ -13,9 +20,10 @@ const Anchor = styled.div`
     outline: none;
   }
 `
-type PopoverChild = {
-  type?: { displayName?: string }
-} & ReactNode
+type PopoverSplit = {
+  anchorElement: ReactNode
+  childArray: ReactNode[]
+}
 
 export type Props = {
   /* Popover placement relative to anchor */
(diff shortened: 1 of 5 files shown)


libraries/core-react/src/Popover/Popover.tsx
```diff
@@ -52,30 +60,22 @@ export const Popover = forwardRef<HTMLDivElement, Props>(function Popover(
...
-      if (child.type && child.type.displayName === 'eds-popover-anchor') {
-        anchorElement = child
-      } else {
-        // Add the remaining children to a new array to display inside <PopoverItem/>
-        childArray.push(child)
+
+  const { anchorElement, childArray } = React.Children.toArray(children).reduce(
+    (acc: PopoverSplit, child): PopoverSplit => {
+      if (isValidElement(child) && child.type === PopoverAnchor) {
+        return {
```
mimarz, line 67: using spread here resulted in empty `childArray` so reassignment like this had to be done instead... 🤷‍♂️

Does not work
```jsx
       return {
          anchorElement: child,
         ...acc,
        }
```

Works 🤯
```jsx
       return {
          anchorElement: child,
          childArray: acc.childArray,
        }
```

libraries/core-react/src/Popover/Popover.tsx
```diff
@@ -52,30 +60,22 @@ export const Popover = forwardRef<HTMLDivElement, Props>(function Popover(
...
+          ...acc,
+          anchorElement: child,
+        }
```
pomfrida, line 70: Satisfying ❤️

pomfrida (approved):
LGTM 👍

wenche (approved):
LGTM 👍

PR #703, opened 2020-10-19
fix: Accordion can't handle null value as child

SvSven:
Adds a guard clause to the rendering of AccordionItems to handle `null`, `undefined`, `false` values passed as Accordion children

Resolves #688

PR #731, opened 2020-10-23
Bugfix in develop moved to typescript branch

### LINKED ISSUE #709: Cherry-pick bug-fixes from develop into typescript
When we fix bugs in the `develop` branch we need to make sure we get those bug fixes into the `typescript` branch as well. It says “cherry-pick” in the title, but we will have to do this manually in cases where Git thinks a `js-file` is replaced with a `ts-file` instead of being renamed.

pomfrida:
added fix: Accordion can't handle null value as child #703 to typescript branch
resolves #709


mimarz (approved):
LGTM 👍

PR #734, opened 2020-10-23
Clean up spacings in CardMedia

### LINKED ISSUE #717: Clean up use of spacings in Card/CardMedia
` tokens.spacings.left` is passed and used for left, right and top spacings in `Card.CardMedia`, aswell as being passed as a prop to the "styled" component, which it doesn't need to...

This needs to be cleaned up to avoid bugs and errors if spacings are changed in tokens/figma in the future.

* Use the correct top, left, right, bottom spacings tokens in css.
* Remove spacing as prop for "styled" component

nicholasdalhaug:
Resolves #717


pomfrida (approved):
LGTM 👍  Thanks for contributing!

Issue #735, opened 2020-10-23
As a developer I need the stories to be migrated to typescript for the prop tables in Storybook to work


wenche:
Add a tick when you _start_ on a component so it's easier to see what other people are doing
- [x] Accordion
- [x] Avatar
- [x] Banner
- [x] Breadcrumbs
- [x] Button
- [x] Card
- [x] Chip
- [x] Dialog
- [x] Divider
- [x] Icon
- [x] List
- [x] Pagination
- [x] Popover
- [x] Progress
- [x] Scrim
- [x] Search
- [x] Selection controls
- [x] SideSheet
-  [x] Slider
- [x] Table
- [x] TOC
- [x] Textfield
- [x] Tooltip
- [x] Top bar
- [x] Typography
- [x] Tabs
- [x] Snackbar
- [x] Menu

mimarz:
I'll take Typography

pomfrida:
Accordion ✋

PR #738, opened 2020-10-23
[WIP] Typescript Story: Accordion

pomfrida:
Aa a part of #735


wenche (commented):
LGTM 👍

libraries/core-react/src/Accordion/AccordionHeader.tsx
```diff
@@ -72,7 +72,7 @@ const StyledIcon = styled(
   chevronPosition === 'left' ? { marginRight: '32px' } : { marginLeft: '16px' },
 )
 
-type AccordionHeaderTitleProps = Props & AccordionProps
+type AccordionHeaderTitleProps = AccordionHeaderProps & AccordionProps
```
mimarz, line 75: This can be simplified as this component does not use all these props (my bad). The following should be enough:
`type AccordionHeaderTitleProps = Pick<
  AccordionHeaderProps,
  'disabled' | 'isExpanded'
>
`

mimarz (changes requested):
Its hard to tell which properties are used for this and sub-components as a lot of props are used internally. Suggest enabling `stripInternals` here and label `@internal`.

Think we'll need @vnys to pitch in on which props are supposed to be set externally also..

pomfrida:
Still **WIP**, unsure if to keep `AccordionHeader`,
  `AccordionHeaderTitle`,
  `AccordionPanel`, as subcomponents to show in PropTable in Storybook or not 🤔  None of them have external props but have a ton of internal

PR #739, opened 2020-10-23
Banner story with typescript

wenche:
Part of #735


wenche:
Stories with multiple components are a [separate thing](https://storybook.js.org/docs/react/workflows/stories-for-multiple-components) with Storybook. The concept of control doesn't work for children.

wenche:
Also, I think the Banner is a candidate for some alignment.. It has children as explicit props atm

pomfrida (approved):
LGTM 👍

libraries/core-react/src/Banner/index.ts
```diff
@@ -1,18 +1,18 @@
...
 
-const Banner = BaseBanner as BannerProps
+const Banner = BaseBanner as AllBannerProps
 
 Banner.BannerIcon = BannerIcon
 Banner.BannerMessage = BannerMessage
 Banner.BannerActions = BannerActions
 
-export { Banner }
+export { Banner, BannerProps }
```
vnys, line 18: Shouldn’t the `AllBannerProps` type be exported here? But as `BannerProps`? 🤔 So maybe import `BannerProps as BaseBannerProps`, and then rename `AllBannerProps` to `BannerProps`? Or am I wrong?

vnys (changes requested):
Should have been a “maybe request changes" option here

libraries/core-react/src/Banner/index.ts
vnys: Hard to know with these compound components… But the components has children, and those children have to be typed, and then I assume that compound type is the type for the component. This is a bit easier to grasp when this is done in the actual component and not when re-exporting. 

```tsx
type ComponentProps = {
  /** Some other colour */
  color?: string
  /** Some foo */
  foo?: string
  /** @ignore */
  children: React.ReactNode
}

type CompoundComponent = React.ForwardRefExoticComponent<ComponentProps> & {
  OtherComponent: typeof OtherComponent
}

/**
 * This is the Component description
 */

const Component = forwardRef<HTMLDivElement, ComponentProps>(function Component(
  { children, color = 'violet' },
  ref,
) {
  return (
    <div style={{ color }} ref={ref}>
      {children}
    </div>
  )
}) as CompoundComponent

Component.OtherComponent = OtherComponent
```

…except the props disappeared from the Storybook when doing it that way…

wenche: The AllBannerProps was just a creative rename since it previously was named BannerProps 😅

wenche: But yes, we need to figure out what is the best way of reexporting both types and props with compound components

wenche:
Tried to export with * in the outer `index.ts `file to get rid of annoying warning. I had to combine it with a separate type export in Banner `index.ts` to make it work

pomfrida (approved):
LGTM 👍

vnys:
> Tried to export with * in the outer `index.ts `file to get rid of annoying warning. I had to combine it with a separate type export in Banner `index.ts` to make it work

That’s right, the `*` will just re-export whatever is already exported.

libraries/core-react/src/Banner/index.ts
```diff
@@ -1,18 +1,19 @@
...
 
-type BannerProps = typeof BaseBanner & {
+type AllBannerProps = typeof BaseBanner & {
   BannerIcon: typeof BannerIcon
   BannerMessage: typeof BannerMessage
   BannerActions: typeof BannerActions
 }
 
-const Banner = BaseBanner as BannerProps
+const Banner = BaseBanner as AllBannerProps
```
vnys, line 12:
```suggestion
const Banner = BaseBanner as BannerProps
```

libraries/core-react/src/Banner/index.ts
```diff
@@ -1,18 +1,19 @@
-import { Banner as BaseBanner } from './Banner'
+import { Banner as BaseBanner, BannerProps as Props } from './Banner'
 import { BannerIcon } from './BannerIcon'
 import { BannerMessage } from './BannerMessage'
 import { BannerActions } from './BannerActions'
 
-type BannerProps = typeof BaseBanner & {
+type AllBannerProps = typeof BaseBanner & {
```
vnys, line 6:
```suggestion
type BannerProps = typeof BaseBanner & {
```

libraries/core-react/src/Banner/index.ts
```diff
@@ -1,18 +1,19 @@
...
 
-const Banner = BaseBanner as BannerProps
+const Banner = BaseBanner as AllBannerProps
 
 Banner.BannerIcon = BannerIcon
 Banner.BannerMessage = BannerMessage
 Banner.BannerActions = BannerActions
 
 export { Banner }
+export type BannerProps = Props
```
vnys, line 19:
```suggestion
export type BannerProps
```

vnys (changes requested):
I think we should re-export the compound type, right?

wenche:
@vnys Not sure if I understood you correctly, but like this?

vnys:
> @vnys Not sure if I understood you correctly, but like this?

I think you could have just accepted the suggestions in Github and they would have been committed

vnys:
…but yes, that’s the way I think it must be done 😊

vnys (approved):
LGTM 👍

wenche:
> > @vnys Not sure if I understood you correctly, but like this?
> 
> I think you could have just accepted the suggestions in Github and they would have been committed

I was thinking of that, but I did a couple of things more than you did :/ With the initial import at least

PR #742, opened 2020-10-23
Typescript Story: Avatar

pomfrida:
Part of #735


apps/storybook-react/stories/Avatar.stories.tsx
```diff
@@ -20,10 +20,13 @@ const Wrapper = styled.div`
 export default {
   title: 'Components/Avatar',
   component: Avatar,
-  decorators: [withKnobs],
-}
+} as Meta
 
-export const Examples = () => (
+export const Default: Story<AvatarProps> = (args) => (
+  <Avatar src={image} {...args}></Avatar>
```
vnys, line 26: Shouldn’t this be self-closing?

```suggestion
  <Avatar src={image} {...args} />
```

vnys (commented):
I think the `<Avatar />` component should be self closing, but apart from that it LGTM 👍

apps/storybook-react/stories/Avatar.stories.tsx
pomfrida: Yes! I'll fix!

wenche (approved):
LGTM 👍

pomfrida:
Changed to **WIP** again due to this module 

```
declare module '*.jpg' {
  const value: string
  export = value
}
```

Needed for the image import in `Avatar.stories.tsx`. This module is currently only reachable from the story when the `custom.d.ts` file is open locally, that's the only way the import error is cleared.

I've tried adding `compilerOptions: typeRoots` in `.storybook/main.js`,` tsconfig.base.json` (root) and in core-react `tsconfig` without any luck. 

Check out Stackoverflow link about [TS image import](https://stackoverflow.com/questions/51100401/typescript-image-import)

mimarz:
> Changed to **WIP** again due to this module
> 
> ```
> declare module '*.jpg' {
>   const value: string
>   export = value
> }
> ```
> 
> Needed for the image import in `Avatar.stories.tsx`. This module is currently only reachable from the story when the `custom.d.ts` file is open locally, that's the only way the import error is cleared.
> 
> I've tried adding `compilerOptions: typeRoots` in `.storybook/main.js`,` tsconfig.base.json` (root) and in core-react `tsconfig` without any luck.
> 
> Check out Stackoverflow link about [TS image import](https://stackoverflow.com/questions/51100401/typescript-image-import)

Suggest just screwing it and use this link for the 😺 : https://i.imgur.com/UM3mrju.jpg
We have better things to work on 🙈

pomfrida:
Did the quick fix @mimarz suggested to avoid TS error on image import.

PR #744, opened 2020-10-23
🎨 Improve story for Divider

vnys:
- The Divider component now uses a named function instead of an arrow function
- The component and it’s type is re-exported as `*`, which gets rid of the error message that type cannot be found

I’ve added some Eslint exceptions to the story, but these issues are something we have to discuss

Part of #735

### DIFF
--- a/apps/storybook-react/stories/Divider.stories.tsx
+++ b/apps/storybook-react/stories/Divider.stories.tsx
@@ -1,5 +1,9 @@
-import React from 'react'
-import styled from 'styled-components'
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
+/* eslint-disable @typescript-eslint/no-unsafe-call */
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+import React, { FunctionComponent, HTMLProps } from 'react'
+import styled, { StyledFunction, ThemedStyledFunction } from 'styled-components'
+
 import { Divider, DividerProps } from '@equinor/eds-core-react'
 import { Story, Meta } from '@storybook/react/types-6-0'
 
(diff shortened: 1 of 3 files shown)


apps/storybook-react/stories/Divider.stories.tsx
```diff
@@ -1,3 +1,6 @@
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
+/* eslint-disable @typescript-eslint/no-unsafe-call */
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
```
vnys, line 3: This seems a bit silly. How do we deal with styled components in the stories? Should we use the same eslint config for the stories that we use in Core React?

libraries/core-react/src/Divider/index.ts
```diff
@@ -1 +1 @@
-export { Divider, DividerProps } from './Divider'
+export * from './Divider'
```
vnys, line 1: By re-exporting `*` we don’t get error messages about the types not being exported.

apps/storybook-react/stories/Divider.stories.tsx
mimarz: I vote yes 👍

mimarz (approved):
LGTM 👍

pomfrida (approved):
LGTM 👍

libraries/core-react/src/Divider/index.ts
wenche: Just did kinda the same for Banner

wenche (approved):
LGTM 👍

PR #745, opened 2020-10-26
Typescript for button story

wenche:
Part of #735


wenche:
I don't think the `as` prop is easy to understand, and we should indeed do something about how we do this.
Skjermbilde 2020-10-26 kl  14 56 15

wenche:
I'm struggling a bit with the `type` prop as well, it should have a default to 'button'? But the code is a bit strange here 😓 
Skjermbilde 2020-10-26 kl  16 19 53

wenche:
I solved the default type props by using @default in the comment 🤷‍♀️  But the code feels a bit fuzzy here.

mimarz (approved):
LGTM 👍

vnys:
> I solved the default type props by using @default in the comment 🤷‍♀️ But the code feels a bit fuzzy here.

Didn’t it work with `someProp = 'default value'`?

wenche:
> > I solved the default type props by using @default in the comment 🤷‍♀️ But the code feels a bit fuzzy here.
> 
> Didn’t it work with `someProp = 'default value'`?

Not the way it was solved, ref. "fuzzy code" 👍

PR #746, opened 2020-10-26
Typescript Story: Breadcrumbs

pomfrida:
Part of #735


libraries/core-react/src/Breadcrumbs/index.ts
```diff
@@ -1,4 +1,7 @@
-import { Breadcrumbs as BaseComponent } from './Breadcrumbs'
+import {
+  Breadcrumbs as BaseComponent,
+  BreadcrumbsProps as Props,
+} from './Breadcrumbs'
 import { Breadcrumb } from './Breadcrumb'
 
 type BreadcrumbsTypes = typeof BaseComponent & {
```
wenche, line 7: In other component with sub component we have exported the merged types - in your case `BreadcrumbsTypes` as type. See e.g. Banner 👍

wenche (changes requested):

wenche (approved):
LGTM 👍

Issue #297, opened 2020-04-15
Update Table component


mimarz:
**Description**
Update the `Table`, with missing states and latest changes from Figma

**Documentation**

- [The storefront](https://eds.equinor.com/components/table/)
- [Figma](https://www.figma.com/file/0bGXR2sCwMVSDNyyzu5BXrO5/UI%E2%80%94User-Interface?node-id=1%3A25)
- [WAI-ARIA Authoring Practices 1.1](https://www.w3.org/TR/wai-aria-practices-1.1/#table)

**Sub tasks**

- [x] Write tests
- [x] Create set of tokens (from base tokens)

**Notes**
#148

vnys:
- Add separate types for `th` and `td`, possibly by using function overload (specifically scope)
- Don’t use `th-tokens` for `thead`
- Rewrite sticky header example in story to use` position: fixed` on `th` (See [CSS Tricks article](https://css-tricks.com/position-sticky-and-table-headers/))

mimarz:
> * Don’t use `th-tokens` for `thead`

We use `th-tokens` for `thead` because there is no tokens `thead`. Figma does not have this concept. We need to use the `background` and `border` from `th-tokens` for when the table is stretched to fit width and `th` not.

mimarz:
blocked by #830

PR #747, opened 2020-10-26
♻️ Write table stories in typescript and fix props table

vnys:
part of #735


vnys:
The cell component now takes `scope` as an optional parameter, which is something we’ll burn in hell for 🔥, so I’ve added that to the almighty [fix the table component issue](https://github.com/equinor/design-system/issues/297#issuecomment-717852095), in addition to a couple of other things.

libraries/core-react/src/Table/Cell.tsx
```diff
@@ -42,19 +42,23 @@ const TableBase = styled.td`
...
-  as: 'td' | 'th'
+  as?: 'td' | 'th'
   /** Specifies which variant to use */
-  variant: 'text' | 'icon' | 'numeric' | 'input'
-} & HTMLAttributes<HTMLTableCellElement>
+  variant?: 'text' | 'icon' | 'numeric' | 'input'
+  /** Is the header cell scoped to column or row? */
+  scope?: 'row' | 'col'
+  /** @ignore */
+  children: JSX.Element[] | JSX.Element | string
```
mimarz, line 53: `children` should be of type `ReactNode`

mimarz (changes requested):
Just one type change, else it looks good for now... #297

libraries/core-react/src/Table/Cell.tsx
```diff
@@ -42,19 +42,23 @@ const TableBase = styled.td`
...
 `
 
-type Props = {
+type CellProps = {
   /** Specifies which td or th to use */
-  as: 'td' | 'th'
+  as?: 'td' | 'th'
   /** Specifies which variant to use */
-  variant: 'text' | 'icon' | 'numeric' | 'input'
-} & HTMLAttributes<HTMLTableCellElement>
```
mimarz, line 50: Just thought of one thing. How will removal of `HTMLAttributes<HTMLTableCellElement>` affect other `react` props as `style`, `className` or `forwardedAs`?

vnys: Good point, I’ll add those manually.

mimarz: What was the reason for removing `HTMLAttributes<HTMLTableCellElement>`? There are a plethora of [other attributes ](https://developer.mozilla.org/en-US/docs/Web/API/HTMLTableCellElement) we miss out on by not having this.

vnys: The reason was that the more I dug into this the more I realised the component needs a more thorough rewrite, so I focused on getting the story to work, since we’ll have to refactor the component anyway before we merge typescript into develop. It doesn’t make sense for a table cell to have the scope attribute for example.

vnys: image

This is what I get if I use `HTMLAttributes<HTMLTableCellElement>`. So ideally the types should be based on the value of the `variant` prop, which I started looking into using function overload, but decided to save for the [refactoring issue](https://github.com/equinor/design-system/issues/297#issuecomment-717852095).

mimarz: I did some testing. Since we have the option to have a cell as `th` or `td`, what we are talking about is a union type. So then something along the lines of this makes more sense:

```typescript
type Props = {
  /** Specifies which td or th to use */
  as: 'td' | 'th'
  /** Specifies which variant to use */
  variant: 'text' | 'icon' | 'numeric' | 'input'
} & (
  | TdHTMLAttributes<HTMLTableDataCellElement>
  | ThHTMLAttributes<HTMLTableHeaderCellElement>
)
```

vnys: That seems to do the trick 👍 But what’s the difference then between `HTMLAttributes<HTMLTableCellElement>` and `TdHTMLAttributes<HTMLTableDataCellElement>`? 🤔 I find this confusing…

wenche: As far as I have understood the latter is more specific, it extends from [HTMLAttributes and adds the td specific attributes in addition.](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts)

vnys: I see… and it has [all the same props as ThHTMLAttributes](ThHTMLAttributes), so then that will probably suffice.

mimarz: > I find this confusing…

Me to 🙈  

You can find out a lot by looking at the types. Looking at how [JSX.Intrinsict](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/5454930b2034878b7cf0d749a1272a5e37f41bfa/types/react/index.d.ts#L2985) can tell you a lot  

I think it was wrong of us to use `HTMLAttributes<>` on many of the components to begin with, as they don't actually don extend the specific attributes for the dom types we pass into them, the type is only applied to dom callback handlers. 

The more specific Attributes such as `TdHTMLAttributes` adds the correct attributes ontop of setting the correct dom callbacks.

mimarz (approved):
LGTM 👍

libraries/core-react/src/Table/Cell.tsx
vnys: I see, we should add something about that in the discussions thread then, I can do that later.

PR #750, opened 2020-10-26
🔧 Lint stories

vnys:
Extends the eslintconfig from root, but turns off the default exports error since storybook relies on default exports in the stories.


wenche (approved):
LGTM 👍

pomfrida (approved):
👍

PR #752, opened 2020-10-27
♻️ Rewrite Dialog stories in Typescript

vnys:
Part of #735


apps/storybook-react/stories/Dialog.stories.tsx
```diff
@@ -0,0 +1,178 @@
...
+    </Fragment>
+  ),
+}
+
+const ACTION_CHOICES = {
+  none: null,
+  buttons: (
+    <TempButtonWrapper>
+      <Button>OK</Button>
+      <Button variant="ghost">Cancel</Button>
```
vnys, line 91: Ok left, cancel right…

apps/storybook-react/stories/Dialog.stories.tsx
```diff
@@ -0,0 +1,178 @@
...
+}
+
+export const Default: Story<DialogProps> = () => (
+  <Dialog>
+    <Title>{TITLE_CHOICES['text']}</Title>
+    <CustomContent>{CUSTOM_CONTENT_CHOICES['description']}</CustomContent>
+    <Actions>
+      <TempButtonWrapper>
+        <Button>Cancel</Button>
+        <Button>OK</Button>
```
vnys, line 103: Cancel left, OK right… 🤔 I think this has gone unnoticed in the original story…

vnys:
Discovered something in the story that I removed since it’s different from what we do elsewhere. If we _don’t_ export the type explicitly, then we can still do `type DialogProps = typeof Dialog` in the story just like we do in the index.ts exports 🤷

apps/storybook-react/stories/Dialog.stories.tsx
vnys: image

vnys: So affirmative actions always on the left, destructive actions on the right.

mimarz: Is it not the other way around? 🙈

vnys: This is from Figma, but I don’t know _who_ made that in Figma. Maybe there’s something in the Storefront…

vnys: Can confirm that this is correct

pomfrida (approved):
LGTM 👍

libraries/core-react/src/Dialog/index.ts
```diff
@@ -3,16 +3,16 @@ import { Actions } from './Actions'
...
 
-const Dialog = BaseComponent as DialogTypes
+const Dialog = BaseComponent as DialogProps
 
 Dialog.Actions = Actions
 Dialog.Title = Title
 Dialog.CustomContent = CustomContent
 
-export { Dialog }
+export { Dialog, DialogProps }
```
mimarz, line 18: Didn't we agree to export the actual `DialogProps` here? @wenche @vnys

mimarz (changes requested):
Looks ok, but this it does not follow the same pattern we have for our other compound components types...

libraries/core-react/src/Dialog/index.ts
vnys: Yes we did

vnys:
> this it does not follow the same pattern we have for our other compound components types

That was not intentional, it should now

mimarz (approved):
LGTM 👍

PR #753, opened 2020-10-27
Typescript for list stories

wenche:
Part of #735


libraries/core-react/src/List/List.tsx
```diff
@@ -41,6 +41,6 @@ const List = forwardRef<HTMLUListElement | HTMLOListElement, Props>(
   },
 )
 
-// List.displayName = 'List'
+List.displayName = 'List'
```
mimarz, line 44: Is `displayName` needed again?

wenche: Hmm, no, I was just investigating another bug. 👍

mimarz (changes requested):
Setting `displayName` seems unnecessary?

mimarz (approved):
LGTM 👍

PR #759, opened 2020-10-27
🐛 Fix divider story

vnys:
Forgot an export, and the story was missing args for some reason

### DIFF
--- a/apps/storybook-react/stories/Divider.stories.tsx
+++ b/apps/storybook-react/stories/Divider.stories.tsx
@@ -10,18 +10,8 @@ export default {
   title: 'Components/Divider',
   component: Divider,
   argTypes: {
-    color: {
-      control: {
-        type: 'radio',
-        options: ['lighter', 'light', 'medium'],
-      },
-    },
-    variant: {
-      control: {
-        type: 'radio',
-        options: ['medium', 'small'],
-      },
-    },
+    color: { control: 'radio' },
+    variant: { control: 'radio' },
   },
 } as Meta
 
--- a/apps/storybook-react/stories/Divider.stories.tsx
+++ b/apps/storybook-react/stories/Divider.stories.tsx
@@ -9,6 +9,20 @@ import { Story, Meta } from '@storybook/react/types-6-0'
 export default {
   title: 'Components/Divider',
   component: Divider,
+  argTypes: {
+    color: {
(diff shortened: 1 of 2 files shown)


apps/storybook-react/stories/Divider.stories.tsx
```diff
@@ -9,6 +9,20 @@ import { Story, Meta } from '@storybook/react/types-6-0'
...
+        options: ['lighter', 'light', 'medium'],
+      },
+    },
+    variant: {
+      control: {
+        type: 'radio',
+        options: ['medium', 'small'],
+      },
+    },
+  },
```
mimarz, line 25: Should this not be derived from `DividerProps`?

mimarz: All other stories have this auto-generated from types

vnys: It is, but I want the controls to be of type radio button and not a drop down, and then you have to specify it. As I mentioned somewhere else, you only have to do that with the types you want to have specific controls for. So if I was fine with a dropdown for variant I would just omit that from the config and it would still be visible in the props table.

vnys: I had this originally, but it was removed in [this commit](https://github.com/equinor/design-system/commit/87bdcc776fa5c6908a7aa0ca27b0ad4ce07bbea4) for some reason.

mimarz: Fair enough. I think it is possible to just pick the values from the `DividerProps`, like so; `DividerProps['color']`. Maybe you'll need a `keyof` also. Then we can guard us for bugs if we introduce new variants or colors

vnys: Aha, I see what you mean. That would be better, I’ll see if it works.

vnys: Not possible to use types as values, but this is at least an improvement, and the result is the same: 

```
export default {
  title: 'Components/Divider',
  component: Divider,
  argTypes: {
    color: { control: 'radio' },
    variant: { control: 'radio' },
  },
} as Meta
```

mimarz (approved):
LGTM 👍

PR #760, opened 2020-10-28
Textfield typescript story

mimarz:
Props table does not work, and I cant figure out why.

Submitting as is because we need to get on with things and maybe we can figure this out later... 🤔


libraries/core-react/src/TextField/Icon/Icon.tsx
```diff
@@ -38,7 +38,7 @@ const StyledIcon = styled.div`
   ${Variation}
 `
 
-type Props = {
+type TextfieldIconProps = {
```
wenche, line 41: I suppose we don't need to export internal types right?

libraries/core-react/src/TextField/Input/Input.tsx
```diff
@@ -92,63 +92,66 @@ type Props = {
...
-      ...spacings,
-      input: {
-        ...spacings.input,
-        right: '32px',
-      },
+} & InputHTMLAttributes<HTMLInputElement>
+
+const TextFieldInput = React.forwardRef<HTMLInputElement, TextfieldInputProps>(
+  function TextFieldInput(props, ref) {
+    const {
```
wenche, line 99: I think we found out that this way of doing things doesn't work properly with argsTable because it can't figure out default props

wenche (changes requested):
I think we should change it to not using destructing at a separate line. Otherwise good!

libraries/core-react/src/TextField/Input/Input.tsx
mimarz: Which way is that?  Prefixed with `TextField`?

libraries/core-react/src/TextField/Icon/Icon.tsx
mimarz: I suppose not as they are not used externally?

wenche: Hopefully 😓

libraries/core-react/src/TextField/Input/Input.tsx
wenche: Destructing inside the function parentheses :)

wenche: Below [ update august header](https://github.com/equinor/design-system/discussions/521)

mimarz: Changed it but it didn't seem to work...  but I can change it alas.

wenche (approved):
LGTM 👍

PR #761, opened 2020-10-28
Typescript for Progress stories

wenche:
Part of #735


wenche:
I don't think it's possible to avoid the uncontrolled vs controlled component react error message in dev console when we use the control option for value 🤷‍♀️ 
Skjermbilde 2020-10-28 kl  10 25 53

apps/storybook-react/stories/Progress Indicators/hooks/useProgress.tsx
```diff
@@ -0,0 +1,21 @@
+import { useState, useEffect } from 'react'
```
mimarz, line 1: This should maybe be `.ts`?

wenche: Good catch! Not used to writing .tsx 👍

mimarz (approved):
LGTM 👍

PR #795, opened 2020-10-28
Typescript for Chip story

wenche:
Part of #735


wenche:
The `Chip.tsx` file needs to have more types for eslint to be happy. We must take a look at this, and add types for event functions.

pomfrida (approved):
LGTM 👍

PR #797, opened 2020-10-29
♻️  Popover story to Typescript

pomfrida:
Part of #735


libraries/core-react/src/Popover/index.ts
```diff
@@ -1,21 +1,19 @@
-import { Popover as BaseComponent } from './Popover'
+import { Popover as BaseComponent, PopoverProps as Props } from './Popover'
 import { PopoverTitle } from './PopoverTitle'
 import { PopoverAnchor } from './PopoverAnchor'
 import { PopoverContent } from './PopoverContent'
-import { PopoverItem } from './PopoverItem'
 
 type PopoverTypes = typeof BaseComponent & {
```
wenche, line 6: Same comment as Breadcrumbs, we've been exporting the merged types (And hopefully that is correct 🙈 )

wenche (changes requested):
Looks good, except for the small issue with exporting type :)

wenche (approved):
LGTM 👍

Issue #449, opened 2020-07-07
Create documentation for Search


vnys:
Write documentation for [Search](https://eds.equinor.com/components/search/)

**Sub tasks**

- [ ] Write documentation
- [ ] Create examples
- [ ] Add props-table
- [ ] Add live preview

PR #799, opened 2020-10-29
Search typescript story

mimarz:


libraries/core-react/src/Search/Search.tsx
```diff
@@ -168,134 +174,139 @@ const InsideButton = styled.div<InsideButtonProps>`
...
-    const clearedValue = ''
-    setReactInputValue(input, clearedValue)
-    setState({ ...state, isActive: false })
-  }
-  const setIsActive = (newValue) =>
-    setState({ ...state, isActive: newValue !== '' })
+export type SearchProps = InputHTMLAttributes<HTMLInputElement>
+
+export const Search = React.forwardRef<HTMLInputElement, SearchProps>(
+  function EdsSearch(
```
pomfrida, line 186: Change:
```suggestion
  function Search(
```

pomfrida (changes requested):
Just the function name change from `EdsSearch` to `Search`, looks good otherwise 👍

apps/storybook-react/stories/Search.stories.tsx
```diff
@@ -33,19 +38,19 @@ const StyledSearch = styled(Search)`
 
 export default {
   title: 'Components/Search',
-  component: Search,
```
wenche, line 36: Shouldn't remove the component key?

wenche (changes requested):
Looks good, but didn't quite understand all the nitty gritty 🙈 

I thought the component = Component was needed for the argsTable to show up?

apps/storybook-react/stories/Search.stories.tsx
mimarz: We have no `argsTable` as all props are "native" input props. 
So I was left with: 
image

We could start adding value, defaultValue and other native props to table, but we would have to just replicate whats already defined in React, which didn't seem right. 

To remove the "empty" argsTable, you are supposed to remove the `component` key, according to storybook docs.

wenche: I see!

mimarz: I do see that [Material still documents "native" props](https://material-ui.com/api/text-field/). Which maybe we should also do for better clarity on how its supposed to be used. But maybe thats a discussion for when we start documentation work #449

wenche: Yes, I was thinking of the same with my selection controls, feels like some important stuff was missing in the props table because it was native

pomfrida (approved):
LGTM 👍

wenche (approved):
LGTM 👍

PR #802, opened 2020-10-30
♻️  Scrim story to Typescript

pomfrida:
Part of #735


libraries/core-react/src/Scrim/Scrim.tsx
```diff
@@ -29,10 +22,23 @@ const ScrimContent = styled.div`
...
 `
 
-export const Scrim = forwardRef<HTMLDivElement, Props>(function EdsScrim(
+export type ScrimProps = {
+  /** Whether scrim can be dismissed with esc key and outside click
+   */
+  isDismissable?: boolean
+  /** function to handle closing scrim */
+  onClose?: (event: MouseEvent | KeyboardEvent, open: boolean) => void
+} & HTMLAttributes<HTMLDivElement>
```
pomfrida, line 31: Note to self: Import `HTMLAttributes` from React instead of using `React.HTMLAttributes` directly. Somehow this f***s up the prop table. 🤯

wenche: It was the other way around last time 🤪

wenche (approved):
LGTM 👍

PR #803, opened 2020-10-30
♻️  Sidesheet story to Typescript

pomfrida:
Part of #735


pomfrida:
Since we now have controls in story, I think an example page for each variant is a bit overkill - so I removed a lot of code.

apps/storybook-react/stories/SideSheet.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export default {
+  title: 'Components/SideSheet',
+  component: SideSheet,
+  argTypes: {
+    variant: {
+      defaultValue: 'medium',
+      control: {
+        type: 'select',
+        options: ['small', 'medium', 'large', 'xlarge'],
```
wenche, line 22: Can you avoid duplicating this here? I _think_ @vnys  did something here for the, hmm can't remember exactly which component, table maybe?

wenche (changes requested):
Just a small comment on tight coupling

apps/storybook-react/stories/SideSheet.stories.tsx
pomfrida: Yup, this code was pretty much useless, just removed it 😄

wenche (approved):
LGTM! 👍

PR #805, opened 2020-10-30
Fix focus-visible & button focus

mimarz:
* Turns out `focus-visible` does not work being loaded via the `rollup-plugin-polyfill` anymore / with typescript?
* Fixed missing default tabIndex for `Button`


wenche (approved):
Strange... 🤷‍♀️  LGTM

PR #806, opened 2020-10-30
♻️  Slider story to Typescript

pomfrida:
Part of #735


libraries/core-react/src/Slider/Slider.tsx
```diff
@@ -227,27 +238,28 @@ export const Slider = forwardRef<HTMLDivElement, Props>(function EdsSlider(
...
   }
-  const handleKeyUp = (event) => {
+  const handleKeyUp = (event: KeyboardEvent) => {
     if (event.keyCode === 37 || event.keyCode === 39) {
       handleCommitedValue(event)
     }
   }
 
-  const handleCommitedValue = (event) => {
+  const handleCommitedValue = (event: MouseEvent | KeyboardEvent) => {
```
wenche, line 247: Why did this work previously? 🤔

wenche (approved):
Some comments, you had to make a lot of changes?

Looks good otherwise 🤷‍♀️

libraries/core-react/src/Slider/MinMax.tsx
```diff
@@ -23,12 +23,12 @@ const StyledMinMax = styled.span`
   }
 `
 
-type Props = {
+type MinMaxProps = {
   /** Children is required */
-  children: ReactNode
```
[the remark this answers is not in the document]
pomfrida: Actually I can change back to `ReactNode`, this was a change I did because of this error in `Slider.tsx`. But I fixed it by changing return type from `void` to `number | string` instead. 
Screenshot 2020-11-03 at 10 29 24

libraries/core-react/src/Slider/Slider.tsx
```diff
@@ -183,9 +190,9 @@ export type Props = {
   minMaxValues?: boolean
   /** Disabled */
   disabled?: boolean
-} & JSX.IntrinsicElements['div']
+} & Omit<HTMLAttributes<HTMLDivElement>, 'onChange'>
```
pomfrida, line 193: Got a mismatch between our `onChange` type and `HTMLDivElement` `onChange` type... So had to remove it from `HTMLDivElement` type import to clear TS errors

mimarz: Can we change the sliders `onChange` to match the native behaviour?

libraries/core-react/src/Slider/Slider.tsx
mimarz: Did you try `React.MouseEvent` and `React.KeyboardEvent`? The types you have her are the global ones...

pomfrida: They are importet from React @mimarz

libraries/core-react/src/Slider/Slider.tsx
pomfrida: @mimarz I've been playing a lot with this, check out this weird error when I update our `onChange` type to match native type. No matter how our `onChange` type looks like, we get the error `'(event: blablabla is not assignable to type '((event: ...` It looks like they get extra brackets somewhere!  
Screenshot 2020-11-03 at 15 02 19

apps/storybook-react/stories/Slider.stories.tsx
```diff
@@ -43,7 +57,7 @@ export const Examples = () => {
         <span id="range-slider-label">Range slider</span>
         <Slider
           value={value}
-          onChange={onChange}
+          onChange={() => changeHandler}
```
pomfrida, line 60: `onChange` expected a function, this fixes mentioned TS error

libraries/core-react/src/Slider/Slider.tsx
mimarz: No wonder it does not work with native `onChange` because it expects only one parameter, not two, which we have override/changed 🙈

image

pomfrida: @mimarz 🤯  I don't have that error in my branch

pomfrida: Btw all errors cleared now after latest commit, merge or do you wan't to review first @mimarz ?

mimarz: I'll have a look @pomfrida  :)

mimarz (approved):
LGTM 👍

PR #808, opened 2020-11-02
Typescript migration of snackbar stories

wenche:
Part of #735


wenche:
I had to add a `z-index: 10` to the snackbar because the default example would be positioned under e.g the action example otherwise.

vnys:
Why 10? Not a big fan of arbitrary numbers in css 🙄

wenche:
Did struggle a bit with the suggested approach of exporting the merged type in `index.ts` for a story with multiple components. If I do it like this
```
import { SnackbarAction } from './SnackbarAction'
import { Snackbar as BaseComponent, SnackbarProps } from './Snackbar'

type SnackbarType = typeof BaseComponent & {
  SnackbarAction: typeof SnackbarAction
}

const Snackbar = BaseComponent as SnackbarType

Snackbar.SnackbarAction = SnackbarAction

export { Snackbar }
export type { SnackbarProps }
```
I can use the component in a story like this
Skjermbilde 2020-11-02 kl  10 44 47

If I followed the suggested approach with
```
import { SnackbarAction } from './SnackbarAction'
import { Snackbar as BaseComponent } from './Snackbar'

type SnackbarProps = typeof BaseComponent & {
  SnackbarAction: typeof SnackbarAction
}

const Snackbar = BaseComponent as SnackbarProps

Snackbar.SnackbarAction = SnackbarAction

export { Snackbar, SnackbarProps }
```
then I'm stuck with
Skjermbilde 2020-11-02 kl  11 09 09

The on hover editor suggestions also differs (same order as above)
Skjermbilde 2020-11-02 kl  11 10 14
Skjermbilde 2020-11-02 kl  11 09 35

wenche:
> Why 10? Not a big fan of arbitrary numbers in css 🙄

Had to choose something. :) I don't like it when z-index is too hight. Do we have a default z-index number?

wenche:
Just like with the `Card` component I'm struggling with the [renaming of the first tab to `Story`](https://github.com/storybookjs/storybook/issues/12357) instead of `Snackbar` . I can't see why this works with e.g. the `Banner` or the `TopBar` component 🤷‍♀️  I've tried to write the Snackbar function itself in approx. 100 different ways, but no luck.

mimarz:
What if you just drop the `Button` and have open fixed to `true` for the controlled example? 

Having it open and close is a separate story I would say

wenche:
> What if you just drop the `Button` and have open fixed to `true` for the controlled example?
> 
> Having it open and close is a separate story I would say

I don't like the idea of that as it would collide with the other snackbar examples and it should be one of the things one could experiment with. 

But my main concern was more with how this component behaves in a real life use case?

apps/storybook-react/stories/Snackbar.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+export default {
+  title: 'Components/Snackbar',
+  component: Snackbar,
+  subcomponents: { SnackbarAction },
+} as Meta
+
+export const Default: Story<SnackbarProps> = (args) => {
+  const { open } = args
+  const [visible, setVisible] = useState(open)
```
pomfrida, line 15: You can remove `const { open } = args` here and let the user trigger only at button click, that way the user can play with the other `Snackbar` examples too without interference... But it's weird how we get errors on it because this method has worked before 😕

wenche: That is the my main issue - how do we know that we are not just masking errors by just removing code that doesn't work? And then I'll have to turn off the control for open as well 😬 

For the record, I'm doing the same thing with the Progress indicators - but because they is a single component with a single type it just works. But obviously not with the composed component approach

pomfrida: `Open` is right there! 
Screenshot 2020-11-03 at 08 34 45
🤔

pomfrida: I also did it on Popover with subcomponents

libraries/core-react/src/Snackbar/Snackbar.tsx
```diff
@@ -34,25 +35,25 @@ const StyledSnackbar = styled.div.attrs(() => ({
...
-  /* Callback fired when the snackbar is closed by auto hide duration timeout */
+  /** Callback fired when the snackbar is closed by auto hide duration timeout */
   onClose?: () => void
-  /* Media query from which the snackbar will be horizontal centered */
+  /** Media query from which the snackbar will be horizontal centered */
   leftAlignFrom?: string
 } & HTMLAttributes<HTMLDivElement>
 
-export const Snackbar = ({
+export const Snackbar: FC<SnackbarProps> = ({
```
pomfrida, line 49: @wenche Why `FC` here? I thought we should use `forwardRef` and named functions instead

wenche: See [commit message](https://github.com/equinor/design-system/pull/808/commits/0f002dd56cd81e0868f0a3e3c0bc2c8365c91a15). It didn't have forwardRef and adding it will be a breaking change.

pomfrida: Oh.. So we should add `FC` to those who didn't have `forwardRef`? What breaking change will adding `forwardRef` make? I'm pretty sure I've added forwardRef's to some of my components so I should probably go back to do it your way.. 😬

pomfrida (approved):
LGTM 👍

PR #812, opened 2020-11-03
Icon typescript story

mimarz:
* Had to override `SvgProps` `ref` with non-Legacy ref to match `forwardRef` ref.
* Was not able to get `select` control to work on table for name prop (for easier testing of switching icons)


pomfrida (approved):
LGTM 👍

PR #814, opened 2020-11-03
Migrate tabs stories to typescript

wenche:
Part of #735


wenche:
I'm having trouble with the panels example. I tried to add type `number` to the callback function to avoid `any` type. Gives me this
Skjermbilde 2020-11-03 kl  16 08 08
The props are actually like this
Skjermbilde 2020-11-03 kl  16 10 06

So where does the `e: React.FormEvent<HTLMDivElement>` comes from? Is it because of the naming `onChange`? 🤯

pomfrida:
> So where does the `e: React.FormEvent<HTLMDivElement>` comes from? Is it because of the naming `onChange`? 🤯

I have the same problem in #806 
When I was about to change the `children` type after your review this same error came up from nowhere (or Visual Studio was not catching them before for some reason). Use `Omit` to remove `onChange` from `HTMLDivElement` type import. Have no idea why they don't match though.

wenche:
> > So where does the `e: React.FormEvent<HTLMDivElement>` comes from? Is it because of the naming `onChange`? 🤯
> 
> I have the same problem in #806
> When I was about to change the `children` type after your review this same error came up from nowhere (or Visual Studio was not catching them before for some reason). Use `Omit` to remove `onChange` from `HTMLDivElement` type import. Have no idea why they don't match though.

I ❤️  you @pomfrida

wenche:
One last issue with `Tabs`. When I use the controls to play with the component, I'll get a wall of errors in the developer tools console.
Skjermbilde 2020-11-04 kl  08 27 06

I suppose this is because of the mismatch between
```
export type TabsProps = {
  /** The index of the active tab */
  activeTab?: number
  /** The callback function for selecting a tab */
  onChange?: (index: number) => void
  /** Sets the width of the tabs. Tabs can have a maximum width of 360px */
  variant?: Variants
} & Omit<HTMLAttributes<HTMLDivElement>, 'onChange'>
```

and 
```
const Tabs = forwardRef<HTMLDivElement, TabsProps>(function Tabs(
  { activeTab = 0, onChange, onBlur, onFocus, variant = 'minWidth', ...props },
  ref,
)
```
Why is not `onBlur` and `onFocus` in the type? Is this intentionally? @vnys

mimarz:
> One last issue with `Tabs`. When I use the controls to play with the component, I'll get a wall of errors in the developer tools console.
> Skjermbilde 2020-11-04 kl 08 27 06
> 
> I suppose this is because of the mismatch between
> 
> ```
> export type TabsProps = {
>   /** The index of the active tab */
>   activeTab?: number
>   /** The callback function for selecting a tab */
>   onChange?: (index: number) => void
>   /** Sets the width of the tabs. Tabs can have a maximum width of 360px */
>   variant?: Variants
> } & Omit<HTMLAttributes<HTMLDivElement>, 'onChange'>
> ```
> 
> and
> 
> ```
> const Tabs = forwardRef<HTMLDivElement, TabsProps>(function Tabs(
>   { activeTab = 0, onChange, onBlur, onFocus, variant = 'minWidth', ...props },
>   ref,
> )
> ```
> 
> Why is not `onBlur` and `onFocus` in the type? Is this intentionally? @vnys

Because they are defined in `React.HTMLAttributes<HTMLDivElement>`. Why this fails tho 🤷‍♂️

pomfrida:
> > Why is not `onBlur` and `onFocus` in the type? Is this intentionally? @vnys
> 
> Because they are defined in `React.HTMLAttributes<HTMLDivElement>`. Why this fails tho 🤷‍♂️

Could this be the same issue as in #806 where `onChange` expected a function?

wenche:
> Because they are defined in `React.HTMLAttributes<HTMLDivElement>`. Why this fails tho 🤷‍♂️
Hmm, but when we explicit use a function we will normally define it in types even if it is a native prop? Since we use it we should define it? Feels like a mismatch between other components

mimarz:
I had a sneak edit before, but heres the problem @wenche @pomfrida 

We have in some places used to define `onFocus`, `onBlur` and `onChange` properties with default NoOp functions, to avoid defined checking in the code, which is maybe a bit lazy as these are optional props... Suggest we check if its defined before executing in `Tabs`

vnys:
> Why is not `onBlur` and `onFocus` in the type? Is this intentionally? @vnys

I was about to say they were [defined as defaultProps](https://github.com/equinor/design-system/blob/develop/libraries/core-react/src/Tabs/Tabs.jsx#L66-L72) originally, so then I would use default parameters in the function signature or check if they exist as @mimarz suggests.

wenche:
Yes, that would be my suggestion as well. But even optional props used in this way should be declared with ? in types imho?

wenche:
> 
> I was about to say they were [defined as defaultProps](https://github.com/equinor/design-system/blob/develop/libraries/core-react/src/Tabs/Tabs.jsx#L66-L72) originally, so then I would use default parameters in the function signature or check if they exist as @mimarz suggests.

OK, makes sense, so a typescript migration "feature" then

mimarz (approved):
LGTM 👍

PR #817, opened 2020-11-04
Stepper variant update

lucasveil:


BeckyBrekke:
Looks good. Does Bendik need to look at it too, or can I merge it in?

PR #829, opened 2020-11-04
♻️  Conform element types

### LINKED ISSUE #818: Conform typescript element types
- [x] Go through all components and check if proptable is active in story, list those who don't and make new issues to fix the prop tables
- [x] Update all element types to match HTMLAttributes syntax and make sure this doesn't affect proptables or creates errors inside each component

pomfrida:
Resolves #818


wenche (approved):
This PR has a lot of things going on 😓 

Looks good as far as I can see, but we need a throughout testing at the end 👍

PR #846, opened 2020-11-05
Get the storefront up and running again after Typescript migration

### LINKED ISSUE #819: Validate Storefront against typescript core-react
- Remove props table if it does not work
- Check if it runs 🤷‍♂️

wenche:
Resolve #819


wenche:
When running `pnpn run storefront` I'm stuck with this
Skjermbilde 2020-11-05 kl  15 27 53

I initially hoped this was due to the fact that the storefront folder was missing the `babel-plugin-styled-component` for the [gatsby styled-component plugin](https://www.gatsbyjs.com/plugins/gatsby-plugin-styled-components/) where the function `withConfig` is defined (as far as I have found out... ) but adding that didn't solve the problem.

The problem is not in the `cjs.js` dist file but with the Gatsby generated `public/render-page.js` file for ssr as far as I can see

wenche:
> When running `pnpn run storefront` I'm stuck with this

> I initially hoped this was due to the fact that the storefront folder was missing the `babel-plugin-styled-component` for the [gatsby styled-component plugin](https://www.gatsbyjs.com/plugins/gatsby-plugin-styled-components/) where the function `withConfig` is defined (as far as I have found out... ) but adding that didn't solve the problem.
> 
> The problem is not in the `cjs.js` dist file but with the Gatsby generated `public/render-page.js` file for ssr as far as I can see

To solve this problem I had to add some config to the styled-component plugin in Gatsby
```
  {
      resolve: `gatsby-plugin-styled-components`,
      options: {
        displayName: false,
      },
    },
```
Since the issue appeared in the built file where displayName was added, I disabled this option. Shouldn't be necessary thought, I doesn't feel very futureproof 🙈

wenche:
I disabled the props table. We need to look more into this. It should be possible to use the real Docz Props table with Typescript somehow, but right now there's a lot of errors. I recommend that we look more into this at a later stage.

vnys:
But did you try to use the components from npm? Then you wouldn’t have to import the cjs bundle… If the propstable isn’t working anyway, and we’re not using components in the storefront that’s not already published, then I think that would be better. And we _can_ publish an alpha version of the typescript components just to test in the storefront.

wenche:
> But did you try to use the components from npm? Then you wouldn’t have to import the cjs bundle… If the propstable isn’t working anyway, and we’re not using components in the storefront that’s not already published, then I think that would be better. And we _can_ publish an alpha version of the typescript components just to test in the storefront.

Yes, I did in a separate branch. Can do this here as well if we prefer that approach 👍

pomfrida (approved):
LGTM 👍

Issue #826, opened 2020-11-04
Check viral ts-config settings


mimarz:

mimarz:
Maybe [create-react-library](https://www.npmjs.com/package/create-react-library) template/boilerplate can have some clues as to good ts-config for libraries

vnys:
Suggest closing this for now. Too many chain reactions.

mimarz:
Agreed @vnys beta-testing should uncover needs

Issue #822, opened 2020-11-04
Set up beta/next release for typescript


mimarz:
- [x] Set `link-workspace-packages = false` in `.npmrc`
- [x] Publish beta/next release
- [x] Create a CRA app in `/app/` and test the published components
- [x] Draft an announcement for Slack

wenche:
@vnys have created 3 releases:

- `@equinor/eds-tokens: 0.5.1-beta.1`
- `@equinor/eds-icons: 0.5.0-beta.1`
- `@equinor/eds-core-react: 0.7.0-beta.1`

pomfrida:
✔️  Posted announcement on the EDS Slack community and created a [Github discussion](https://github.com/equinor/design-system/discussions/867) to collect test reports

PR #848, opened 2020-11-06
✨ Add demo app

vnys:
Adds a create react app that can be run from the root with `pnpm run demo`
- I’ve updated Eslint in the root to the same version used by the demo-app
- The start-script has been renamed to `dev` like we do everywhere else
- I’ve set `allowSyntheticDefaultImports` and `esModuleInterop` to false in the tsconfig for the demo app since that’s the route we want to go with #826 
- I’ve installed `react-refresh` as a `devDependency` to make it work with PNPM

Partly solves #822


mimarz (approved):
LGTM 👍

PR #855, opened 2020-11-09
eds-core-react bundle improvements 

### LINKED ISSUE #627: Optimize eds-core-react bundle size
Bundle size for `eds-core-react` has gotten a lot bigger, 1.8mb with just a few typescript components, expected to increase by much more when done with typescript.

Look at ways to reduce bundle size

~* Bundle all typescript declarations into one file `d.ts` file?~ withdrawn
~* Do we need all modules types still (umd/cjs/es)? Looking at you umd 👀~  withdrawn
* Seperate `.map` files from published and dev bundle

mimarz:
partially resolves #627

### DIFF
--- a/libraries/core-react/package.json
+++ b/libraries/core-react/package.json
@@ -6,9 +6,9 @@
   "publishConfig": {
     "main": "dist/core-react.cjs.js",
     "module": "dist/core-react.es.js",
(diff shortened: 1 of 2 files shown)


pomfrida (approved):
LGTM 👍

PR #859, opened 2020-11-10
🔧 Update React imports

### LINKED ISSUE #826: Check viral ts-config settings

vnys:
This pr started out as an attempt to remove `allowSyntheticDefaultImports` and `esModuleInterop` from the config – but we already consume modules ourselves which require us to keep these settings. 

For example, if we remove these, we have to change the lodash/uniqueId import to `import createId = require('lodash/uniqueId')` – which then results in eslint errors because `eslint-plugin-import` does not yet support this syntax. 

See https://github.com/benmosher/eslint-plugin-import/issues/1244 for details. 

However, since the React imports are all now on separate lines, it will be easier for us to remove these once we upgrade to React 17, so I propose we still merge this pr as it is.

I also updated all the lock-files, allthough strictly not part of this issue.

resolves #826

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,3 +1,9 @@
+const pkg = require('./libraries/core-react/package.json')
+
+const {
+  devDependencies: { react: reactVersion },
+} = pkg
+
 module.exports = {
   root: true,
   extends: [
@@ -25,7 +31,7 @@ module.exports = {
   },
   settings: {
     react: {
-      version: 'detect',
+      version: reactVersion,
     },
   },
   globals: {
--- a/apps/figma-broker/pnpm-lock.yaml
+++ b/apps/figma-broker/pnpm-lock.yaml
@@ -10,7 +10,7 @@ dependencies:
   prettier: 2.0.5
   ramda: 0.27.0
   svgo: 1.3.2
-lockfileVersion: 5.1
+lockfileVersion: 5.2
 packages:
   /@koa/router/9.0.1:
     dependencies:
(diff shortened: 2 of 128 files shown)


.eslintrc.js
```diff
@@ -25,7 +31,7 @@ module.exports = {
   },
   settings: {
     react: {
-      version: 'detect',
+      version: reactVersion,
```
vnys, line 34: This setting was set to 'detect' which did not work because React is not installed in the root. Eslint then guesses that we’re using the latest version of React, which worked fine up until React 17. The only drawback from fetching the version from the core-react package, is that we now assume we’re using the same version of React in all the projects.

mimarz:
I ran to some edge-cases with tree-shaking and lodash aswell. Saw some people recommend using [lodash-es](https://www.npmjs.com/package/lodash-es) instead. Maybe this would also help you. Suggest we make a separate spike changing `lodash` with `lodash-es`

mimarz (approved):
Thats a lot of changes, but it looked ok as far as I could see 😅 

If I understood correctly `import * as React from 'react'` will be removed with a script for React v17?

wenche (approved):
Agree with @mimarz , lot's of file edits 😓 

LGTM 👍

vnys:
> If I understood correctly `import * as React from 'react'` will be removed with a script for React v17?

Yes there are scripts but a simple search and replace will do the job as well.

vnys:
> why don't we make our own `uniqueId` function

well that’s both the beauty and the curse of npm 😉 lodash is the swiss army knife of single purpose modules, so as long as they _work_, I don’t see why we should re-invent them, but in this case if it’s the _only_ module holding us back then I agree.

PR #862, opened 2020-11-11
eds-core-react dependencies 

mimarz:
Moved correct dependencies to `dev` and change `react` to be a minimum instead of up to minor releases.

This was done due to problems with resolving react versions and leaking dependencies used for development to consumers. 

image

So in short the following packages was installed for any user installing `eds-core-react`.... 🙈
  *  @testing-library/react-hooks
  * @types/lodash
   * csstype
   * focus-visible
   * lodash


wenche (approved):
#YOLO 🤪

PR #863, opened 2020-11-11
Add data visualisation as a top level

BeckyBrekke:
Moving data visualisation up from under patterns so that it is a top level category


apps/storefront/docs/patterns/data-visualisation/overview.mdx
```diff
@@ -1,22 +1,22 @@
 ---
-title: Data visualisation
-metaTitle: Data visualisation
+title: Overview
+metaTitle: Overview
```
vnys, line 3: I would change these back to "Data visualisation". The `title` becomes the main heading on the page, which is important for search engine optimalisation - if someone is looking for information about the EDS and data visualisation [in Google for example](https://www.google.com/search?q=eds+data+visualisation) and we want that document to rank high in their search results. 

The `metaTitle` is what's displayed in the tab in your browser and what ends up in people's bookmark list. If you look at https://eds.equinor.com/contributing/designers/ for example, you'll see that the main heading says "Designers" - but the tab says "Contributing for designers | Equinor Design System" - probably truncated since it's so long.

Ideally these should be the same or similar, but in this case we add more information to the `metaTitle` because if you have a long list of bookmarks it is more descriptive that is says something about the context and not just "Designers". The "| Equinor Design System" is added in code btw. 

The raw file for that page is https://raw.githubusercontent.com/equinor/design-system/develop/apps/storefront/docs/contributing/designers/overview.mdx.

vnys (changes requested):
I would change the title and metaTitle, apart from that it looks good 👍

apps/storefront/docs/patterns/data-visualisation/overview.mdx
BeckyBrekke: None of the other pages had the same name as the top level menu so I wasn't sure if that would look weird or what. Won't it come up twice then in the menu, Data visualisation, data visualisation?

vnys: Good point. What if you call it “Data visualisation in Equinor” then? Would be good to have “data visualisation” in the page header for SEO purposes.

BeckyBrekke:
I changed it now, can you check @vnys ?

vnys (approved):
LGTM 👍

PR #868, opened 2020-11-13
Simplify dependencies and package type

mimarz:
* `tokens` and `icons` have `cjs` bundle as main now.
* Removed `type` flag on `icons` & `tokens` as this flag is meant for consuming applications as[ i've understood](https://nodejs.org/docs/latest-v15.x/api/esm.html#esm_enabling). Meaning if you make a new application and set `type="module"` in `package.json`, it will then load the bundle under `"module"` in our library.
* Set all peer-deps to have a min.version approach

### DIFF
--- a/libraries/core-react/package.json
+++ b/libraries/core-react/package.json
@@ -82,8 +82,8 @@
     "styled-components": ">=4.2"
   },
   "dependencies": {
-    "@equinor/eds-icons": "0.5.0-beta.1",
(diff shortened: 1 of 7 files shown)


wenche (approved):
LGTM 👍 

Should we remove prop-types in this PR or wait?

vnys (approved):
LGTM 👍

mimarz:
> Should we remove prop-types in this PR or wait?

Suggest we do it separately

PR #877, opened 2020-11-16
🔧  Align lockfiles after pnpm settings update

wenche:
Changed link-workspace-packages  in .npmrc from true to false

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -11,7 +11,7 @@ devDependencies:
   eslint-plugin-testing-library: 1.5.0_eslint@7.11.0
   prettier: 2.0.5
   typescript: 4.0.2
-lockfileVersion: 5.2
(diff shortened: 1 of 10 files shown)


pnpm-lock.yaml
```diff
@@ -11,7 +11,7 @@ devDependencies:
   eslint-plugin-testing-library: 1.5.0_eslint@7.11.0
   prettier: 2.0.5
   typescript: 4.0.2
-lockfileVersion: 5.2
+lockfileVersion: 5.1
```
vnys, line 14: Something strange here… Maybe it has to do with different versions of pnpm?

wenche:
I had version `5.9.3` of pnpm installed when I ran into these issues. Upgraded to 5.12.0 and the problem disappeared. 🤷‍♀️

Issue #509, opened 2020-08-24
Create React Select component using native Select


mimarz:
Anita Update 02.12.2020
This US has high priority. Not sure why it is on the low pri. milestone. 
Spike US can be done first, if it is needed. 

Create React component for Select. Consider using https://react-select.com/ or other librariers due to complexity of component

mimarz:
blocked by #655

wenche:
We have created [a discussion at Github](https://github.com/equinor/design-system/discussions/861)

wenche:
We decided to split this task:
1. Native select (native version looking as much like the design as possible, but not 100% due to browser constraints)
2. Select (styled version like the Figma design)
3. Advanced version with some sort of search

wenche:
To clarify this split and to avoid this issue to be around until forever and a day, I create separate issues for the more complex variants of select

PR #879, opened 2020-11-16
Native select

wenche:
Part of #509


wenche:
I don't know how this ends up at the bitter end, but for the time being I just do the same as MUI and call the new component NativeSelect.

wenche:
Discussed the `TextField` and this component with @mimarz , we agreed it would be a good idea to reuse the `Label` and maybe some other components from the `TextField` in this component. Need to restructure the code slightly to make this work by just moving `Label` up to the top level. In the same way that `Paper` works by not reexport to the outside.

wenche:
We should probably change the inner `<label>` tags to span since `<p>` tags just feel… wrong.

wenche:
I had some issues with the typographyTemplate as it adds a black color. I had to use the cascade and add color below typographyTemplate in code. I don't know if this is intentionally or not 🤷‍♀️

mimarz:
> I had some issues with the typographyTemplate as it adds a black color. I had to use the cascade and add color below typographyTemplate in code. I don't know if this is intentionally or not 🤷‍♀️

Are we using the correct typygraphy token? Else we have normally overriden this in the component token file to avoid having to resort to cascading css. @wenche

wenche:
> > I had some issues with the typographyTemplate as it adds a black color. I had to use the cascade and add color below typographyTemplate in code. I don't know if this is intentionally or not 🤷‍♀️
> 
> Are we using the correct typygraphy token? Else we have normally overriden this in the component token file to avoid having to resort to cascading css. @wenche

Yes, I've triple checked, it's supposed to be `Text + static icons/⚪️/Tertiary` and input/text for Typography.

libraries/core-react/src/Select/NativeSelect/NativeSelect.tsx
```diff
@@ -0,0 +1,70 @@
...
+      id,
+      meta,
+      disabled = false,
+      multiple = false,
+      ...other
+    },
+    ref,
+  ) {
+    const containerProps = {
+      ref,
```
mimarz, line 43: Shouldn't the ref be passed to the `Select`component?

libraries/core-react/src/Select/NativeSelect/Select/Select.tokens.ts
```diff
@@ -0,0 +1,69 @@
...
+    border: {
+      outline: { color: string; width: string }
+      bottom: { color: string; width: string }
+    }
+    focus: {
+      border: {
+        width: string
+        color: string
+      }
+    }
```
mimarz, line 39: I think we have a type for `Border` and `Focus` in tokens types. Maybe try using them? Small-nude to conform tokens 🙈

libraries/core-react/src/Select/NativeSelect/Select/index.js
```diff
@@ -0,0 +1 @@
+export { Select } from './Select'
```
mimarz, line 1: This should be `.ts`?

mimarz (changes requested):
I like it! nice and native 🤘

Some nitpicking in regards to conformity and what about tests?

wenche:
> I like it! nice and native 🤘
> 
> Some nitpicking in regards to conformity and what about tests?

Haha, too much typescript lately! Completely forgot 🙈

libraries/core-react/src/Select/NativeSelect/Select/Select.tokens.ts
wenche: Did copy and paste from Text field - it feel's like that component is lagging behind

libraries/core-react/src/Select/NativeSelect/NativeSelect.tsx
wenche: To get i consistent I did it in the same way as the `<TextField>` component with one ref for the container and a separate ref for the select element 🤷‍♀️

libraries/core-react/src/Select/NativeSelect/Select/Select.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+const StyledSelect = styled.select`
+ border: none;
+ border-radius: 0;
+  box-shadow: inset 0 -${tokens.default.border.bottom.width} 0 0 ${
+  tokens.default.border.bottom.color
+};
+  ${spacingsTemplate(tokens.spacings.input)}
+  ${typographyTemplate(tokens.typography)}
+  color: ${tokens.color};
```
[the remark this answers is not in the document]
wenche: But you mean outside this PR to avoid confusion and too much stuffs right? @mimarz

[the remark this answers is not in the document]
mimarz: You should fix it in `Select` in this PR I would say.

mimarz (approved):
LGTM 👍

Issue #871, opened 2020-11-13
Fix typescript testing errors 


mimarz:
> Upgrade was pretty straightforward. 
> Tested on
> ````
>     "react": "^17.0.1",
>     "react-dom": "^17.0.1",
>     "typescript": "~4.0.2",
>     "@equinor/eds-core-react": "^0.7.0-beta.3",
>     "@equinor/eds-icons": "^0.5.0-beta.1",
>     "@equinor/eds-tokens": "^0.5.1-beta.1",
> ````
> 
> Textfield deprecated property errorMsg, using helperText instead in combination with variant. 
> 
> Some unexpected changes: 
> * TextField has no longer any rows property (when multiline is true)
> * Icon has no longer variant property. 
> ![image](https://user-images.githubusercontent.com/48285335/98955052-c2988280-24fe-11eb-908b-647628760e2b.png)
> 

_Originally posted by @cbuv in https://github.com/equinor/design-system/discussions/867#discussioncomment-127844_

blocked by #866

PR #885, opened 2020-11-17
📝 Added 'rows' to TextField prop table 

pomfrida:
Solves #871

Notes:

- `rows` prop was always included in `InputHTMLAttributes` but I added it manually to the Story to have it documented. It's probably not intuitive for all users that these default attributes are included as well as the custom ones.

- Regarding the `errorMsg` prop, I don't find that it deprecated on Typescript migration. It was already labeled `helperText`. I didn't take more time to research this, but I guess they have used this prop before and now after TS support it's not allowed.

- `variant` prop in Icon has never been a case before either, so I guess the same happened as with the `errorMsg` mentioned above

So the only 'fix' here is adding rows prop to `TextField` story to document multiline input.

### DIFF
--- a/apps/storybook-react/stories/TextField.stories.tsx
+++ b/apps/storybook-react/stories/TextField.stories.tsx
@@ -45,6 +45,13 @@ const ICONS = {
 export default {
   title: 'Components/TextField',
   component: TextField,
+  argTypes: {
+    rows: {
+      control: 'number',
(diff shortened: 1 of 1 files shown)


mimarz (approved):
LGTM 👍

mimarz (approved):
LGTM 👍

PR #888, opened 2020-11-17
🐛 Fixed CardActions type error

mimarz:
Type error in CardActions is causing `core-react` build to fail now for some reason...


vnys (approved):
LGMT 🍓

PR #889, opened 2020-11-17
[𝓌𝒾𝓅] 👷 Fix Storybook Docker build

### LINKED ISSUE #886: Storybook does not build in Docker after typescript merger
Looks like we are hitting heap size limits or something now.

Maybe we need to add more multi-build stages with all the libraries being built at once 

* `Dockerfile` and `azurepipelines` needs to be updated

vnys:
resolves #886


vnys:
The build is quite huge! I assume this is part of the build problem in Docker…

```sh
WARN asset size limit: The following asset(s) exceed the recommended size limit (244 KiB).
WARN This can impact web performance.
WARN Assets:
WARN   main.45c31fe3a30104a9f2cc.bundle.js (1.36 MiB)
WARN   vendors~main.45c31fe3a30104a9f2cc.bundle.js (4.53 MiB)
WARN entrypoint size limit: The following entrypoint(s) combined asset size exceeds the recommended limit (244 KiB). This can impact web performance.
WARN Entrypoints:
WARN   main (5.89 MiB)
WARN       runtime~main.45c31fe3a30104a9f2cc.bundle.js
WARN       vendors~main.45c31fe3a30104a9f2cc.bundle.js
WARN       main.45c31fe3a30104a9f2cc.bundle.js
```

vnys:
This is actually _down_ from the pre-typescript branch, so maybe I’m barking up the wrong tree 🤔 

```sh
asset size limit: The following asset(s) exceed the recommended size limit (244 KiB).
WARN This can impact web performance.
WARN Assets:
WARN   static/media/cat.8d35c0d4.jpg (536 KiB)
WARN   main.c591e4a1b5d1ca0ebcfc.bundle.js (2.18 MiB)
WARN   vendors~main.c591e4a1b5d1ca0ebcfc.bundle.js (4.21 MiB)
WARN entrypoint size limit: The following entrypoint(s) combined asset size exceeds the recommended limit (244 KiB). This can impact web performance.
WARN Entrypoints:
WARN   main (6.4 MiB)
WARN       runtime~main.c591e4a1b5d1ca0ebcfc.bundle.js
WARN       vendors~main.c591e4a1b5d1ca0ebcfc.bundle.js
WARN       main.c591e4a1b5d1ca0ebcfc.bundle.js
```

mimarz:
@vnys seems like this is a [know problem...](https://github.com/storybookjs/storybook/issues/12348)

vnys:
I added fish and neovim to the Dockerfile to make it easer to debug with `docker run -it `. So the status now is that if `pnpm m i` and `pnpm build:storybook` is run as part of the Docker build, then the build fails. However, if I comment out those lines, build it, and then run it interactively and do those same steps manually from inside the container, then everything works. 🤷 I also tried `npm@7`, which now has workspaces, but got an issue with conflicting versions of some dependencies, which we don’t get with pnpm – since pnpm doesn’t have any problems with that. So the short summary is that there’s still no solution, and unfortunately I’m going to have to put this on hold for a couple of days to take care of other things on my list.

mimarz:
> I added fish and neovim to the Dockerfile to make it easer to debug with `docker run -it `. So the status now is that if `pnpm m i` and `pnpm build:storybook` is run as part of the Docker build, then the build fails. However, if I comment out those lines, build it, and then run it interactively and do those same steps manually from inside the container, then everything works. 🤷 I also tried `npm@7`, which now has workspaces, but got an issue with conflicting versions of some dependencies, which we don’t get with pnpm – since pnpm doesn’t have any problems with that. So the short summary is that there’s still no solution, and unfortunately I’m going to have to put this on hold for a couple of days to take care of other things on my list.

Okidoki, seems like pain. 
Suggest we remove `storybook-react` pipeline from PR checking until this is resolved then @vnys

vnys:
Yeah that’s probably a good idea. We can also build locally and deploy to the container registry manually as a last resort so we get storybook with types at least.

vnys:
Had another stab at this in #937 🔪

PR #892, opened 2020-11-18
✨ Variants export support

mimarz:
Side-effect from spiking for theming, I found out we could export all the variants created in Figma which can be used to conform on property names in code or some other cool features 😎


wenche (approved):
LGTM 👍

PR #893, opened 2020-11-18
⬆️ Use @next in the Storefront

vnys:
- Installs dependencies from npm in Docker instead of hard linking internally
- Uses the @next release of the EDS components


mimarz (approved):
LGTM 👍

PR #898, opened 2020-11-19
Restructure core-react for future bugfixes and improvments

### LINKED ISSUE #887: Restructure core-react
Life has progressed, lessons learned and features grown,  its time to restructure for cleaner code.

Suggest restructuring `src` into something like this for now:

* `src`
  * `/components`
     * _contains all the react components, external and internal_
  * `/utils`
     * _different utils used that are not a hook_
  * `/hooks`
    * _hooks used in components_

mimarz:
resolves #887 

**new structure:**
image

* Added alias `@hooks` & `@utils` for easier access (no longer need to dot your way up)
* Removed `useKeyPress` as is was not used
* Had to replicate alias mapping in jest and storybook config for _reasons_. Good to remember when adding new aliases in the future.
* Added `.size-snapshot.json` to ignore for now
* Minor type fixes to list after move


pomfrida (approved):
Sweet 🥇

wenche (approved):
👍

PR #899, opened 2020-11-20
♻️  Fix some of the minor issues discovered by working on the Select

wenche:
Avoiding height of 37px with border. Add the correct color to the tokens. The color in use was not the correct one from Figma, so changed that as well.


mimarz (approved):
Niceness! Good we remember to fix the small inconsistencies 👍

PR #906, opened 2020-11-23
Single and Multi select

### LINKED ISSUE #896: Create React Select component
This issue is for creating a Select component with the complete design from Figma with styled options and multiple items selected.

Blocked by #655 as it's probably difficult to create this component all by ourself.

wenche:
Resolve #896

### DIFF
--- a/apps/storybook-react/stories/Select/MultiSelect.stories.tsx
+++ b/apps/storybook-react/stories/Select/MultiSelect.stories.tsx
@@ -22,3 +22,11 @@ export const Default: Story = (args) => (
     ></MultiSelect>
   </Container>
 )
+export const Disabled: Story = () => (
+  <MultiSelect
+    label="Choose an element"
+    meta="km/t"
+    items={items}
+    disabled
+  ></MultiSelect>
+)
--- a/apps/storybook-react/stories/Select/MultiSelect.stories.tsx
+++ b/apps/storybook-react/stories/Select/MultiSelect.stories.tsx
@@ -10,7 +10,7 @@ export default {
 } as Meta
 
 const Container = styled.div`
-  margin-bottom: 150px;
+  margin-bottom: 350px;
 `
 
 export const Default: Story = (args) => (
--- a/apps/storybook-react/stories/Select/MultiSelect.stories.tsx
+++ b/apps/storybook-react/stories/Select/MultiSelect.stories.tsx
@@ -53,8 +53,6 @@ export const UseOnChangeHandler: Story<MultiSelectProps> = () => {
     changes: UseMultipleSelectionStateChange<string>,
   ) {
     setSelectedItems(changes.selectedItems)
-
-    //setSelectedItem(changes.selectedItems)
   }
   return (
     <Container>
@@ -77,8 +75,6 @@ export const Controlled: Story<MultiSelectProps> = () => {
     changes: UseMultipleSelectionStateChange<string>,
   ) {
     setSelectedItems(changes.selectedItems)
-
-    //setSelectedItem(changes.selectedItems)
   }
   return (
     <Container>
--- a/apps/storybook-react/stories/Select/MultiSelect.stories.tsx
+++ b/apps/storybook-react/stories/Select/MultiSelect.stories.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { MultiSelect } from '@equinor/eds-core-react'
+import { MultiSelect, MultiSelectProps } from '@equinor/eds-core-react'
 import { Story, Meta } from '@storybook/react/types-6-0'
 import styled from 'styled-components'
 import { items } from './data'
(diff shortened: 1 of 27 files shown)


wenche:
There were some issues and bugs by using the whole wrapped checkbox with label for the multi select options list. Especially when you clicked several times at the same item. The example from the Downshift approach uses a single input for the option list, without the label. Since it's supposed to be good on a11y, it felt safer to follow the same approach 🤷‍♀️  It also seems like the click maniac bug disappeared…

wenche:
The ghost button doesn't actually fit, even if I killed some kittens and add a size of 24px. There's so much markup and styles inside it, so it's overdoing things in this scenario. Maybe we should consider to create a smaller variant of an icon button? 🤔

wenche:
I don't actually know where to put the `ref` element in this case. I'm also a bit worried about declaring it as a select element with `SelectHTMLAttributes<HTMLSelectElement>` in types. We don't actually have somewhere to spread the {…other} in this case as it's not _actually_ a select element. What to do?

wenche:
I'm not quite sure about the best way to solve a controlled version of `selectedItems`.

wenche:
I added a few examples to get a feeling on how it's like to use this in the wild. We need to discuss where we add spread props and what this component _actually is_ in regard of element. I don't think we actually should use `SelectHTMLAttributes`

wenche:
I attempted to support the `itemToString` prop as I thought it would be nice to have arrays with `[{id: 'id', value: 'My string value}]`. But it's complicated and I didn't managed to fulfill this within a reasonable amount of time. 😞

wenche:
Still unsure if we should make `items` optional or not, my initial idea was to provide the opportunity to fetch on demand but it might be over engineering.

wenche:
Add the` {...other} ` on the input element. I'm still not quite sure about what is the correct approach here, as we are still using the `SelectHTMLAttributes` and what do we want people to do anyway 🤯

wenche:
We are struggling at bit with the ARIA representation of selected items, especially with MultiSelect. However, hopefully this will be fixed by the Downshift library as [this is how I would expect it to work.](https://github.com/downshift-js/downshift/issues/935#issuecomment-735506327)

wenche:
This is v1 for single and multi select. As this is pretty complicated stuff, we don't want to make the scope too big. Known limitations:
- If the options are an array of object, you'll need to handle a string representation of that outside of `Select` and `MultiSelect`
- Due to ARIA complexity, this version don't provide an indeterminate checkbox "Select all"
- In a multi select with lot's of options, it might be difficult to quickly get an overview of the selected items.

libraries/core-react/src/components/Select/commonStyles.ts
```diff
@@ -0,0 +1,62 @@
...
+  box-shadow: ${tokens.boxShadow};
+  overflow-y: scroll;
+  max-height: 300px;
+  padding: 0;
+  border-radius: ${tokens.borderRadius};
+  margin-top: 4px;
+  position: absolute;
+  right: 0;
+  left: 0;
+  z-index: 200;
```
pomfrida, line 40:
```suggestion
  z-index: 50;
```
I think Select option list should have lower priority than Popover (100) and Menu (150)

vnys (approved):
LGTM 👍

apps/storybook-react/stories/Select/MultiSelect.stories.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+export const UseOnChangeHandler: Story<MultiSelectProps> = () => {
+  const [selectedItems, setSelectedItems] = useState<string[]>(initial)
+
+  function handleSelectedItemsChange(
+    changes: UseMultipleSelectionStateChange<string>,
+  ) {
+    setSelectedItems(changes.selectedItems)
+
+    //setSelectedItem(changes.selectedItems)
```
pomfrida, line 57:
```suggestion
  
```

apps/storybook-react/stories/Select/MultiSelect.stories.tsx
```diff
@@ -0,0 +1,202 @@
...
+}
+export const Controlled: Story<MultiSelectProps> = () => {
+  const [selectedItems, setSelectedItems] = useState<string[]>(initial)
+
+  function handleSelectedItemsChange(
+    changes: UseMultipleSelectionStateChange<string>,
+  ) {
+    setSelectedItems(changes.selectedItems)
+
+    //setSelectedItem(changes.selectedItems)
```
pomfrida, line 81:
```suggestion
    
```

libraries/core-react/src/components/Select/MultiSelect/MultiSelect.tsx
```diff
@@ -0,0 +1,223 @@
...
+    const {
+      getDropdownProps,
+      addSelectedItem,
+      removeSelectedItem,
+      selectedItems,
+    } = useMultipleSelection(multipleSelectionProps)
+    const getFilteredItems = (items: string[]) =>
+      items.filter((item) =>
+        // Remove selected items from the list
+        /* selectedItems.indexOf(item) < 0 && */
```
pomfrida, line 100: Comment
```suggestion
       
```

libraries/core-react/src/components/Select/MultiSelect/MultiSelect.tsx
```diff
@@ -0,0 +1,223 @@
...
+
+        // Can be used if we need filter the list on first letters aka starts with search
+        // item.toLowerCase().startsWith(inputValue.toLowerCase()),
```
pomfrida, line 103:
```suggestion
     
```

libraries/core-react/src/components/Select/SingleSelect/SingleSelect.tsx
```diff
@@ -0,0 +1,139 @@
...
+    const [inputItems, setInputItems] = useState(items)
+    const isControlled = selectedOption !== undefined ? true : false
+
+    let comboboxProps: UseComboboxProps<string> = {
+      items: inputItems,
+      onSelectedItemChange: handleSelectedItemChange,
+      onInputValueChange: ({ inputValue }) => {
+        setInputItems(
+          items.filter((item) =>
+            // item.toLowerCase().startsWith(inputValue.toLowerCase()),
```
pomfrida, line 76:
```suggestion
      
```

pomfrida (approved):
Awesome job. I just found some comments we can remove

Issue #678, opened 2020-10-15
Conform typescript types on tokens


mimarz:
With types its apparent that there are some deviations on how base values are used such as typography, clickbounds, hover, focus, borders etc. 

- [ ] Conform `<component>.token.ts` files with defined types for base types such as  typography, clickbounds, hover, focus, borders etc. 
- [ ] Asses if `figma-broker` needs changes    

Look to #521 discussion for typescript token syntax and approach

PR #910, opened 2020-11-24
Base Table tokens

mimarz:
Lots of changes 🙈

* Had to do a bigger change in the `Cell` and `Head` sub-components to support new base tokens
* Partially solved #678 to figure out how to do different variants/validations for `DataCell`
  * Added some common types in tokens which was a result from this.
  * Type `ComponentToken` is meant to be used for all component tokens.
  * Improved templates for better use with typescripts type-safety. (We should use `styled-components` with an object instead for string template for type-safety)
  * Improved base types for tokens
* Partially solved #297 as I had to refactor `Cell` because of different typescript types for cell header and data. 
* Changed `package.json` links to workspace versions as I expect this PR to be merged in after #912


pomfrida (approved):
LGTM 👍

wenche (approved):
LGTM 👍

PR #922, opened 2020-11-26
Don't show Tooltip if title is empty

### LINKED ISSUE #920: Tooltip: Hide component
If title="", do not show anything. Alternativly have an disabled prop. This will be useful when you only want to show the tooltip sometimes, for instance only when a button is disabled

pomfrida:
Resolves #920


vnys (approved):
LGTM 👍

PR #937, opened 2020-11-30
🐛 Fix storybook

### LINKED ISSUE #886: Storybook does not build in Docker after typescript merger

vnys:
Things are working now ~but all the props in the entire galaxy are in the props table. This sounds like a familiar problem, so I’m counting on somebody else to tell me what the problem is 😉~

A couple of things: 
- I renamed `storybook-react` ⭢ `storybook` and the folders under libraries are now prefixed with `eds-`. I think this makes more sense, since that’s the actual names of the packages – and also because it’s convenient when creating aliases to just rename `libraries` ⭢ `@equinor`. 
- In Docker we now copy `libraries` into the storybook instead of using pnpm’s hard-linking which is causing problems for webpack
- That’s it – a couple means two

I have forgotten stuff – because my head is spinning. ~The azure pipelines come to mind.~ Anything else? Shoot from the hip 🔫

Oh, and this resolves #886


mimarz (changes requested):
Sooo many changes, but I was not able to build `eds-core-react`, probably due to some config files not being updated to new folder name yet? 

A quick search I found `.eslintrc.js` and `'lint:core-react'` script in `package.yaml` still had references to old folder names. `.esignore` in storefront also had old reference to `core-react`

vnys:
Fixed and rebased on updated develop, but there’s still a build problem with `eds-core-react` that seems to be related to `import type`…

vnys:
…the reason being that `icons` and `tokens` have to be built first, or else there are no types…

mimarz:
> …the reason being that `icons` and `tokens` have to be built first, or else there are no types…

🤦‍♂️ But ofcourse @vnys 

But do we want to do this restructure now that we have #921 and #906 PRs in the works? This can cause a lot of frustration making such a big change while people are working on stuff.

Can we fix this without renaming and maybe do that bit later/after the PRs are done?

vnys:
> 🤦‍♂️ But ofcourse @vnys

🤨

vnys:
> But do we want to do this restructure now that we have #921 and #906 PRs in the works? This can cause a lot of frustration making such a big change while people are working on stuff.

That’s the problem of long living branches. At what point are we not working on stuff? 

> Can we fix this without renaming and maybe do that bit later/after the PRs are done?

We _can_ – it’s just a lot of work going reverting that now. Let’s discuss it in the standup.

apps/storefront/.eslintignore
```diff
@@ -2,6 +2,6 @@
 /../../apps/storybook-html/**
 /../../apps/storybook-react/**
 /../../libraries/icons/**
-/../../libraries/core-react/**
-../../libraries/core-react/dist/core-react.umd.js
+/../../libraries/eds-core-react/**
+../../libraries/eds-core-react/dist/core-react.umd.js
```
mimarz, line 6: missed revert changes?

libraries/core-react/azure-pipelines.yml
```diff
@@ -47,10 +47,10 @@ steps:
...
     displayName: 'Install dependencies'
 
-  - script: pnpm --filter ./libraries/core-react test
+  - script: pnpm --filter ./libraries/eds-core-react test
     displayName: 'Run tests'
 
   - script: >-
       ./node_modules/eslint/bin/eslint.js
-      ./libraries/core-react
+      ./libraries/eds-core-react
```
mimarz, line 55: missed revert name changes?

libraries/core-react/package.json
```diff
@@ -27,7 +27,6 @@
     "dist/*"
   ],
   "scripts": {
-    "prepare": "npm run build",
```
mimarz, line 30: suggest we change this to `pnpm` instead of removing altogether to avoid human errors when publishing to npm (god, we need to fix publishing this with pipelines soon 🙈)

libraries/icons/azure-pipelines.yml
```diff
@@ -40,7 +40,7 @@ steps:
       install
     displayName: 'Install dependencies'
 
-  - script: pnpm --filter ./libraries/icons run build
+  - script: pnpm --filter ./libraries/eds-icons run build
```
mimarz, line 43: missed revert name changes?

mimarz (changes requested):
Looks like there was still some files missed with reverting the names change?

I suggest we keep the `prepare` and fix it to pnpm instead of removing, just to avoid that human factor when publishing 👨

libraries/core-react/package.json
vnys: Yes we do, some day…

libraries/core-react/azure-pipelines.yml
vnys: Yup, that’s the problem with reverting this manually – and also why it will probably not be renamed again 😉

apps/storybook-react/.storybook/main.js
```diff
@@ -1,7 +1,13 @@
+const path = require('path')
+
 module.exports = {
-  // typescript: {
-  //   reactDocgen: 'react-docgen',
```
mimarz, line 3: remove dead code?

libraries/tokens/package.json
```diff
@@ -25,9 +25,9 @@
     "commonjs/*"
   ],
   "scripts": {
-    "prepare": "npm run build",
```
mimarz, line 28: missing `"prepare": "pnpm run build"` by the looks of it?

apps/storybook-react/.storybook/main.js
```diff
@@ -19,4 +25,34 @@ module.exports = {
       },
     },
   ],
+  webpackFinal: async (config, { configType }) => {
```
mimarz, line 28: Where is the value for `configType` set/decided?

mimarz (changes requested):

apps/storybook-react/.storybook/main.js
vnys: Not sure what you mean, it’s already removed

mimarz: I need more coffee ☕ 🙈

apps/storybook-react/.storybook/main.js
vnys: Read the docs

libraries/tokens/package.json
vnys: *sigh* I’m fed up with this task, I need more coffee too…

mimarz (approved):
LGTM 👍

PR #963, opened 2020-12-09
🐛 Fixed icon button horizontal clickbound

### LINKED ISSUE #595: Update Ghost button hover size
Ghost button was updated with reduced hover size from 48px to 40px. Clickbound remains the same at 48px.

[Figma](https://www.figma.com/file/0bGXR2sCwMVSDNyyzu5BXrO5/UI%E2%80%94User-Interface?node-id=1%3A4)

Quick review of needed changes
* Changes needs to be done in figma-broker for supporting width offset (as clickbound stays the same size)
* Update Button with support for offset on x-axis (width)

mimarz:
fixes #595

### DIFF
--- a/libraries/core-react/src/components/Button/tokens/icon.ts
+++ b/libraries/core-react/src/components/Button/tokens/icon.ts
@@ -19,9 +19,11 @@ const {
   clickbounds: { default__base: clicboundHeight },
 } = tokens
 
+const buttonSize = 40
+
 export const primary: ButtonToken = R.mergeDeepRight(button, {
-  height: '40px',
-  width: '40px',
+  height: `${buttonSize}px`,
+  width: `${buttonSize}px`,
   border: {
     radius: '50%',
   },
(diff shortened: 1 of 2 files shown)


libraries/core-react/src/components/Button/tokens/icon.ts
```diff
@@ -26,6 +27,12 @@ export const primary: ButtonToken = R.mergeDeepRight(button, {
   },
   background: 'transparent',
   spacings: { left: '0', right: '0' },
+  clickbound: {
+    width: clicboundHeight,
+    offset: {
+      left: `${(parseInt(clicboundHeight) - 40) / 2}px`,
```
wenche, line 33: What is 40? Size?

wenche (approved):
LGTM 👍 

Perhaps we should make a variable of 40 just to remember what it is? E.g. if it's size `const buttonSize = 40`

Issue #965, opened 2020-12-09
Release libraries


wenche:
Release day 👏 

Release new versions of eds-core-react and eds-tokens

PR #967, opened 2020-12-09
Release eds-tokens

wenche:
Part of #965 

### Changed

- Removed component tokens in favour of the base tokens approach
  - `Table`
  - `Button`
    - Danger
    - Disabled
    - Primary
    - Secondary

### DIFF
--- a/libraries/tokens/CHANGELOG.md
+++ b/libraries/tokens/CHANGELOG.md
@@ -5,6 +5,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.5.3] - 2020-12-09
+
+### Changed
+
+- Removed component tokens in favour of the base tokens approach
+  - `Table`
+  - `Button`
+    - Danger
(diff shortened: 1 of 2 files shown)


libraries/tokens/CHANGELOG.md
```diff
@@ -5,6 +5,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.5.3] - 2020-12-09
+
+### Changed
```
mimarz, line 10: I would say this should be categorised as a "Removed" according to [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).

mimarz (changes requested):

mimarz:
@wenche maybe this change is a minor version bump after all... 🤔

libraries/tokens/CHANGELOG.md
wenche: True that.

wenche:
> @wenche maybe this change is a minor version bump after all... 🤔

If we were on 1.x.x then yes

mimarz (approved):
LGTM 👍

PR #968, opened 2020-12-09
:bug: Fix typo in component name

wenche:


mimarz (approved):
👍

vnys:
Is the only difference here that the changelog is updated? Then I don’t think we should make a new release just because of that.

wenche:
Decided to not create a new release with this typo fix.

PR #971, opened 2020-12-10
Tests for table component

wenche:
Part of #297

### DIFF
--- a/libraries/core-react/src/components/Table/Table.test.tsx
+++ b/libraries/core-react/src/components/Table/Table.test.tsx
@@ -4,6 +4,7 @@ import { render, cleanup, screen } from '@testing-library/react'
 import '@testing-library/jest-dom'
 import 'jest-styled-components'
 import { Table } from '.'
+import styled from 'styled-components'
 
 const { Caption, Cell, Head, Row, Body } = Table
 
@@ -17,6 +18,19 @@ describe('Caption', () => {
       </Table>,
     )
     expect(container.querySelector('Caption')).toBeInTheDocument()
+  })
+  it('Prints the caption', () => {
+    const text = "I'm da caption"
+    const { getByText } = render(
+      <Table>
+        <Head>
+          <Row>
+            <Cell as="th">{text}</Cell>
+          </Row>
+        </Head>
+      </Table>,
+    )
+    expect(getByText(text)).toBeDefined()
   })
   it('Renders a table with caption, and caption-side set default to top', () => {
     const { container } = render(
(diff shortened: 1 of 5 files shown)


wenche:
Don't test implementation details they say 🙈

libraries/core-react/src/components/Table/Table.test.tsx
```diff
@@ -41,3 +55,72 @@ describe('Caption', () => {
...
+          <Row>
+            <Cell variant="numeric">{text}</Cell>
+          </Row>
+        </Body>
+      </Table>,
+    )
+
+    expect(getByText(text)).toHaveStyleRule(
+      'font-feature-settings',
+      "'tnum' on,'lnum' on",
```
mimarz, line 114: This should be tested against the `DataCell` tokens value for numeric cells

libraries/core-react/src/components/Table/Table.test.tsx
```diff
@@ -41,3 +55,72 @@ describe('Caption', () => {
...
+          <Row>
+            <Cell>{body}</Cell>
+          </Row>
+        </Body>
+      </Table>,
+    )
+    const headElement = getByText(header).closest('thead')
+    const regularContent = getByText(body)
+    expect(headElement).not.toHaveStyleRule('background', 'rgba(255,255,255,1)')
+    expect(regularContent).toHaveStyleRule('background', 'rgba(255,255,255,1)')
```
mimarz, line 97: If I am not mistaken the Table head also has a thicker bottom border that can be asserted on.

mimarz (changes requested):
We still have some missing features in `Table` that needs to have testes further on but its a start :)

libraries/core-react/src/components/Table/Table.test.tsx
wenche: Yes, probably... But it gives 
Skjermbilde 2020-12-14 kl  10 27 59

wenche: Tried to fix this in figma-broker, but it didn't work that well... Seems like prettier is re adding the whitespace or something 🤯  Cheated and  trimmed the text. I'm not sure if the world is a better place or not after this change 🤷‍♀️  @mimarz

libraries/core-react/src/components/Table/Table.test.tsx
wenche: I added that as well. Using the tokens as assertion was just too complicated so I just added the border as a string. But things are slightly better by the usage of tokens for the white colour at least

mimarz (approved):
LGTM 👍

PR #975, opened 2020-12-11
🔖 Release new versions of the libraries

vnys:


libraries/core-react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@equinor/eds-core-react",
-  "version": "0.8.0",
+  "version": "0.8.2",
```
pomfrida, line 3: Why not 0.8.1 here? 🤔

wenche: Because we burned it 😂  Ref. issues with the non-working package

wenche: Had to unpublish 0.8.1 from npm, and because of security npm doesn't allow you to republish with the same number. :)

pomfrida: Ahaa that's right 🤣

pomfrida (approved):
👍 LGTM

PR #976, opened 2020-12-14
🐛 Remove unintended button width

mimarz:


wenche (approved):
LGTM 👍 

Did see that when upgrading, but I wasn't sure if it had been like that before or not

mimarz:
> LGTM 👍
> 
> Did see that when upgrading, but I wasn't sure if it had been like that before or not

yeah, same 🙈

Issue #962, opened 2020-12-09
Combine pages in Figma to create a better structure and hierarchy


benkeq:
Sort components in categorized pages such as "Navigation", and "Inputs".

**Data display**
- Chip
- Divider (MUI have added it under data display, is that the best place for it?)
- Lists (not a component)
- Popover
- Table
    * Header
    * Cell
- Tooltip

**Feedback/Notifications** 
- Banner
- Dialog
- Progress indicators
    * Linear
    * Circular
    * Dots
    * Star
- Scrim
- Snackbar

**Inputs/Data entry/Forms**
- Button
- Date picker
- Search
    * Search
    * Scope
- Select
- Selection controls
    * Checkbox
    * Radio button
    * Switch
- Slider
    * Dot
    * Handle
    * Indicator
    * Track
- Text fields
- Time picker
- Toggle button

**Navigation**
- App launcher
- Breadcrumbs
- Drawer
    * Listing
    * Container
    * Subtitle
    * Destination item
- Link (not a component)
- Menu
    * Menu
    * Container
    * Item
    * Subtitle
- Pagination
- Stepper
- Table of Contents
    * Item
    * Listing
- Tabs
- Top bar

**Surfaces**
- Accordion
- Card
    * Container
    * Title
    * Rich media
    * Supplemental actions
- Side sheet
- Toolbar

vnys:
Looks good 👍 We do have some components in code that we don’t have in Figma – `<Typography>` for example. I think that would fit under “display”. So maybe add that to the list with “code only” in parentheses since we’ll use the same structure in the Storybook. How about `<Avatar>`, don’t we have that in Figma?

mimarz:
Looks good. Maybe a stupid question, but does everything need to be categorised?  

> ... We do have some components in code that we don’t have in Figma – `<Typography>` for example. I think that would fit under “display”. So maybe add that to the list with “code only” in parentheses since we’ll use the same structure in the Storybook...

Typography is defined as a design token in Figma, and implemented as component in code. Which is not wrong by any mean but dunno.  🤷‍♂️

> ...How about <Avatar>, don’t we have that in Figma?

I think its time we add this to Figma to for consistency :D Maybe some notes on sizing and such for avatars?

vnys:
> Typography is defined as a design token in Figma, and implemented as component in code. Which is not wrong by any mean but dunno. 🤷‍♂️

[...]

PR #977, opened 2020-12-14
🚀 Deploy storefront

vnys:
118 images updated 🤔 
Does this sound correct @equinor/eds-ux?

### DIFF
--- a/apps/storefront/docs/components/accordion/overview.mdx
+++ b/apps/storefront/docs/components/accordion/overview.mdx
@@ -6,8 +6,9 @@ tabs:
   - Usage
 toc:
   - Guidelines
-  - Variations
+  - Variant
   - Interaction
+  - Demo
 mode: publish
 route: /components/accordion/
 ---
--- a/apps/storefront/docs/components/chips/overview.mdx
+++ b/apps/storefront/docs/components/chips/overview.mdx
@@ -7,7 +7,8 @@ tabs:
 toc:
   - Guidelines
   - Variations
-  - Examples
+  - Variant
+  - Demo
 mode: publish
 route: /components/chips/
 ---
--- a/apps/storefront/docs/components/divider/overview.mdx
+++ b/apps/storefront/docs/components/divider/overview.mdx
@@ -6,7 +6,7 @@ tabs:
   - Usage
 toc:
   - Guidelines
-  - Variations
+  - Variant
   - Demo
 mode: publish
 route: /components/divider/
--- a/apps/storefront/docs/components/menu/overview.mdx
+++ b/apps/storefront/docs/components/menu/overview.mdx
@@ -11,21 +11,6 @@ mode: publish
 route: /components/menu/
 ---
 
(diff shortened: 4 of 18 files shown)


mimarz:
@vnys have you looked at the images? Curious how this looks now with changes done in #962

lucasveil:
I haven't uploaded nor updated any image on my last documentation update for variants. I have updated the figma links though, does that count as an image in the storefront?

vnys:
> have you looked at the images?

Haven’t checked all the images, but I’ve browsed quickly through a few pages at https://eds-storefront-development.azurewebsites.net

vnys:
> I haven't uploaded nor updated any image on my last documentation update for variants. I have updated the figma links though, does that count as an image in the storefront?

If you make changes in tokens or components in Figma then that results in an updated image when we run the Figma Broker, so maybe the re-organisation is the reason there are so many updated images…

mimarz:
> > I haven't uploaded nor updated any image on my last documentation update for variants. I have updated the figma links though, does that count as an image in the storefront?
> 
> If you make changes in tokens or components in Figma then that results in an updated image when we run the Figma Broker, so maybe the re-organisation is the reason there are so many updated images…

true true, its just based of the unique id that figma has for its images, these changes for each version figma has of a file if something within that image is changed. So either the re-organization or variants...

mimarz:
@vnys A lot of the pages are missing images. Buttons, lists and toc to name a few... Maybe this calls for an issue to investigate...

mimarz:
> I haven't uploaded nor updated any image on my last documentation update for variants. I have updated the figma links though, does that count as an image in the storefront?

@lucasveil These are the images exported when using <FigmaImage /> :) 
```
<FigmaImage
  url="https://www.figma.com/file/0bGXR2sCwMVSDNyyzu5BXrO5/Desktop-UI?node-id=1%3A4"
  alt="Figma design of primary buttons"
/>
```

vnys:
> @vnys A lot of the pages are missing images. Buttons, lists and toc to name a few... Maybe this calls for an issue to investigate...

Sounds like it does, regression tests would have been nice – too many risks for human error the way things are now. I’ll look into it next year – don’t have the time this week 😉

mimarz:
> > @vnys A lot of the pages are missing images. Buttons, lists and toc to name a few... Maybe this calls for an issue to investigate...
> 
> Sounds like it does, regression tests would have been nice – too many risks for human error the way things are now. I’ll look into it next year – don’t have the time this week 😉

I compared the urls for these, and its because of the re-organization(#962) that they are missing, because they got new urls now

mimarz (changes requested):
As we agreed to drop the updating of images, we should revert changes done in [update images commit](https://github.com/equinor/design-system/pull/977/commits/0a55a663e27c93205e89e9568cc584cb303b7630)

vnys:
> drop the updating of images

That commit has now been delegated to the history disposal unit and only exists in a parallel universe

mimarz (approved):
LGTM 👍

PR #978, opened 2020-12-14
📝 Replace non-working React code

nilsml:
In the Readme and in the storefront docs the code for creating a React button was wrong


mimarz (approved):
🤦‍♂️ 

Thanks for the fix! 👍

PR #982, opened 2020-12-16
Table: Hover & Error state on rows

### LINKED ISSUE #940: Add missing table states
States for active and error cell in table. 
How to show active and hover on both row and cell?

Edit:
~- [ ] Active state, row only~ Replaced by #981
- [x] Hover state, row only
- [x] Error (Validation) state, cell only

mimarz:
resolves #940

### DIFF
--- a/apps/storybook-react/stories/Table.stories.tsx
+++ b/apps/storybook-react/stories/Table.stories.tsx
@@ -1,7 +1,9 @@
 import React from 'react'
 import styled from 'styled-components'
-import { Story, Meta } from '@storybook/react/types-6-0'
-import { Table, TableProps, Typography, Tooltip } from '@equinor/eds-core-react'
+import { Story, Meta } from '@storybook/react'
+import { Table, TableProps, Typography, Icon } from '@equinor/eds-core-react'
+import { warning_outlined } from '@equinor/eds-icons'
+import { tokens } from '@equinor/eds-tokens'
 import './../style.css'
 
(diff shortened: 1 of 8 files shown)


wenche (changes requested):
I think we should remove the Tooltip until we have a better example? Perhaps an i-icon or something just to indicate that there's _something_ to explore

mimarz:
> I think we should remove the Tooltip until we have a better example? Perhaps an i-icon or something just to indicate that there's _something_ to explore

Replaced with an warning icon in first col as this display example is also viable if you have many many columns

wenche:
> > I think we should remove the Tooltip until we have a better example? Perhaps an i-icon or something just to indicate that there's _something_ to explore
> 
> Replaced with an warning icon in first col as this display example is also viable if you have many many columns

Sry, but I still don't understand this example 😬  And empty thead cell? 🤔

mimarz:
> > > I think we should remove the Tooltip until we have a better example? Perhaps an i-icon or something just to indicate that there's _something_ to explore
> > 
> > 
> > Replaced with an warning icon in first col as this display example is also viable if you have many many columns
> 
> Sry, but I still don't understand this example 😬 And empty thead cell? 🤔

Should have been an icon in thead cell maybe, but removed it alltogether, saving it for later when we have more features that work together for a better example then :)

wenche (approved):
LGTM 👍

Issue #823, opened 2020-11-04
Decide on compact mode component api


mimarz:
How do users or developers trigger compact mode on components which support this? 

Food for thought
* Do we have a boolean prop which triggers compact mode for the respective component
* Or do we have a global variable which triggers compact mode for all components
* Or some additional context switcher on the component for users to decide mode

mimarz:
blocks #811

wenche:
MUI's approach
![Skjermbilde 2020-12-17 kl  08 49 38](https://user-images.githubusercontent.com/1568252/102459171-8ac4b380-4045-11eb-85bb-aa4e8dfd8443.png)

wenche:
We discussed this on today's standup.

In Figma it's designed as a boolean at component level. This feature is located in the top menu. We decided to keep it as simple as possible for the time being, and add a prop density on the `<Table>` component.

wenche:
![Skjermbilde 2020-12-17 kl  10 14 01](https://user-images.githubusercontent.com/1568252/102467581-c022ce80-4050-11eb-8034-740b715860b4.png)

mimarz:
> We discussed this on today's standup.
> 
> In Figma it's designed as a boolean at component level. This feature is located in the top menu. We decided to keep it as simple as possible for the time being, and add a prop density on the `<Table>` component.

`density` sounds good. Suggest we have atleast ` "comfortable" | "compact"` as values. A `"default"`value could be applicable so that users can save the preference in the localstorage or so, but maybe save that for later 🤪

PR #985, opened 2020-12-17
✨  Add compact table 

### LINKED ISSUE #811: Table compact mode
Related to #297

wenche:
Resolve #811

### DIFF
--- a/apps/storybook-react/stories/Table.stories.tsx
+++ b/apps/storybook-react/stories/Table.stories.tsx
@@ -1,7 +1,7 @@
 import React from 'react'
 import styled from 'styled-components'
 import { Story, Meta } from '@storybook/react'
-import { Table, TableProps, Typography, Icon } from '@equinor/eds-core-react'
+import { Table, TableProps, Typography } from '@equinor/eds-core-react'
 import './../style.css'
 
 const { Caption, Body, Row, Cell, Head } = Table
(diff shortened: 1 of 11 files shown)


libraries/core-react/src/components/Table/Cell/utils.ts
```diff
@@ -0,0 +1,18 @@
+import { TableCellToken } from './DataCell.tokens'
+import { TableHeadToken } from './HeaderCell.tokens'
+import { Density } from '../Table.types'
+
+type Cell = TableCellToken | TableHeadToken
+type ApplyDensity<T> = (density: Density, token: Partial<Cell>) => T
```
mimarz, line 6: Forgot to remove unused type 😅

libraries/core-react/src/components/Table/index.tsx
```diff
@@ -1,19 +1,28 @@
...
+import { Table as BaseTable, TableProps } from './Table'
 import { Body } from './Body'
 import { Cell } from './Cell'
 import { Head } from './Head'
 import { Row } from './Row'
 import { Caption } from './Caption'
+import { TableProvider } from './Table.context'
 
-type TableCompoundProps = typeof BaseComponent & {
+const TableWrapper: FC<TableProps> = (props) => (
```
mimarz, line 11: Does this still show up as `TableWrapper` in Storybook props table? Maybe change the const name or displayname to `Table`

wenche: Hmm, yes. And still no props, forgot to check that

apps/storybook-react/stories/Table.stories.tsx
```diff
@@ -122,3 +131,42 @@ export const FixedTableHeader: Story<TableProps> = () => {
...
 }
+
+export const CompactTable: Story<TableProps> = () => {
+  const data = [
+    { number: '1', name: 'Banana', colour: 'Yellow' },
+    { number: '2', name: 'Orange', colour: 'Orange' },
+    { number: '4', name: 'Kiwi', colour: 'Greenish' },
+  ]
+  return (
+    <Table density="compact">
```
vnys, line 142: Did you consider `<Table compact>` instead of `<Table density="compact">` since there are only two possible values? 🤔 Also since `compact` is the exception to the rule 🤷🏻

vnys (commented):
Looks good, but I would consider using `<Table compact>` – or have you already discussed that?

apps/storybook-react/stories/Table.stories.tsx
mimarz: We considered using just `compact`, but felt it lacking in terms of future compatibility and component api design. Part of discussion is here: #823. 

If someone has `compact` as default for their application, then `comfortable` would be the exception to that applications rule. Having a `density` prop also makes it easier for users to implement a global toggle for setting density application wide.

vnys (approved):
LGTM 👍🏻

Issue #987, opened 2020-12-18
Unwanted hover-effect on icons wrapped in ghost button


ErlendLygre:
**Describe the bug**

Hover state stays visible on ghost buttons after user has tapped on mobile

**Steps to reproduce the bug**

1. Go to https://eds-storybook-react.azurewebsites.net/?path=/docs/components-selection-controls-checkbox--default
2. Change to "toggle device toolbar" in chrome inspect mode. Select any phone
3. Press the checkbox and remove "finger" from device,

**Expected behavior**

A ripple effect or no effect at all when pressing the checkbox, such as here https://material-ui.com/components/checkboxes/

**Specifications**

- Version: 
- Browser: 
- OS: 

**Additional context**

Add any other context or screenshots about the bug.

PR #988, opened 2021-01-04
🐛 Stuck hover effects on touch devices

pomfrida:
Solves #987 

Using [CSS media query](https://medium.com/@mezoistvan/finally-a-css-only-solution-to-hover-on-touchscreens-c498af39c31c) approach to check if device has a mouse before applying hover styles


wenche (commented):
LGTM 👍 

Lot's of fancy things one could do if IE is not an issue :)

wenche (approved):

Issue #955, opened 2020-12-07
Storybook crashes when changing icon name


vnys:
**Describe the bug**

As [mentioned on Slack](https://equinor.slack.com/archives/CJT20H1B9/p1607340075219400) – when a user change the name value on the[ icon examples](https://eds-storybook-react.azurewebsites.net/?path=/docs/components-icon--icon-examples) the page crashes.

**Steps to reproduce the bug**

1. Go to https://eds-storybook-react.azurewebsites.net/?path=/docs/components-icon--icon-examples
2. Change the value of the name property

**Expected behavior**

It should show the icon you put into the name field

**Specifications**

- Version: n/a
- Browser: Microsoft Edge
- OS: macOS Big Sur

mimarz:
This is due to our check if provided icon name exists in our library [here](https://github.com/equinor/design-system/blob/7ac17da28b7d98dba9649944c51e8779cd6ce5e3/libraries/core-react/src/components/Icon/Icon.tsx#L84). 

We had weird errors getting this option to work with Storybooks select option in props table for some reason during [conversion of story](https://github.com/equinor/design-system/pull/812). Might work now since we have updated Storybook since then.

PR #991, opened 2021-01-04
📝 Updated Icon story with controls

pomfrida:
Solves #955 

Adding a selection of icons in the control solves crash when changing icon name in demo.
Did not include all Icons. Is it ok with only a selection or should we add all available EDS icons to this list?

### DIFF
--- a/apps/storybook-react/stories/Icon.stories.tsx
+++ b/apps/storybook-react/stories/Icon.stories.tsx
@@ -56,19 +56,7 @@ export default {
     name: {
       control: {
         type: 'select',
-        options: [
-          'layers',
-          'layers_off',
-          'grid_off',
-          'compare',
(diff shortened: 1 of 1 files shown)


apps/storybook-react/stories/Icon.stories.tsx
```diff
@@ -54,7 +54,25 @@ export default {
   component: Icon,
   argTypes: {
     name: {
+      control: {
+        type: 'select',
+        options: [
```
mimarz, line 59: Suggest changing options to use `Object.keys(icons)`, so it will be updated dynamically if we change icons used in example.

mimarz (changes requested):

mimarz (approved):
LGTM 👍

PR #995, opened 2021-01-05
Change from cricle to circle

### LINKED ISSUE #993: Wrong icon name
`close_cricle_outlined` should be `close_circle_outlined` [here](https://github.com/equinor/design-system/blob/e9c1c64fdf4de744e2dec456b7cab8da06260cb6/libraries/icons/src/data.ts#L882)

maiaht:
Closes #993


mimarz (changes requested):
Much love ❤️ for providing a fix for this typo 😊

You missed renaming the static svg file for close_circle_outlined [here](https://github.com/equinor/design-system/blob/develop/libraries/static/icons/system-icons/ui-action/close_cricle_outlined.svg) which also needs to be renamed.

mimarz:
Did you fix this using the [assets script](https://github.com/equinor/design-system/blob/develop/apps/figma-broker/Commands.httpie) in figma-broker or just changed files manually? 

All our assets are generated dynamically from Figma, so it will need to be fixed there before we can approve this fix, or else its just going to re-appear next time we run the assets script 🙈

maiaht:
The files were changed manually

mimarz (approved):
Thanks for your contribution! 👍

PR #996, opened 2021-01-06
📝  Document required TS version

### LINKED ISSUE #931: EDS libraries does not work for older than v3.8 Typescript
We are using the `import type` syntax for type-only imports which was introduces in [Typescript 3.8](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html)

Using Typescript older than 3.8 gives an error when trying  to read the declaration files generated for our EDS libraries.

Example when using core-react.
image

pomfrida:
Resolves #931

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,10 +101,13 @@ The [storefront][] is the official documentation for the EDS. As we’re getting
 This is the main implementation of the EDS, and will eventually contain accessible React components of each and every EDS component in Figma.
 
 ### Installation
-Make sure you have Typescript v3.8 or newer
 
 ```sh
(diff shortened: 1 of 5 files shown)


vnys (changes requested):
Since Typescript is optional, I suggest rephrasing this a little bit: 

### Installation

```sh
npm install @equinor/eds-core-react styled-components
```

If you use Typescript, make sure you have `typescript >= 3.8` as a devDependency:

```sh
npm install typescript --save-dev
```

wenche (approved):
LGTM! 👍

vnys (approved):
LGTM 👍🏻

Issue #997, opened 2021-01-06
Deprecate as="th" on Table Cell


mimarz:
With smart use of Context API, we can know when a `<Cell />` needs to be rendered as `td` or `th`. 

1. We make a separate "content" context which is used by `<Cell />`.
2. We use the context provider in `<Head />` and `<Body />` and set a "variant" enum to be either `head` or `body` for the context value. 
3. The `<Cell />` can then determine which type it needs to render based on variant context value.

Notes: 
* Material-ui's [implementation]( https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/TableHead/TableHead.js)

PR #998, opened 2021-01-06
Support for Table active and sorting styling

### LINKED ISSUE #983: Should table header have a default sort state
Should table headers have a "default" state to indicate that a column is sortable without hover/interacting. 

Icons & state for: ascending, descending and none.

Active styling is only applied for asc and desc.

Related to #653 

Sketches: https://www.figma.com/file/luNobpDIWTuQ9rhyovGLBI/Desktop?node-id=624%3A0

mimarz:
resolves #983 

* Added `sort` on `<Cell/>` that applies `aria-sort` and appropriate active/hover styling when this props is defined.
* Refactored `Table.stories` to use the same data and columns for all examples.
* Added sortable table example with custom sorting logic.
* Examples with `react-table` _will have to wait due_ to lacking typescript support in v7.
* ⚠️ Missing icon on hover for sortable Table headers, as we need to remove the `as` prop for this to be done correctly. Will be fixable after #997 is resolved.

### DIFF
--- a/apps/storybook-react/stories/Table.stories.tsx
+++ b/apps/storybook-react/stories/Table.stories.tsx
@@ -180,101 +180,135 @@ export const CompactTable: Story<TableProps> = () => {
   )
 }
 
-export const Sortable: Story<TableProps> = (args) => {
-  const columns: Column<HackHead>[] = React.useMemo(
-    () => [
-      {
-        Header: 'Number',
-        accessor: 'number',
-      },
-      {
-        Header: 'Name',
-        accessor: 'name',
-      },
-      {
-        Header: 'Colour',
-        accessor: 'colour',
-      },
-    ],
-    [],
-  )
-
-  const data: HackHead[] = [
-    { number: '1', name: 'Banana', colour: 'Yellow' },
-    { number: '2', name: 'Orange', colour: 'Orange' },
-    { number: '4', name: 'Kiwi', colour: 'Greenish' },
+export const Sortable: Story<TableProps> = () => {
+  const data = [
+    {
+      number: '123-456',
+      description: 'Pears',
+      origin: 'Europe',
+      price: 1.5,
+    },
+    {
+      number: '234-567',
+      description: 'Apples',
+      origin: 'Africa',
+      price: 1.2,
+    },
+    {
+      number: '45-6789',
+      description: 'Oranges',
+      origin: 'South America',
+      price: 1.8,
+    },
   ]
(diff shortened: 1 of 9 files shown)


libraries/core-react/src/components/Typography/Typography.tsx
```diff
@@ -52,7 +52,7 @@ const findTypography = (
 }
 
 const findColor: (a: ColorVariants | string) => string = (
-  inputColor = null,
+  inputColor = 'currentColor',
```
pomfrida, line 55: ❗ Note to self: Major change

mimarz: Reverted this change, and passed `currentColor` via the `color` prop on `<Typography />` instead as this change had unintended consequences.

apps/storybook-react/stories/Table.stories.tsx
```diff
@@ -1,11 +1,104 @@
...
+          ($)
+        </Typography>
+      </>
+    ),
+    accessor: 'price',
+    sortDirection: 'none',
+  },
+]
+
+// type PrepareData<T extends typeof data> = (data: T, columns: Column) => T[][]
```
wenche, line 91: Should be removed?

libraries/core-react/src/components/Table/Cell/HeaderCell.tsx
```diff
@@ -1,44 +1,82 @@
...
+  const { token } = props
   const { background, height, typography, spacings } = token
+  const activeToken = token.states.active
+  const ariaSort = props['aria-sort']
+  let sortStylingHover = css({})
+  let sortStylingActive = css({})
+
+  if (ariaSort) {
+    sortStylingHover = css`
+      &:hover {
```
wenche, line 23: Ref. lack of hover state on touch devices, shouldn't we wrap in a media query here as well?

wenche (changes requested):
Just a minor thing... Looks good! 👍

apps/storybook-react/stories/Table.stories.tsx
mimarz: Fixed

libraries/core-react/src/components/Table/Cell/HeaderCell.tsx
mimarz: Added :)

wenche (approved):

PR #1002, opened 2021-01-11
Remove addon-knobs

### LINKED ISSUE #757: Replace knobs with control addon in Storybook
Controls seems to be the replacement for knobs with their new docs format. https://github.com/storybookjs/storybook/tree/next/addons/controls#how-will-this-replace-addon-knobs

Controls docs: https://storybook.js.org/docs/react/essentials/controls

Its also part of their "essentials" addon package, so maybe update it with that? 

https://www.npmjs.com/package/@storybook/addon-essentials

wenche:
As this is kind of deprecated with the new control prop

Resolve #757


wenche:
Since the removal of addon-knobs is a part of new features of storybook, we decided that it makes sense to do some other upgrade things in the same PR 🤫

pomfrida (approved):
LGTM 👍

PR #1005, opened 2021-01-12
Table cell improvements

### LINKED ISSUE #997: Deprecate as="th" on Table Cell

mimarz:
resolves #997 

* Created a `Inner.context`, which determines which variant `Cell` within that context are rendered as. 
* ⚠️ Removed property "as", as its no longer needed and causes bugs when trying to extend `Cell` with styling using `styled()`
* Exposed typings for the other `Table` components as needed when "extending" our components with additional styling. 
* Added icon on hover for sortable cells in example

### DIFF
--- a/apps/storybook-react/stories/Table.stories.tsx
+++ b/apps/storybook-react/stories/Table.stories.tsx
@@ -224,12 +224,24 @@ export const Sortable: Story<TableProps> = () => {
   const onSortClick = (sortCol: Column) => {
     const updateColumns = state.columns.map((col) => {
       if (sortCol.accessor === col.accessor) {
-        const sortDirection: SortDirection =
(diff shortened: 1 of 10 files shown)


pomfrida (approved):
LGTM 👍

wenche (approved):
LGTM! 👍 

Much easier to use

PR #1014, opened 2021-01-12
🎨  Update Table to use forwardRef

### LINKED ISSUE #810: Conform on the use of forwardRef and {...rest}
Some component (at least the `Snackbar`, `Banner` and `Table` components) are missing forwardRef and spreading of rest props. This should be aligned across all components.

Also, some component use {...props} at the last parameter, other as the first as some logic depends on another order.

### LINKED ISSUE #1007: Update Table to use forwardRef
Update Table and its sub components to use, React's `forwardRef` so its on par with our other components

pomfrida:
Resolves #1007

### DIFF
--- a/libraries/core-react/src/components/Table/Head.tsx
+++ b/libraries/core-react/src/components/Table/Head.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react'
-import { FunctionComponent } from 'react'
+import { ReactNode } from 'react'
 import styled from 'styled-components'
 import { token } from './Cell/HeaderCell.tokens'
 import { bordersTemplate } from '@utils'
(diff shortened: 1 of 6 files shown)


libraries/core-react/src/components/Table/Head.tsx
```diff
@@ -10,12 +10,12 @@ const StyledTableHead = styled.thead`
   background: ${token.background};
 `
 
-export const Head: FunctionComponent = ({ children, ...props }) => {
+type HeadProps = { children: ReactNode }
```
mimarz, line 13: Should this not be `React.HTMLAttributes<HTMLTableHeaderCellElement>` type?

libraries/core-react/src/components/Table/Head.tsx
```diff
@@ -10,12 +10,12 @@ const StyledTableHead = styled.thead`
...
+
+export const Head = ({ children, ...props }: HeadProps): JSX.Element => {
```
mimarz, line 15: Missing use of `forwardRef` here

mimarz (changes requested):
Seems to be some missing changes in `Cell` and `Body` for `forwardRef`?

pomfrida:
@mimarz I had some problems with the `ref` and the `TableHeaderCellElement` type in `Head`. It suggested to use `HTMLTableSectionElement` instead. Last commit with `Body` and `Head` fix 👍

mimarz (approved):
LGTM 👍

PR #1016, opened 2021-01-12
Update table example in storefront

### LINKED ISSUE #1006: Update Table code examples in Storefront
Even if it's cumbersome to change several places, we need to update the example in the storefront as well when we have started down that path

wenche:
Resolve #1006

### DIFF
--- a/apps/storefront/docs/components/table/overview.mdx
+++ b/apps/storefront/docs/components/table/overview.mdx
@@ -13,7 +13,7 @@ mode: publish
 route: /components/table/
 ---
 
-import { Table } from '@equinor/eds-core-react'
+import { Table, Typography } from '@equinor/eds-core-react'
 
 _Tables_ display data in a structured format.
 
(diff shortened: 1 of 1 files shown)


wenche:
OK, so the stupid thing here is that we are using package names and not workspace for the Storefront app. 🤦‍♀️  I can't remember why exactly, hopefully for some good reason? But the consequence is that the features I wanted to display is not available in the version of the React component we currently depend on.

That means that we'll have to remember to bump the storefront `package.lock` file when we release packages.... Do something about the storefront anybody? 🤯 

Rant aside, for this particular issue I decided to just update as if the updated package is available so that the examples will work eventually when the package is updated. Until then, no runtime errors but missing style features.

vnys:
> OK, so the stupid thing here is that we are using package names and not workspace for the Storefront app. 🤦‍♀️ I can't remember why exactly, hopefully for some good reason?

Because we want the package to be installed from NPM and not from the workspace, because if we install from the workspace [we get the typescript files](https://github.com/equinor/design-system/blob/develop/libraries/core-react/package.json#L5-L10) – and that doesn’t work with the storefront.

vnys:
> That means that we'll have to remember to bump the storefront `package.lock` file when we release packages

Do we still have to do that if we set version to `"latest"`?

wenche:
> > That means that we'll have to remember to bump the storefront `package.lock` file when we release packages
> 
> Do we still have to do that if we set version to `"latest"`?

Yeah, did consider that, but since I didn't remembered why we did it in the first place, I didn't actually tried it.

But I guess it will depend on the order? 🤔  We'll need the correct package at build time, so that means the Storefront must be built and deployed after a package is updated. So probably more automation will help us out here? Even if it's far from an ideally solution :)

pomfrida:
Should we also do #1028 here, and add a line in Storefront about Table Caption: 'if caption is top, then it can be any of the headings, else its paragraph/caption' ? 🤔

apps/storefront/docs/components/table/overview.mdx
```diff
@@ -78,29 +78,83 @@ If a Cell input is used in a column made out of Cell monospaced numeric componen
...
+        <Table.Cell scope="col">
+          Description
+        </Table.Cell>
+        <Table.Cell scope="col">
+         Origin
+        </Table.Cell>
+        <Table.Cell scope="col">
+         <>
+        Price 
+        <Typography group="input" variant="label" color="currentColor">
```
mimarz, line 139: you can add ` as="span"` to get correct element type here.

mimarz:
> Should we also do #1028 here, and add a line in Storefront about Table Caption: 'if caption is top, then it can be any of the headings, else its paragraph/caption' ? 🤔

Suggest we solve it in a seperate PR for that issue, as there may, be or not be more changes that just adding a line ;)

apps/storefront/docs/components/table/overview.mdx
```diff
@@ -78,29 +78,83 @@ If a Cell input is used in a column made out of Cell monospaced numeric componen
...
+        <Table.Cell scope="col">
+          Description
         </Table.Cell>
-        <Table.Cell as="th" scope="col">
-          Kill count
+        <Table.Cell scope="col">
+         Origin
+        </Table.Cell>
+        <Table.Cell scope="col">
+         <>
```
vnys, line 96: What’s this fragment doing here?

wenche: Copy'n paste gone wrong 🤪

mimarz (approved):
LGTM 👍

PR #1029, opened 2021-01-13
Verify input elements in compact table

### LINKED ISSUE #1027: Verify use of EDS components in compact Table
Make example of using compact Table with the components used in [Figma desktop examples ](https://www.figma.com/file/luNobpDIWTuQ9rhyovGLBI/Desktop?node-id=624%3A0)

- [x] Selection controls
- [x] Ghost button
- [x] Icon
- [x] Menu (opened from ghost button)
- [x] Textfield
- [x] SingleSelect

wenche:
resolve #1027 

This is not actually code we want to document in the Table stories for end users and it also added a lot of code. That's why I created a new menu item called Playground. As this is just an exercise for us to find pain points, we don't actually have a good way of doing this 🤷‍♀️

### DIFF
--- a/apps/storybook-react/stories/DataTable.stories.tsx
+++ b/apps/storybook-react/stories/DataTable.stories.tsx
@@ -1,0 +1,306 @@
+import React from 'react'
+import { Story, Meta } from '@storybook/react'
+import {
+  Table,
+  TableProps,
+  Typography,
+  Icon,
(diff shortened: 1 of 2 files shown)


mimarz (approved):
LGTM 👍

Will be good to use for validation of our components used in `Table`

PR #1030, opened 2021-01-13
🐛  Add addon-actions back in

wenche:
#757 removed addon-actions as the package is a part of addon-essentials. However, I missed that we use i directly in some components, like in the `Menu`. I didn't catch this error previously. 😊  

I can't see how the [new way to use actions](https://storybook.js.org/docs/react/essentials/actions) will solve the way we use them, since we're not on the actual parent component. Feels like we still need what storybook mentions as Advanced / legacy usage. 🤔


mimarz (approved):
LGTM 👍

pomfrida (approved):
Nice catch 👍

PR #1039, opened 2021-01-15
✨ Added sticky option for Table head

### LINKED ISSUE #1031: Implement sticky option to Table
resolves #1001

Suggest we introduce a `sticky` prop to Table, Head or Row, which applies the necessary css for it to be sticky. 

~Requirements for sticky are sticky header and sticky columns. Columns can also be stickied to the left or right.~

**Edit:** 

Turns out having sticky `th` cells along with sticky `td` cells does not work as intended using table html elements.

* If we had a sticky header and a fixed column, the columns header would ignore the sticky `th` cell and scroll up along with content. 
* Another implementation stopper is having a sticky column you need to set the `left/top/right/bottom` in accordance with how far from the edge it should stick. We can not predict column widths (and code to detect would be cumbersome) as this is a usage detail for our users. 

Research also shows that most people who provide this feature, are complexed tables and using `div` with table/grid role for correctly stickiness. As we only provide a semantic table, this is not an option for us at this time.

Notes: 
* https://css-tricks.com/position-sticky-and-table-headers/
* https://styled-components.com/docs/api#forwardedas-prop

mimarz:
resolves #1031

### DIFF
--- a/apps/storybook-react/stories/Table.stories.tsx
+++ b/apps/storybook-react/stories/Table.stories.tsx
@@ -53,6 +53,18 @@ const data: Data[] = [
     origin: 'South Africa',
     price: 2.5,
   },
+  {
+    number: '89-012',
+    description: 'Pineapple',
+    origin: 'Paraguay',
+    price: 1.9,
+  },
+  {
+    number: '89-012',
+    description: 'Pomegranate',
+    origin: 'Persia',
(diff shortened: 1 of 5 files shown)


pomfrida (approved):
LGTM 👍

vnys (approved):
LGTM 🍰

PR #1040, opened 2021-01-15
EdsProvider for density control

mimarz:
solves #1032 

Introduce `EdsProvider` for contextually setting components behaviour within eds. 

### Component development usage: 
In any component you can tap into the context by using `useEds()` from `EdsProvider`. This will give any of our components the ability to change behaviour when children on any level for `EdsProvider`

One given use-case is to have components (`Checkbox`, `TextField`, `Button` etv) be forced as compact within `Table` when its density is set to compact.


vnys (approved):
I think this looks good! It gives the users the ability to have more compact interfaces without us having to sacrifice our commitment to creating accessible components. Good job 👏🏻

mimarz:
> I think this looks good! It gives the users the ability to have more compact interfaces without us having to sacrifice our commitment to creating accessible components. Good job 👏🏻

I see we can remove the `Table.context` and some other minor clean code improvements. Will fix this before merging.

PR #1050, opened 2021-01-19
🐛 Fix pnpm version

### LINKED ISSUE #999: The production build of the storefront fails
The latest production build of the storefront fails when being deployed. We need to set up a pre-production stage to catch errors before deploying to production, and to find out what fails and why.

vnys:
Npm installed an old version of pnpm in Docker. Added latest to version to fix it. Also bumped required version in package.json.

Resolves #999


mimarz (approved):
LGTM 👍

PR #1059, opened 2021-01-19
Compact table with density switcher

### LINKED ISSUE #1042: Create compact table story with topbar and menu
To make it easier for developers to “do the right thing” we should create a story in the Storybook with an example of how to use the `<TopBar>` and `<Menu>` components to let the user choose the density of `<Table>`.

mimarz:
resolves #1042 

- Found an edge-case when using `EdsProvider` and changing values that was not being updated correctly which I fixed here
- Update existing "Compact table" story with `Topbar` and `Menu` for switching `density` on `Table`

### DIFF
--- a/apps/storybook-react/stories/Table.stories.tsx
+++ b/apps/storybook-react/stories/Table.stories.tsx
@@ -7,8 +7,11 @@ import {
   Typography,
   Icon,
   CellProps,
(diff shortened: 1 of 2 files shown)


pomfrida (approved):
LGTM 👍

PR #1066, opened 2021-01-20
Fix SingleSelect rerender (equinor#1064)

### LINKED ISSUE #1064: Select - items list to be triggered on stateChange outside Select component

Trying to use SingleSelect  , passing in an array of strings to items. 
However, this array is created during render, and when click on the SingleSelect, the pop-down list is empty. 
If I add a character (preferably 'a') in the SingleSelect field, the list populates, but not before. 
I´ve tried UseEffect in order to get a re-render when the array is populated, but this has no effect on the SingleSelect list. The only way I get the list populated is when the list is hardcoded  as a constant as in the example you show in the StoryBook. 

So I think this occurs because SingleSelect (or any of the EDS selects) does not trigger of state changes after initial render.

The ability to pass in a list of items in Select which is created during render and stored in a useState variable, and control re-render with useEffect (as possible with material UI select.)

The workaround (not elegant) I need to do now to get the list populated is to delay render of the Select until I know that the items list is ready.
```
{displayFields.length > 0 && 
     <SingleSelect id={"fields"} label={"Fields"} items={displayFields} handleSelectedItemChange={onChangeFieldFilter} />
}
```

The solution I would like would be something like this:
```
useEffect(() => {
  if (fields) {
      const list = fields.map((field) => field.name);
      setDisplayFields(list);
    }
  }, [])};

return (
    <SingleSelect id={"fields"} label={"Fields"} items={displayFields} handleSelectedItemChange={onChangeFieldFilter} />
)
```
**Additional context**

### LINKED ISSUE #1065: Single Select - Populate the whole list, not only the last selected item when reselecting

When a user selects one item in the list, this is the only value populated if the user decides to reselect looking for another value.

So if the user want to change value, she has to delete manually the selected value, in order to get the full list of items populated again...

image
image

After having selected a value, the user should get the whole list of items as default populated if she decides to look for another value.

Removing the selected value manually is cumbersome..., a slightly better alternative is an "x" delete option just to quickly clear the input field. However, the best alternative is that the whole list is populated always.

**Additional context**

cbuv:
fixes #1064 with useEffect to update local state inputItems

### DIFF
--- a/libraries/core-react/src/components/Select/SingleSelect/SingleSelect.tsx
+++ b/libraries/core-react/src/components/Select/SingleSelect/SingleSelect.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react'
-import { forwardRef, SelectHTMLAttributes, useState } from 'react'
+import { forwardRef, SelectHTMLAttributes, useEffect, useState } from 'react'
 import {
   useCombobox,
   UseComboboxProps,
@@ -66,6 +66,10 @@ export const SingleSelect = forwardRef<HTMLDivElement, SingleSelectProps>(
   ) {
     const [inputItems, setInputItems] = useState(items)
     const isControlled = selectedOption !== undefined ? true : false
+
+    useEffect(() => {
+      setInputItems(items)
+    }, [items])
 
     let comboboxProps: UseComboboxProps<string> = {
       items: inputItems,


mimarz (approved):
Awesomesauce! 

LGTM 👍

pomfrida (approved):
🏅

PR #1068, opened 2021-01-21
Use state improvements

### LINKED ISSUE #1062: Fix usages of useState in React
We have been abit lazy on usage of the hook `useState`. 🙈 This is not to critical, but should be done very soon to avoid future bugs or unwanted use-cases we dont know about.

 https://reactjs.org/docs/hooks-reference.html#functional-updates
image

Most places we have just merged the current state with new values when updating state. As `useState` setter is async, this may lead to new values being overwritten by old values when the setter is called multiple times in succession.

mimarz:
resolves #1062 

Fixed `useState` where objects are used for storing multiple values and updates are done to avoide race conditions because of `useState` async nature


wenche (approved):
LGTM 👍 

Just these three components?

mimarz:
> LGTM 👍
> 
> Only these three components were affected?

As far as i could tell, only the places where we have multiple values stored in state this is a problem for. The rest was just single value state handling (mostly bool flipping)

PR #1069, opened 2021-01-21
❌  Disabled Table density functionality

mimarz:


wenche (approved):
LGTM 👍 

Guess it's easier to reintroduce when it is just commented out

Issue #1046, opened 2021-01-19
Spike storybook as storefront


mimarz:
Try to get an idea what it takes to move the content from the Storefront to the storybook to replace the storefront.

There are other component libraries using only Storybook for documentation, see e.g [Circuit](https://circuit.sumup.com/) and [Orbit](https://sg-orbit.netlify.app/)

- [x] What can be achieved by styling
- [x] Recreate the front page with graphics
- [x] Test Figma addon
- [x] Use Markdown for a page instead of the default storybook template

Merged into the branch `spike-storybook-as-storefront`

~Deployed to https://edsstoragedev.z16.web.core.windows.net/~

wenche:
A simple Markdown approach, based on copy'n paste from original Storefront markdown files.
<img width="1788" alt="Skjermbilde 2021-01-19 kl  19 43 28" src="https://user-images.githubusercontent.com/1568252/105079741-20fa5800-5a90-11eb-8d94-03f9fd0d587f.png">
<img width="1016" alt="Skjermbilde 2021-01-19 kl  19 43 39" src="https://user-images.githubusercontent.com/1568252/105079748-248ddf00-5a90-11eb-899f-246ae9cc4981.png">
<img width="887" alt="Skjermbilde 2021-01-19 kl  19 43 49" src="https://user-images.githubusercontent.com/1568252/105079761-28b9fc80-5a90-11eb-833d-7b6990b6f80d.png">

wenche:
1. impression: 
I like the flexibility it gains with placement of the props table and also that is generates unique urls based on each story, while we can place them where we want inside the text. We can also design and develop a few components for a better overview and structure of the page. E.g. an Figma link component, a better do's and don't component and so on.

As long as we only provides React component, I think it's easier for the end user to get an overview of the component by mixing text and live previews. But it might be harder to edit for non-developers as you don't want to accidentally break a code block. I might also provide a better overall impression of EDS since design and actual components  will feel more aligned.

I find it more error prone and harder to work with mdx instead of js. The tooling is lacking behind (automatic code formatting, ts support ++) and if something fails, it’s more difficult to debug. 

The pages will be quite long, but maybe we should remove some text?

mimarz:
Great work!👍  @wenche 

Maybe we could also test the [figma-addon](https://medium.com/storybookjs/storybook-figma-addon-7a399f094a0a) for storybook?

vnys:
[...]

PR #1076, opened 2021-01-22
👷 Move storybook, use Github actions and AZ Storage

### LINKED ISSUE #1010: Setup github actions for storybook releases
- [x] Move Storybook to core-react folder
- [x] Remove superfluous files
- [x] Deploy Storybook to Azure Storage Blob
- [x] Make sure storybook files don't leak into npm package

vnys:
- Installed fresh storybook under core-react
- Moved stories from apps/storybook-react
- Added placeholder for screens (app examples)

To make screen examples, uncomment the relevant line in .storybook/main.js

Tested build and deploy from my own branch, currently running at https://edsstoragestorybookdev.z16.web.core.windows.net/

~I just used the standard introduction page, so there are a couple of more things to move over, but the Github action works as expected, and so does building core-react – no story types in the build.~

There are a lot of files in this pr, as you would expect when moving all the stories. No need to spend time on the stories, they are unchanged for the most part except how we import core-react into the stories, which is now using a webpack alias, e.g. `import { Avatar, AvatarProps, Typography } from '@components'`

I have resisted the tempation to make changes to the stories as much as possible, as that’s outside the scope of this pr – but we need to do some spring cleaning.

resolves #1010


vnys:
Note to self: Remove svg icons in stories/assets

vnys:
…also added an action to deploy #1046 to https://edsstoragedev.z16.web.core.windows.net/

mimarz (changes requested):
* There seems to be a lot of new svgs under `/storybook/assets` that are not used. Part of the default storybook installation?

* We plan on hosting this from `edsstoragedev` no?

vnys:
> * There seems to be a lot of new svgs under `/storybook/assets` that are not used. Part of the default storybook installation?

That’s correct. I left them there because I thought we could use them ourselves in the documentation stories, but it’s probably better to use our own icons.

vnys:
> * We plan on hosting this from `edsstoragedev` no?

No. We need a dev and prod environment. You can only have one static site for each account, so the cleanest way to do that is to have one storage account for each environment.

mimarz:
> > * We plan on hosting this from `edsstoragedev` no?
> 
> No. We need a dev and prod environment. You can only have one static site for each account, so the cleanest way to do that is to have one storage account for each environment.

Yeah, maybe we should also clean up some of the web-services we have running that aren't used also 🙈

mimarz (approved):
LGTM 🇵🇱

vnys:
> Yeah, maybe we should also clean up some of the web-services we have running that aren't used also 🙈

Yes, we’ll have to go through that after the transition. Don’t think there are a lot of costs involved, but some spring cleaning is much needed.

PR #1085, opened 2021-01-26
Adjustments for elements inside button

### LINKED ISSUE #992: Fix the Button Icon to the sides
Button Icon is currently centered with the Button text, it should be fixed to the sides instead

mimarz:
resolves #992 

Tried solutions:
* Different CSS Grid variants using js to place elements in correct areas.
* Different Flex-variants

Not tried:

Having a `prefix/suffix` prop where we can pass a component for rendering icons and progress indicators would be a solution here, but it goes against our component-design philosophy of not passing components/elements as props. 

This was the best solution I could make work, using flexbox:

It works unless you have them full-width and stacked with different directions, its not optimal, but better that today atleast 🤷‍♂️
image

### DIFF
--- a/libraries/core-react/src/components/Button/Button.tsx
+++ b/libraries/core-react/src/components/Button/Button.tsx
@@ -42,13 +42,28 @@ const getToken = (variant: Variants, color: Colors): ButtonToken => {
   }
 }
 
-// display:grid; does not work on Webkit browser engine, so we have to wrap content in element where css-grid works
+const ButtonInnerText = styled.span`
+  text-align: center;
+`
+
 const ButtonInner = styled.span`
-  display: grid;
-  grid-gap: 8px;
-  grid-auto-flow: column;
+  display: flex;
   align-items: center;
   height: 100%;
+
+  > * {
+    flex: 1;
+  }
+
+  > svg:last-child {
+    flex: 0 24px;
+    margin-left: 8px;
+  }
+
+  > svg:first-child {
+    flex: 0 24px;
+    margin-right: 8px;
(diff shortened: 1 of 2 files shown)


wenche (approved):
So much work for so little code 😓

vnys:
How about something like this? https://codepen.io/vnys/pen/jOMgXNP

mimarz:
> How about something like this? https://codepen.io/vnys/pen/jOMgXNP Unless there’s something I haven’t thought of… But here you don’t need to know what the user sends in, just that the combinations are either icon + label, icon +. label + icon or label + icon.

Changed to use the css-grid approach you suggested. Had to add a `:only-child` and manipulate the `grid-template-columns` to only be one to work for `ghost_icon` variants of `Button`

pomfrida (approved):
LGTM 👍

vnys (changes requested):
I think we should avoid `left` and `right` since that doesn’t take languages like Arabic into account. Suggest replacing them with `start` and `end`.

vnys:
As just mentioned on Slack, I didn’t think of the Ghost Icon button – then flex works better after all. Another day, another codepen → https://codepen.io/vnys/pen/ZEBzvrK

mimarz:
> As just mentioned on Slack, I didn’t think of the Ghost Icon button – then flex works better after all. Another day, another codepen → https://codepen.io/vnys/pen/ZEBzvrK

Ok, will test it out 👍

vnys (approved):
Smashing pumkins 👊🏻 🎃

PR #1098, opened 2021-02-02
✨ Added size to Dots progress indicator

### LINKED ISSUE #417: `size` property for Progress components

No.

I'd like to be able to adjust the size of the `Progress` components, similar to how `Icon`s may be resized.

@vnys suggested overriding the style directly, which works well.
The downside, is that I would have to set both the width, and height to ensure the aspect ratio remains constant.

**Additional context**

*Some background*
When the dotted progress "bar" is the main component on a site (e.g. the data for the page is loading), it seemed rather small on a full desktop.

mimarz:
resolves #417 

After talking with @equinor/eds-ux we decided to a `size` option for `Dots` progress so that it may better suit loading states for pages or bigger widgets.


libraries/core-react/src/components/Progress/Dots/DotProgress.test.tsx
```diff
@@ -36,4 +36,14 @@ describe('DotProgress', () => {
...
+  it('has correct default width ', () => {
+    const { container } = render(<StyledProgress />)
+    const progress = container.firstChild
+    expect(progress).toHaveAttribute('width', '32')
+  })
+  it('has correct width when size is 64', () => {
+    const { container } = render(<StyledProgress size={64} />)
+    const progress = container.firstChild
+    expect(progress).toHaveAttribute('width', '64')
+  })
```
wenche, line 48: Assuming that we shouldn't have knowledge of implementation details, would it be better to test for height as well? 🤔

wenche (approved):

libraries/core-react/src/components/Progress/Dots/DotProgress.test.tsx
mimarz: Its 1/4 of height, which we could test.. but then again, its kinda implementation detail and kinda not 🤔 @wenche

wenche: Yeah... Maybe it could be more of a "has height set" approach to make sure it's a circle

mimarz: Updated tests @wenche

wenche: 👍

PR #1099, opened 2021-02-02
🐛 Fixed centering of content when its not text in Button

### LINKED ISSUE #1093: Fix alignment of button with only a number as text
The numbers inside buttons in `PaginationItem` are left aligned.
Skjermbilde 2021-02-02 kl  10 10 50

This probably happened in the button refactoring, and might be as simple as [this line of code
](https://github.com/equinor/design-system/blob/b8b11caa5f84c24fd737167d584898761f9df7cc/libraries/core-react/src/components/Button/Button.tsx#L214)

mimarz:
fixes #1093 
Inversed operator for when an child gets put in a element inside `Button` since react elements are `objects`,


wenche (approved):
LGTM 👍

PR #1103, opened 2021-02-04
Accordion - return <> if props hidden is true

### LINKED ISSUE #1102: Accordion - return <> if props hidden is true

We use the accordion component in a hierarchy context where the DOM gets very lagre if all elements are expanded. To reduce the size of DOM we start out with having the components collapsed. 
However the Accordion as by now does not take this into account and renders  the whole DOM structure, which affects the responsiveness of our web app significantly.

Accordion needs to check the hidden props and retur <> if true

The solution is fairly simple. Accordion should return <> if property hidden is true. 
My colleague Tor Christian will create a PR with a solution :-)

tfrausin:
resolves #1102

### DIFF
--- a/libraries/core-react/src/components/Accordion/AccordionPanel.tsx
+++ b/libraries/core-react/src/components/Accordion/AccordionPanel.tsx
@@ -38,6 +38,9 @@ const StyledAccordionPanel = styled.div.attrs(
 
 const AccordionPanel = forwardRef<HTMLDivElement, AccordionPanelProps>(
   function AccordionPanel({ id, headerId, hidden, children, ...props }, ref) {
+    if (hidden) {
+      return null
+    }
     return (
       <StyledAccordionPanel
         headerId={headerId}


libraries/core-react/src/components/Accordion/AccordionPanel.tsx
```diff
@@ -38,6 +38,9 @@ const StyledAccordionPanel = styled.div.attrs(
 
 const AccordionPanel = forwardRef<HTMLDivElement, AccordionPanelProps>(
   function AccordionPanel({ id, headerId, hidden, children, ...props }, ref) {
+    if (hidden) {
+      return null;
```
mimarz, line 42: Eslint check failed on `;` as we don't use semi-colons 🤓 Please remove

mimarz (changes requested):
Super duper! 😄 

Just a small linting error that needs to be fixed before we can approve.

mimarz (changes requested):
tenor (2)


Soo... having slept on this, I realise this would break the `Accordion` component in some use-cases 😴🕵️

Having the content not being rendered/mounted when closed would break SEO and server-side React rendering for this component. We still need the default behaviour to be that content is mounted as is behaviour today. 

We do realise however that there is a need for content not being rendered when closed, so we suggest introducing a optional `unmountOnCollapse` boolean property on `<Accordion>`, that lets users opt-in on content not being mounted when closed. If you could also write some testes to verify that its working as intended that would be great :) 

TDLR;
- [ ] We need a `unmountOnCollapse` boolean property on `<Accordion>`, if `true` and hidden=`true`, `AccordionPanel` return `null`. 
- [ ] Write tests to verify that everything works when the prop is either set or not.

cbuv:
I was hoping AccordionHeader's toggleExpanded callback could be used to solve this issue and similar ones by storing the expanded state  outside the eds accordian components. 

But, I tested the toggleExpanded, and it does not do anything. 
What is the status of toggleExpanded? @vnys

mimarz:
@tfrausin We would love to have this feature in our next release, any idea when you can do these changes? Do you need any assistance? :)

tfrausin:
@mimarz After discussing this within the team, we have unfortunately decided that we do not have the capacity to prioritize this to the extent necessary to fulfill your requirements. Hence, we're withdrawing our pull request.

PR #1104, opened 2021-02-04
Fix menu height

### LINKED ISSUE #1087: Menu item has wrong height
Menu items have a height of 61px in the storybook, correct height according to Figma should be 48px.

Skjermbilde 2021-01-26 kl  14 58 13

wenche:
Resolve #1087  

Seems like this is a classic user error. 🤓  I updated the markup used in storybook to something that makes more sense, and is aligned with Figma. As a magical side effect, the height is now 48px. 🥳


mimarz (approved):
magic

Issue #1107, opened 2021-02-09
Label as separate React component


mimarz:
notes #1015

wenche:
Add an example in Storybook about aria-label for TextField without visible label

wenche:
One obvious shortcoming of the current Label component @mimarz  is that you can't use it to wrap an input component.
```
<label>
  <input>
</label>
```
I don't know if we should fix this as well or delay this? 🤔

mimarz:
> One obvious shortcoming of the current Label component @mimarz is that you can't use it to wrap an input component.
> 
> ```
> <label>
>   <input>
> </label>
> ```
> 
> I don't know if we should fix this as well or delay this? 🤔

Don't we put children inside a `label` element inside `Label`? 😅

wenche:
Nope, only the label text and meta

wenche:
I think my vote goes to fix it (if it's a low hanging fruit). If we are supposed to reuse the `<Label>` component for selection controls we're gonna need it anyway 🤷‍♀️

wenche:
It was not a low hanging fruit, more like in the middle of the tree or something 😓  Beyond adding the `children` we need to rethink the styling. My suggestion would be to keep it like this for now, and attach the problem more holistically when we do the selection controls.

Issue #308, opened 2020-04-17
Update TextField component


mimarz:
**Description**

Update `TextField` with missing states

**Documentation**

_Edit the following links to point to relevant information_

- [The storefront](https://eds.equinor.com/components/text-fields/)
- [Figma](https://www.figma.com/file/0bGXR2sCwMVSDNyyzu5BXrO5/UI%E2%80%94User-Interface?node-id=1%3A26)
- [WAI-ARIA Authoring Practices 1.1](https://www.w3.org/TR/wai-aria-practices-1.1/#)

**General guidelines**

See [COMPONENT_GUIDELINES.md](https://github.com/equinor/design-system/blob/develop/COMPONENT_GUIDELINES.md) for general guidelines like browser support and common implementation details

**Sub tasks**

- [ ] Read only state

_Relevant creation issue (see #146 )_

mimarz:
replaced by #1162

PR #1110, opened 2021-02-09
Refactor TextField

### LINKED ISSUE #1108: Input as separate React component
notes #1015

wenche:
Resolve #1108  and #1107 

We need to eat this elephant bit by bit 😓  This PR solves the first step toward DataTable and more flexible form fields by exposing the `Label` and `Input` component used internally by `Select` and `TextField` into externally exposed components.


wenche:
I tried a new approach by clarifying the accessibility stuff for the Label component, don't know if it's a good idea or not
Skjermbilde 2021-02-09 kl  16 32 53

wenche:
It would have been nice if the color tokens for helper text + icon and icon inside the input field were the same, but they are only like 70% the same. So we'll need some extra code to take this into account
Skjermbilde 2021-02-12 kl  08 10 23

mimarz (approved):
Good work! 👍

This has a lot of changes, a good step in the right direction for a better world! <3 We have some follow-up issues that needs to be solved before these changes can be released (which we need to prioritise)

Follow up issues: #1131, #1130, #1111 and #308

PR #1113, opened 2021-02-10
 ✨ Refactor Menu, Popover and Tooltip to use Popper (Old Spike PR)

### LINKED ISSUE #673: Conform popup logic
* Menu
* Popover
* Tooltip

These componenets are all using seperate implementations of logic for attaching and popping up on an anchor element. 

Find a way to conform this logic using a popover hook.

Notes:
* https://reactjs.org/docs/portals.html
* https://medium.com/better-programming/create-a-tooltip-component-using-react-hooks-cb9fd00a2ff0
* https://www.npmjs.com/package/react-popover-hook
* https://github.com/popperjs/popper-core

Features:
* Requested feature for popup placement #952
* Related feature #984 that could be fixed

### LINKED ISSUE #844: 🐛 Fix weird behaviour on Menu controlled docs
Menu placement is not working as expected in controlled docs. 

**May** be fixed by #673 but created this issue for transparency

### LINKED ISSUE #952: Menu: Choose position of where menuitem drops
image

This, instead of this for example

image

### LINKED ISSUE #984: Tooltip support on table cells
We have examples of usages with Tooltip on Table cells: https://www.figma.com/file/luNobpDIWTuQ9rhyovGLBI/Desktop?node-id=624%3A0

The current implementation of tooltip wraps the "anchor element" with a `div` to use as an anchor point. This approach does not work in tables where you can't have `div` as children of at `tr`

Either we fix this internally within the table cell somehow or expand tooltip component to have a `anchoEl` for anchoring (like in `Menu`)

Notes: This needs to be looked on together with issue #673

### LINKED ISSUE #1112: Spike popperjs as popup component
In regards to conforming on popup logic, we need to test existing libraries. #673

- [x] Test implementation in React
- [x] Test on `Tooltip`
- [x] What props do we need to introducs to support popperjs features
- [x] Test if it works on #984
- [x] Do we need seperate anchor element or can an element be passed to it

pomfrida:
Resolves #1112, resolves #673, resolves #952, resolves #984, resolves #844

Spiking [react-popper](https://popper.js.org/react-popper/), a lightweight placement hook for tooltips / popovers / menus / etc
So this PR started as a spike and evolved to become a giant change PR instead.. Sorry about that! 
Popper was pretty straight forward and easy to use, and it's helping us by a lot by outsourcing all the placement calculations that gave us a lot of headaches! It's lightweight and I think it gives us a lot of value. 

### Some quick notes 
Why **not** use CSS or naive JS tooltip:

- 	 Clipping in regard to viewport 
- 	 No automatic flipping possible
- 	Interactive HTML difficult to create without UX issues
- 	No virtual positioning (mouse cursor, context menu)
- 	Lack of extensibility
- 	Scrolling containers (hard to ensure the tooltip stays with the reference while scrolling)
- 	Often require tooltips outside of its original DOM
- 	Often lack advanced configurability
- 	Relatively large in size
- 	Runtime performance issues

**Why use Popper:**

-  Popper solves all the problems listed above
- 	Lightweight, ~7kB regular popper and ~3 kB library if we use popper-lite @popperjs/core/lib/popper-lite'
- 	Used by Bootstrap, Foundation, Material UI
- 	35M/monthly downloads
- 	Used by 1.4m users on Github
- 	Active Github community
- 	React wrapper react-popper https://popper.js.org/react-popper/v2/
- 	Type definitions


pomfrida:
### Some notes along the way: 

- 	Created a hook for reusability and less code
- 	Popper automatically repositions popovers when users scroll the page, this is a neat feature!
- 	 Simplified our components by deprecating the old popover placement logic
- 	 State from inside the components are moved to outside control
- 	 Tooltips are now supported inside Table Cells! 
- 	 The only extra `prop` we needed to add to our components was the anchor ref (`MutableRefObject` type)
- 	 Placement prop is optional in Popper, because Popper uses `auto` placement (calculates the placement best for the window), but it's added to give the users more options (i.e Menu didn't have this prop before)

### Some roadblocks along the way:
- 	Had to rewrite the components to fit Popper properly
- 	 Popper didn't support `SVG` arrows (Old logic used arrows from figma), so we had to make a `HTML` arrow instead (arrow placement calculations by Popper needed, Popper only supported HTML elements). These arrows aren't 100% identical to Figma design at the moment.
- 	 Arrow positioning is not 100% consistent in all the different placements (Tooltip, Popover)
- 	 Popover arrow had a `drop-shadow` filter to make the arrow more visible on a white background, it doesn't work as expected with Popper. Will need to find a way this shadow doesn't overlap the actual popover
- Had to do a workaround by wrapping divs around disabled elements that doesn't trigger mouse events (Tooltip example)
- Need to rewrite the tests to fit the refactoring

### Breaking changes, need to inform users before release: 

- the deprecation of `PopoverAnchor` component
- change in Tooltip, Menu and Popover that users need to add refs to their anchor elements, and to send it as `anchorEl` prop in the respective components
- changed old `placement` values (in Tooltip and Popover) to fit Popper, users already using this prop will have to change the value
- added `forwardRef` to the Cell consumer, was needed to forward refs to TableHeaderCell and TableDataCell correctly

libraries/core-react/src/hooks/useOutsideClick.ts
```diff
@@ -2,10 +2,15 @@ import { useEffect, MutableRefObject } from 'react'
...
   ref: MutableRefObject<HTMLElement>,
+  anchorEl: HTMLElement,
   callback: () => void,
 ): void => {
   const handleClick = (e: MouseEvent) => {
-    if (ref.current && !ref.current.contains(e.target as Node)) {
+    if (
+      ref.current &&
+      !ref.current.contains(e.target as Node) &&
+      !anchorEl.contains(e.target as Node)
```
pomfrida, line 12: ` !anchorEl.contains(e.target as Node)` was added to avoid close on first render

libraries/core-react/src/components/Popover/Popover.tsx
```diff
@@ -2,100 +2,166 @@ import * as React from 'react'
...
+import { usePopper, useOutsideClick } from '@hooks'
+import { popover as tokens } from './Popover.tokens'
 
-const Container = styled.div`
-  position: relative;
-  display: inline-flex;
-  justify-content: center;
+type StyledPopoverProps = Pick<PopoverProps, 'open'>
+
+const StyledPopover = styled(Card)<StyledPopoverProps>`
```
pomfrida, line 22: To do: Do not use `Card` here 🚫

pomfrida:
Tests finally complete and successful ✅  
Just a couple of things before this is ready for review:
- check why `eds-core-react` build fails
- Popover should use `Paper` instead of `Card` 
- maybe a few more tests on `Tooltip`

pomfrida:
Build finally resolved ✅  
@equinor/eds-dev You can look over the PR while I rewrite `Popover` to use `Paper` instead of `Card`..
Also let me know if you have any suggestions to add more tests. Tooltip tests are kind of scarce. Testing `Popper` positioning is tricky because it's not hardcoded anywhere (ref. old tests)

mimarz:
We can't review it when its still a draft @pomfrida :)

pomfrida:
You can't or won't? @mimarz 😅

pomfrida:
🤔  Working with Menu stories and was just thinking if we really need `MenuProvider` logic at the top of the component, or if we could move it. Would be really neat to utilize Storybook's dynamic control functionalities and expose the `Menu` props.. `MenuProvider` kind of ruins this possibility the way it's used at the moment. 

`Popover` and `Tooltip` stories both have dynamic story where users can play around with the `placement` prop. We can add an example in `Menu` with hard coded examples of all the placement values but it's not as neat.

mimarz:
> 🤔 Working with Menu stories and was just thinking if we really need `MenuProvider` logic at the top of the component, or if we could move it. Would be really neat to utilize Storybook's dynamic control functionalities and expose the `Menu` props.. `MenuProvider` kind of ruins this possibility the way it's used at the moment.
> 
> `Popover` and `Tooltip` stories both have dynamic story where users can play around with the `placement` prop. We can add an example in `Menu` with hard coded examples of all the placement values but it's not as neat.

`MenuProvider` is just a provider for the context, and it shouldn't be in the way of dynamic control functionality, so not sure how its in the way?

pomfrida:
Updated `Menu` story with an example with `SingleSelect` for users to play around with the new `placement` prop. This is just a temporary workaround because Storybook doesn't support controlled subcomponents. Look into issue https://github.com/equinor/design-system/issues/1149 at some time later to utilize this functionality.

libraries/core-react/src/components/Table/Cell/index.tsx
```diff
@@ -16,19 +16,27 @@ export type CellProps = {
   | TdHTMLAttributes<HTMLTableDataCellElement>
   | ThHTMLAttributes<HTMLTableHeaderCellElement>
 )
+export const Cell = forwardRef<
+  HTMLTableDataCellElement | HTMLTableHeaderCellElement,
+  CellProps
+>(function Cell({ className, ...rest }, ref) {
+  const props = {
+    ...rest,
+    className,
```
mimarz, line 25: This className should not need to be deconstructed... Suggest just passing props as before and set `ref` on components

wenche:
I would suggest to split the single story "Examples" into four stories in the Menu stories. Easier to find the relevant code in the View code and it also becomes more like how we do it elsewhere in the storybook @pomfrida 

Skjermbilde 2021-03-09 kl  09 57 04

libraries/core-react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,175 +1,118 @@
...
 `
 
-const StyledTooltip = styled.div`
+const StyledTooltip = styled.div<Pick<TooltipProps, 'open'>>`
   ${typographyTemplate(tokens.typography)}
   ${spacingsTemplate(tokens.spacings)}
+  z-index: 350;
   background: ${tokens.background};
-  fill: #333;
+  fill: ${tokens.background};
```
mimarz, line 29: again, why is fill set to background color? 🙈

libraries/core-react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,175 +1,118 @@
...
+    transform: rotate(45deg);
+    background: #333;
   }
 `
 
-const StyledTooltip = styled.div`
+const StyledTooltip = styled.div<Pick<TooltipProps, 'open'>>`
   ${typographyTemplate(tokens.typography)}
   ${spacingsTemplate(tokens.spacings)}
+  z-index: 350;
```
mimarz, line 27: Doesn't popper handle z-indexes?

libraries/core-react/src/hooks/useOutsideClick.ts
mimarz: This seems as a weird behaviour... What is causing this, because we aren't clicking anywhere on first render...

libraries/core-react/src/hooks/useOutsideClick.ts
```diff
@@ -2,10 +2,16 @@ import { useEffect, MutableRefObject } from 'react'
 
 export const useOutsideClick = (
   ref: MutableRefObject<HTMLElement>,
+  anchorEl: HTMLElement,
```
mimarz, line 5: Do we really need this? I would like to avoid this complexity as hooks are intended to solve simple problems. Can this be solved differently? Maybe inside the individual components?

libraries/core-react/src/hooks/usePopper.ts
```diff
@@ -0,0 +1,53 @@
...
+    | 'top-end'
+    | 'bottom'
+    | 'bottom-start'
+    | 'bottom-end'
+    | 'right'
+    | 'right-start'
+    | 'right-end'
+    | 'left'
+    | 'left-start'
+    | 'left-end',
```
mimarz, line 22: Suggest making a seperate type for `Placements` and export this for use in the individual components instead of duplicating string literal.

libraries/core-react/src/utils/test-utils.tsx
```diff
@@ -1,17 +1,21 @@
 import * as React from 'react'
-import { ElementType, ReactElement, ReactNode } from 'react'
+import { ElementType, ReactElement, ReactNode, MutableRefObject } from 'react'
```
mimarz, line 2: MutableRefObject doesn't seem to be used?

wenche:
There's something missing in the arrow styling of popover, you can see the shadow on the arrow is wrong 😄 
Skjermbilde 2021-03-09 kl  10 19 18

libraries/core-react/src/utils/test-utils.tsx
```diff
@@ -1,17 +1,21 @@
...
   }, [anchorRef])
 
   const updatedChildren = React.Children.map(children, (child) =>
-    React.cloneElement(child as ReactElement, { anchorEl }),
+    React.cloneElement(child as ReactElement, {
+      anchorEl,
+    }),
   ) as ReactNode
 
   return (
```
mimarz, line 21: Good job on making this a shared testing util 👍

I suggest we some more changes, as its not very evident in testing files what `test-utils` does...

We can make a `test` folder under `utils` and add a file called for example `withAnchorEl`. 

This way its more explicit when using the custom render in test files such as;
```typescript
import { render, cleanup, screen, fireEvent } from '@utils/test/withAnchorEl'
```
instead of;
```typescript
import { render, cleanup, screen, fireEvent } from '../../utils/test-utils'
```

mimarz (changes requested):
This is a big big PR so I have noted down some oddities and suggestions so far. 

I have not looked at the stories yet, but make it seperate so its easier todo after suggested changes

wenche:
We should probably assess the use of something that traps focus. I don't think any of the examples for Menu or Popover (or Dialog for that matter 😄  ) actually are accessible. 

We probably don't want to add this to code? But I think we should at least address it in storybook, e.g. with a story that highlight this issue. I've used [focus-trap-react](https://github.com/focus-trap/focus-trap-react) in previous projects, that could be one option.

wenche:
I suggest to add some dummy text like "`Some tooltip for `Africa" to clarify that tooltips in table cells are not supposed to repeat the content of the cell.
Skjermbilde 2021-03-09 kl  10 35 05

libraries/core-react/package.json
```diff
@@ -95,7 +95,10 @@
   "dependencies": {
     "@equinor/eds-icons": "workspace:*",
     "@equinor/eds-tokens": "workspace:*",
-    "downshift": "^6.0.6"
+    "downshift": "^6.0.6",
+    "@popperjs/core": "2.9.0",
+    "react-fast-compare": "3.2.0",
```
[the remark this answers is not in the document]
pomfrida: https://github.com/popperjs/react-popper/issues/345

I know it's weird but had to install it to fix some build errors (https://github.com/equinor/design-system/runs/2029484993?check_suite_focus=true#logs)

libraries/core-react/src/components/Tooltip/Tooltip.tsx
pomfrida: This was from before popper, when the the arrow was an svg

libraries/core-react/src/hooks/useOutsideClick.ts
pomfrida: I sat with this problem for a while, and had to add this check because on first _click_ (to open menu) it closed before it had the chance to open. Had to click the button twice to open the menu after first render. 

I think it's because Popper uses anchorEl to calculate the positioning.

wenche:
Is it possible to position the `tooltip` on top center instead? It doesn't behave 100% correct in this example 🤔 
Skjermbilde 2021-03-09 kl  11 19 18

libraries/core-react/src/components/Menu/Menu.context.tsx
```diff
@@ -3,104 +3,39 @@ import { useState, useContext, ReactNode, MouseEvent } from 'react'
...
-}
-
-const calculatePosition = (
-  { left, top, height }: ClientRect,
-  offset: number,
-) => ({
-  top: top + height + offset,
-  left: left,
-})
-
```
wenche, line 65: This feels so good! 🔥

libraries/core-react/src/components/Popover/Popover.tsx
```diff
@@ -1,101 +1,162 @@
...
+  &[data-popper-placement^='bottom'] > .arrow {
+    top: -4px;
+  }
+
+  &[data-popper-placement^='left'] > .arrow {
+    right: -4px;
+  }
+
+  &[data-popper-placement^='right'] > .arrow {
+    left: -4px;
```
wenche, line 52: Where does this number '4px' come from, can it be some `spacings` token instead?

libraries/core-react/src/components/Popover/Popover.tsx
```diff
@@ -1,101 +1,162 @@
...
 
-const Anchor = styled.div`
-  &:focus {
-    outline: none;
+const StyledCloseButton = styled(Button)`
+  position: absolute;
+  top: 8px;
+  right: 16px;
+  height: 32px;
+  width: 32px;
```
wenche, line 61: Same as above, can we use token here to clarify what this is? 🤔  It would be a lot easier to conform on `ComponentType` later on

libraries/core-react/stories/components/Popover.stories.tsx
```diff
@@ -1,8 +1,11 @@
-import React from 'react'
+import React, { useRef, useState } from 'react'
 import styled from 'styled-components'
 import { Typography, Button, Popover, PopoverProps, Card } from '@components'
 import { Meta, Story } from '@storybook/react'
 
+const { Title, Content } = Popover
```
wenche, line 6: We should avoid the use of object destructing and conform on writing the stories by using` Popover.Title` as in other stories. 👍  In this way it's easy to see what this sub component is in "View source" in the story.

libraries/core-react/stories/components/Tooltip.stories.tsx
```diff
@@ -34,104 +21,283 @@ const TextWrapper = styled.div`
...
+    origin: 'Paraguay',
+    price: 1.9,
+  },
+  {
+    number: '89-012',
+    description: 'Pomegranate',
+    origin: 'Persia',
+    price: 4.5,
+  },
+]
```
wenche, line 188: Perhaps this table data could be shared between several stories in a data folder?

wenche (changes requested):

libraries/core-react/src/hooks/useOutsideClick.ts
pomfrida: Not sure if I get what you mean by adding a hook for anchorEl?

libraries/core-react/src/components/Tooltip/Tooltip.tsx
pomfrida: We should still follow our own z-index hierarchy here https://eds-storybook-react.azurewebsites.net/?path=/story/documentation-z-index--z-index

pomfrida:
Solved #1134 while I was working with `Breadcrumbs`, just thought this PR was too small 🤡

libraries/core-react/stories/components/hooks/toCellValues.ts
```diff
@@ -0,0 +1,10 @@
+import { Data, Column } from '../helpers/data'
+
+export const toCellValues = (data: Data[], columns: Column[]): string[][] =>
+  data.map((item) =>
+    columns.map((column) =>
+      typeof item[column.accessor] !== 'undefined'
+        ? (item[column.accessor] as string)
+        : '',
+    ),
+  )
```
mimarz, line 10: Why was this moved to `hooks` ? Its not a hook and its very specific to made up example of how a `Table`  _could_ be done.

libraries/core-react/src/components/Breadcrumbs/Breadcrumb.tsx
```diff
@@ -38,10 +38,20 @@ type BreadcrumbProps = {
   onClick?: (e: MouseEvent | KeyboardEvent) => void
   /** Children is breadcrumb text */
   children: string
-} & HTMLAttributes<HTMLDivElement>
+} & React.AnchorHTMLAttributes<HTMLAnchorElement>
 
-export const Breadcrumb = forwardRef<HTMLDivElement, BreadcrumbProps>(
+export const Breadcrumb = forwardRef<HTMLAnchorElement, BreadcrumbProps>(
   function Breadcrumb({ children, maxWidth, ...other }, ref) {
+    const [openState, setOpenState] = React.useState(false)
```
mimarz, line 45: Suggest renaming this property for what it controls, which is the `Tooltip`, so maybe `openTooltip` or something like that?

libraries/core-react/stories/components/hooks/toCellValues.ts
pomfrida: I needed it for `Tooltip` in Table Cells example, but will move it to `helpers` instead  👍

mimarz (changes requested):

libraries/core-react/src/components/Menu/index.tsx
```diff
@@ -14,7 +13,7 @@ type MenuType = typeof BaseMenu & {
   Section: typeof MenuSection
 }
 
-const MenuWrapper: FC<MenuProps> = ({ children, ...rest }) => (
+const MenuWrapper = ({ children, ...rest }: MenuProps): JSX.Element => (
```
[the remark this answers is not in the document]
pomfrida: No :/

mimarz:
Snippets from our testing earlier @pomfrida 

hook
```typescript
import { useEffect, MutableRefObject } from 'react'

export const useOutsideClick = (
  ref: MutableRefObject<HTMLElement>,
  callback: (e: MouseEvent) => void,
): void => {
  const handleClick = (e: MouseEvent) => {
    if (ref.current && !ref.current.contains(e.target as Node)) {
      callback(e)
    }
  }

  useEffect(() => {
    document.addEventListener('click', handleClick)

    return () => {
      document.removeEventListener('click', handleClick)
    }
  })
}

```

usage
```typescript
useOutsideClick(popperRef, (e) => {
  if (open && onClose !== null && !anchorEl.contains(e.target as Node)) {
    onClose()
  }
})
```

libraries/core-react/stories/components/Tooltip.stories.tsx
```diff
@@ -53,14 +53,16 @@ export const Default: Story<TooltipProps> = (args) => {
...
+        onBlur={handleClose}
       >
         Hover me!
       </Button>
       <Tooltip
-        {...args}
         open={openState}
         id="tooltip"
         anchorEl={referenceElement.current}
+        {...args}
```
mimarz: 🙈

mimarz (approved):
LGTM 👍

Good job on this mothertrucker! 

nod yes

libraries/core-react/stories/components/Tooltip.stories.tsx
pomfrida: Haha played with some things to see if it could fix the weird code preview in Storybook

libraries/core-react/src/components/Menu/index.tsx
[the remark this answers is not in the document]
mimarz: This will be solved in #1164

wenche (approved):
Too much code for a quick overview 😅  Lot's of great work on popovers, I think it will be a lot easier to use these components now 👍

PR #1119, opened 2021-02-18
[WIP]: Spike CSS variables

### LINKED ISSUE #1118: Spike CSS variables
Spike the use of CSS variables for theming.

wenche:
Even if we are not supposed to merge this experimental stuffs into develop, we need a place to collect information and discuss the different approaches.

Resolve #1118


wenche:
Alright, I'm pretty confused 😓 

Things to do:
- How should we export` tokens.css` and `elements.css` from '@equinor/eds-tokens'
- How should the component in `@equinor/eds-core-react` access some global style variables as there no shared code at the moment. (We don't have the App.tsx thing in `@equinor/eds-core-react`)
- How can we support both plain CSS and styled-components/css-in-js

🤷‍♀️

wenche:
Better IDE support with the use of CSS variables for colors is a good thing
Skjermbilde 2021-02-18 kl  14 29 41

mimarz:
Its also possible to add a new property, say `cssVar`, in figma-broker that contains the css variable name, for example in color transform [here](https://github.com/equinor/design-system/blob/e980d235bf979316d381ae65e1566def075c5e97/apps/figma-broker/files/design-tokens/color.js#L48). Then all we would have to do is change the deconstruct we have in most token files from `rgb` to `cssVar`. 

This would be easiest and quickest in terms of conversion straight up, but we don't get syntax help from IDE tho. But maybe it could be a by step job we need to do... 🤔

wenche:
- When colors from tokens, we provide both hsla, rgba and hex. Which color format should we use for variables?

vnys:
> * Which color format should we use for variables?

I think we should use hsla, it’s the only human readable colour format of those three alternatives, and it’s supported in all major browsers.

vnys:
> * How should we export `tokens.css` and `elements.css` from '@equinor/eds-tokens'

Good question. If we replace all the token values with css-variables, then `tokens.css` will be required for the components to function properly – but then overriding those variables becomes a hassle. Which is partly why I suggested this morning that we use the tokens as they are today as a fallback, and then use an empty variable as the first choice in a `var()` function.

```js
color: {
  default: `var(--does-not-exist-yet, ${typography.color})`,
  disabled: `var(--neither-does-this, ${disabledColor})`,
  activated: `var(--or-this, ${activatedColor})`,
},
```

Then if `tokens.css` is not available, you get light mode like now. And if it exists, because you imported it or made your own 😱 then the css variables take precedence.

wenche:
A slightly different approach. The idea is to use a fallback inside the actual React component itself. That will give the light theme out of the box by using the tokens in the same way as we do today, but you can override by proving a CSS file with variables.

With variables (salmon to see the different better)
Skjermbilde 2021-02-19 kl  10 53 35
Skjermbilde 2021-02-19 kl  10 54 32

And then by removing the global CSS file (in the Storybook config as this is probably something you'll do at app level)
Skjermbilde 2021-02-19 kl  10 54 46
Skjermbilde 2021-02-19 kl  10 56 43
Skjermbilde 2021-02-19 kl  10 58 09
Skjermbilde 2021-02-19 kl  10 58 44

vnys:
Looks, good – that’s pretty much how I imagined it too with a fallback to the javascript tokens. But should we add a local variable too based on the same model that’s used in [pattern-fly elements](https://medium.com/patternfly-elements/patternfly-elements-theming-hooks-using-css-variables-in-a-component-4e8c481f86b6). That would mean something like `color: var(--eds-AccordionHeader-disabledColor, var(--eds-infographicColor-01, ${headerColor.disabled}));`. Then the default is a local hook that has no value, but it’s there if we need it, the next fallback is a global variable (which of course does not have a colour in it’s name 😉), and then finally a fallback to whatever the value is in the tokens – which is `hsla(0, 0%, 75%, 1)` in this case.

vnys:
We use PascalCasing for the components, so I think it makes sense to do that in the css as well. [SuitCSS](https://github.com/suitcss) uses the following naming convention, I think we should do something similiar (I simplified things a bit in [the example above](https://github.com/equinor/design-system/pull/1119#issuecomment-782005598)), but still prefixing the variables with `--eds-` to prevent conflicts with user defined variables. Then we have a reckognizable pattern, so it’s easy to see what’s a component and what’s a decendant or modifier.

**Component variables**

```css
:root {
  --ComponentName-backgroundColor
  --ComponentName-descendant-backgroundColor
  --ComponentName--modifier-backgroundColor
  --ComponentName-onHover-backgroundColor
  --ComponentName-descendant-onHover-backgroundColor
}
```

**Theme variables**

```css
:root {
  --fontSize: 16px;
  --fontFamily: sans-serif;
  --lineHeight: 1.4;

  --spaceSmall: 10px;
  --spaceMedium: 15px;
  --spaceLarge: 20px;
}
```

SvSven:
From an end-user's perspective I think this should work pretty well. Makes it library agnostic
Might get a bit confusing if you end up doing something like:

```css
:root {
  --eds_infographic_substitute__blue_ocean: rgba(255, 0, 0, 1);
}
```
But honestly it's your own fault for overwriting it at that point 😬 

I assume the idea would be to declare said variables globally on `:root` rather than on a per component basis?

wenche:
Yes @SvSven, at least the initial approach as (we believe) it makes it easier. But we discussed this, and the downside is ofc that the components are less standalone and encapsulated. But it should be possible to change where the variables are loaded later on.
Could be something like 
`<Wrapper style={{css variables from somewhere here scoped to this component}} />`

vnys:
> Might get a bit confusing if you end up doing something like

Yes, but the colour pallette should be fixed. So you wouldn’t change the value of that colour variable, you would change the value of the variable that uses that colour variable.

wenche:
Yes, something like this variable `--eds_interactive_secondary__highlight` would be a better example, but I needed a variable with a color that was easy to find, hence salmon

Issue #1120, opened 2021-02-18
AccordionHeader toggleExpanded is broken


cbuv:
**Describe the bug**
AccordionHeader toggleExpanded does not work when it's used with Accordion and AccordionItem. 

**Steps to reproduce the bug**

1. see codesandbox (paste example from storybook in codesandbox)

**Expected behavior**
click on AccordionHeader should fire callback

**Specifications**

- Version: 
- Browser: 
- OS: 

**Additional context**
https://codesandbox.io/s/elastic-montalcini-bwg36
#1102

cbuv:
I've looked more into this and possible solution is to don't mix the internal state with external props. 
Rename public toggleExpanded to onToggle (optional), and call this method in AccordionHeader onClick and onKeyDown should resolve this issue.

pomfrida:
Thank you @cbuv for reaching out to us. We will take a look. But it looks like you have a solution here, if you don't mind opening a PR with your solution we would very much appreciate the contribution. 🥇

PR #1121, opened 2021-02-19
Fix AccordionHeader toggleExpanded

cbuv:
**problem:** 
AccordionItem shadows external toggleExpanded. The internal and external props are mixed.
A common use case is to defer  render of components that are not visible to the user. I many of the inhouse application do have a lot of data and hidden components may be expensive to render. 

**solution:** 
add a new onToggle prop to support toggleExpanded when it's wrapped in AccordionItem. 

#1120

### DIFF
--- a/libraries/core-react/src/components/Accordion/Accordion.test.tsx
+++ b/libraries/core-react/src/components/Accordion/Accordion.test.tsx
@@ -81,6 +81,17 @@ describe('Accordion', () => {
     fireEvent.click(header)
     expect(header).toHaveAttribute('aria-expanded', 'true')
   })
+  it('triggers onToggle callback', () => {
+    const mockOnToggle = jest.fn()
+    render(
+      <AccordionItem isExpanded>
+        <AccordionHeader onToggle={mockOnToggle}>Summary 1</AccordionHeader>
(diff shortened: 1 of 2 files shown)


cbuv:
what's the secret to make the react-tests build pass?  
all tests are passing.

wenche:
> what's the secret to make the react-tests build pass?
> all tests are passing.

Seems like an eslint issue, do you run eslint and prettier locally? @cbuv 

Skjermbilde 2021-02-19 kl  13 53 24

mimarz:
> what's the secret to make the react-tests build pass?
> all tests are passing.

Its the magic comma! 🙈

Don't worry, we are switching to github actions at this exact moment so it'll be more transparent in the future :)

wenche:
We will add Husky for precommit hooks in another project we are involved in - could probably be useful for EDS as well. Sorry about that!

cbuv:
thanks, I actually did pnpm run lint:all first, but didnt notice the 401. and then I couldn't access the logs. 
And I'm used to prettier default config. 🤣

mimarz (approved):
LGTM 👍

Thanks for your contribution in making EDS better! 👏

Issue #1129, opened 2021-02-23
Add support for CSS variables in React components


wenche:
This is a step on the way to provide theming in EDS. 

- [ ] Add a CSS variable with the current naming convention for all color definition inside components. Use current solution with js tokens as fallback.
- [ ] Suggest a new naming convention
- [ ] Rename variables to the new convention
- [ ] Add component variables to each color definition. This will have no value from EDS, but will provide users with a way to override variables on per component level, if needed.

Issue #801, opened 2020-10-30
Conform on prop types descriptions & names


mimarz:
When using our React components, prop types, names, descriptions should sound the same and be informative.

For example, we have a wide variety of `disabled` being described in different ways across components.

akvam:
@mimarz is this US still relevant? Any part of it that is already done?

mimarz:
This is still very much relevant. It was a blocker for some cleaning up in #2336.

Parts of this need to be also fixed in Figma. Hopefully it will be done with #2355 :)

PR #1132, opened 2021-02-24
Some cleanup for CSS variables

wenche:
Part of #1129 

Edit: The initial approach of this PR was to rewrite the React components to use CSS vars. it quickly became evident that it was too difficult. The approach was changed to rewrite the tokens of the components, but that was cumbersome as well. The final approach was to add CSS color variables inside the tokens, see #1137 . 

This PR features some tokens cleanup to make it easier to add CSS vars and a test file for CSS vars used for testing inside Storybook

### DIFF
--- a/libraries/core-react/src/components/Accordion/AccordionHeader.tsx
+++ b/libraries/core-react/src/components/Accordion/AccordionHeader.tsx
@@ -8,6 +8,7 @@ import { Icon } from '../Icon'
 import { AccordionHeaderTitle } from './AccordionHeaderTitle'
 import { accordion as tokens } from './Accordion.tokens'
 import type { AccordionProps } from './Accordion.types'
+// import { bordersTemplate } from '@utils'
 
 Icon.add({ chevron_down, chevron_up })
 
@@ -58,7 +59,10 @@ const StyledAccordionHeader = styled.div.attrs<StyledAccordionHeaderProps>(
     borderBottom: border,
     borderLeft: border,
     boxSizing: 'border-box',
-    color: (disabled && headerColor.disabled) || headerColor.default,
+    color:
+      (disabled &&
+        `var(--eds_interactive_disabled__text, ${headerColor.disabled})`) ||
+      `var(--eds_text_static_icons__default, ${headerColor.default})`,
     outline: 'none',
     '&[data-focus-visible-added]:focus': {
       outline,
--- a/libraries/core-react/src/components/Accordion/AccordionHeader.tsx
+++ b/libraries/core-react/src/components/Accordion/AccordionHeader.tsx
@@ -8,7 +8,6 @@ import { Icon } from '../Icon'
 import { AccordionHeaderTitle } from './AccordionHeaderTitle'
 import { accordion as tokens } from './Accordion.tokens'
 import type { AccordionProps } from './Accordion.types'
-// import { bordersTemplate } from '@utils'
 
 Icon.add({ chevron_down, chevron_up })
 
--- a/libraries/core-react/src/components/Accordion/AccordionHeader.tsx
+++ b/libraries/core-react/src/components/Accordion/AccordionHeader.tsx
@@ -48,6 +48,7 @@ const StyledAccordionHeader = styled.div.attrs<StyledAccordionHeaderProps>(
     fontFamily: typography.fontFamily,
     fontSize: typography.fontSize,
     fontWeight: typography.fontWeight as number,
+    backgroundColor: headerBackground.default,
     lineHeight: typography.lineHeight,
     textAlign: typography.textAlign as CSSObject['textAlign'],
     margin: 0,
@@ -59,10 +60,7 @@ const StyledAccordionHeader = styled.div.attrs<StyledAccordionHeaderProps>(
     borderBottom: border,
     borderLeft: border,
     boxSizing: 'border-box',
-    color:
-      (disabled &&
-        `var(--eds_interactive_disabled__text, ${headerColor.disabled})`) ||
-      `var(--eds_text_static_icons__default, ${headerColor.default})`,
+    color: (disabled && headerColor.disabled) || headerColor.default,
     outline: 'none',
     '&[data-focus-visible-added]:focus': {
       outline,
--- a/libraries/core-react/src/components/Banner/Banner.tokens.ts
+++ b/libraries/core-react/src/components/Banner/Banner.tokens.ts
@@ -26,17 +26,12 @@ const {
 } = tokens
 
 export type BannerToken = ComponentToken & {
-  entities: {
-    icon: ComponentToken & {
-      variants: {
-        info: ComponentToken
-        warning: ComponentToken
-      }
-    }
+  entities?: {
+    icon: ComponentToken
   }
 }
 
-export const banner: BannerToken = {
+export const primary: BannerToken = {
   typography: {
     ...messageTypography,
   },
@@ -57,19 +52,27 @@ export const banner: BannerToken = {
         width: 0,
         color: 'transparent',
       },
-      variants: {
-        info: {
-          background: infoBackground,
-          typography: {
-            color: infoColor,
-          },
-        },
-        warning: {
-          background: warningBackground,
-          typography: {
-            color: warningColor,
-          },
-        },
+    },
+  },
+}
+
+export const info: BannerToken = {
+  entities: {
+    icon: {
+      background: infoBackground,
+      typography: {
+        color: infoColor,
+      },
+    },
+  },
+}
+
+export const warning: BannerToken = {
+  entities: {
+    icon: {
+      background: warningBackground,
+      typography: {
+        color: warningColor,
       },
     },
   },
--- a/libraries/core-react/src/components/Banner/Banner.tokens.ts
+++ b/libraries/core-react/src/components/Banner/Banner.tokens.ts
@@ -1,12 +1,13 @@
 /* eslint-disable camelcase */
 import { tokens } from '@equinor/eds-tokens'
+import type { ComponentToken } from '@equinor/eds-tokens'
 
 const {
   typography: {
     paragraph: { body_long: messageTypography },
   },
   spacings: {
-    comfortable: { medium: spacingMedium },
+    comfortable: { medium },
   },
   colors: {
     infographic: {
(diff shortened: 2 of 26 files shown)


wenche:
I did try to fix the Accordion tokens type but it was too complicated for a quick fix. 🙈

wenche:
@mimarz  I don't quite understand the use of `Borders/Border type`. In the <Chip> tests I had to temp. remove this:
`//expect(chip).toHaveStyleRule('border-color', rgbaTrim(error.border.color))`. Gives me this:
Skjermbilde 2021-02-26 kl  14 44 12
🤯

libraries/core-react/src/components/Accordion/AccordionHeader.tsx
```diff
@@ -8,6 +8,7 @@ import { Icon } from '../Icon'
 import { AccordionHeaderTitle } from './AccordionHeaderTitle'
 import { accordion as tokens } from './Accordion.tokens'
 import type { AccordionProps } from './Accordion.types'
+// import { bordersTemplate } from '@utils'
```
mimarz, line 11: suggest we just remove the comment if its not used

libraries/core-react/src/components/Banner/Banner.tokens.ts
```diff
@@ -15,33 +16,57 @@ const {
...
+    warning: {
+      background: string
+      color: string
+    }
+    shape: {
+      minHeight: string
+      minWidth: string
+      borderRadius: string
+    }
+  }
```
mimarz, line 43: I found putting `icon` under `entities` to be better for conveying sub-entites that are styled or tokenized. The type for icon should also be `ComponentToken` so its "conformed". 

I have found putting on different colors/variants on root level to be easier to maintain. 

Checkout [tokens](https://github.com/equinor/design-system/blob/c6433df9f32fbb2c88bca0e768eda58940551baf/libraries/core-react/src/components/Progress/Circular/CircularProgress.tokens.ts#L20) for `CircularProgress` on suggestion on how to do both of these

libraries/core-react/src/components/Banner/BannerActions.tsx
```diff
@@ -13,11 +11,11 @@ type StyledBannerActionsProps = {
 }
 
 const StyledBannerActions = styled.div<StyledBannerActionsProps>`
-  margin-left: ${enabled.spacings};
+  margin-left: ${tokens.spacings.left};
   grid-column: ${({ placement }) => (placement === 'bottom' ? '1/-1' : 'auto')};
   ${({ placement }) =>
     placement === 'bottom' && {
-      marginTop: enabled.spacings,
+      marginTop: tokens.spacings.left,
```
mimarz, line 18: shouldn't this point to a `tokens.spacings.top` property instead of `.left`?

libraries/core-react/src/components/Card/Card.tokens.ts
```diff
@@ -17,7 +18,21 @@ const {
   },
 } = tokens
 
-export const card = {
+export type CardToken = ComponentToken & {
+  shape: {
+    minHeight: string
+    minWidth: string
+    borderRadius: string
+  }
```
mimarz, line 26: Should we have `shape` instead of `height`/`width` on `ComponentToken`? We are making our own deviation here with something that maybe we need to expand `ComponentToken` with 🤔

libraries/core-react/src/components/Card/Card.tokens.ts
```diff
@@ -17,7 +18,21 @@ const {
...
+  backgroundVariants: {
+    default: string
+    info: string
+    danger: string
+    warning: string
```
mimarz, line 31: different colors I found works best as seperate tokens variables as mention earlier

libraries/core-react/src/components/Chip/Chip.tokens.ts
```diff
@@ -29,66 +30,93 @@ const {
...
-    typography: {
-      ...chipTypography,
-      color: primaryColor,
-    },
-    icon: {
-      height: medium,
-      width: medium,
-      border: {
-        radius: borderRadius,
+export type ChipToken = ComponentToken & {
```
mimarz, line 33: Checkout  `CircularProgress` [tokens](https://github.com/equinor/design-system/blob/c6433df9f32fbb2c88bca0e768eda58940551baf/libraries/core-react/src/components/Progress/Circular/CircularProgress.tokens.ts#L20) on suggestion on how to make this easier

libraries/tokens/src/types/index.ts
```diff
@@ -28,10 +28,12 @@ export type TypographyTokens = {
...
 
 export type Spacing = {
-  left: string
-  right: string
+  left?: string
+  right?: string
   top?: string
   bottom?: string
+  x?: string
+  y?: string
```
mimarz, line 36: If we are adding new properties to `Spacing` type you need to also update the [spacingsTemplate](https://github.com/equinor/design-system/blob/c6433df9f32fbb2c88bca0e768eda58940551baf/libraries/core-react/src/utils/templates/index.ts#L51) to support  these properties to avoid bugs in the future where people think they are supported in the template and aren't.  

I would also suggest maybe adding a secondary parameter to the template which specifies spacings `type?: "padding" | "margin"`. It be set to `"padding"`  as default value.

libraries/core-react/stories/testing/GlobalStyles.ts
```diff
@@ -0,0 +1,182 @@
+// Testing global variables when migrating
+
+import { createGlobalStyle } from 'styled-components'
+
+export const GlobalStyleDark = createGlobalStyle`
+:root {
```
mimarz, line 6: Can we interpolate this string with reading the tokens.css content directly? easier to update changes from figma-broker instead of copy paste content from one file to another?

mimarz (changes requested):
Lots of good work here on cleaning up stuff and paving way for the future!

Added some comments and suggestions based on my experience with working on "token conform". 


I would also suggest updating the `ComponentToken` type with a new property I used a lot (but did not figure out how to type correctly) with `  entities: Record<string, ComponentToken>`

libraries/core-react/src/components/Banner/Banner.tokens.ts
wenche: Maybe we should consider to add a "variant" to the ComponentToken 🤔  We have several components with variants like `warning`, `info` and `error`.

mimarz: Are they variants tho? 🤪  I started a discussion on this in #801 as i ran into the same questions myself when doing progress indicators @wenche

wenche: I don't know what they are, 🤪  but def. not states like active, disabled and so on. Just put them under variants for now, should be easy to rename to something else. I think we'll see the whole picture with TextArea, Cards ++

mimarz: They are under both color and variants in other components so its a weird one. I am leaning towards that they are _not_ variants, as its just coloring of the component. Variant makes more sense for use-cases such as button `contained` vs. `ghost_icon` where there is a clear visual difference. 

But yeah. lets not change it from what its today, better to do that in one change/release. Albeit we should not colors to variants in future changes imho.

libraries/core-react/src/components/Banner/BannerActions.tsx
wenche: Yup!

libraries/core-react/src/components/Card/Card.tokens.ts
wenche: Yes, it might be as this was values from the `shape` objects from tokens. But on the other hand, it wasn't used as minHeight and minWidth in code, so I just moved it to height, width and borders accordingly. But let's see if this is the case in other components as well or not. I don't think we use shape that much, we probably have used it more in code 🤔

mimarz: We don't have much min/max, height/width values. Maybe less that we should... We should have support for it in the `ComponentToken` type, question is in what manner 🤔

wenche: Rome wasn't built in a day 😂

wenche: Skjermbilde 2021-03-02 kl  09 55 12
`Card` uses a minHeight, but this doesn't actually makes sense in code as the padding is 32px, so a 4px something added 🤔  But why

wenche: @mimarz  Instead of adding `minHeight` and `minWidth` to token types because of `Card`, I suggest to remove minHeight and minWidth from Card styles. 😨  A min width of 112px (magic px number 😓 ) doesn't actually work anyway, see image of a Card with width of 112px
Skjermbilde 2021-03-02 kl  10 06 17

mimarz: if it doesn't work remove it 😅

wenche:
I wish `default` wasn't a reserved word in js 😓

libraries/tokens/src/types/index.ts
wenche: > I would also suggest maybe adding a secondary parameter to the template which specifies spacings `type?: "padding" | "margin"`. It be set to `"padding"` as default value.

Let's see if we need it. This is probably a bigger change in a lot of components, might be a separate issue. 🤔

wenche: I removed the x and y options as it probably complicate things. Changed the `margin` to `padding` in the Breadcrumb, it's not exactly wrong IMHO. But we might consider your suggestion of a better separation between margin and padding, but let's do that later.

libraries/core-react/stories/testing/GlobalStyles.ts
wenche: Not for the moment at least, but probably when we have added variables to the library. It's just for testing purposes anyway, so no need to overcomplicate things

libraries/core-react/src/components/Banner/Banner.tokens.ts
wenche: Did change it, I think it works.

libraries/core-react/src/components/Banner/Banner.tokens.ts
```diff
@@ -15,32 +16,63 @@ const {
...
     circle: { minHeight, minWidth, borderRadius },
   },
 } = tokens
 
-export const banner = {
-  enabled: {
-    typography: {
-      ...messageTypography,
+export type BannerToken = ComponentToken & {
+  entities?: {
```
mimarz, line 29: Probably dont need this optional flag?

wenche: I know, that was the one thing I noticed in VS Code after the commit, hoped no one would notice 😂

wenche: ftfy

libraries/core-react/src/components/Card/Card.tsx
```diff
@@ -28,24 +27,35 @@ const StyledCard = styled.div<StyledCardProps>`
...
+  })
+
   const cursor = onClick ? 'pointer' : 'default'
 
   const props = {
     ...rest,
     className,
     ref,
-    background: tokens.background[variant],
+    background: cardVariant && cardVariant.backgroundColor,
```
mimarz, line 58: This feels abit of an overkill for a simple object lookup?

My suggestion would be something like this maybe?
``` 
const lookupKey = variant === 'default' ? 'primary' : variant

const token = tokens[lookupKey]

const props = {
...
background: token.background,
...
}

```

libraries/core-react/src/components/Card/Card.tsx
```diff
@@ -28,24 +27,28 @@ const StyledCard = styled.div<StyledCardProps>`
...
-  min-height: ${shape.minHeight};
   cursor: ${({ cursor }) => cursor};
-
-  ${spacingsTemplate(spacings)}
+  ${bordersTemplate(primary.border)};
+  ${spacingsTemplate(primary.spacings)}
 `
 
+type backgroundVariants = { variant: string; backgroundColor: string }[]
+
```
mimarz, line 36: remove unsued type?

libraries/tokens/src/types/component.ts
```diff
@@ -22,6 +22,7 @@ export type ComponentToken = {
     hover?: ComponentState
     pressed?: ComponentState & { pressed?: Pressed }
   }
+  entities?: Record<string, ComponentToken>
```
mimarz, line 25: 😍

mimarz (approved):
Lots of changes here, looks ok as far as I can tell 👍

Chip.tokens could be improved/simplified with using `R.mergeDeepRight`  but that can be an improvement for some other time...

mimarz:
resolves #1129

PR #1144, opened 2021-03-04
 ♻️  Update TextField  to use EDS Icon component

### LINKED ISSUE #1130: Use Icon component for TextField
Inside the TextField story we have a comment
`// TODO replace with eds react icon component`

We should do this. 😓  Also dig into if we when using EDS icon still need the Icon component inside TextField

wenche:
Resolve #1130

### DIFF
--- a/libraries/core-react/src/components/TextField/TextField.test.tsx
+++ b/libraries/core-react/src/components/TextField/TextField.test.tsx
@@ -1,9 +1,13 @@
 /* eslint-disable no-undef */
 import * as React from 'react'
-import { render, cleanup, fireEvent } from '@testing-library/react'
+import { render, cleanup, fireEvent, screen } from '@testing-library/react'
 import 'jest-styled-components'
 import styled from 'styled-components'
 import { TextField } from '.'
+import { Icon } from '../Icon'
+import { save } from '@equinor/eds-icons'
+
+Icon.add({ save })
 
 afterEach(cleanup)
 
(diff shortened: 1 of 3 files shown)


wenche:
We need to keep the old way with svg for backward compatibility in `Icon.tsx` but we should probably mark it as deprecated. But it don't adds that much in code anyway 🤷‍♀️

libraries/core-react/src/components/TextField/TextField.test.tsx
```diff
@@ -50,6 +54,40 @@ describe('TextField', () => {
...
+    const labelText = 'Label for test'
+    render(
+      <TextField
+        id="test"
+        label={labelText}
+        inputIcon={<Icon name="save" />}
+      />,
+    )
+    const inputNode = screen.getByLabelText(labelText).parentNode
+    expect(inputNode.querySelectorAll('svg')).toHaveLength(1)
```
mimarz, line 75: Suggest adding `title` and assertion on this as we are using that in the stories examples

mimarz (changes requested):
Looks good 👍

Suggest one addition to for tests to match expected behaviour in stories/examples we have made

libraries/core-react/src/components/TextField/TextField.test.tsx
wenche: I copied that Icon from Accordion and we use `title` there. I'm not sure what we actually recommends here? Is it for a11y? If so, wouldn't it be better to  use aria-label on the icon instead of title? Or in combination? Maybe we should highlight several a11y options in the storybook? 🤔  I'm not sure if it think it's worth testing for, should probably be tested in the Icon component @mimarz

mimarz: Not sure if its needed when its _inside_ the `TextField`, but its recommended a11y when the svg/icon is [standalone](https://css-tricks.com/accessible-svgs/). We are using the actual `<title>` element inside here.

mimarz: @wenche maybe remove the title from the stories then?

wenche: I think we should either use aria-label="something" or aria-hidden={true} @mimarz

mimarz: How do you suggest we do this? [Because we are using it inside Icon already](https://github.com/equinor/design-system/blob/31eac3c103362dd416ffe58d62e0175f2fca7e09/libraries/core-react/src/components/Icon/Icon.tsx#L111)

mimarz: <Label>'s id would probably be best to set as `aria-labelledby` on `<Icon>` in this use-case maybe (given title is not set)? @wenche

wenche:
The example for error text input fields (Variants header) is not following best practices with form validation and a11y. And it's also impossible to achieve with TextField as it is right now. Let's create a separate issue for this.

mimarz (approved):
LGTM 👍

PR #1153, opened 2021-03-12
⬆️ Update dependencies

mimarz:
* `babel-eslint` [is deprecated](https://www.npmjs.com/package/babel-eslint) so updated to new recommened approach
* Removed old token story as it didn't provide much anymore (was still jsx)
* Update lots of other dependencies to max


wenche (approved):
Nice to be up to date 👍

PR #1157, opened 2021-03-15
🔧  Update config files to exclude test and stories files

mimarz:
For a long time we had typings for test being included in `core-react`. 

Excluded files including `.test` and `.stories` (so its ready for moving story files 👌)


wenche (approved):
LGTM 👍

PR #1168, opened 2021-03-22
Fix wrong use of cache in actions

### LINKED ISSUE #1166: Missing dist folder in publish tokens github action
When "publish to npm" job runs the dist folder is missing, causing malformed package to be published to npm.

mimarz:
fixes #1166 

The problem was due to a cache being used multiple times, expecting it to be updated.  Caches in github actions are fixed once they have been greated. You have to save it under a different key if you want to add new files.


pomfrida (approved):
LGTM 👍

vnys (approved):
LGTM 👍🏻

PR #1181, opened 2021-03-24
Update Tooltip story

### LINKED ISSUE #1154: Show tooltips with a delay
Currently the EDS tooltip show up immediately, which can be distracting.  We believe that the user experience would benefit from delayed tooltips.  Desktop operating systems typically also use a delay before showing tooltips.  Material-ui tooltips support a "enterDelay" and "leaveDelay" parameters to control this behaviour.

### LINKED ISSUE #1174: Clarify tooltip examples
Remove examples of using tooltip on buttons (as its against [guidelines](https://eds.equinor.com/components/tooltip/) 🙈)

Also add delay on open, to fix feature request #1154

pomfrida:
Story fixed so it matches the guidelines, resolves #1174 
Also added delay on the examples, resolves #1154


libraries/core-react/stories/components/Tooltip.stories.tsx
```diff
@@ -149,14 +185,19 @@ export const TableCellsWithTooltip: Story<TooltipProps> = () => {
     openCell: null,
   })
 
+  let timer: ReturnType<typeof setTimeout>
```
mimarz, line 188: Should have default value, maybe just set it to null?

mimarz (changes requested):
Suggest changing the <Icon> to using the `data` property ([way 2](https://github.com/equinor/design-system/tree/develop/libraries/core-react/src/components/Icon#usage)) instead of `name`and `Icon.add`, just to make it clear for users that it comes from somewhere else.

libraries/core-react/stories/components/Tooltip.stories.tsx
pomfrida: Those icons are actually Equinor icons 😄

mimarz (changes requested):
I can't see that you changed how the icons are used?

mimarz (approved):
LGTM 👍

PR #1182, opened 2021-03-24
Conform close logic with 'Escape' key

### LINKED ISSUE #1175: Conform close logic on Scrim, Menu and Popover
useEscapePress hook for reusability across components (Menu, Scrim, Popover)

pomfrida:
Resolves #1175

### DIFF
--- a/libraries/core-react/src/hooks/useEscapePress.ts
+++ b/libraries/core-react/src/hooks/useEscapePress.ts
@@ -1,12 +1,12 @@
 import { useEffect } from 'react'
 
-export const useEscapePress = (callback: () => void): void => {
+export const useEscapePress = (callback: (e: KeyboardEvent) => void): void => {
   const handleGlobalKeyPress = (e: KeyboardEvent) => {
     const { key } = e
 
     switch (key) {
       case 'Escape':
-        callback()
+        callback(e)
         break
       default:
         break
@@ -19,5 +19,5 @@ export const useEscapePress = (callback: () => void): void => {
     return () => {
       document.removeEventListener('keydown', handleGlobalKeyPress, true)
     }
-  })
+  }, [])
 }
--- a/libraries/core-react/src/hooks/useEscapePress.ts
+++ b/libraries/core-react/src/hooks/useEscapePress.ts
@@ -1,0 +1,26 @@
+import { useEffect, MutableRefObject } from 'react'
+
+export const useOutsideClick = (
+  ref: MutableRefObject<HTMLElement>,
+  callback: () => void,
+): void => {
+  const handleGlobalKeyPress = (e: KeyboardEvent) => {
+    const { key } = e
+
+    switch (key) {
+      case 'Escape':
+        callback()
+        break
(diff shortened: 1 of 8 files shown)


libraries/core-react/stories/components/Popover.stories.tsx
```diff
@@ -51,13 +51,25 @@ export const Default: Story<PopoverProps> = (args) => {
...
+  }
+
   return (
     <div style={{ margin: '10em' }}>
       <Button
         id="default-popover-anchor"
         aria-controls="default-popover"
         ref={referenceElement}
         onClick={handleToggle}
+        onKeyDown={onKeyPress}
```
mimarz, line 72: This wouldn't work globally tho?

libraries/core-react/stories/components/Popover.stories.tsx
```diff
@@ -51,13 +51,25 @@ export const Default: Story<PopoverProps> = (args) => {
...
+    const { key } = e
+    switch (key) {
+      case 'Escape':
+        handleToggle()
+        break
+      default:
+        break
+    }
+  }
+
```
mimarz, line 64: This approach differs from how we have done it in `Menu`, where we do this internally in the component and call the `onClose` callback when key is triggered. Suggest we revert this change and make a seperate PR for having this internally in the component like `Menu`

mimarz (changes requested):
The delay example looks ok, but I would leave out the 'esc' key handling as it needs to be done internally in popover to conform with how we have done it in our other components (Scrim, Menu comes to mind atleast)

libraries/core-react/stories/components/Popover.stories.tsx
pomfrida: Oops, I just continued with the change on this PR. But renamed the task and PR

libraries/core-react/src/hooks/useEscapePress.ts
```diff
@@ -0,0 +1,23 @@
+import { useEffect } from 'react'
+
+export const useEscapePress = (callback: (e: KeyboardEvent) => void): void => {
```
mimarz, line 3: I suggest changing the signature of this hook so its abit more applicable for other keys (not just "Escape"). 

If you change it to be more generic such as `useGlobalKeyPress` and also take in a parameter for which Key (as first parameter). Then it will be easier for us to re-use this hook for other use-case and keys.

Since we are using typescript you can also make an [enum type](https://www.typescriptlang.org/docs/handbook/enums.html) for KEY so that we can get type-safety for [keycodes](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values) on the provided parameter for which key.

example of enum:
```typescript 
enum KEY = {
Escape = 'Escape' 
}
```

mimarz (changes requested):
Suggestion we do some slight changes so that we can use this awesome hook on other stuff 👍

mimarz (approved):
hook happy

PR #1197, opened 2021-03-31
♻️ Optional & nullable anchorEl

### LINKED ISSUE #1196: Optional anchorEl on components
`anchorEl` should be optional or `null` in `Menu`, `Popover` & `Tooltip` for easier use of wiring up with dom elements and use of `useRef` in React.

mimarz:
resolves #1196


pomfrida (approved):
LGTM 👍

PR #1198, opened 2021-03-31
♻️ Clarified use of Banner.Actions in story

### LINKED ISSUE #1188: Clean up Banner story
Banner story's `Wrapper` component is confusing as it extends `Banner.Actions`. Remove extends `styled(Banner.Actions)` and only use our components directly in stories to avoid confusion.

It also looks like display, flex/grid should be applied internally in the `Banner.Actions` removing the need for `Wrapper`

mimarz:
resolves #1188


pomfrida (approved):
LGTM 👍

PR #1200, opened 2021-04-06
♻️ Moved state and anchor-managment inside Tooltip

### LINKED ISSUE #1195: Tooltip component api and usage
*This needs to be looked at before release `core-react@0.10.0` as we have breaking changes for `Tooltip` in upcoming release.*

There is currently a lot of boilerplate needed for showing a tooltip.

We are using the same approach as `Popover` and `Menu` where state needs to be controlled on the outside of the component, but the usage for `Tooltip` is a lot simpler, so this might be overkill. 

It should not be up to our users to need to create hacks for supporting "onHover" on disabled elements (`onPointerEnter` with `div` surrounding anchor element) for tooltips in Safari vs Chrome, etc...

Action
* Should we simplify usage of `Tooltip`?
* Should it be possible to encapsulate the anchor element with a tooltip and do the definition of  `anchorEl` and open close state internally in `Tooltip`?

Notes:
* https://material-ui.com/components/tooltips/
* https://react.carbondesignsystem.com/?path=/story/components-tooltip--default-bottom
* https://eufemia.dnb.no/uilib/components/tooltip

mimarz:
resolves #1195 
Co-authored-by: @pomfrida


pomfrida (approved):
LGTM 👍

PR #1210, opened 2021-04-09
Typography link focus

### LINKED ISSUE #1203: Missing focus outline on Typography links
Move focus outline logic from Breadcrumb to Typography

pomfrida:
Resolves #1203

### DIFF
--- a/libraries/core-react/src/components/Typography/Typography.test.tsx
+++ b/libraries/core-react/src/components/Typography/Typography.test.tsx
@@ -1,12 +1,19 @@
 /* eslint-disable no-undef */
 import * as React from 'react'
-import { render, cleanup, screen } from '@testing-library/react'
+import {
+  render,
+  cleanup,
+  screen,
+  fireEvent,
+  act,
+  prettyDOM,
+} from '@testing-library/react'
 import '@testing-library/jest-dom'
 import 'jest-styled-components'
 import styled from 'styled-components'
 import { Typography } from '.'
 import { tokens } from '@equinor/eds-tokens'
-import { colors } from './Typography.tokens'
+import { colors, link } from './Typography.tokens'
 import type { Typography as TypographyType } from '@equinor/eds-tokens'
 
 const StyledTypography = styled(Typography)`
@@ -147,5 +154,28 @@ describe('Typography', () => {
 
     expect(typography).toHaveStyleRule('margin-top', '16px')
     expect(typography).toHaveStyleRule('margin-bottom', '32px')
+  })
+  it('can focus links', async () => {
+    render(
+      <Typography link href="#">
+        Link
+      </Typography>,
+    )
+
+    const typography = screen.getByText('Link')
+
+    typography.focus()
+    await act(() => new Promise((r) => setTimeout(r, 100)))
+    console.log(prettyDOM(typography, 99999))
+    expect(typography).toHaveStyleRule(
+      'outline',
+      `${
+        link.states.focus.outline.width +
+        ' ' +
+        link.states.focus.outline.type +
+        ' ' +
+        link.states.focus.outline.color
+      }`,
+    )
   })
 })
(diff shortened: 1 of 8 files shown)


libraries/core-react/src/components/Typography/Typography.tsx
```diff
@@ -90,6 +91,14 @@ const StyledTypography = styled.p<StyledProps>`
         text-overflow: ellipsis;
       }
     `}
+  ${({ link }) =>
+    link &&
+    css`
+      width: max-content;
```
mimarz, line 97: Why are we setting `width`? Not exactly normal behaviour for an `a`element

libraries/core-react/src/components/Typography/Typography.tsx
```diff
@@ -90,6 +91,14 @@ const StyledTypography = styled.p<StyledProps>`
         text-overflow: ellipsis;
       }
     `}
+  ${({ link }) =>
+    link &&
```
mimarz, line 95: How does the css look when link is `false` or `undefined`?

libraries/core-react/src/components/Typography/Typography.tokens.ts
```diff
@@ -51,11 +52,26 @@ type ColorVariants =
...
 
+const link: ComponentToken = {
+  states: {
+    focus: {
+      outline: {
+        width: '1px',
+        color: focus,
+        style: 'dashed',
+        type: 'outline',
+        offset: '6px',
```
mimarz, line 63: Should the offset be `6px`?

mimarz (changes requested):
Nice to standarize 👍 Did you check with @equinor/eds-ux if it looks ok?

libraries/core-react/src/components/Typography/Typography.tsx
pomfrida: Width adjusted to the parent width, so focus was full width instead of wrapping the link itself

libraries/core-react/src/components/Typography/Typography.tsx
pomfrida: As expected

libraries/core-react/src/components/Typography/Typography.tokens.ts
pomfrida: Tbh I just copied this from the Breadcrumb component. Looking for the Typography component in Figma turned out to be harder than expected. Do you know where it is? 🤔

BirteThornquist:
Focus state is missing in Figma design (will add an issue). Use the common focus border with 2 px spacing for the links.

libraries/core-react/src/components/Typography/Typography.tsx
mimarz: This sounds like its something thats specific for when its used in `Breadcrumbs`? So maybe it should be applied there instead? We should avoid changing native behaviour unless its intended.

pomfrida: @mimarz Yes grid was just missing `justify-items` so that fixed it 👍  Ready again for review now

mimarz:
@pomfrida The cake links does not lead to cake. I feel lied to! 🍰 😢 /denied

libraries/core-react/src/components/Typography/Typography.test.tsx
```diff
@@ -148,4 +148,16 @@ describe('Typography', () => {
...
+    render(
+      <Typography link href="#">
+        Link
+      </Typography>,
+    )
+
+    const typography = screen.getByText('Link')
+
+    typography.focus()
+    expect(typography).toHaveFocus()
```
mimarz, line 161: This test does not test much... It would still be green even if the outline was wrong or not present. Maybe add some more assertions on style?

mimarz (approved):
LGTM 👍

PR #1213, opened 2021-04-12
🐛 Fix wrong offset Menu

### LINKED ISSUE #1212: Wrong offset on menu

pomfrida:
Resolves #1212

### DIFF
--- a/libraries/core-react/src/components/Menu/Menu.tsx
+++ b/libraries/core-react/src/components/Menu/Menu.tsx
@@ -100,7 +100,7 @@ export const Menu = React.forwardRef<HTMLUListElement, MenuProps>(function Menu(
     containerRef.current,
     null,
     placement,
-    0,
+    4,
   )
 
   const props = {
--- a/libraries/core-react/src/components/Menu/Menu.tsx
+++ b/libraries/core-react/src/components/Menu/Menu.tsx
@@ -100,6 +100,7 @@ export const Menu = React.forwardRef<HTMLUListElement, MenuProps>(function Menu(
     containerRef.current,
     null,
     placement,
+    0,
   )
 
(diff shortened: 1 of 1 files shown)


libraries/core-react/src/components/Menu/Menu.tsx
```diff
@@ -100,6 +100,7 @@ export const Menu = React.forwardRef<HTMLUListElement, MenuProps>(function Menu(
     containerRef.current,
     null,
     placement,
+    0,
```
mimarz, line 103: `0` looks very "tight" did you check with @equinor/eds-ux that this was ok value? I think value was more like 2 or 4px before... 🤔

mimarz (changes requested):

libraries/core-react/src/components/Menu/Menu.tsx
pomfrida: I used the context menu example [here](https://www.figma.com/file/luNobpDIWTuQ9rhyovGLBI/Desktop?node-id=977%3A27105) and it's 0px from anchor to menu

mimarz: 4px 
image

vs 

2px 
image

vs 

0px 
image

mimarz: I would say maybe `2px`? What do you guys think @equinor/eds-ux ?

BirteThornquist: @mimarz @pomfrida I think it shall be 4px space between the button and the drop menu. I guess the clickbound take care of that?

mimarz (approved):
LGTM 👍

PR #1231, opened 2021-04-15
🐛 Fixed decleration types does not support aliases

### LINKED ISSUE #1229: Missing intellisense values on placement property on typescript components
**Describe the bug**

1.
2.
3.
4.

**Expected behavior**

**Specifications**

- Version: 
- Browser: 
- OS: 

**Additional context**

mimarz:
fixes #1229

### DIFF
--- a/libraries/core-react/src/hooks/usePopper.ts
+++ b/libraries/core-react/src/hooks/usePopper.ts
@@ -1,25 +1,22 @@
 import type { CSSProperties } from 'react'
 import { usePopper as reactPopper } from 'react-popper'
 
-export const PlacementValues = [
-  'auto',
-  'auto-start',
-  'auto-end',
-  'top',
-  'top-start',
-  'top-end',
-  'bottom',
-  'bottom-start',
-  'bottom-end',
-  'right',
-  'right-start',
-  'right-end',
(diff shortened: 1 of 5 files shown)


libraries/core-react/src/hooks/usePopper.ts
```diff
@@ -1,25 +1,22 @@
 import type { CSSProperties } from 'react'
 import { usePopper as reactPopper } from 'react-popper'
 
-export const PlacementValues = [
```
pomfrida, line 4: If the props table in Popover still works without this, you can remove the argTypes from `Popover.stories`:

```
argTypes: {
    placement: {
      control: {
        type: 'select',
        options: PlacementValues,
        defaultValue: 'auto',
      },
    },
  },
```

mimarz: hmmm, I forgot to test the stories. 🙈 Let me check it out!

mimarz: Removed the `argsTypes` as it works without it now @pomfrida  :)

pomfrida (approved):
LGTM 👍

PR #1241, opened 2021-04-16
🐛  Storybook decorators disrupts code examples

### LINKED ISSUE #1238: Storybook decorators disrupts code examples
Disrupts first code examples in the Storybook**
Because this first example (with controls) is apparently compiled differently than the other examples. 
How to fix: remove the decorators from the Meta object in stories and define it on each example instead (except the first one though..)

Example of the decorator: 

The unwanted result in code example (should show the Slider with props...): 

Blocked by #1156

pomfrida:
Resolves #1238 

NB! Slider was the only story with component decorator


mimarz (approved):
We really need to find a better way of handling this, as its very different from story to story how we handle this 🤪 

Still better than it was 👍

PR #1243, opened 2021-04-19
🔥🚨 Remove typescript aliases & activated typescript type rules

### LINKED ISSUE #1242: Remove typescript aliases
As nice as having aliases are, they are proving to be cumbersome with linting, type declarations and other transformations that relate to paths.

To avoid future bugs and error we decided to remove typescript aliases, as they are nice and clean, they have proved to give more headache than they solved. This means more dotting, but vscode solves most of this when moving stuff and its a "better" con to live with than potential errors in type declarations or not being able to type lint.

related bugs #1232, #1233

mimarz:
resolves #1242


pomfrida (approved):
LGTM 👍

PR #1244, opened 2021-04-19
Export subcomponent types

### LINKED ISSUE #1127: Export missing types for child components of compound components
E.g. for Cards we only export CardProps type from the library. We must export types for the sub components as well for all compound components

Edit: Might be a good idea to do it at the same time as we remove the now deprecated name pattern for sub components? #1088

pomfrida:
Resolves #1127


mimarz (approved):
LGTM 👍

PR #1293, opened 2021-04-21
♻️ Updated Label tokens

### LINKED ISSUE #1270: Conform Label tokens

pomfrida:
Resolves #1270


mimarz (changes requested):
We are using `rgba` values on all other components, suggest we also fix that now, since we are conforming tokens.

mimarz (approved):
LGTM 👍

PR #1299, opened 2021-04-21
♻️ Conform Popover tokens

### LINKED ISSUE #1265: Conform Popover tokens

pomfrida:
Resolves #1265

### DIFF
--- a/libraries/core-react/src/components/Popover/Popover.tokens.ts
+++ b/libraries/core-react/src/components/Popover/Popover.tokens.ts
@@ -1,5 +1,5 @@
 import { tokens } from '@equinor/eds-tokens'
-import type { Typography, Spacing } from '@equinor/eds-tokens'
+import type { Typography, Spacing, ComponentToken } from '@equinor/eds-tokens'
 
 const {
   typography: {
@@ -24,9 +24,7 @@ const {
   elevation: { overlay: elevation },
 } = tokens
 
-type Popover = {
-  header: Typography
-  background: string
+type Popover = ComponentToken & {
   elevation: string
   gridGap: string
   popover: {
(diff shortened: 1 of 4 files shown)


libraries/core-react/src/components/Popover/Popover.tokens.ts
```diff
@@ -24,61 +24,50 @@ const {
   elevation: { overlay: elevation },
 } = tokens
 
-type Popover = {
-  header: Typography
-  background: string
+type Popover = ComponentToken & {
   elevation: string
```
mimarz, line 28: Maybe we need that `boxShadow` property on `ComponentToken` 🙈 Suggest using `Paper` component in `Popover` nonetheless

libraries/core-react/src/components/Popover/PopoverTitle.tsx
```diff
@@ -1,24 +1,22 @@
...
 import { Divider } from '../Divider'
-import { typographyTemplate } from '../../utils'
 
 import { popover as tokens } from './Popover.tokens'
 
 const StyledPopoverTitle = styled.div`
-  ${typographyTemplate(tokens.header)}
-  margin-right: ${tokens.closeButton.width};
-  max-width: calc(${tokens.popover.maxWidth} - ${tokens.closeButton.width});
+  margin-right: ${tokens.entities.closeButton.width};
```
mimarz, line 8: This looks the wrong use of a token value? Should this not be a `spacings.right` value on `closeButton` instead?

libraries/tokens/src/types/clickbound.ts
```diff
@@ -1,7 +1,7 @@
 export type Clickbound = {
-  height: string
+  height?: string
   width?: string
-  offset: {
+  offset?: {
```
mimarz, line 4: Not sure I can see why these were set to optional?

mimarz (changes requested):
Some nitpicking from good o'l Mickey 🤪

libraries/core-react/src/components/Popover/Popover.tokens.ts
pomfrida: We're actually using `Paper` so elevation on `Popover` seem to be unused, will remove it 🔥

libraries/core-react/src/components/Popover/PopoverTitle.tsx
pomfrida: This margin is to leave space for the button, had to do this when I moved the button into `Popover` instead of `PopoverTitle` a while ago.

libraries/tokens/src/types/clickbound.ts
pomfrida: True, I found out I didn't need it at some point but forgot to remove it 🔥

mimarz (approved):
LGTM 👍

PR #1300, opened 2021-04-21
♻️  Updated Divider tokens

### LINKED ISSUE #1273: Conform Divider tokens

pomfrida:
Resolves #1273


mimarz (changes requested):
So far we have been exporting variants as two separate tokens, suggest doing the same here for "conformity" as they are not really entities 😅

mimarz (approved):
These color values and variants are kinda confusing 🙈 

but yeah, token conversion done atleast! 👍

PR #1301, opened 2021-04-21
♻️ Updated Snackbar tokens

### LINKED ISSUE #1255: Conform snackbar tokens

pomfrida:
Resolves #1255

### DIFF
--- a/libraries/core-react/src/components/Snackbar/Snackbar.tokens.ts
+++ b/libraries/core-react/src/components/Snackbar/Snackbar.tokens.ts
@@ -10,29 +10,25 @@ const {
       background__overlay: { rgba: background },
     },
     text: {
-      static_icons__primary_white: { hex: color },
+      static_icons__primary_white: { rgba: color },
     },
     interactive: {
-      link_in_snackbars: { hex: buttonColor },
+      link_in_snackbars: { rgba: buttonColor },
     },
   },
   spacings: {
     comfortable: { medium: spacingMedium, x_large: spacingXLarge },
   },
-  elevation: { overlay: boxShadow },
   clickbounds: { default__base: clickbounds },
   shape: {
     button: { borderRadius: radius },
   },
 } = tokens
 
-type Snackbar = ComponentToken & {
-  boxShadow: string
-}
+type Snackbar = ComponentToken
 
 export const snackbar: Snackbar = {
   background,
-  boxShadow,
   minHeight: clickbounds,
   border: {
     type: 'border',
--- a/libraries/core-react/src/components/Snackbar/Snackbar.tokens.ts
+++ b/libraries/core-react/src/components/Snackbar/Snackbar.tokens.ts
@@ -1,5 +1,5 @@
 import { tokens } from '@equinor/eds-tokens'
-import type { Typography, ComponentToken } from '@equinor/eds-tokens'
+import type { ComponentToken } from '@equinor/eds-tokens'
 
 const {
   typography: {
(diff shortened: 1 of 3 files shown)


libraries/core-react/src/components/Snackbar/Snackbar.tokens.ts
```diff
@@ -17,46 +17,48 @@ const {
     },
```
mimarz, line 17: Suggest we also fix the use of `hex` to `rgba` values

libraries/core-react/src/components/Snackbar/Snackbar.tokens.ts
```diff
@@ -17,46 +17,48 @@ const {
...
+    type: 'border',
+    width: 0,
+    radius,
+  },
+  entities: {
+    actions: {
+      spacings: {
+        left: spacingXLarge,
+        top: '10px',
+        bottom: '10px',
```
mimarz, line 47: These spacings values look to be wrong? They are also abit superflous as we are setting `minHeight`.

libraries/core-react/src/components/Snackbar/Snackbar.tokens.ts
```diff
@@ -17,46 +17,48 @@ const {
...
   clickbounds: { default__base: clickbounds },
+  shape: {
+    button: { borderRadius: radius },
+  },
 } = tokens
 
-type Snackbar = {
-  background: string
+type Snackbar = ComponentToken & {
   boxShadow: string
```
mimarz, line 30: We don't want to "mutate" `ComponentToken` like this. Suggest using (a styled)`Paper` component inside `Snackbar` with the correct `elevation` set.

libraries/core-react/src/components/Snackbar/SnackbarAction.tsx
```diff
@@ -4,9 +4,9 @@ import { snackbar as tokens } from './Snackbar.tokens'
 
 const StyledSnackbarAction = styled.div`
   display: inline-flex;
-  margin-left: ${tokens.spacings.actionSpace};
-  margin-top: -10px;
-  margin-bottom: -10px;
+  margin-left: ${tokens.entities.actions.spacings.left};
+  margin-top: -${tokens.entities.actions.spacings.top};
+  margin-bottom: -${tokens.entities.actions.spacings.bottom};
```
mimarz, line 9: You can use `spacingsTemplate` here?

mimarz (changes requested):
Saw some discrepancies we might as well fix while we are changing this 🐛

libraries/core-react/src/components/Snackbar/Snackbar.tokens.ts
pomfrida: These are negative top/bottom values in `SnackbarAction` - spacing buttons to `Snackbar` is `6px`, but since padding already is `16px` without buttons, we need `-10px` on the buttons 🤔 And apparently negative values wasn't valid in the token

libraries/core-react/src/components/Snackbar/SnackbarAction.tsx
pomfrida: It was because of the negative values mentioned above but I'll try with aligning with flex 👍

pomfrida: Seem like it has to be this way because we don't want `16px` padding on the buttons 🤔

mimarz (approved):
LGTM 👍

PR #1305, opened 2021-04-26
♻️ Updated Dialog tokens

### LINKED ISSUE #1274: Conform Dialog tokens

pomfrida:
Resolves #1274

### DIFF
--- a/libraries/core-react/src/components/Dialog/Dialog.tokens.ts
+++ b/libraries/core-react/src/components/Dialog/Dialog.tokens.ts
@@ -1,5 +1,5 @@
 import { tokens } from '@equinor/eds-tokens'
-import type { Typography, ComponentToken } from '@equinor/eds-tokens'
+import type { ComponentToken } from '@equinor/eds-tokens'
 
 const {
   spacings: {
@@ -9,7 +9,6 @@ const {
     ui: { accordion_header },
     paragraph: { body_long },
   },
-  elevation: { above_scrim: boxShadow },
   colors: {
     ui: {
       background__default: { rgba: background },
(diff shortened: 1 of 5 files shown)


libraries/core-react/src/components/Dialog/Dialog.tokens.ts
```diff
@@ -20,24 +19,41 @@ const {
...
+      typography: accordion_header,
+    },
+    content: {
+      minHeight: '80px',
+      height: '104px',
+      typography: body_long,
+    },
+    actions: {
+      minHeight: '48px',
+      height: '8px',
```
mimarz, line 56: This just looks weird, `minHeight` is `48px` and `height` is `8px`? 😂

Do we need to do trickery when no children i present? Can't it just be `0` in the code? Not sure this is a value worth having in token as its very niche specific.

mimarz (changes requested):
Maybe the value for when no children a present does not need to be tokenized as its very specific for how its solved technically

libraries/core-react/src/components/Dialog/Dialog.tokens.ts
pomfrida: Yes I agree 🤣  I'll have a look. But if we still need it I can move `height: '8px'` to `children` where it's more fitting I think

pomfrida: `!children` ..Or not 🤔

pomfrida: I found out that the `height` was to avoid grid filling up space whenever a user forgot to add content inside `Title` or `Actions`

mimarz (approved):
LGTM 👍

PR #1307, opened 2021-04-27
♻️ Conform Tabs tokens

### LINKED ISSUE #1254: Conform Tabs tokens

mimarz:
resolves #1254

### DIFF
--- a/libraries/core-react/src/components/Tabs/Tabs.tokens.ts
+++ b/libraries/core-react/src/components/Tabs/Tabs.tokens.ts
@@ -4,16 +4,16 @@ import type { ComponentToken } from '@equinor/eds-tokens'
 const {
   colors: {
     text: {
-      static_icons__tertiary: { hex: defaultColor },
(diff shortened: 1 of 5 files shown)


libraries/core-react/src/components/Tabs/Tabs.tokens.ts
```diff
@@ -15,67 +16,96 @@ const {
       primary__hover: { hex: activeHoverColor },
```
pomfrida, line 16: Should change hex to rgba

pomfrida (changes requested):
Nice rewrite 👍  Just change hex to rgba in tokens then it's good to go

libraries/core-react/src/components/Tabs/Tabs.tokens.ts
mimarz: Fixed

pomfrida (approved):
LGTM 👍

PR #1311, opened 2021-04-29
🐛 Check if title is set before opening

### LINKED ISSUE #1292: Tooltip: Displays when title is ""
Should not be displayed when title is ""/blank. This is an old issue that reappeared with 0.10.1.

Also, why was the "open" property removed from tooltip? It was very handy

mimarz:
fixes #1292


pomfrida (approved):
LGTM 👍

PR #1320, opened 2021-04-30
♻️ Conform Input tokens

### LINKED ISSUE #1271: Conform Input tokens

pomfrida:
Resolves #1271

### DIFF
--- a/libraries/core-react/src/components/Input/Input.tokens.ts
+++ b/libraries/core-react/src/components/Input/Input.tokens.ts
@@ -57,128 +57,93 @@ export const baseInput: InputToken = {
 
 export const input: InputVariantsToken = {
   default: {
-    entities: {
-      border: {
-        border: {
-          type: 'bordergroup',
-          bottom: {
-            style: 'solid',
-            color: static_icons__tertiary.rgba,
-            width: '1px',
-          },
-        },
-      },
-      outline: {
-        border: {
-          type: 'border',
+    boxShadow: 'inset 0px -1px 0px 0px ' + static_icons__tertiary.rgba,
+    states: {
+      active: {
+        outline: {
+          type: 'outline',
           color: 'transparent',
           width: '1px',
+          style: 'solid',
+          offset: '0px',
         },
       },
-    },
-    states: {
       focus: {
         outline: {
           type: 'outline',
           width: '2px',
           color: primary__resting.rgba,
           style: 'solid',
+          offset: '0px',
         },
       },
     },
   },
   error: {
-    entities: {
-      border: {
-        border: {
-          type: 'bordergroup',
-          bottom: {
-            style: 'solid',
-            color: 'transparent',
-            width: '1px',
-          },
-        },
-      },
-      outline: {
-        border: {
-          type: 'border',
-          style: 'solid',
+    boxShadow: 'inset 0px -1px 0px 0px transparent',
+    states: {
+      active: {
+        outline: {
+          type: 'outline',
           color: danger__resting.rgba,
           width: '1px',
+          style: 'solid',
+          offset: '0px',
         },
       },
-    },
-    states: {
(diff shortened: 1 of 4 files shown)


libraries/core-react/src/components/Input/Input.tsx
```diff
@@ -97,17 +107,17 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(function Input(
   ref,
 ) {
   const as: ElementType = multiline ? 'textarea' : 'input'
-  const inputVariant = tokens[variant]
-  const spacings = tokens.spacings.comfortable
+  const inputVariant = input[variant]
+  const spacings = tokens.comfortable.spacings
```
pomfrida, line 111: Not sure why we had `compact` in tokens, too. Doesn't seem like any other component uses Input tokens.. But I left it since we'll work on compact next .

mimarz (changes requested):
I am abit confused by the component token structure... Why is `border` and `outline` as `entities` when they have specific fields for them in `ComponentToken`?

pomfrida:
> I am abit confused by the component token structure... Why is `border` and `outline` as `entities` when they have specific fields for them in `ComponentToken`?
> 
> You are also not using `borderTemplate` or `outlineTemplate` when it looks like we want to use them?

Yes, I think this is a tricky one because in this component we use `outline` both with and without the focus state, and border is actually a `box-shadow`.  I'll change to `outlineTemplate`, that was a miss 👍

mimarz:
> > I am abit confused by the component token structure... Why is `border` and `outline` as `entities` when they have specific fields for them in `ComponentToken`?
> > You are also not using `borderTemplate` or `outlineTemplate` when it looks like we want to use them?
> 
> Yes, I think this is a tricky one because in this component we use `outline` both with and without the focus state, and border is actually a `box-shadow`. I'll change to `outlineTemplate`, that was a miss 👍

We can expand on `borderTemplate` to support `boxShadow` or make a seperate template for it then :)

libraries/core-react/src/components/Input/Input.tokens.ts
```diff
@@ -1,125 +1,169 @@
...
+export type InputVariantsToken = {
+  default: ComponentToken
+  error: ComponentToken
+  warning: ComponentToken
+  success: ComponentToken
 }
 
-export const input: InputProps = {
-  background: colors.ui.background__light.hex,
+export const baseInput: InputToken = {
```
mimarz, line 36: suggest naming this `input`

libraries/core-react/src/components/Input/Input.tokens.ts
```diff
@@ -1,125 +1,169 @@
...
+        color: disabled__text.rgba,
       },
-      border: {
-        width: '2px',
-        color: colors.interactive.primary__resting.hex,
+    },
+  },
+}
+
+export const input: InputVariantsToken = {
```
mimarz, line 58: ...and this `inputVariants`?

libraries/core-react/src/components/Input/Input.tsx
mimarz: Think it was just some leftovers from internal testing 🙈

mimarz (changes requested):
Just a suggestion on renaming some consts used in tokens to as I got confused what was what.

Else it looked good!

libraries/core-react/src/components/Input/Input.tokens.ts
pomfrida: Fixed 👍

libraries/core-react/src/components/Input/Input.tokens.ts
pomfrida: Fixed 👍

mimarz (approved):
LGTM 👍

PR #1351, opened 2021-05-12
✨  rowsMax and autoresize to multiline TextField

### LINKED ISSUE #1336: rowsMax on TextField

I want to make it so that a text field expands as the content takes up to much space in the TextField, and not have a scroll-bar.

Include rowsMax on TextField.

Switch to TextField from material-ui, which is boring to have to do just for rowsMax.

**Demo how I want it to work**

https://codesandbox.io/s/material-demo-forked-us7hp

pomfrida:
Resolves #1336


pomfrida:
Getting linting error   `169:7  error  React Hook "useAutoResize" is called conditionally. React Hooks must be called in the exact same order in every component render  react-hooks/rules-of-hooks`

But the condition is just if the component is a multiline, so it would not break the hook laws I think 🤔 It would be called either 0 if it's not a multiline and the same amount of times if it is a multiline

mimarz:
You Are using the hook inside an 'if' statement, this is why you are getting the error

mimarz (approved):
LGTM, but we should also get #1357 done now to seperate concern as there is some mixing of input and textarea stuff now 🙈👍

PR #1353, opened 2021-05-12
♻️ CombinedRefs update

### LINKED ISSUE #1347: Improve useCombinedRefs to avoid potential bugs
During #1220 we discovered that the hook `useCombinedRefs` could lead to potential bugs. 

We are currently creating another `ref` inside `useCombinedRefs` with the "merge" function inside a `useEffect`. This is not the best way to solve this as it can cause unintended consequences with the additional `ref` and rules/lifecycle of hooks inside `useEffect` .

It would be better to use `useCallback` instead and just call the refs sequentially inside the "merger" function.

Actions
* `useCallback` instead of `useEffect`
* remove the superfluous `targetRef`

Notes
* https://github.com/gregberge/react-merge-refs
* https://github.com/jaredLunde/react-hook/tree/master/packages/merged-ref

mimarz:
resolves #1347

### DIFF
--- a/libraries/core-react/src/components/Menu/MenuItem.tsx
+++ b/libraries/core-react/src/components/Menu/MenuItem.tsx
@@ -121,10 +121,11 @@ export const MenuItem = memo(
     return (
       <ListItem
         {...props}
-        ref={useCombinedRefs<HTMLLIElement>(
-          ref,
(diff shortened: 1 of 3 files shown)


pomfrida (approved):
LGTM 👍

PR #1359, opened 2021-05-19
Textarea component

### LINKED ISSUE #1357: Make (internal) Textarea component
Textarea is currently set as an `as` attribute inside `Input`. We need to move this logic into its own component to avoid lint errors and/or unnecessary calculations.

pomfrida:
Resolves #1357

### DIFF
--- a/libraries/core-react/src/components/Textarea/Textarea.tsx
+++ b/libraries/core-react/src/components/Textarea/Textarea.tsx
@@ -1,6 +1,6 @@
-import { forwardRef } from 'react'
+import { forwardRef, useState, TextareaHTMLAttributes } from 'react'
 import styled, { css } from 'styled-components'
-import * as tokens from '../Input/Input.tokens'
+import * as inputTokens from '../Input/Input.tokens'
 import type { InputToken } from '../Input/Input.tokens'
 import {
   typographyTemplate,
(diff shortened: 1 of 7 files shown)


libraries/core-react/src/components/Textarea/Textarea.tsx
```diff
@@ -0,0 +1,139 @@
...
+      disabled = false,
+      type = 'text',
+      rowsMax = 2,
+      ...other
+    },
+    ref,
+  ) {
+    const inputVariant = inputVariants[variant]
+    const spacings = tokens.comfortable.spacings
+    const [inputEl, setInputEl] = useState<HTMLTextAreaElement>(null)
```
mimarz, line 113: Suggest renaming this to appropriate name since its a `textarea` now 🤪

pomfrida: 👍  Done

mimarz (changes requested):
We do have some cleaning up to do, but I suggest we can do some renaming so its easier to understand whats going on until then?

`InputWrapper.ts` -> `Field.ts` ?
`StyledInput` -> `StrippedInput`
`StyledTextarea` -> `StrippedTextArea`

libraries/core-react/src/components/Textarea/Textarea.tsx
```diff
@@ -0,0 +1,139 @@
...
+export const Textarea = forwardRef<HTMLTextAreaElement, TextareaProps>(
+  function Textarea(
+    {
+      variant = 'default',
+      disabled = false,
+      type = 'text',
+      rowsMax = 2,
+      ...other
+    },
+    ref,
```
mimarz, line 109: ref is not used, we need to combine this using the `useCombinedRef` hook

pomfrida: 👍  Done

mimarz (approved):
LGTM 👍

PR #1361, opened 2021-05-20
Added filterOptions prop on SingleSelect

mariush2:
- Adds optional prop on SingleSelect component to disabled filtering of options

Solves #1065

### DIFF
--- a/libraries/core-react/src/components/Select/SingleSelect/SingleSelect.tsx
+++ b/libraries/core-react/src/components/Select/SingleSelect/SingleSelect.tsx
@@ -40,7 +40,7 @@ export type SingleSelectProps = {
   /** Callback for the selected item change
    * changes.selectedItem gives the selected item
    */
-  /** Allow for filtering of options to be turned off */
+  /** Allow for filtering of options when one is selected to be turned off */
   disableFilter?: boolean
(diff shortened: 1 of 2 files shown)


libraries/core-react/src/components/Select/SingleSelect/SingleSelect.tsx
```diff
@@ -40,6 +40,8 @@ export type SingleSelectProps = {
   /** Callback for the selected item change
    * changes.selectedItem gives the selected item
    */
+  /** Allow for filtering of options to be turned off */
```
mimarz, line 43: Suggest updating the description to include that its filtering of options when an option is already selected. Just to make it clear thats its not the filtering all together 🙈

mimarz (changes requested):
Looks good! Just a nitpick on description so its more detailed what the prop actually does!

mimarz:
The PR was withdrawn after testing together on call. When typing in new values in input it would not filter the list at all making the behaviour very similar to our existing `NativeSelect`. 

A possible solution could be to have "current" filtered item be highlighted with an active color in the list on re-opening, but this feature would entail a bit more work, and not feasible given time constraints at this moment

PR #1369, opened 2021-05-25
Fix rowsMax bug

### LINKED ISSUE #1367: Textfield does not follow rowsMax rules on initial render
Reported in FR issue here: https://github.com/equinor/design-system/issues/1336#issuecomment-845973674

pomfrida:
Resolves #1367

### DIFF
--- a/libraries/core-react/src/hooks/useAutoResize.ts
+++ b/libraries/core-react/src/hooks/useAutoResize.ts
@@ -15,14 +15,14 @@ export const useAutoResize = (
         newHeight = Math.max(targetEl.scrollHeight, newHeight)
         if (maxHeight) {
           newHeight = Math.min(maxHeight, newHeight)
+          if (targetEl.scrollHeight > maxHeight) {
+            console.log(targetEl.scrollHeight, newHeight, maxHeight)
+            targetEl.style.overflow = 'auto'
+          }
         }
         if (newHeight > targetEl.clientHeight) {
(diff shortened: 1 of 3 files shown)


libraries/core-react/src/hooks/useAutoResize.ts
```diff
@@ -14,13 +14,22 @@ export const useAutoResize = (
         newHeight = Math.max(targetEl.scrollHeight, newHeight)
         if (maxHeight) {
```
mimarz, line 15: Is this check necessary as you are already checking it on line 13?

mimarz (changes requested):
Suggest we remove `targetEl.style.height = 'auto'` from the autoReisze hook as its no longer needed?

libraries/core-react/src/hooks/useAutoResize.ts
pomfrida: `targetEl.style.height = 'auto'` is needed here too, for it to autoresize correctly when user deletes text. It's to rewrite the `height` when it's set before.
 
The nested `maxHeight` check is also needed, as you can see its `if` parent is `!maxHeight || maxHeight > newHeight`, this logic is needed for whenever `maxHeight > newHeight` only

pomfrida: Improvements fixed and tested so ready again for review  @mimarz

mimarz (approved):
LGTM 👍

PR #1373, opened 2021-05-26
Split fullWidth Button and regular Button

### LINKED ISSUE #1364: Wrong empty spaces in Button when an icon is used
When we introduced `fullWidth` we had to do some trickery internally with `flex` & `margins` to get icons to float out to the sides. This introduces some unwanted conseques with large empty spaces in our "normal" `Button` when using an `Icon` on one side. We thought we could live with it but we can't!

Design in [Figma](https://www.figma.com/file/0bGXR2sCwMVSDNyyzu5BXrO5/UI%E2%80%94User-Interface?node-id=17394%3A47297)

Actions
* We should introduce a separate "clone" of `Button` called `ButtonFullWidth` (or something like that) which we use instead when property `fullWidth` is set. This will lets us contain the h4xory for floating icons to the sides 🙈

pomfrida:
Resolves #1364

### DIFF
--- a/libraries/core-react/src/components/Button/ButtonFullWidth.tsx
+++ b/libraries/core-react/src/components/Button/ButtonFullWidth.tsx
@@ -1,0 +1,58 @@
+import { forwardRef, Children as ReactChildren, HTMLAttributes } from 'react'
+import styled from 'styled-components'
+
+const ButtonCenterContent = styled.span`
+  text-align: center;
+  flex: 1;
+`
+
+const ButtonInner = styled.span`
+  height: 100%;
+  display: flex;
+  align-items: center;
+
+  > img:first-child,
+  > svg:first-child {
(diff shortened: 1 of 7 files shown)


libraries/core-react/src/components/Button/ButtonFullWidth.tsx
```diff
@@ -0,0 +1,58 @@
...
+  > span:only-child {
+    margin-right: 0;
+    margin-left: 0;
+  }
+`
+
+export const ButtonFullWidth = forwardRef<
+  HTMLSpanElement,
+  HTMLAttributes<HTMLSpanElement>
+>(function ButtonFullWidth({ children, ...other }, ref) {
```
mimarz, line 47: Remove unused parameters

mimarz (approved):
LGTM 👍

PR #1378, opened 2021-05-27
🐛  Don't trigger autoResize if maxHeight is null

### LINKED ISSUE #1375: Textfield multiline resizes on input when having a height set
I have this multiline textfield that before 0.12.0 stayed this heigh after input
image

Not it looks like this when starting to type: 

image

I have set the height CSS element on <TextField to a specific height, I can see that once i start typing, it is set to Auto instead

pomfrida:
Resolves #1375


mimarz (approved):
LGTM 👍

PR #1398, opened 2021-06-03
✨ Compact Checkbox

### LINKED ISSUE #1033: Compact Checkbox
- [x] Remove paddings from the span to fit compact

- [x] Pressed overlay height is 40px but compact table height is 32px

- [x] Investigate use of multiple checkboxes in a table could use a single label header. (a11y)

pomfrida:
Resolves #1033


pomfrida:
@mimarz Not sure to leave the label as _not_ required, or if there's a smart way to allow a blank label when people are using `aria-labelledby `instead? 🤔  Should probably be careful with this as people may misuse or misunderstand

libraries/core-react/src/components/Checkbox/Checkbox.tsx
```diff
@@ -16,12 +16,15 @@ const StyledCheckbox = styled.label<StyledCheckboxProps>`
...
-  ${typographyTemplate(tokens.typography)}
+  ${typographyTemplate(tokens.typography)};
+  ${spacingsTemplate(tokens.entities.label.spacings)}
 `
 
 export type CheckboxProps = {
-  /** Label for the checkbox */
-  label: string
+  /** Label for the checkbox
+   * Do _not_ leave this blank unless you are using `aria-labelledby` instead!
```
mimarz, line 25: its also valid to use `aria-label`... I suggest we remove this reminder as its something thats part of the documentation or guidelines. Property descriptions should be short and descriptive of the property

mimarz (changes requested):
Looks good! Just a small suggested change on property description :)

libraries/core-react/src/components/Checkbox/Checkbox.tsx
pomfrida: Removed 🔥

mimarz (approved):
LGTM 👍

PR #1407, opened 2021-06-04
🐛  Pagination unwanted onChange type 

### LINKED ISSUE #1397: Pagination onChange callback has strange types
**Describe the bug**

When using the pagination component we've been using the callback to change page. This has broken at some point, probably because we didn't implement full typescript in EDS until recently. When checking the type of the callback it says it's either a Mouse or Keyboard Event AND a FormEventHandler. (see attached image).  

1. Use the pagination component
2. Try to use the onChange callback

**Expected behavior**

I assumed it would expect one type.

**Specifications**

- Version: EDS 0.12.1
- Browser: Chrome
- OS: Windows

**Additional context**

image

pomfrida:
Resolves #1397

### DIFF
--- a/libraries/core-react/src/components/Pagination/Pagination.tsx
+++ b/libraries/core-react/src/components/Pagination/Pagination.tsx
@@ -82,7 +82,7 @@ export type PaginationProps = {
   onChange?: (event: MouseEvent | KeyboardEvent, page: number) => void
   /** Default start page */
   defaultPage?: number
-} & HTMLAttributes<HTMLElement>
+} & Omit<HTMLAttributes<HTMLElement>, 'onChange'>
 
 export const Pagination = forwardRef<HTMLElement, PaginationProps>(
   function Pagination(


libraries/core-react/src/components/Pagination/Pagination.tsx
```diff
@@ -82,7 +82,7 @@ export type PaginationProps = {
   onChange?: (event: MouseEvent | KeyboardEvent, page: number) => void
```
mimarz, line 82: Suggest using the Reacts type for `event` in our custom event handler.

mimarz (changes requested):

libraries/core-react/src/components/Pagination/Pagination.tsx
pomfrida: This caused some problems so we agreed to let this fix be good for now. We should look at these types to conform with the native event types in a few more of our components.

mimarz (approved):
LGTM 👍

PR #1410, opened 2021-06-04
Update Menu hover state

### LINKED ISSUE #1363: Update hover state for the menu in dev
According to issue #1048 the hover state of the text in the menu has now be changed to bold

pomfrida:
Resolves #1363

### DIFF
--- a/libraries/core-react/src/components/Menu/Menu.stories.tsx
+++ b/libraries/core-react/src/components/Menu/Menu.stories.tsx
@@ -173,7 +173,7 @@ export const Complex: Story<MenuProps> = () => {
             size={16}
             color={colors.text.static_icons__tertiary.hex}
           />
-          <Typography group="navigation" variant="menu_title">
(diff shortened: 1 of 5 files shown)


pomfrida:
So this update ended up a bit more complex than expected. We have auto width menus so the width gets auto adjusted when text becomes bold on hover. I added a hack to avoid the auto width adjustment but this only works for simple menus (when there's only one child and the child is a string). 

Example of menu working as expected:
https://user-images.githubusercontent.com/31915755/120807460-41f7ad00-c548-11eb-9a96-fc0e9215701c.mov

Example of complex menu not working ideal (as there's more than one child and combination of different types):
https://user-images.githubusercontent.com/31915755/120807595-6a7fa700-c548-11eb-8d6d-5554a4bea310.mov

pomfrida:
Blocked by #1426

pomfrida:
Changes done as suggested in #1426 and ready for review now

mimarz (approved):
LGTM 👍

PR #1412, opened 2021-06-07
✨ Compact SingleSelect & MultiSelect

### LINKED ISSUE #1035: Compact Select
- [ ] Adjust height to fit compact

- [ ] Empty label needed to fit to compact, can column name act as label for screen readers?

- [ ] How to fix selects in narrow columns

mimarz:
resolves #1035

### DIFF
--- a/libraries/core-react/src/components/Select/Select.tokens.ts
+++ b/libraries/core-react/src/components/Select/Select.tokens.ts
@@ -15,7 +15,7 @@ const {
       medium_small: spacingMediumSmall,
       medium: spacingMedium,
       large: spacingLarge,
-      x_small,
+      xx_small,
     },
   },
   elevation: { temporary_nav: boxShadow },
@@ -64,7 +64,6 @@ export const select: ComponentToken = {
       },
       entities: {
         button: {
-          height: '24px',
           spacings: {
             left: spacingSmall,
             right: spacingSmall,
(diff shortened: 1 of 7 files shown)


libraries/core-react/src/components/Select/Select.tokens.ts
```diff
@@ -62,4 +83,12 @@ export const multiSelect: ComponentToken = R.mergeDeepRight(select, {
     left: spacingMediumSmall,
     right: spacingLarge,
   },
+  modes: {
+    compact: {
+      spacings: {
+        top: '2px',
```
pomfrida, line 89: Should we probably add `xx_small` (2px) here?

libraries/core-react/src/components/Select/Select.tokens.ts
```diff
@@ -53,6 +54,26 @@ export const select: ComponentToken = {
...
+    compact: {
+      spacings: {
+        left: spacingSmall,
+        right: spacingSmall,
+        top: spacingSmall,
+        bottom: spacingSmall,
+      },
+      entities: {
+        button: {
+          height: '24px',
```
pomfrida, line 67: No shape token for compact button? 🤔

pomfrida (approved):
Just some minor questions, LGTM otherwise 👍

libraries/core-react/src/components/Select/Select.tokens.ts
mimarz: Good point! Fixed 👌

libraries/core-react/src/components/Select/Select.tokens.ts
mimarz: Its actually not used. So removed :)

PR #1425, opened 2021-06-09
🐛  TextField toggling inputIcon prop breaks inputRef

### LINKED ISSUE #1405: TextField - toggling inputIcon prop breaks inputRef
**Describe the bug**

The TextField component break input refs when toggling the `inputIcon` prop. When toggling the prop, the component will internally in `Field` use conditional rendering to switch between different input tags. When this occurs, input refs are broken as the DOM element registered to the ref is unmounted and replaced with a new input tag. 

In our use case we would only want to display an input icon when the input is in an invalid state. Attempts at circumventing the conditional rendering by toggling between two icons, where one is hidden by css, resolves the issue with the ref, as the actual input element in the DOM never changes. However, it breaks the styling of the component. After the component has been focused once, it never loses its focus styling but combines it with its default non-focus styling.

1. Toggle the `inputIcon` prop of `<TextField>` on some condition, e.g., `inputIcon={invalid ? <Icon ... /> : undefined}`

I've made a [codesandbox](https://codesandbox.io/s/mystifying-tharp-918d1?file=/src/App.js) highlighting the issue. The first text field with label `focus issues` toggles between an icon and undefined and has the breaking ref problem. The second text field with label `styling issues` toggles between two icons, where one is hidden by css, and has the styling issue.

**Expected behavior**

(1) Toggling the `inputIcon` prop between an icon and undefined should not break the `inputRef`, or (2) toggling between two icons should not break the components styling.

**Specifications**

- Version: 12.0.1
- Browser: Google Chrome Version 91.0.4472.77
- OS: macOS Version 10.15.7

**Additional context**

There is also a visual bug in the `TextField` component with `default` variant and an input icon. The box shadow beneath the icon has a different color than the box shadow beneath the input element. See attached screenshot. Containing div gets style `box-shadow: inset 0 -1px 0 0 #007079` while the input element inside the div has style `box-shadow: inset 0px -1px 0px 0px var(--eds_text__static_icons__tertiary,rgba(111,111,111,1))`.
image

pomfrida:
Resolves #1405

### DIFF
--- a/libraries/core-react/src/components/TextField/Field.tsx
+++ b/libraries/core-react/src/components/TextField/Field.tsx
@@ -58,20 +58,27 @@ const StrippedTextarea = styled(Textarea)`
   }
 `
 
-type InputWithAdornmentsType = {
+type InputWrapperType = {
   isFocused: boolean
   isDisabled: boolean
   variant: string
   token: TextFieldToken
+  inputIcon?: ReactNode
+  unit?: string
 }
 
-export const InputWithAdornments = styled.div<InputWithAdornmentsType>`
-  display: flex;
-  align-items: center;
-  ${{
-    background: textfield.background,
-    paddingRight: textfield.spacings.right,
-  }}
+export const InputWrapper = styled.div<InputWrapperType>`
+  ${({ inputIcon, unit }) =>
+    (inputIcon || unit) &&
+    css`
+      display: flex;
+      align-items: center;
+      ${{
+        background: textfield.background,
+        paddingRight: textfield.spacings.right,
+      }}
+    `}
+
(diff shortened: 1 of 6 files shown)


libraries/core-react/src/components/TextField/Field.tsx
```diff
@@ -170,45 +186,30 @@ export const Field = forwardRef<
   }
 
   return (
-    <>
-      {inputIcon || unit ? (
```
pomfrida, line 174: This condition was the reason for the re-render of input when error handling was used

mimarz (approved):
LGTM 👍

PR #1448, opened 2021-06-21
Compact switch

### LINKED ISSUE #1432: Compact switch
edit: After reviewing `Switch`, the `small` variant/size will be used for density=compact.

Actions
* Update `Switch` small to be have correct height/width and clickbounds as per new "compact" rules

mimarz:
resolves #1432


.eslintrc.js
```diff
@@ -115,6 +115,12 @@ module.exports = {
         ],
         '@typescript-eslint/restrict-plus-operands': ['warn'],
         '@typescript-eslint/ban-ts-comment': ['warn'],
+        '@typescript-eslint/no-empty-interface': [
```
vnys, line 118: Why do we need this?

vnys (approved):
LGMT 👍

.eslintrc.js
mimarz: This is needed for the declaration file, [styled.d.ts](https://github.com/equinor/design-system/blob/746f719b3267e173bf5f4f18e5719fff840cae30/libraries/core-react/src/components/styled.d.ts) where we extend the type for `styled-components` `ThemeProvider`. https://styled-components.com/docs/api#create-a-declarations-file

PR #1449, opened 2021-06-22
🐛  Multiline TextField double box-shadow

### LINKED ISSUE #1447: Double box-shadow on multiline TextField (Textarea) in default state
This bug was apparently caused in the last fix https://github.com/equinor/design-system/pull/1425
Since that fix wasn't released yet, we need to prioritise fixing this bug before next release.

pomfrida:
Resolves #1447


mimarz (approved):
LGTM 👍

PR #1450, opened 2021-06-22
fix: Accordion displayName typo

karex:
Fixing a typo affecting the Accordion components' displayName property.


pomfrida:
Nice catch @karex and thanks a lot for contributing 👏

pomfrida (approved):
LGTM 👍

PR #1457, opened 2021-06-24
Optional label on Switch

### LINKED ISSUE #1445: Optional label in Switch
Make `label` optional so that it can be used with `aria-labeledby`.

Actions:
* Make property `label` optional
* If not `label` is provided, wrap the `Switch` in a `span` (instead of `label`). Maybe use styled `as` or `forwardedAs`

Part of #1436

mimarz:
resolves #1445


pomfrida (approved):
LGTM 👍

PR #1458, opened 2021-06-25
Optional Label for Checkbox

### LINKED ISSUE #1453: Optional label in Checkbox
We added optional label in Checkbox in #1033, but we need to also make sure that the enclosing element is a `span` and not `label` when no "label" is provided.

A part of #1436

pomfrida:
Optional label was already done so I just updated the Storybook with an example
Resolves #1453


mimarz (approved):
LGTM 👍

PR #1463, opened 2021-07-01
Disabled Link Button

### LINKED ISSUE #1456: button cannot be disabled if href is passed
**Describe the bug**

The `disabled` property on the `<Button>` does not work if a `href` is passed and the button is used as a link.

See: https://codesandbox.io/s/modest-grass-nz1ie?file=/src/App.js

**Expected behavior**

Button should either be able to be disabled when used as a link, or documentation should reflect this behaviour

pomfrida:
Resolves #1456 

~~Updated Storybook with an example of a workaround for disabled links~~

### DIFF
--- a/libraries/core-react/src/components/Button/Button.stories.tsx
+++ b/libraries/core-react/src/components/Button/Button.stories.tsx
@@ -271,6 +271,29 @@ export const Link: Story<ButtonProps> = () => (
   </Wrapper>
 )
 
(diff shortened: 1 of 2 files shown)


mimarz (commented):
Should this be a separate story? 🤔 Seems like a very specific edge-case...

pomfrida:
Now will a disabled button link become a native button so the component is consistent all over.

mimarz (approved):
LGTM 👍

PR #1464, opened 2021-07-01
[WIP] Stacking bug in Menu with React Portal

### LINKED ISSUE #1462: Version 0.12.0 breaks the stacking context of the Menu component
**Describe the bug**

Version 0.12.0 of EDS changed the rendering behavior of the `Menu` component. It is now attached to `document.body` using `React.Portal`. This changes the stacking context of the component and it will no longer be part of the stacking context where it is defined. E.g., if we now use `Menu` in conjunction with `TopBar`, where a header action opens a `Menu`, the header will now overlap and hide part of the `Menu` component.

1. Make an action in `TopBar` component open a `Menu` component. See CSB https://codesandbox.io/s/eds-stacking-context-bug-m1lmt

**Expected behavior**

`Menu` respects the stacking context where it is defined. I.e., `Menu` opens on top of and overlaps the `TopBar` component.

**Specifications**

- Version: 0.12.1
- Browser: Google Chrome Version 91.0.4472.114 
- OS: macOS Big Sur Version 11.4

pomfrida:
Resolves #1462

### DIFF
--- a/libraries/core-react/src/components/Menu/Menu.tsx
+++ b/libraries/core-react/src/components/Menu/Menu.tsx
@@ -120,17 +120,15 @@ export const Menu = forwardRef<HTMLUListElement, MenuProps>(function Menu(
 
   return (
     <>
-      {isMounted &&
-        ReactDom.createPortal(
-          <ThemeProvider theme={token}>
-            <MenuPaper elevation="raised" ref={setContainerEl} {...props}>
-              <MenuProvider>
-                <MenuContainer {...menuProps} ref={ref} />
-              </MenuProvider>
-            </MenuPaper>
-          </ThemeProvider>,
-          document.body,
-        )}
+      {isMounted && (
+        <ThemeProvider theme={token}>
+          <MenuPaper elevation="raised" ref={setContainerEl} {...props}>
+            <MenuProvider>
+              <MenuContainer {...menuProps} ref={ref} />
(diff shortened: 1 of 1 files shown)


pomfrida:
Looking at PR [#1360](https://github.com/equinor/design-system/pull/1360), the portal was probably introduced in regard to some accessibility issues, but when I removed it now all the checks in Storybook work. Might be something I am overlooking so can't wait 'till Mickey is back and can help me on Monday. Hopefully we don't need the portal so the stacking context bug is an easy fix. If not, we should probably try to render the `Menu`/portal in the same container as its parent and not in `document.body`.

mimarz:
> Looking at PR [#1360](https://github.com/equinor/design-system/pull/1360), the portal was probably introduced in regard to some accessibility issues, but when I removed it now all the checks in Storybook work. Might be something I am overlooking so can't wait 'till Mickey is back and can help me on Monday. Hopefully we don't need the portal so the stacking context bug is an easy fix. If not, we should probably try to render the `Menu`/portal in the same container as its parent and not in `document.body`.

Removing `z-index` on `Topbar` would also solve this problem. So question is which approach would be best in this use-case 🤔

mimarz:
@pomfrida I suggest we fix this by tweaking our `z-index` values (which are wrong for some components) #1466

PR #1470, opened 2021-07-06
ThemeProvider in TextField

### LINKED ISSUE #1469: Include ThemeProvider in TextField
@mimarz found a better solution for implementing compact theme so this should be consistent across all compact components.

pomfrida:
Resolves #1469

### DIFF
--- a/libraries/core-react/src/components/TextField/Field.tsx
+++ b/libraries/core-react/src/components/TextField/Field.tsx
@@ -122,9 +122,9 @@ const Adornments = styled.div<AdornmentsType>`
   & div:nth-child(2) {
     margin-left: ${textfield.spacings.left};
   }
-  margin-top: ${({ theme }) => theme.spacings.top};
-  ${({ multiline }) =>
+  ${({ multiline, theme }) =>
     multiline && {
+      marginTop: theme.spacings.top,
       alignSelf: 'start',
     }}
 `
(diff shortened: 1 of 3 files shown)


libraries/core-react/src/components/TextField/Field.tsx
```diff
@@ -125,13 +122,10 @@ const Adornments = styled.div<AdornmentsType>`
...
   }
-  ${({ multiline, density }) =>
+  margin-top: ${({ theme }) => theme.spacings.top};
+  ${({ multiline }) =>
     multiline && {
       alignSelf: 'start',
-      marginTop:
-        density === 'compact'
-          ? inputToken.modes.compact.spacings.top
-          : textfield.spacings.top,
```
pomfrida, line 134: Not sure why the `InputToken` is used as compact token here, while textfield token didn't have any compact modes? Added mode to textfield token because that's needed in the `useToken` hook.

libraries/core-react/src/components/TextField/Field.tsx
```diff
@@ -125,13 +122,10 @@ const Adornments = styled.div<AdornmentsType>`
   & div:nth-child(2) {
     margin-left: ${textfield.spacings.left};
   }
-  ${({ multiline, density }) =>
+  margin-top: ${({ theme }) => theme.spacings.top};
```
mimarz, line 125: This looks like it might cause a visual bug? `margin-top` was only applied if `multiline` was true...

mimarz (changes requested):

libraries/core-react/src/components/TextField/Field.tsx
pomfrida: Yes, you are right. Corrected 👍

mimarz (approved):
LGTM 👍

PR #1478, opened 2021-07-06
ThemeProvider in Input

### LINKED ISSUE #1477: Include ThemeProvider in Input

pomfrida:
Resolves #1477

### DIFF
--- a/libraries/core-react/src/components/Input/Input.tsx
+++ b/libraries/core-react/src/components/Input/Input.tsx
@@ -11,6 +11,8 @@ import type { Variants } from '../TextField/types'
 import { useEds } from '../EdsProvider'
 import { useToken } from '../../hooks'
 
+const { input } = tokens
+
 const Variation = ({ variant, theme }: StyledProps) => {
   if (!variant) {
     return ``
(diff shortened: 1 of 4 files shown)


libraries/core-react/src/components/Input/Input.tsx
```diff
@@ -9,10 +9,11 @@ import {
 } from '../../utils'
 import type { Variants } from '../TextField/types'
 import { useEds } from '../EdsProvider'
+import { useToken } from '../../hooks'
 
 const { input, inputVariants } = tokens
 
-const Variation = ({ variant, token, density }: StyledProps) => {
+const Variation = ({ variant, token }: StyledProps) => {
```
mimarz, line 16: `token` is the same as `theme` here, so suggest changing it to use theme instead of token.

libraries/core-react/src/components/Input/Input.tsx
```diff
@@ -103,6 +98,7 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(function Input(
 ) {
   const inputVariant = inputVariants[variant]
   const { density } = useEds()
+  const token = useToken({ density }, input)()
```
mimarz, line 101: last parameter here should be the `inputVariant` and not `input` as its whats used as "token" on line 108.

mimarz (changes requested):
Some superfluous code that can be removed now that we use `Themeprovider`

libraries/core-react/src/components/Input/Input.tsx
pomfrida: `input` is the token that has `compact`, `inputVariant` only has tokens in regard to the four different variants.

libraries/core-react/src/components/Input/Input.tsx
pomfrida: I tried to replace `token` with `theme` but it won't compile.. 🤔  I guess it could be because token=`inputVariant` and theme=`input`..?

pomfrida: Merged variant tokens with input token, so it works now

mimarz (changes requested):
Since we have already done so much refactoring on how the token is used I suggest we update `StyledInput` to use the function approach with `theme` we started adopting elsewhere. [Example in Button](https://github.com/equinor/design-system/blob/70b1231ece67ac0a8b9c0bdedb61f183a90b6ebe/libraries/core-react/src/components/Button/Button.tsx#L56). Should not be to much work as we can just copy the `Variation` function.

libraries/core-react/src/components/Input/Input.tsx
```diff
@@ -9,10 +9,9 @@ import {
...
 import { useEds } from '../EdsProvider'
+import { useToken } from '../../hooks'
 
-const { input, inputVariants } = tokens
-
-const Variation = ({ variant, token, density }: StyledProps) => {
+const StyledInput = styled.input(({ variant, theme }: StyledProps) => {
   if (!variant) {
     return ``
   }
```
mimarz, line 17: Suggest we remove this as we merged all css into one class and this escape is redundant now.

libraries/core-react/src/components/Input/Input.tsx
```diff
@@ -101,18 +84,22 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(function Input(
   { variant = 'default', disabled = false, type = 'text', ...other },
   ref,
 ) {
-  const inputVariant = inputVariants[variant]
+  const actualVariant = variant === 'default' ? 'input' : variant
+  const inputVariant = tokens[actualVariant]
```
mimarz, line 88: I am getting a type error here? 
image

mimarz (changes requested):

libraries/core-react/src/components/Input/Input.tsx
pomfrida: Lol me too! See last build for lint error

mimarz (approved):
sheep

PR #1485, opened 2021-07-13
Stacking Snackbar

### LINKED ISSUE #1172: Improve snackbar centering
Using pixels as default for centering an element is bad pratice. 

Use dynamic values such as, 50% or 50vw

Using just left makes placement wrong. Use portals on body instead?

### LINKED ISSUE #1173: Improve snackbars stacking context
The snackbar is not on its top-most layer. 

For example when testing in storybook it falls under other stories. 

image

Notes: 
* Maybe this is due to [stacking content](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context)? Since we are using transform on snackbar... 🤔

pomfrida:
Resolves #1173, resolves #1172

### DIFF
--- a/libraries/core-react/src/components/Snackbar/Snackbar.test.tsx
+++ b/libraries/core-react/src/components/Snackbar/Snackbar.test.tsx
@@ -2,39 +2,58 @@
 import {
   render,
   cleanup,
+  screen,
   waitForElementToBeRemoved,
 } from '@testing-library/react'
 import '@testing-library/jest-dom'
 import 'jest-styled-components'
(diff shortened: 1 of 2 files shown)


mimarz (changes requested):
`leftAlignFrom` does not seem to when using pixels vales aswell as `left` spacing not being applied? 

image

A follow-up after this PR would also be #1486

pomfrida:
This PR also fixes #1172

mimarz (approved):
Nice! 🍭

PR #1491, opened 2021-07-15
🔥  Remove internal EdsProvider in Table

### LINKED ISSUE #1489: Uniform usage of density in Table
Update `Table` to have same approach as other components for applying density using `EdsProvider`.

Actions:
* Remove internal `EdsProvider`
* Remove `density` property on `Table`
* Update `Table` examples with `EdsProvider` as parent element

mimarz:
resolves #1489


pomfrida (approved):
LGTM 👍

PR #1492, opened 2021-07-15
♻️ Improve Snackbar timeout

### LINKED ISSUE #1486: Improve Snackbar timeout
During `Tooltip` refresh we found a better way to handle timeouts (using `useRef`) in our components to avoid dropped timeouts. 

Actions:
* `Snackbar` should be updated to the improved way of handling timeouts

Notes:
* Checkout `Tooltip` for how this was [done]( https://github.com/equinor/design-system/blob/f35ac5202317c94fe2dbb8208fdbe9c017b5d1e0/libraries/core-react/src/components/Tooltip/Tooltip.tsx#L113)

pomfrida:
Resolves #1486


mimarz (approved):
LGTM 🍦

PR #1504, opened 2021-07-28
⚡ Spike using css variables in Button

### LINKED ISSUE #1494: Css variables theming
Test using *pure css* (css variables, with css tokens and css classes) for styling components in `eds-core-react`, instead of todays *css in javascript* with `styled-components`. Testing with `Button` component.

Approaches:
1 "Pure CSS"
2 "CSS in Javascript"
3 Hybrid 

### 1. Pure CSS
PRO's: 
* Its pure css 
  * No additional styling library dependency for our users such as `styled-components`
* Somewhat simpler and cleaner implementation since its only using known native tooling
   * New users don't need to learn how we write and use component tokens , how to use `Themeprovider` and `styled-components` that needs to be learned.
* CSS variables are powerful 
* Easier to start using web-components 
* Less code
* CSS utility classes can be used by our users

Middle:
* Naming conventions for css variables and css classes 🤔
  * How we use css variables can become messy with different people
  * BEM-like convention worked
* Maybe easier for theming?
* Maybe easier for other web technologies to use styling?
   * CSS classes are still somewhat tailor towards how react components are put together.
* Design tokens conversion to css will need to be updated and improved
* CSS-modules?

CON's: 
* No typesafety for tokens
   * This can be somewhat helped/mitigated with VsCode extensions such as; [CSS Variable Autocomplete](https://marketplace.visualstudio.com/items?itemName=vunguyentuan.vscode-css-variables)
* Combination of variants & colors can become fineky with having to use additional css classes or selectors. 
  * One example is `Button`; "contained + secondary" vs. "outlined + secondary"
* IE11 does not support css-variables (even though we don't support it)
* Postprocessing & autoprefixing will need to be setup
* Styling can not be tested (vs. css in javascript)

### 2. CSS in javascript
PRO's:
* Typesafety for tokens
* Easier style based on properties
* `styled-components` does post-processing and autoprefixing for us
* Styling is "closer" to components instead of using classes
* Styling can be tested*

Middle: 
* It works
* Some variants of components become bloated with CSS

CON's:
* Our users are forced to use `styled-components`
* Locked in to our react-components
* We have a token structure outside of components that needs to be learned
* Lots more javascript code is needed to achieve same effect as css variables

### 3. Hybrid (NOT TESTED)

Using todays token structure we could introduce a the component css-variables (as done in pure-css example) and just override their values the same way we do in pure-css but still using styled components and js tokens.

### Notes: 

* https://codepen.io/vnys/pen/vYNgGRX
* https://epicreact.dev/css-variables/

mimarz:
resolves #1494 

This PR serves as a spike for using plain css with tokens in css, instead of `styled-components` with tokens in js.


mimarz:
Moved to `eds-lab` for further spiking in #1841

PR #1523, opened 2021-08-17
Conform use of forwardRef and rest props

### LINKED ISSUE #810: Conform on the use of forwardRef and {...rest}

pomfrida:
Resolves #810


mimarz (approved):
LGTM 👍

PR #1524, opened 2021-08-17
♻️  Render breadcrumb as span when href is undefined

### LINKED ISSUE #1505: Render breadcrumb without <a> element when href is not set

The `<a>` element rendered by Breadcrumb href does not work with react-router and causes a full page reload.
To make it work, we use the NavLink component from react-router-dom inside the Breadcrumb component, instead of setting href on Breadcrumb.
The issue is that Breadcrumb renders an `<a>` element even when hrewf is not set, and with NavLink inside Breadcrumb, the `<a>` rendered by NavLink becomes a child of the `<a>` rendered by Breadcrumb.

A way to control if Breadcrumb should render an empty `<a>` element. Two options come to my mind:
1. Implicit: Do not wrap childer in `<a>` if href is empty
2. Explicit: a new property, e.g. alwaysRenderAnchor, to control if `<a>` should be rendered or not on empty href. Guess default should be true in order to keep current behaviour and prevent possible breaking changes for existing apps.
3. Implement **as="span"** the same way as done with the Button componenet

**Additional context**

Source
```html
<Breadcrumbs.Breadcrumb>
   <NavLink to={routeWithParams(routes.app, { appName })}>
     {appName}
   </NavLink>
</Breadcrumbs.Breadcrumb>
```
becomes
```html
<li class="sc-kYPZxB fTccnM">
  <a class="sc-eCApnc dlHHwl sc-jHcXXw dwudBu">
    <a aria-current="page" class="active" href="/applications/radix-job-demo">radix-job-demo</a>
  </a>
</li>
```

pomfrida:
Resolves #1505

### DIFF
--- a/libraries/core-react/src/components/Breadcrumbs/Breadcrumb.tsx
+++ b/libraries/core-react/src/components/Breadcrumbs/Breadcrumb.tsx
@@ -39,12 +39,12 @@ export const Breadcrumb = forwardRef<HTMLAnchorElement, BreadcrumbProps>(
       maxWidth,
     }
     const showTooltip = maxWidth > 0
-    const isValidLink = href !== undefined
+    const isLinkDefined = href !== undefined
 
     const crumb = (
       <StyledTypography
-        link={isValidLink}
-        as={isValidLink ? null : 'span'}
+        link={isLinkDefined}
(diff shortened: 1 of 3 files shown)


mimarz (approved):
LGTM 👍

PR #1526, opened 2021-08-20
Fix a11y issues in example page

### LINKED ISSUE #1513: Fix a11y issues on test page in Storybook

pomfrida:
Resolves #1513

### DIFF
--- a/libraries/core-react/stories/playground/DataTable.tsx
+++ b/libraries/core-react/stories/playground/DataTable.tsx
@@ -187,7 +187,9 @@ export const DataTable: Story<TableProps> = (args) => {
             Favourite
           </Cell>
           {columns.map((col) => (
-            <Cell key={`head-${col.accessor}`}>{col.name}</Cell>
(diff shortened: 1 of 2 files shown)


pomfrida:
Compact table form components does have some known accessibility issues, as we have discussed before when we decided to remove labels for compact form elements so they can fit nicely into the compact table. There's workarounds of course, using attributes `aria-labelledby` or `aria-label`, but it doesn't look like the a11y tester in Storybook cares about these. At least the warnings doesn't clear even though these attributes are defined...

mimarz:
> Compact table form components does have some known accessibility issues, as we have discussed before when we decided to remove labels for compact form elements so they can fit nicely into the compact table. There's workarounds of course, using attributes `aria-labelledby` or `aria-label`, but it doesn't look like the a11y tester in Storybook cares about these. At least the warnings doesn't clear even though these attributes are defined...

@pomfrida Does this also affect the `SingleSelect`? 
image

mimarz (approved):
LGTM 👍

PR #1528, opened 2021-08-24
✨ Snackbar placement

### LINKED ISSUE #1488: Snackbar: Replace leftAlignFrom prop with a placement prop
`leftAlignFrom` should be replaced by a `placement` prop as this makes more sense

pomfrida:
Resolves #1488


mimarz (changes requested):
Suggest sending `placement` prop and make a function thats used in the `StyledSnackbar` for positioning and css. Currently the logic is split into two separate logical evaluations making it harder to maintain and understand.

mimarz (approved):
LGTM 👍

PR #1534, opened 2021-09-01
🐛 Popover guard for onEscape

### LINKED ISSUE #1529: Guard check onEscape - Popover
From Slack:

> After the popover was upgraded, I was not aware of the escape behaviour, until an e2e test failed for no sane reason. Pressing escape in our app now opens a ton of popovers, because onClose was implemented as a toggle. Since this behaviour is global, could it be guarded by an if open check? I guess the intention is to close all popover on escape press.

pomfrida:
Resolves #1529


mimarz (approved):
LGTM 👍

PR #1536, opened 2021-09-01
 ♻️ Update Table tokens with compact Typography tokens

### LINKED ISSUE #1452: Compact Typography
Needed to align items better in a compact `Menu`

A compact `Typography` with `group="navigation"` and `variant="label"` should have height `12px` (is currently 15). (Or line height 1)

See `DataTable.stories.tsx` or https://github.com/equinor/design-system/pull/1451 for details, and remove the inline styles when this is implemented in Typography.

Update: After discussion with design we decided that line-height must be lowered to 1 for all Typography in compact. This has to be added to tokens: #1454

mimarz:
resolves #1452

### DIFF
--- a/libraries/core-react/src/components/Table/DataCell/DataCell.tokens.ts
+++ b/libraries/core-react/src/components/Table/DataCell/DataCell.tokens.ts
@@ -7,6 +7,7 @@ const {
       cell_text: cellTypography,
       cell_numeric_monospaced: cellNumericTypography,
     },
+    _modes: { compact: compactTypography },
   },
   colors: {
     text: {
@@ -105,9 +106,7 @@ const tableCell: ComponentToken = {
   modes: {
     compact: {
       height: '32px',
-      typography: {
-        lineHeight: '1rem',
-      },
+      typography: compactTypography.table.cell_text,
     },
   },
 }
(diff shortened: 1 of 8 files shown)


libraries/core-react/src/components/Table/DataCell/DataCell.tokens.ts
```diff
@@ -7,6 +7,7 @@ const {
       cell_text: cellTypography,
       cell_numeric_monospaced: cellNumericTypography,
     },
+    _modes: { compact: compactTypography },
```
pomfrida, line 10: Also add the compact cell_numeric_monospaced, here and in variants below

pomfrida (changes requested):
Missing compact `cell_numeric_monospaced`

libraries/core-react/src/components/Table/DataCell/DataCell.tokens.ts
mimarz: Its not missing as the merge function [here](https://github.com/equinor/design-system/blob/6f82d0f52b5fea513eb2cae4f76b7a2bc3e42d20/libraries/core-react/src/components/Table/DataCell/DataCell.tsx#L23) retains the `fontFeature` thats unique for numeric. But its maybe not to obvious, I can try to see if there is a more transparent way of doing this. Problem is we have multiple permutations with compact now 🤔

mimarz: I have altered the way the variants is applied. Check it out @pomfrida

pomfrida (approved):
LGTM 👍

PR #1571, opened 2021-09-20
🔨 Keep pnpm after prebuild

vnys:
When installing pnpm globablly using npm, it’s not available when the environment is prebuilt. Using a minimal dockerfile with pnpm installed with homebrew will (hopefully) fix the problem.

### DIFF
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -1,0 +1,5 @@
+FROM gitpod/workspace-full
+
+USER gitpod
+
+RUN brew install pnpm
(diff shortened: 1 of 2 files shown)


pomfrida (approved):
LGTM 👍

nilsml:
Homebrew in Docker? I guess then it works only on a Mac? @vnys

mimarz:
> Homebrew in Docker? I guess then it works only on a Mac? @vnys

Guess its included in the docker image `gitpod/workspace-full`. External OS has no affect on Docker, only the image you use inside the docker file. @nilsml

nilsml:
Yeah, I was just wondering if external OS could have something to do when actually building the image. Since Homebrew is not usually used inside Linux.

PR #1578, opened 2021-09-23
Fix/input readonly styles

### LINKED ISSUE #1162: Add missing read-only styling to Textfield
Textfield is missing the readonly styling when property `readOnly` is set to true.

oddvernes:
Fixes #1162 

Should ComponentToken be updated with a readOnly state? the only style change is transparent background and no box-shadow to make an underline so maybe it isn't neccessary.. 
I followed the behaviour of read only text input in material ui, where it still gets a visual focus

### DIFF
--- a/libraries/core-react/src/components/Input/Input.tokens.ts
+++ b/libraries/core-react/src/components/Input/Input.tokens.ts
@@ -54,6 +54,10 @@ export const input: InputToken = {
         color: disabled__text.rgba,
       },
     },
+    readOnly: {
+      background: 'transparent',
+      boxShadow: 'none',
+    },
     active: {
       outline: {
         type: 'outline',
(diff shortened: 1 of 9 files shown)


mimarz (changes requested):
I suggest updating `ComponentToken` with `readOnly` state as its a valid state for html elements (even when its just `input` & `textarea`)

One other suggestions is that could this be solved using [:read-only](https://developer.mozilla.org/en-US/docs/Web/CSS/:read-only)? This way we can avoid all the ternaries with `isReadOnly`.

oddvernes:
I tried :read-only already but it conflicts with disabled, that is element with disabled counts as :read-only. &[readOnly] does work but i don't know if that is preferred

mimarz:
> I tried :read-only already but it conflicts with disabled, that is element with disabled counts as :read-only. &[readOnly] does work but i don't know if that is preferred

ah, didn't know `:read-only` also styles `:disabled`. I would say using the attribute selector is valid. What do you think @oddvernes ?

mimarz (approved):
LGTM 👍

PR #1600, opened 2021-10-01
✨ Combobox data-model support with data property

### LINKED ISSUE #1416: Support different data-models in Single-/MultiSelect components
In 11/10 cases our selectionlists are based on name+id models `{ "name": "Select me", "id": "long-ass-guid" }`, but EDS only has support for simple strings. 

Being able to send in a different datatype, or to send in an Option render-prop where we can connect the id to the name ourselves.

mimarz:
resolves #1416

### DIFF
--- a/libraries/core-react/src/components/Combobox/Combobox.stories.tsx
+++ b/libraries/core-react/src/components/Combobox/Combobox.stories.tsx
@@ -1,18 +1,12 @@
 /* eslint-disable @typescript-eslint/no-unsafe-return */
 import { useState } from 'react'
-import {
-  Combobox,
-  ComboboxProps,
-  ComboboxChanges,
-  Button,
-  Typography,
-  EdsProvider,
-} from '../..'
+import { Combobox, ComboboxProps } from '../..'
 import { Story, Meta } from '@storybook/react/types-6-0'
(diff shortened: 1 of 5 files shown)


oddvernes:
Everything seems to be in working order. One thing though perhaps is screen reader is abit janky (windows narrator on microsoft edge) it reads every line twice and says 'selected' which is because the line is focused but not actually selected as in checked, toggling state between selected (checked) and not selected (unchecked) in a multiple the reader remains mute

oddvernes:
Testing voiceover on mac, and it is considerably better here. It reads each line only once, it reads when you deselect an option in multiple but not when you select an option

mimarz:
> Everything seems to be in working order. One thing though perhaps is screen reader is abit janky (windows narrator on microsoft edge) it reads every line twice and says 'selected' which is because the line is focused but not actually selected as in checked, toggling state between selected (checked) and not selected (unchecked) in a multiple the reader remains mute

Hmm, I only get it read once on MacOs (Voiceover). This has however been a challenge with downshift and comboboxes in general. The same faults can be found on `Single` & `Multi` selects today.

oddvernes (approved):
I think this is good enough, considering the limitations with downshift

PR #1622, opened 2021-10-14
Switch refactor label

### LINKED ISSUE #1613: Refactor Switch to without wrapped label when no label is present
Look to mui's checkbox for inspiration: https://mui.com/components/radio-buttons/#standalone-radio-buttons

Notes:
* Return container (`span`)  when `label` is not defined
* If `label` is defined we wrap container with a `label` and provided text

Actions:
* Create story for `Switch` with `Tootltip` and standalone `Label` as example

oddvernes:
resolves #1613


oddvernes:
> Maybe we should make an issue to replace the prop-drilling for `isDisabled` with an react context 🤔 What do you think?

Sounds like a potential case for it 👍

Issue #1665, opened 2021-10-29
📝Spike using storybook-addon-jest


oddvernes:
trying out this addon to see if it is useful during development

oddvernes:
The question is how to make this useful during development. I tried running storybook and test:watch at the same time but it gets into an infinite loop of rebuilding storybook

mimarz:
After testing the addon in storybook I agree that its kinda cumbersome to use with a lot of manual steps in Storybook. So no use for us as it can't beat having just having `test:watch` running in the terminal.

PR #1666, opened 2021-10-29
✨ addon-jest added to storybook

oddvernes:
ticket: #1665 

Had some problems getting the addon panel to appear in storybook. Turns out i had to click the "Change addons orientation" button and then it showed up. Maybe this is a bug in storybook.


mimarz:
Rejected as the plugin does not give us anything beneficial in having the test in storybook.

PR #1669, opened 2021-11-01
🗑️ Tabs replace lodash with useId

### LINKED ISSUE #1667: Replace Tabs lodash uniqueId with our own useId
Lodash can then be removed from dependencies as this is the only remaining place it is used

oddvernes:
Removes lodash dependency from project
resolves #1667


mimarz (approved):
LGTM 👍

Issue #1186, opened 2021-03-29
Collection state management hook


mimarz:
blocks #605 

We are currently maintaining two duplicate implementations of managing states when navigating a collection of items and which one is in focus/current. With navigation drawer also needing the same logic its counterintuitive to have yet another separate implementation of this.

Suggest creating or using an existing state hook for managing a collection of items and which is in focus as well as provide correct aria-props that should be applied

Affected components: 
* Navigation Drawer (not started)
* Menu
* Tabs

Inspiration:
* https://react-spectrum.adobe.com/react-stately/useTreeState.html
* https://github.com/sparksuite/react-accessible-dropdown-menu-hook

PR #1671, opened 2021-11-02
🐛 Fixed keyboard navigation when menuSection is present

### LINKED ISSUE #1668: Menu Item onClick not running correctly in scrolling container
**Describe the bug**

Clicking `Menu.Item` doesn't run onClick event as expected when dynamically creating Menu component with max-height and overflow: auto.

1. Create Menu component with `overflow: auto` and `max-height: 'some-value'`
2. Create Menu.Item's inside Menu.Section(s) (same bug appears if all the items are in the same section) 
3. Click Menu.Item
4. Menu scrolls to bottom instead of running onClick

**Expected behavior**

Clicking Menu.Item's inside Menu.Section's which are in turn inside a styled Menu-component with max-height and overflow should just run onClick event and not scroll to bottom on first click.

**Specifications**

- Version: 0.14.2
- Browser: Chrome 95.0.4638.54
- OS: macOS

**Additional context**

Codesandbox: [https://codesandbox.io/s/focused-wind-7flk0?file=/src/App.js](https://codesandbox.io/s/focused-wind-7flk0?file=/src/App.js)

Video:

https://user-images.githubusercontent.com/24507952/139652458-0fc5e07b-124f-4eb5-982b-66fa7e2fdaea.mov

u

oddvernes:
Resolves #1668 

The internal logic to add indexes to focusable children and then make them focusable with keyboard did not work at all with menuSections present. 
I have changed the code now to add index to menuItems inside sections, as well as menuItems that are not in sections. I then  made a separate flat array of children to build the focusableIndexs.

I also added event.preventDefault when pressing up/down arrows inside menu to stop scrolling the viewport.

On a sidenote it does not appear that memo works on menuItems, I guess because of the function props, so there are probably some more optimizations that can be done on this component

### DIFF
--- a/libraries/core-react/src/components/Menu/MenuList.tsx
+++ b/libraries/core-react/src/components/Menu/MenuList.tsx
@@ -47,13 +47,14 @@ export const MenuList = forwardRef<HTMLUListElement, MenuListProps>(
             (grandChild: ReactNode) => {
               index++
               return cloneElement(grandChild as MenuChild, { index })
-            })
+            },
+          )
           return cloneElement(child as MenuChild, null, updatedGrandChildren)
         } else {
           index++
           return cloneElement(child as MenuChild, { index })
         }
-      }
+      },
     )
 
     const flattenedChildren = ReactChildren.map(
@@ -64,7 +65,8 @@ export const MenuList = forwardRef<HTMLUListElement, MenuListProps>(
         } else {
           return child
         }
-      })
+      },
+    )
 
     const focusableIndexs: number[] = ((flattenedChildren as MenuChild[]) || [])
       .filter((x) => !x.props.disabled)
--- a/libraries/core-react/src/components/Menu/MenuList.tsx
+++ b/libraries/core-react/src/components/Menu/MenuList.tsx
@@ -33,48 +33,39 @@ type MenuChild = ReactElement<MenuItemProps> & ReactElement<MenuSectionProps>
 
 type Direction = 'down' | 'up'
 
+function isIndexable(item: MenuChild) {
+  if (isValidElement(item) && !item.props.disabled && item.type === MenuItem)
+    return true
+  return false
+}
+
 export const MenuList = forwardRef<HTMLUListElement, MenuListProps>(
   function MenuList({ children, focus, ...rest }, ref) {
     const { focusedIndex, setFocusedIndex } = useMenu()
 
     let index = -1
+    const focusableIndexs: number[] = []
+
     const updatedChildren: Array<MenuChild> = ReactChildren.map(
       children,
-      (child: ReactNode) => {
-        if ((child as MenuChild).type === MenuSection) {
+      (child: MenuChild) => {
+        if (child.type === MenuSection) {
           const updatedGrandChildren = ReactChildren.map(
-            (child as MenuChild).props.children,
-            (grandChild: ReactNode) => {
+            child.props.children,
+            (grandChild: MenuChild) => {
               index++
-              return cloneElement(grandChild as MenuChild, { index })
+              if (isIndexable(grandChild)) focusableIndexs.push(index)
+              return cloneElement(grandChild, { index })
             },
           )
-          return cloneElement(child as MenuChild, null, updatedGrandChildren)
+          return cloneElement(child, null, updatedGrandChildren)
         } else {
           index++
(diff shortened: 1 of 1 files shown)


mimarz (changes requested):

libraries/core-react/src/components/Menu/MenuList.tsx
```diff
@@ -45,17 +37,38 @@ export const MenuList = forwardRef<HTMLUListElement, MenuListProps>(
...
+          )
+          return cloneElement(child as MenuChild, null, updatedGrandChildren)
+        } else {
+          index++
+          return cloneElement(child as MenuChild, { index })
+        }
+      },
+    )
+
+    const flattenedChildren = ReactChildren.map(
```
mimarz, line 60: Memo?

libraries/core-react/src/components/Menu/MenuList.tsx
```diff
@@ -45,17 +37,38 @@ export const MenuList = forwardRef<HTMLUListElement, MenuListProps>(
...
-
+    let index = -1
     const updatedChildren: Array<MenuChild> = ReactChildren.map(
-      pickedChildren,
-      (child: ReactNode, index: number) =>
-        cloneElement(child as MenuChild, { index }),
+      children,
+      (child: ReactNode) => {
+        if ((child as MenuChild).type === MenuSection) {
+          const updatedGrandChildren = ReactChildren.map(
```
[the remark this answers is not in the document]
oddvernes: Since i need the original structure to pass on as children, I changed this to build the indexable list directly in the first loop instead.

libraries/core-react/src/components/Menu/MenuList.tsx
oddvernes: got rid of flattenedChildren entirely

libraries/core-react/src/components/Menu/MenuList.tsx
```diff
@@ -45,17 +37,38 @@ export const MenuList = forwardRef<HTMLUListElement, MenuListProps>(
   function MenuList({ children, focus, ...rest }, ref) {
     const { focusedIndex, setFocusedIndex } = useMenu()
 
-    const pickedChildren = isFragment(children)
-      ? (children as MenuChild).props.children
-      : children
-
+    let index = -1
     const updatedChildren: Array<MenuChild> = ReactChildren.map(
```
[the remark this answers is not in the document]
oddvernes: I'll try this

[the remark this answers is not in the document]
oddvernes: I added useMemo and now updatedChildren is only run when menu is opened and closed but not during scrolling as it did before

libraries/core-react/src/components/Menu/MenuList.tsx
```diff
@@ -41,28 +34,41 @@ type MenuChild = ReactElement<MenuItemProps> & ReactElement<MenuSectionProps>
...
-    const pickedChildren = isFragment(children)
-      ? (children as MenuChild).props.children
-      : children
-
-    const updatedChildren: Array<MenuChild> = ReactChildren.map(
-      pickedChildren,
-      (child: ReactNode, index: number) =>
-        cloneElement(child as MenuChild, { index }),
+    let index = -1
+    const focusableIndexs: number[] = useMemo<number[]>(() => [], [])
```
mimarz, line 48: Just curious, whats the benefit of putting memonizing this constant when is used from a memonized function?

mimarz (approved):
LGTM 👍

Its almost as if we could extract some of logic to a seperate hook later to fulfil #1186
`focusableIndexs` could have lived inside the react context in a hook/provider 🤔

libraries/core-react/src/components/Menu/MenuList.tsx
oddvernes: The linter told me to do it:
`The 'focusableIndexs' array makes the dependencies of useMemo Hook (at line 70) change on every render. To fix this, wrap the initialization of 'focusableIndexs' in its own useMemo() Hook.`

mimarz: aha, probably because its declared in the function, which means that it would be re-declared on every render 🤔

Issue #1672, opened 2021-11-03
Be able to change z-index of Menu component


jorgenholme96:
**Is your feature request related to a problem? Please describe.**

We need to change the z-index of the Menu component. However, we can't change the css of its parent element, MenuPaper, where the z-index is set.  Any styling set on the Menu component will be applied to the MenuContainer, which is wrapped by the MenuPaper.

**Describe the solution you'd like**

Could you make it possible to change this z-index?

PR #1673, opened 2021-11-03
✨Menu: className and style props are put on outermost element

oddvernes:
resolves: #1672 

This will allow users to overwrite styles on an element that is also used by popper to set position via `position`, `inset (top, left, bottom, right)` and `transform`, but I think it's fine, users will quickly realize if they try that..

### DIFF
--- a/libraries/core-react/src/components/Menu/Menu.tsx
+++ b/libraries/core-react/src/components/Menu/Menu.tsx
@@ -105,11 +105,9 @@ export const Menu = forwardRef<HTMLUListElement, MenuProps>(function Menu(
     4,
   )
 
-  const combinedStyles = { ...styles.popper, ...style }
-
   const props = {
     open,
-    style: combinedStyles,
+    style: { ...styles.popper, ...style },
     className,
     ...attributes.popper,
   }
--- a/libraries/core-react/src/components/Menu/Menu.tsx
+++ b/libraries/core-react/src/components/Menu/Menu.tsx
@@ -89,7 +89,7 @@ export type MenuProps = {
 } & HTMLAttributes<HTMLUListElement>
 
 export const Menu = forwardRef<HTMLUListElement, MenuProps>(function Menu(
-  { anchorEl, open, placement = 'auto', ...rest },
+  { anchorEl, open, placement = 'auto', style, className, ...rest },
   ref,
 ) {
   const [containerEl, setContainerEl] = useState<HTMLElement>(null)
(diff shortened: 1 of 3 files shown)


oddvernes:
Actually, is there a reason why `...rest` is sent to the list inside in the first place?

mimarz:
> Actually, is there a reason why `...rest` is sent to the list inside in the first place?

I guess its for the components props to get where they are supposed to as they are are used on different parts in the component-tree. Question is where should our users expect changes to occur when they supply their own `className` and `style`. I am thinking it should be the outermost elements so that our users can set margins, css-grid, flex etc used for positioning, but this might ofcourse limit the possibility of overriding other styling. I have a gut feeling we need to check all our components to make sure our components are behaving correctly in this manner 🙈

oddvernes:
> I guess its for the components props to get where they are supposed to as they are are used on different parts in the component-tree. Question is where should our users expect changes to occur when they supply their own `className` and `style`. I am thinking it should be the outermost elements so that our users can set margins, css-grid, flex etc used for positioning, but this might ofcourse limit the possibility of overriding other styling. I have a gut feeling we need to check all our components to make sure our components are behaving correctly in this manner 🙈

As long as the outermost element is selectable with css then everything inside is selectable as well with > div > ul etc, of course with the  caveat that internal refactors from us will make breaking changes if you do this. But then again I hope this isn't a commonplace thing to do.

mimarz:
> > I guess its for the components props to get where they are supposed to as they are are used on different parts in the component-tree. Question is where should our users expect changes to occur when they supply their own `className` and `style`. I am thinking it should be the outermost elements so that our users can set margins, css-grid, flex etc used for positioning, but this might ofcourse limit the possibility of overriding other styling. I have a gut feeling we need to check all our components to make sure our components are behaving correctly in this manner 🙈
> 
> As long as the outermost element is selectable with css then everything inside is selectable as well with > div > ul etc, of course with the caveat that internal refactors from us will make breaking changes if you do this. But then again I hope this isn't a commonplace thing to do.

I've made an issue #1674 for us to investigate this. If it was wrong in `Menu`, it might be wrong elsewhere also 🤔

libraries/core-react/src/components/Menu/Menu.tsx
```diff
@@ -105,9 +105,12 @@ export const Menu = forwardRef<HTMLUListElement, MenuProps>(function Menu(
     4,
   )
 
+  const combinedStyles = { ...styles.popper, ...style }
```
mimarz, line 108:
```suggestion
```

libraries/core-react/src/components/Menu/Menu.tsx
```diff
@@ -105,9 +105,12 @@ export const Menu = forwardRef<HTMLUListElement, MenuProps>(function Menu(
...
+
   const props = {
     open,
-    style: styles.popper,
+    style: combinedStyles,
```
mimarz, line 112:
```suggestion
    style: { ...styles.popper, ...style },
```

mimarz (approved):
LGTM, one suggestion on just putting the style merging in the props declaration.

PR #1701, opened 2021-11-16
🐛  Fix focusIndex on close for Menu

### LINKED ISSUE #1681: Menu items focus jump on subsequent open
Found a small visual bug when testing using keyboard navigation and focused elements. 

Steps:
* `Menu` has been opened
* Focus item (higher than 0 index)
* Close `Menu`
* Reopen `Menu`
* Focus frame jumps from previous focused item to first item

Notes:
* Should focus stay on item when `Menu` is opened/closed?

Actions:
* This should be fixable by setting the `focusedIndex` to -1 in the context when `Menu` is closed. Might be doable using the `useEffect` in `MenuContainer` to set the `focusedIndex` when property `open` is `false`

mimarz:
fixes #1681

### DIFF
--- a/libraries/core-react/src/components/Menu/Menu.context.tsx
+++ b/libraries/core-react/src/components/Menu/Menu.context.tsx
@@ -35,7 +35,11 @@ export const MenuProvider = ({ children }: ProviderProps): JSX.Element => {
   }
 
   const setOnClose: UseMenu<State>['setOnClose'] = (onClose) => {
-    setState((prevState) => ({ ...prevState, onClose }))
+    const onCloseHelper = () => {
+      onClose()
+      setFocusedIndex(-1)
+    }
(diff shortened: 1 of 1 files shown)


oddvernes (approved):
LGTM 👍 

On a sidenote, should focus be returned to the button after pressing escape in the menu? Right now it ends up somewhere else and i have to press tab repeadedly to get back to the button in storybook

mimarz:
> > LGTM 👍
> > On a sidenote, should focus be returned to the button after pressing escape in the menu? Right now it ends up somewhere else and i have to press tab repeadedly to get back to the button in storybook
> 
> Good catch! According to [best practices](https://www.w3.org/TR/wai-aria-practices-1.1/#menu) it should return to the element that opened the Menu

Created buge issue #1704

laliq (approved):
LGTM 👍

PR #1703, opened 2021-11-16
Added central aligment for TopBar Actions

### LINKED ISSUE #1680: Tobar.Actions center align items
When using the `Topbar.Actions`, items inside are not aligned center. Most evident when using different item types such as `Button` + `Search` or `Icon` together.

Suggest we add a center alignment for ease of use.

laliq:
resolve #1680

### DIFF
--- a/libraries/core-react/src/components/TopBar/Actions.tsx
+++ b/libraries/core-react/src/components/TopBar/Actions.tsx
@@ -6,6 +6,8 @@ export type TopbarActionsProps = HTMLAttributes<HTMLDivElement>
 const StyledActions = styled.div`
   grid-area: right;
   text-align: right;
+  align-items: center;
+  display: flex;
 `
 
 export const Actions = forwardRef<HTMLDivElement, TopbarActionsProps>(


oddvernes:
Is this about vertical alignment? the original issue doesn't specify

mimarz:
> Is this about vertical alignment? the original issue doesn't specify

Horziontal alignment. So that items put directly in Actions are aligned horizontally with eachother.

oddvernes:
Ok because align-items is vertical alignment when flex-direction is row. Not sure what is best, force justify-content: space-between + gap and such on the users or let them define the container behaviour themselves with their own css?

vnys:
I’m a bit puzzled too… the icons in the actions can’t be aligned horizontally when they’re aligned to the same vertical axis, that would make them overlap 🤔 So then I’m assuming you mean vertically. But that doesn’t make any sense either since all the icons have equal heights. So what are we trying to achieve?

oddvernes (approved):
LGTM! 👍

PR #1710, opened 2021-11-18
📝  Add compact mode stories

### LINKED ISSUE #1707: Don’t use the density prop directly in Storybook
In the stories in Storybook we use `compact` directly as a value for the `density` prop. This is just the thing we _don’t_ want users to do – so these stories need to be updated asap.

Example: https://eds-storybook-react-development.azurewebsites.net/?path=/docs/components-button--default#compact

vnys:


mimarz:
Looks like [@storybook/client-api ](https://github.com/storybookjs/storybook/tree/next/lib/client-api) is deprecated, or I am i reading it wrong?

vnys:
> Looks like [@storybook/client-api ](https://github.com/storybookjs/storybook/tree/next/lib/client-api) is deprecated, or I am i reading it wrong?

It looks like it _will_ be deprecated, as that is on the `next` branch.

mimarz:
> > Looks like [@storybook/client-api ](https://github.com/storybookjs/storybook/tree/next/lib/client-api) is deprecated, or I am i reading it wrong?
> 
> It looks like it _will_ be deprecated, as that is on the `next` branch.

Looking at the documentation for `@storybook/client-api` it looks like `@storybook/api` provides the same functionality.

libraries/core-react/src/components/EdsProvider/EdsProvider.stories.mdx
```diff
@@ -0,0 +1,25 @@
...
+  component={EdsProvider}
+  args={{  density: 'comfortable' }} />
+
+export const Abbr = styled.abbr`
+  cursor: help;
+`
+
+# EdsProvider
+
+Use the `<EdsProvider />` to provide users with the option of compressing the user interface to see more information at the same time without having to scroll.
```
mimarz, line 16: We have not talked to much on how we formulate our stories in Storybook, but I think this is to much text to have as an intro to the component, i felt like being hit by a wall of text and had to strive to read through it.

This feels like more of an explanation of _why_ we have a density property in EDS and not just the components, but ofcourse, there are other reasons for why this is written here not instead 🙈

Maybe the intro could be short and consice that the `EdsProvdier` is a provider for setting "settings" for the eds react components, and then have a new heading with "Accessibility" (which we should have on all our stories in the future imho) where we explain the importancy of switcher for `density`

libraries/core-react/src/components/EdsProvider/EdsProvider.stories.mdx
```diff
@@ -0,0 +1,25 @@
+import { Meta, Story, Canvas, ArgsTable } from '@storybook/addon-docs'
+import styled from 'styled-components'
+import { EdsProvider } from '../..'
+import * as stories from './EdsProvider.stories.tsx'
+
+<Meta title="Components/EdsProvider"
+  component={EdsProvider}
+  args={{  density: 'comfortable' }} />
```
mimarz, line 8: I don't think you need to set the explicitly? Should be derived from default value of props.

libraries/core-react/src/components/EdsProvider/EdsProvider.stories.tsx
```diff
@@ -0,0 +1,187 @@
...
+} from '../..'
+
+const Wrapper = styled.div`
+  margin: 32px;
+  display: grid;
+  grid-gap: 32px;
+  grid-template-columns: repeat(4, fit-content(100%));
+`
+
+export const Default: Story<EdsProviderProps['density']> = (args) => {
```
mimarz, line 22: Why should we hide the `children` property? Its just as important to communicate is required in the props table.

libraries/core-react/src/components/EdsProvider/EdsProvider.stories.tsx
```diff
@@ -0,0 +1,187 @@
...
+  const [_, setState] = useArgs()
```
mimarz, line 23: I assume this was to update the "props table"? Didn't seem to work on my machine as I had no controls for this prop in the props table as we usually have...

libraries/core-react/src/components/EdsProvider/EdsProvider.stories.tsx
```diff
@@ -0,0 +1,187 @@
...
+          <Button variant="ghost_icon">
+            <Icon data={menu} title="Ghost icon menu"></Icon>
+          </Button>
+        </Wrapper>
+      </EdsProvider>
+    </div>
+  )
+}
+
+const sourceCode = `const [state, setState] = useState<{
```
mimarz, line 101: My initial impression is that this is a bad bad idea in terms of maintenance and keep the preview updated, why do this? I get that this was to have a clean sourc-code example but it feels excessive and prone to errors when trying to maintane updated code examples.

Maybe just have the `density` state in the actual code and comment where you set the `setArgs` with that its a storybook thing?

mimarz (changes requested):
We should probably have a talk on how we do the wording on the "developer documentation".   Added some suggestions. We need a "Developer Writer" 😂

Also I could not get the props table to work, making the use of the storybook-api redunant as I it was not interactable for me. 🤔

vnys:
There’s a reason I said in the standup this morning we should do a code review on Teams 😁 There’s much to discuss here. I’ll call for a meeting.

vnys:
> Looking at the documentation for `@storybook/client-api` it looks like `@storybook/api` provides the same functionality.

Good, I’ll update that then.

vnys:
Closing this issue now that Zeroheight is up and running. Made a separate issue for documenting the EdsProvider. (#1814)

PR #1711, opened 2021-11-19
 :recycle:  Card.Media - introducing  grid, adding Card.Content

laliq:
The grid was created. I =n this stage there is question why `Card` component doesn't have`Card.Content` component?

### DIFF
--- a/libraries/core-react/src/components/Card/CardMedia.tsx
+++ b/libraries/core-react/src/components/Card/CardMedia.tsx
@@ -11,8 +11,13 @@ export type CardMediaProps = {
 } & HTMLAttributes<HTMLDivElement>
 
 const StyledCardMedia = styled.div<CardMediaProps>`
-  display: flex;
-  width: 100%;
+  /* MARTA */
+  display: grid;
+  place-items: center;
+  grid-area: yyy;
+  /* MARTA */
+  /* display: flex; */
+  /* width: 100%; */
   &:last-child {
     margin-bottom: 8px;
     /* Last child to have 24px total spacing to bottom */
@@ -21,9 +26,10 @@ const StyledCardMedia = styled.div<CardMediaProps>`
     fullWidth
       ? css`
           > * {
-            width: calc(100% + ${spacings.left} + ${spacings.right});
+            width: 100%;
+            /* width: calc(100% + ${spacings.left} + ${spacings.right});
             margin-left: -${spacings.left};
-            margin-right: -${spacings.right};
+            margin-right: -${spacings.right}; */
           }
 
           &:first-child {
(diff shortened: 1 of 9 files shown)


libraries/core-react/src/components/Card/CardMedia.tsx
```diff
@@ -37,6 +34,7 @@ const StyledCardMedia = styled.div<CardMediaProps>`
           }
         `
       : css`
+          padding: 0 16px;
```
mimarz, line 37: Suggest updating this to use the token `spacings` values so its updates correctly if spacings change (the same way you've done in the other components).

libraries/core-react/src/components/Card/CardMedia.tsx
```diff
@@ -11,21 +11,18 @@ export type CardMediaProps = {
 } & HTMLAttributes<HTMLDivElement>
 
 const StyledCardMedia = styled.div<CardMediaProps>`
-  display: flex;
-  width: 100%;
+  display: grid;
```
mimarz, line 14: nitpicking, but I think `place-items` still works with `display:flex`?

libraries/core-react/src/components/Card/CardMedia.tsx
```diff
@@ -11,21 +11,18 @@ export type CardMediaProps = {
...
+  place-items: center;
   &:last-child {
-    margin-bottom: 8px;
+    padding-bottom: 24px;
```
mimarz, line 17: Should images still have a bottom spacing of `24px` when they are last in the Card? @equinor/eds-ux

mimarz (changes requested):
Looks very good! 👍 

Just some small nitpicking and cleaning up feedbacks. I think we need to figure out what the correct spacing for the last image should be with @equinor/eds-ux

libraries/core-react/src/components/Card/CardMedia.tsx
laliq: `width:auto` was added and, `display:flex` is back. It looks that component doesn't need `place-items`

libraries/core-react/src/components/Card/CardMedia.tsx
laliq: Good point, done.

benteljones:
https://www.figma.com/file/MDkIHZdkMUw23EKI5HT83T/Bente?node-id=169%3A53764

libraries/core-react/src/components/Card/CardMedia.tsx
mimarz: This will be looked at in #1713

mimarz (approved):
Great work! 🎉👍

PR #1714, opened 2021-11-23
🐛♻️ Replacing ul>li with div>button fixes keyboard enter-as-click

### LINKED ISSUE #1702: MenuItem does not active on Enter keypress
When testing the `Menu` component with keyboard navigation, the selected menu item did not close or active on keypress. This is in violation of how expected behaviour for menu should be according to [aria best practices](https://www.w3.org/TR/wai-aria-practices-1.1/#menu)

oddvernes:
resolves #1702 

In the menu, we want keyboard 'enter' to behave the same as clicking an item with the mouse. By replacing the `li` element with a `button` element we get this `onClick` behaviour with keyboard for free

### DIFF
--- a/libraries/core-react/src/components/Menu/MenuItem.tsx
+++ b/libraries/core-react/src/components/Menu/MenuItem.tsx
@@ -28,7 +28,7 @@ type StyleAttrsProps = {
   isFocused: string
 }
 
-const ListItem = styled.button.attrs<StyleAttrsProps>(({ isFocused }) => ({
+const Item = styled.button.attrs<StyleAttrsProps>(({ isFocused }) => ({
   role: 'menuitem',
   tabIndex: isFocused ? -1 : 0,
 }))<StyleProps>`
@@ -125,7 +125,7 @@ export const MenuItem = memo(
     }
 
     return (
-      <ListItem
+      <Item
         {...props}
         ref={useCombinedRefs<HTMLButtonElement>(ref, (el) => {
           if (el !== null && isFocused) {
@@ -143,7 +143,7 @@ export const MenuItem = memo(
         }}
       >
         <Content>{children}</Content>
-      </ListItem>
+      </Item>
     )
   }),
 )
--- a/libraries/core-react/src/components/Menu/MenuItem.tsx
+++ b/libraries/core-react/src/components/Menu/MenuItem.tsx
@@ -134,7 +134,7 @@ export const MenuItem = memo(
         })}
         onFocus={() => toggleFocus(index)}
         onClick={(e) => {
-          if (!disabled && onClick) {
+          if (onClick) {
             onClick(e)
             if (onClose !== null) {
               onClose(e)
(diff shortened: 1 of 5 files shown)


libraries/core-react/src/components/Menu/MenuSection.tsx
```diff
@@ -6,7 +6,7 @@ import { spacingsTemplate } from '../../utils'
 import { Divider } from '../Divider'
 import { Typography } from '../Typography'
 
-const ListItem = styled.li.attrs(() => ({
+const ListItem = styled.div.attrs(() => ({
```
mimarz, line 9: Suggest changing this to `Title` or `Header` something more descriptive as its a container for section name

libraries/core-react/src/components/Menu/MenuItem.tsx
```diff
@@ -124,7 +127,7 @@ export const MenuItem = memo(
     return (
       <ListItem
         {...props}
-        ref={useCombinedRefs<HTMLLIElement>(ref, (el) => {
+        ref={useCombinedRefs<HTMLButtonElement>(ref, (el) => {
           if (el !== null && isFocused) {
             el.focus()
           }
```
mimarz, line 133: Do we still need the `!disabled` check on line 137 now that its a `button`  element, the onClick handler should not be triggered when disabled.

mimarz (changes requested):
Looks good! 👍 

Just some small naming and cleanup now that its button!
Did you check a11y semantic rules for changing `ul>li` to `div>button` (all I could find was that roles `menu`, `menuitem`, `aria-expanded` etc needed to be assigned)?

oddvernes:
> Did you check a11y semantic rules for changing ul>li to div>button (all I could find was that roles menu, menuitem, aria-expanded etc needed to be assigned)?

The roles menu and menuitem are already assigned so I don't think there is anything else needed. I can try with screenreader on the new and old solution later to see if there is a difference

libraries/core-react/src/components/Menu/MenuSection.tsx
oddvernes: agreed, done 👍

libraries/core-react/src/components/Menu/MenuItem.tsx
oddvernes: Good point, we did not need it

libraries/core-react/src/components/Menu/MenuItem.tsx
```diff
@@ -28,10 +28,12 @@ type StyleAttrsProps = {
   isFocused: string
 }
 
-const ListItem = styled.li.attrs<StyleAttrsProps>(({ isFocused }) => ({
+const ListItem = styled.button.attrs<StyleAttrsProps>(({ isFocused }) => ({
```
mimarz, line 31: Rename to `Item`?

mimarz (approved):
LGTM 👍

libraries/core-react/src/components/Menu/MenuItem.tsx
oddvernes: Sure!

PR #1719, opened 2021-11-24
🚸 Menu: Return focus to anchor after escape/enter keypress

### LINKED ISSUE #1704: Menu focus does not return on anchor element when closed
When `Menu` is closed when using keyboard navigation, the anchor element should be focused or next in line for focus according to https://www.w3.org/TR/wai-aria-practices-1.1/#menu

oddvernes:
Resolves #1704

### DIFF
--- a/libraries/core-react/src/components/Menu/Menu.tsx
+++ b/libraries/core-react/src/components/Menu/Menu.tsx
@@ -64,6 +64,13 @@ const MenuContainer = forwardRef<HTMLDivElement, MenuContainerProps>(
     useGlobalKeyPress('Escape', () => {
       if (open && onClose !== null) {
         onClose()
+        anchorEl.focus()
+      }
+    })
+
+    useGlobalKeyPress('Enter', () => {
+      if (open && onClose !== null) {
+        if (window.document.contains(anchorEl)) anchorEl.focus()
       }
     })
 


libraries/core-react/src/components/Menu/Menu.tsx
```diff
@@ -64,6 +64,13 @@ const MenuContainer = forwardRef<HTMLDivElement, MenuContainerProps>(
     useGlobalKeyPress('Escape', () => {
       if (open && onClose !== null) {
         onClose()
+        anchorEl.focus()
+      }
+    })
+
+    useGlobalKeyPress('Enter', () => {
+      if (open && onClose !== null) {
+        if (window.document.contains(anchorEl)) anchorEl.focus()
```
mimarz, line 73: Should we also call the `onClose()` here?

mimarz (changes requested):

libraries/core-react/src/components/Menu/Menu.tsx
oddvernes: This is handled by MenuItem

mimarz: Ah, because its a `button` now right 🙈

mimarz (approved):
LGTM 👍

PR #1721, opened 2021-11-24
🐛 Fix Prettier bug

### LINKED ISSUE #1718: The prettier extension in VS Code does not use prettier from the project
**Describe the bug**

When using formatOnSave there is a particular line that ends up being formatted in a way that prettier complains about after formatting, even though prettier is the formatter. The code block in question is this one:

```tsx
export const Compact: Story<RadioProps> = () => {
  const [density, setDensity] = useState<EdsProviderProps['density']>(
    'comfortable',
  )
}
```

…which when running the format document command in VS Code gets formatted like this: 

```tsx
export const Compact: Story<RadioProps> = () => {
  const [density, setDensity] =
    useState<EdsProviderProps['density']>('comfortable')
}
```

This leads me to believe there are different versions of prettier at play here…

vnys:
- Pins the (root and Figma Broker) prettier version as recommended by the docs
- Removes additional prettier instance in core-react
- Adds resolution setting so the VS Code extension uses the locally installed module

resolves #1718


vnys:
How ironic that it fails on eslint 😁 This will be fun…

oddvernes:
Seems to be working for me now, no more broken formatting on save..

mimarz:
Figma-broker also has a version of prettier installed, its used for formatting the output written to files... @vnys

oddvernes:
..But I also get the same linting errors and warnings as the lint check

vnys:
I don’t think we can avoid loads of reformatted files now… not sure what the best order would be, as it will affect all the other pull requests.

mimarz:
> I don’t think we can avoid loads of reformatted files now… not sure what the best order would be, as it will affect all the other pull requests.

Looks like its only four files with errors, and looking at which files it should be ok to fix linting...

oddvernes:
None of the errors conflict with tree-shaking, but the warning from rollup.config I have already fixed there

mimarz (approved):
LGTM 👍

PR #1730, opened 2021-11-26
🐛  Fix on close Scrim with esc button

laliq:
isDismissable is added to Scrim storybook, to close scrim with esc button.


oddvernes (approved):
LGTM 👍

PR #1735, opened 2021-11-29
♻️Add spacing in Card.Media if fullWidth is false

### LINKED ISSUE #1713: Adjust Card.Media spacing
We discovered inconsistencies for CardMedia in Card when its the last child that we need to adjust.

laliq:
resolves #1713

### DIFF
--- a/libraries/core-react/src/components/Card/CardMedia.tsx
+++ b/libraries/core-react/src/components/Card/CardMedia.tsx
@@ -38,6 +38,12 @@ const StyledCardMedia = styled.div<CardMediaProps>`
         `
       : css`
           padding: 0 ${spacings.right} 0 ${spacings.left};
+          &:first-child {
+            padding: ${spacings.top} ${spacings.right} 0 ${spacings.left};
+          }
+          &:last-child {
+            padding: 0 ${spacings.right} ${spacings.bottom} ${spacings.left};
+          }
           > * {
             width: 100%;
           }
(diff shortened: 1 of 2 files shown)


libraries/core-react/src/components/Card/CardMedia.tsx
```diff
@@ -14,8 +14,10 @@ const StyledCardMedia = styled.div<CardMediaProps>`
   display: flex;
   width: auto;
   &:last-child {
-    padding-bottom: 24px;
-    /* Last child to have 24px total spacing to bottom */
+    img {
+      border-bottom-right-radius: ${border.type === 'border' && border.radius};
+      border-bottom-left-radius: ${border.type === 'border' && border.radius};
+    }
```
mimarz, line 20: This should probably be within the css block for when `fullWidth` is true to stay consistent with when image is top or bottom, since now it has round borders even when `fullWidth` is `false`.

mimarz (changes requested):
The descriptions for the full with stories should be updated as they talk about specifics which are no longer true, such as bottom 24px padding when image is bottom...

Suggest removing `margin-top: -${spacings.top};` on line 29, as this interferes with current spacing model.

Awaiting response for comment [here](https://github.com/equinor/design-system/issues/1713#issuecomment-981528893) wether we need more padding on bottom & top when fullwidth is false.

mimarz (approved):
LGTM 👍

PR #1741, opened 2021-11-30
♻️ Clean redundant code for btn in Search

### LINKED ISSUE #1737: Search clear button
The "clear button" in our `Search` component is not a button and thus fails a11y. Should be updated to use the `Button` so its focusable and readable.

laliq:
resolves #1737

In Search.tokens `button` entity was created,  with defined height, width and spacing.

In Search component:
* `Button` component was imported. 
* New `button` entity was added.

### DIFF
--- a/libraries/core-react/src/components/Search/Search.tsx
+++ b/libraries/core-react/src/components/Search/Search.tsx
@@ -9,6 +9,7 @@ import {
 import styled, { css } from 'styled-components'
 import { search, close } from '@equinor/eds-icons'
 import { search as tokens } from './Search.tokens'
+import { Button } from '../Button'
 import { Icon } from '../Icon'
 import {
   spacingsTemplate,
@@ -25,7 +26,7 @@ const {
   typography,
   border,
   clickbound,
-  entities: { icon, placeholder },
+  entities: { icon, placeholder, button },
   states,
 } = tokens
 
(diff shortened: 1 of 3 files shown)


libraries/core-react/src/components/Search/Search.tsx
```diff
@@ -307,8 +279,14 @@ export const Search = forwardRef<HTMLInputElement, SearchProps>(function Search(
...
       <Icon data={search} aria-hidden size={size} />
       <Input {...inputProps} />
-      <InsideButton {...clearButtonProps} aria-label="Clear search">
-        <Icon data={close} size={size} />
+      <InsideButton
+        {...clearButtonProps}
+        aria-label={'clear search'}
+        title="clear"
+        variant="ghost_icon"
+        type="button"
```
mimarz, line 287: Since we are using an actual `button` now, this type is redundant.

libraries/core-react/src/components/Search/Search.tsx
```diff
@@ -294,7 +267,6 @@ export const Search = forwardRef<HTMLInputElement, SearchProps>(function Search(
   const clearButtonProps = {
     isActive: state.isActive,
     size,
```
mimarz, line 269: Suggest removing `size` property as `Button` does not support it nor, our styled wrapper.

mimarz (changes requested):
Some suggestions on cleaning up redundant variables/properties now that we use a `Button`.

mimarz (approved):
LGTM 👍

PR #1769, opened 2021-12-03
✨Support css import

### LINKED ISSUE #1690: Jest & Rollup css import configuration
We currently don't have support for using `import` on `css` files in our libraries due to missing loaders as this has not been needed.

This gives a faulty build for #1601 as it imports css as part of [react-datepicker](https://reactdatepicker.com/)

Quick investigation shows that both Jest & Rollup would need to be updated with plugins/configuration for support using `import` on css files.

Action:
* As we are in the process of changing our build pipeline as part of #1043 we need to decided on how, if, or when we will support css files in build steps

Notes:
* https://www.npmjs.com/package/rollup-plugin-import-css
* https://stackoverflow.com/questions/54627028/jest-unexpected-token-when-importing-css

oddvernes:
Resolves #1690 

Tested by adding a test.css file with a class selector (to set background-color) in a component and importing it, and using the className in the component.
Tested Jest on the component to make sure snapshot fails but otherwise no other error messages
Tested import css from node_modules by temporarily installing react-datepicker and importing its css into a component

Tested building. this results in a core-react.cjs.css file in /dist and a bundle.css file in /dist/esm

Edit:
Tested pnpm pack and using this in a create-react-app, and it seems you need to add
`import '@equinor/eds-core-react/dist/esm/bundle.css'`
for the css to be inlcuded

### DIFF
--- a/libraries/core-react/jest.config.js
+++ b/libraries/core-react/jest.config.js
@@ -4,6 +4,9 @@ module.exports = {
   transform: {
     '.(js|ts|tsx)': 'babel-jest',
   },
+  moduleNameMapper: {
+    '\\.(css)$': '<rootDir>/src/test/__mocks__/styleMock.js',
+  },
   testRegex: '(/__tests__/.*|\\.(test|spec))\\.(ts|tsx|js)$',
   moduleFileExtensions: ['ts', 'tsx', 'js'],
   testEnvironment: 'jsdom',
--- a/libraries/core-react/package.json
+++ b/libraries/core-react/package.json
@@ -2,7 +2,9 @@
   "name": "@equinor/eds-core-react",
   "version": "0.16.1",
   "description": "The React implementation of the Equinor Design System",
-  "sideEffects": false,
+  "sideEffects": [
+    "**/*.css"
+  ],
   "main": "dist/core-react.cjs.js",
   "publishConfig": {
     "main": "dist/core-react.cjs.js",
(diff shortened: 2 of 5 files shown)


mimarz:
> Resolves #1690
> 
> Tested by adding a test.css file with a class selector (to set background-color) in a component and importing it, and using the className in the component. Tested Jest on the component to make sure snapshot fails but otherwise no other error messages Tested import css from node_modules by temporarily installing react-datepicker and importing its css into a component
> 
> Tested building. this results in a core-react.cjs.css file in /dist and a bundle.css file in /dist/esm
> 
> Edit: Tested pnpm pack and using this in a create-react-app, and it seems you need to add `import '@equinor/eds-core-react/dist/esm/bundle.css'` for the css to be inlcuded

After talking in a call we decided to have another look if there are any other solutions out there so our users don't need to import a css file. This will also affect #1549.

oddvernes:
Testing out an alternate aproach: using rollup-plugin-postcss with extract: false. This way css is stored in js and injected into head when the component loads similar to how styled components does it.
To test: create a css file alongside a component with a class selector that changes something and apply this class to the component. pnpm pack and install locally in a create-react-app etc, use the component and inspect head style tags
(Note i have rebased this branch on develop)

oddvernes:
Tested with react-datepicker and importing its css, this works too. The css.js file is created inside node_modules in /dist/esm

mimarz (approved):
LGTM 👍

PR #1774, opened 2021-12-06
Improvement/ramda named import

### LINKED ISSUE #1768: Ramda: use named imports
https://www.npmjs.com/package/ramda
"Destructuring imports from ramda does not necessarily prevent importing the entire library. You can manually cherry-pick methods like the following, which would only grab the parts necessary for `identity `to work
`import identity from 'ramda/src/identity'`"

In our project we mostly use  import * as R from 'ramda', but now that it is not bundled by us, but instead a dependency, it is up to other projects to tree-shake, so perhaps it is better we do the named import method instead to make sure it is always properly tree shaked. We only use mergeDeepRight, and in a create-react-app this is properly tree-shaked as is, but as their documentation says, perhaps in some setups it does not.

oddvernes:
Resolves #1768 

ramda is moved to devDependencies and files related to mergeDeepRight is now bundled in the build by using named path imports


mimarz (approved):
LGTM 👍

PR #1810, opened 2021-12-15
⚡️Improvement - Menu, Popover: deactivate popper when closed

### LINKED ISSUE #1675: Properly memoize menu and popover
**Describe the bug**

Menu and popover uses the popper hook which sends updated positional data every frame during scrolling, which means these components and some of their children gets rendered every frame (not painted though, just the function runs which can be a bit heavy on itself)

1. console log the functions and children

**Expected behavior**

Better memoization to avoid performance impact during scroll. Also perhaps there is a way to pause popper when the popup is closed

oddvernes:
resolves #1675 

There is room for improvement in menu still, as it does rerender on each frame while scrolling when menu is open. But it is tricky to figure out all the root causes of this and how to resolve them, so I will leave it for now.

### DIFF
--- a/libraries/core-react/src/components/Menu/Menu.tsx
+++ b/libraries/core-react/src/components/Menu/Menu.tsx
@@ -53,7 +53,7 @@ const MenuContainer = forwardRef<HTMLDivElement, MenuContainerProps>(
       if (onClose === null && onCloseCallback) {
         setOnClose(onCloseCallback)
       }
-    })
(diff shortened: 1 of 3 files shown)


oddvernes:
> 
> Maybe you need to clear the `storeAnchorEl` in a cleanup function in the useEffect 🤔

Ah, good point, setting the storeAnchorEl in cleanup seems to do the trick

mimarz (approved):
LGTM 👍

PR #1811, opened 2021-12-15
♻️ Simplify Figma-broker exports

### LINKED ISSUE #1786: Convert figma-broker to use node scripts
There is no point in using a web-server for running figma-broker anymore as its only run locally anways. 

Suggest removing `koa` and just use plain node scripts and [args](https://nodejs.org/docs/latest/api/process.html#processargv) for triggering the update actions. Adding the scripts to our root package.yml so that its streamlined with the rest.

This will also remove the need to have specific extensions and cli's (httpie + extension) installed for running the scripts.

Edit: Update README with how new usage will be

mimarz:
resolves #1786 

Now, if we want to update tokens or assets we just run `export:tokens` or `export:assets`  (with appended `true` to force new data fetch from Figma)


apps/figma-broker/README.md
```diff
@@ -5,15 +5,34 @@
...
+
+### Run
+
+1. Run either `tokens` or `assets` script with a `fileId` and/or flag for forcing to fetch new data from Figma.
+
+#### Example
+
+```sh
+pnpm run tokens 0TbIXrrObWj80Cf7KucKYFL0 true  # Fetches new data from Figma
+pnpm run tokens 0TbIXrrObWj80Cf7KucKYFL0 # Reads local figma json file
```
vnys, line 24: This needs some clarification, it would be helpful to know how to find the fileId, what the local figma json file is, how it works etc. Also, I get an unhandled rejection error if I don’t have a local json file.

apps/figma-broker/README.md
```diff
@@ -5,15 +5,34 @@
 
 Extracts design decisions into [design tokens] from Figma using the [Figma Web API]
 
-# Notes 
+## Usage
+
+### Setup
+
+1. Add an .env file in the root.
```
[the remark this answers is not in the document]
vnys: image

package.yaml
```diff
@@ -12,7 +12,8 @@ scripts:
   build:icons: pnpm --filter @equinor/eds-icons run build
   storybook: pnpm --filter ./libraries/core-react run storybook
   build:storybook: pnpm --filter ./libraries/core-react run build:storybook
-  figma-broker: pnpm --filter @equinor/eds-figma-broker run dev
+  export:tokens: pnpm --filter @equinor/eds-figma-broker run tokens 0TbIXrrObWj80Cf7KucKYFL0
```
vnys, line 15:
```suggestion
  export:tokens: pnpm --filter @equinor/eds-figma-broker tokens 0TbIXrrObWj80Cf7KucKYFL0
```
Remove `run`

package.yaml
```diff
@@ -12,7 +12,8 @@ scripts:
...
+  export:assets: pnpm --filter @equinor/eds-figma-broker run assets BQjYMxdSdgRkdhKTDDU7L4KU
```
vnys, line 16:
```suggestion
  export:assets: pnpm --filter @equinor/eds-figma-broker assets BQjYMxdSdgRkdhKTDDU7L4KU
```
Remove `run`

vnys (changes requested):
I think this is a good improvement. Long term I think we should consider making this a cli, but definitely easier to use now. 

One thing I would change is to put the fileIds into env variables, it looks a little messy to include those in the script in package.yaml. And then use those as defaults if unless a fileID is provided. 

And the README needs a little more clarification.

apps/figma-broker/README.md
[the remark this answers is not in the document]
mimarz: Good point 👍

apps/figma-broker/README.md
mimarz: I'll look into it 🤔

mimarz: What kind of error did you get? Worked fine here, running locally... 🤔 @vnys

vnys: When I run the same command now there are no errors 🤷🏻‍♂️ Not sure why. But I got the error message mentioned earlier today if I didn’t have a local figma json file.

mimarz: You can delete the `raw` folder and try again? @vnys

oddvernes (approved):
Works on windows now 👍

apps/figma-broker/README.md
vnys: I did, but can’t reproduce the error – so just ignore it for now

vnys (approved):
Added a suggestion for a slightly more elaborate explanation for Codespaces users. Apart from that it LGTM 👍🏻

PR #1813, opened 2021-12-16
🩹 Don’t use density prop (cleaned up)

### LINKED ISSUE #1707: Don’t use the density prop directly in Storybook

vnys:
Split things up a bit. This pr is just about fixing stories with direct usage of compact as density value. Also added missing descriptions to some of the stories and fixed some typos along the way.

Resolves #1707


oddvernes (approved):
Looks good to me 👍

PR #1815, opened 2021-12-16
Create Group component

### LINKED ISSUE #1812: Create Group component
Based on #394 we need a "Button group" component. Be it for only for internal and/or external use we need a way of grouping several elements together with a distinct styling to show they are related to one another.

The Button group component should be a simple container component that groups its children and only applies rounded corner to the outer edges.

Notes:
* Container which removes border-radius of its children
* Only applies rounded corners to its outer edges
* No spacing between its children
* Styling of children should work for all elements
* Naming of said component, discuss what makes most sense
  * `Group` ~`ButtonGroup` or  (sub-component)`Button.Group`~

laliq:
Resolves #1812

### DIFF
--- a/libraries/core-react/src/components/Button/Button.stories.tsx
+++ b/libraries/core-react/src/components/Button/Button.stories.tsx
@@ -19,6 +19,12 @@ const FullWidthWrapper = styled.div`
   grid-gap: 16px;
 `
 
+const GroupWrapper = styled.div`
+  margin: 32px;
+  display: grid;
+  grid-template-columns: repeat(4, fit-content(100%));
+`
+
 export default {
   title: 'Components/Button',
   component: Button,
@@ -346,11 +352,13 @@ Compact.parameters = {
   },
 }
 
-export const GroupButtons: Story<ButtonProps> = () => (
-  <Group>
-    <Button>Contained</Button>
-    <Button variant="outlined">Outlined</Button>
-    <Button variant="outlined">Outlined</Button>
-    <Button variant="outlined">Outlined</Button>
-  </Group>
+export const ButtonGroup: Story<ButtonProps> = () => (
+  <GroupWrapper>
+    <Group>
+      <Button>Contained</Button>
+      <Button variant="outlined">Outlined</Button>
+      <Button variant="outlined">Outlined</Button>
+      <Button variant="outlined">Outlined</Button>
+    </Group>
+  </GroupWrapper>
 )
--- a/libraries/core-react/src/components/Button/Button.stories.tsx
+++ b/libraries/core-react/src/components/Button/Button.stories.tsx
@@ -19,12 +19,6 @@ const FullWidthWrapper = styled.div`
   grid-gap: 16px;
 `
 
-const GroupWrapper = styled.div`
-  margin: 32px;
-  display: grid;
-  grid-template-columns: repeat(4, fit-content(100%));
-`
-
 export default {
   title: 'Components/Button',
   component: Button,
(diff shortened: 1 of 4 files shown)


laliq:
I have created `ButtonGroup` in `Button.stories` just to have overview.

libraries/core-react/src/components/Button/Button.stories.tsx
```diff
@@ -17,6 +19,12 @@ const FullWidthWrapper = styled.div`
   grid-gap: 16px;
 `
 
+const GroupWrapper = styled.div`
+  margin: 32px;
+  display: grid;
+  grid-template-columns: repeat(4, fit-content(100%));
```
mimarz, line 25: Don't think we need this css? Maybe you can extend the `Wrapper` with styled and set `grid-gap: 0` if you need that margin for the story,

libraries/core-react/src/components/Button/Button.stories.tsx
```diff
@@ -314,18 +322,27 @@ export const FullWidth: Story<ButtonProps> = () => (
...
+        <Button>Contained</Button>
+        <Button variant="outlined">Outlined</Button>
+        <Button variant="ghost">Ghost</Button>
+        <Button variant="ghost_icon">
+          <Icon data={menu} title="Ghost icon menu"></Icon>
+        </Button>
+      </Wrapper>
+    </EdsProvider>
+  )
+}
```
mimarz, line 345: This should not be a changed in this PR or was this intentional?

libraries/core-react/src/components/Group/Group.tsx
```diff
@@ -0,0 +1,36 @@
+import { forwardRef } from 'react'
+import styled from 'styled-components'
+import { group as tokens } from './Group.tokens'
+
+const { border } = tokens
+
+const GroupBase = styled.div`
```
mimarz, line 7: Suggest changing this to use the function style (`styled.div(() => css``)`), then you only need to do the `border.type === 'border' && border.radius` check once as a const.

libraries/core-react/src/components/Group/Group.tsx
```diff
@@ -0,0 +1,36 @@
...
+  }
+`
+
+export type GroupProps = React.HTMLAttributes<HTMLDivElement>
+
+export const Group = forwardRef<HTMLDivElement, GroupProps>(function Group(
+  { children },
+  ref,
+) {
+  return <GroupBase ref={ref}>{children}</GroupBase>
```
mimarz, line 35: We should still spread the `...rest` on the `GroupBase` to conform with our other components, even when its just internal :)

mimarz (changes requested):
Great work! 

Some small suggestions on cleaner code :)

libraries/core-react/src/components/Button/Button.stories.tsx
laliq: This was related to the conflict. I performed rebased, now everything is in check.

libraries/core-react/src/components/Button/Button.stories.tsx
laliq: Great, thanks

libraries/core-react/src/components/Button/Button.stories.tsx
mimarz: @laliq Are you sure you pulled the latest changes from develop locally before rebasing it onto this branch? We are missing from develop the past 5 days 🤔

laliq: running merge now

laliq: @mimarz looks good to me now

mimarz: Now you have 73 files flagged as changed... 🙈

laliq: I did this...
```
git checkout my-brunch
git merge origin/master
```
I guess it was wrong 😑

mimarz:
```sh
git rebase develop
git push -f
```

mimarz (approved):
Great work! 👏 👍

Have a merry christmas and a happy new year! 🎅 🎄 ✨

PR #1819, opened 2021-12-17
♿️ textfield a11y improvements

### LINKED ISSUE #1145: Support proper validation errors in TextField
While working with the `TextField` component, we realized that the way helperText is implemented don't work with proper validation error feedback for screen readers.

We need support for aria-required and aria-valid and a way to link the actual error message to the input field by aria-describedby. It might be easier if the validation error and icon is a separate sub component, but how do we solve that it's also used for information messages 🤷‍♀️ 
It might also be easier to use the validation features with an external forms library if it's a separate component. 

Take a look at the `Checkbox` form example in the Storybook where react-hook-form is used for some inspiration and guidelines.

oddvernes:
resolves #1145 

Screenreader should now read the helpertext

I implemented aria-invalid when variant="error"

### DIFF
--- a/libraries/core-react/src/components/TextField/Field.tsx
+++ b/libraries/core-react/src/components/TextField/Field.tsx
@@ -180,6 +180,7 @@ export const Field = forwardRef<
   const iconSize = density === 'compact' ? 16 : 24
   const actualVariant = variant === 'default' ? 'textfield' : variant
   const inputVariant = tokens[actualVariant]
+  const isError = actualVariant === 'error' ? true : false
 
   const focusHandler = (
     e: React.FocusEvent<HTMLInputElement | HTMLTextAreaElement>,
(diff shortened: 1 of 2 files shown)


vnys:
Do we have a story where this can be tested? If not, can we make one?

oddvernes:
> Do we have a story where this can be tested? If not, can we make one?

There are already stories with variant="error" and stories with helpertext

I have just tested with windows 10 built in narrator, where it seems to work in edge & firefox but not chrome (which is also the case for mui helpertext)

vnys:
> I have just tested with windows 10 built in narrator, where it seems to work in edge & firefox but not chrome (which is also the case for mui helpertext)

Hm… that’s a problem, that leaves out ChromeVox users on Chromebooks.

oddvernes:
> Hm… that’s a problem, that leaves out ChromeVox users on Chromebooks – unless it’s only Narrator on Chrome. I’ll do some testing.

It works fine on macbook with its default narrator + chrome

vnys (approved):
LGTM 👍🏻

PR #1822, opened 2021-12-20
💥Conform Scrim open logic

### LINKED ISSUE #1176: Scrim improvements
* Conform on open/close behaviour as other components #719. Introduce `open` property for controlling visiblity
* Disable scroll when visible (#1587 )

oddvernes:
resolves #1176 

This introduces breaking changes to scrim:
Use `open` prop to align Scrim open/close logic with other components. It now hides itself (with `display: none` and not rendering children)  (breaking)
Removed the second argument (`open`) in onClose (breaking)

The useHideBodyScroll hook is changed to take a boolean that is watched to determine when to hide scroll (since the Scrim is not neccessarily fully unmounted when closed now)

### DIFF
--- a/libraries/core-react/src/components/Scrim/Scrim.tsx
+++ b/libraries/core-react/src/components/Scrim/Scrim.tsx
@@ -1,11 +1,11 @@
 import { forwardRef, MouseEvent, HTMLAttributes } from 'react'
-import styled from 'styled-components'
+import styled, { css } from 'styled-components'
 import { scrim as tokens } from './Scrim.tokens'
 import { useGlobalKeyPress, useHideBodyScroll } from '../../hooks'
 
 const { height, width, background } = tokens
 
-const StyledScrim = styled.div`
+const StyledScrim = styled.div<StyledScrimProps>`
   width: ${width};
   height: ${height};
   background: ${background};
@@ -15,7 +15,7 @@ const StyledScrim = styled.div`
   left: 0;
   align-items: center;
   justify-content: center;
-  display: flex;
+  ${({ open }) => css({ display: open ? 'flex' : 'none' })}
 `
 
 const ScrimContent = styled.div`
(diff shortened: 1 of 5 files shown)


libraries/core-react/src/components/Scrim/Scrim.tsx
```diff
@@ -15,45 +15,49 @@ const StyledScrim = styled.div`
...
+
 export type ScrimProps = {
   /** Whether scrim can be dismissed with esc key and outside click
    */
   isDismissable?: boolean
+  /** programmatically toggle scrim */
+  open: boolean
   /** function to handle closing scrim */
-  onClose?: (event: MouseEvent | KeyboardEvent, open: boolean) => void
+  onClose?: (event: MouseEvent | KeyboardEvent) => void
```
mimarz, line 35: suggest removing `event: MouseEvent | KeyboardEvent` in `onClose` as we don't have any events in our other `onClose` handlers

libraries/core-react/src/components/Scrim/Scrim.tsx
```diff
@@ -65,7 +69,9 @@ export const Scrim = forwardRef<HTMLDivElement, ScrimProps>(function Scrim(
 
   return (
     <StyledScrim onClick={handleMouseClose} {...props}>
-      <ScrimContent onClick={handleContentClick}>{children}</ScrimContent>
+      {open && (
```
mimarz, line 72: Maybe move this `open` to top level and return `null` if its not open.

mimarz (changes requested):
Looking good 👍

Just some tweaks and I noted a new issue for updated components to [useArgs()](https://storybook.js.org/docs/react/addons/addons-api#useargs) for two-way binding for interact-able components via props-table.

mimarz (approved):
LGTM 👍

Merry Christmas and a Happy New Year! 🎅 🎄 ✨

PR #1824, opened 2021-12-21
🚸 MultiSelect: clear input on blur, clearbutton

### LINKED ISSUE #1664: MultiSelect: be able to clear searchvalue when unfocusing or clicking X
When searching for an item in the dropdown and no results/ or you do not select an option, it should be possible to clear the value entered somehow. I would say it should be cleared when clicking the X, like for instance here (picture under). And also if you have searched for somethig and do not click any options, and then unfocus the component

image

Similar to the SingleSelect component.

Now, we can do this in our application

image

There is now way to clear the searchvalue without manually removing the text

oddvernes:
Resolves #1664


mimarz (approved):
LGTM 👍

Issue #1586, opened 2021-09-29
Create token preview in Storybook


mimarz:
Suggest using a `Card` for each token with relevant information in card content for individual token.
Categories for each token set.

Examples to draw inspiration from: 

* https://components.ruter.as/#/Visual%20Style/Design%20tokens
* https://www.lightningdesignsystem.com/design-tokens/

mimarz:
Prioritising this as we currently are working on improving our developer documentation. This will also be important to have when/if we do any token changes for our developers to get familiar with changes

vnys:
Solved with the design token addon in the EDS 2.0 workshop repo

PR #1843, opened 2021-12-30
🚚  Categorise stories

### LINKED ISSUE #1800: Use same category headers in Storybook as in Figma
A year ago Bendik [categorised all the components](https://github.com/equinor/design-system/issues/962) in Figma to make them easier to find. The plan was to do the same in the Storybook, but we never got around to doing that. I suggest we do that during the holidays when we (hopefully) don’t have any pull requests or feature branches that haven’t been merged into develop.

vnys:
Use the same category headers in Storybook as we do in Figma and Zeroheight.

Resolves #1800


vnys:
**Notes**

- Tabs are missing in Zeroheight
- Checkbox and radiobuttons are grouped as “Selection Controls” in Zeroheight

vnys:
**Components that only exist in code**

- Input
- Avatar
- Group (missing tests and stories)
- EdsProvider (missing tests and stories)
- Icon
- Label
- Paper (missing tests and stories)
- TextArea (missing tests and stories)
- Typography

vnys:
- Token story is missing

mimarz:
> * Token story is missing

#1586

mimarz:
I think we need to have a discussion on this and figure out how its wired up to Zeroheight. I see the benefit of having a red line with categories in Zeroheight and Figma, but I felt it was harder to find components now and see which package you can find them in… 🤔

vnys:
I think the keyword here is consistency. If you look at a component in the Storybook, you should know how to find that in Figma and Zeroheight – and vice versa. Having the same mental map of where the components belong whether you’re in Figma or Storybook will make it easier to navigate between them, and I don’t think that’s going to be an issue once you get used to it.

If there’s a particular component that’s difficult to find, however, it could be that it should have been in a different category, and if so we have to solve that in Figma an Zeroheight as well.

I see your point about which package the components belong to, so that should be addressed – I’ll create an issue for that.

mimarz:
> I think the keyword here is consistency. If you look at a component in the Storybook, you should know how to find that in Figma and Zeroheight – and vice versa. Having the same mental map of where the components belong whether you’re in Figma or Storybook will make it easier to navigate between them, and I don’t think that’s going to be an issue once you get used to it.
> 
> If there’s a particular component that’s difficult to find, however, it could be that it should have been in a different category, and if so we have to solve that in Figma an Zeroheight as well.
> 
> I see your point about which package the components belong to, so that should be addressed – I’ll create an issue for that.

I think we should do a spike on how to write our component stories, which headings to have etc. Installation, usage, accessibility and compact are a few headings that comes to mind which maybe every component story should have 🤔

vnys:
> Installation, usage, accessibility and compact are a few stories that comes to mind which every component story should have 🤔

I agree, we should have a common template.

mimarz:
> > Installation, usage, accessibility and compact are a few stories that comes to mind which every component story should have 🤔
> 
> I agree, we should have a common template.

Made issue #1850 for followup on this.

mimarz (approved):
Its a good start. Brings into light the work we have left on developer documentation 🙈

#1850 #1849

PR #1852, opened 2022-01-03
🎉 Create eds-utils

### LINKED ISSUE #1846: Create eds-utils package
Create a new package `eds-utils` in `packages` for sharing utility and hooks between `eds-core-react` & `eds-lab`.

* Copy `utils` and `hooks` from `eds-core-react`

oddvernes:
resolves #1846


mimarz (changes requested):
Looking good, some small nitpicking.

* You can removed `nginx.conf` as there is nothing here we will host on a webserver standalone.
* Suggest add a `build:utils` and `test:utils` scripts to the root `package.yaml` so we have the same shortcuts as for our other packages.

oddvernes:
test fails btw because there are no tests. So we should either not add test as ci step, or add a test/mock test

mimarz:
> test fails btw because there are no tests. So we should either not add test as ci step, or add a test/mock test

Just make a simple test for for `useMounted` perhaps?

packages/eds-utils/README.md
```diff
@@ -0,0 +1,13 @@
+# @equinor/eds-utils
+
+This package host hooks and utility functions that are used by eds-core-react and eds-lab.
```
vnys, line 3:
```suggestion
This package host hooks and utility functions that are used by [eds-core-react](https://www.npmjs.com/package/@equinor/eds-core-react) and [eds-lab](https://www.npmjs.com/package/@equinor/eds-lab).
```

vnys: Add links to eds-core-react and eds-lab

vnys (changes requested):
Just some missing info. Also, we should add some documentation to the Storybook. Have a look at Circuit UI’s storybook for an example.

vnys:
About the tests… I think we need to agree on _why_ we write tests. I don’t see the value of writing a test just to make some system shut up. If we have 10 modules, and just one test, what’s the point? If we only focus on moving issues to the right on the board, does that mean that the job is done if we cut corners to get there? Or does it mean we have accumulated even more technical debt? If the point is to publish the package as quickly as possible because it’s a blocker, then we should at least create an issue for writing tests and move it to the top of the todo lane. We need to uphold a certain level of quality in EDS, having a good test coverage and writing good documentation is part of that.

mimarz:
> About the tests… I think we need to agree on _why_ we write tests. I don’t see the value of writing a test just to make some system shut up. If we have 10 modules, and just one test, what’s the point? If we only focus on moving issues to the right on the board, does that mean that the job is done if we cut corners to get there? Or does it mean we have accumulated even more technical debt? If the point is to publish the package as quickly as possible because it’s a blocker, then we should at least create an issue for writing tests and move it to the top of the todo lane. We need to uphold a certain level of quality in EDS, having a good test coverage and writing good documentation is part of that.

I think the idea was more to add remaining tests in a separate issue. Added an issue for this: #1868.

mimarz (approved):
LGTM 👍

vnys (approved):
Linting is missing, apart from that it LGTM 👍🏻

PR #1861, opened 2022-01-05
Refactor Popover

### LINKED ISSUE #1365: Refactor Popover
For making our components easier to use we have decided to remove the fixed close button inside the `Popover` component as there are other valid options for using an popover without "that" close button. The documentation serves as a guideline recommendation for the close button on certain types of content. 

Actions
* Remove fixed close button in `Popover`
* Add story with example where said close button is used but composed from existing components
* Update documentation to better reflect when to add a close button inside popover

laliq:
Resolves #1365

### DIFF
--- a/packages/eds-core-react/src/components/Popover/Popover.stories.tsx
+++ b/packages/eds-core-react/src/components/Popover/Popover.stories.tsx
@@ -13,7 +13,7 @@ import {
 import { more_vertical } from '@equinor/eds-icons'
 import { Meta, Story } from '@storybook/react'
 
-const { Title, Content } = Popover
+const { Title, Content, Header } = Popover
 
 const StoryCenter = styled.div({
   display: 'flex',
@@ -27,6 +27,7 @@ export default {
   subcomponents: {
     Title,
     Content,
+    Header,
   },
 } as Meta
 
@@ -252,6 +253,53 @@ Compact.parameters = {
   docs: {
     description: {
       story: 'Compact `Popover` using `EdsProvider` ',
+    },
+  },
+}
+
+export const WithCloseButton: Story<PopoverProps> = () => {
+  const [isOpen, setIsOpen] = useState<boolean>(false)
+  const anchorRef = useRef<HTMLButtonElement>(null)
+
+  const openPopover = () => setIsOpen(true)
+  const closePopover = () => setIsOpen(false)
+
+  return (
+    <EdsProvider>
+      <StoryCenter>
+        <Button
+          id="click-popover-anchor"
+          aria-controls="click-popover"
+          ref={anchorRef}
+          onClick={openPopover}
+        >
+          Click to activate
+        </Button>
+
+        <Popover
+          id="click-popover"
+          aria-expanded={isOpen}
+          anchorEl={anchorRef.current}
+          onClose={closePopover}
+          open={isOpen}
+        >
+          <Popover.Header>
+            <Popover.Title>Title</Popover.Title>
+          </Popover.Header>
+
+          <Popover.Content>
+            <Typography variant="body_short">Content</Typography>
+          </Popover.Content>
+        </Popover>
+      </StoryCenter>
+    </EdsProvider>
+  )
+}
+
+WithCloseButton.parameters = {
+  docs: {
+    description: {
+      story: 'Description comes here ',
     },
   },
 }
--- a/packages/eds-core-react/src/components/Popover/Popover.stories.tsx
+++ b/packages/eds-core-react/src/components/Popover/Popover.stories.tsx
@@ -324,7 +324,8 @@ export const WithCloseButton: Story<PopoverProps> = () => {
 WithCloseButton.parameters = {
   docs: {
     description: {
-      story: 'Description comes here ',
+      story:
+        'In some cases an explicit close button is recommended. Here is an example of how this can be implemented in the Header.',
     },
   },
 }
--- a/packages/eds-core-react/src/components/Popover/Popover.stories.tsx
+++ b/packages/eds-core-react/src/components/Popover/Popover.stories.tsx
@@ -73,7 +73,6 @@ export const Default: Story<PopoverProps> = (args) => {
         <Popover.Header>
           <Popover.Title>Title</Popover.Title>
         </Popover.Header>
-
         <Popover.Content>
           <Typography variant="body_short">Content</Typography>
         </Popover.Content>
--- a/packages/eds-core-react/src/components/Popover/Popover.stories.tsx
+++ b/packages/eds-core-react/src/components/Popover/Popover.stories.tsx
@@ -13,7 +13,7 @@ import {
 import { more_vertical, close } from '@equinor/eds-icons'
 import { Meta, Story } from '@storybook/react'
 
-const { Title, Content, Header } = Popover
+const { Title, Content, Header, Actions } = Popover
 
 const StoryCenter = styled.div({
   display: 'flex',
@@ -28,6 +28,7 @@ export default {
     Title,
     Content,
     Header,
+    Actions,
   },
 } as Meta
 
@@ -76,7 +77,9 @@ export const Default: Story<PopoverProps> = (args) => {
         <Popover.Content>
           <Typography variant="body_short">Content</Typography>
         </Popover.Content>
-        <Button onClick={handleClose}>OK</Button>
+        <Popover.Actions>
+          <Button onClick={handleClose}>OK</Button>
+        </Popover.Actions>
       </Popover>
     </StoryCenter>
   )
@@ -113,7 +116,9 @@ export const ActivateOnClick: Story<PopoverProps> = () => {
         <Popover.Content>
           <Typography variant="body_short">Content</Typography>
         </Popover.Content>
-        <Button onClick={closePopover}>OK</Button>
+        <Popover.Actions>
+          <Button onClick={closePopover}>OK</Button>
+        </Popover.Actions>
       </Popover>
     </StoryCenter>
   )
@@ -172,7 +177,9 @@ export const ActivateOnHover: Story<PopoverProps> = () => {
         <Popover.Content>
           <Typography variant="body_short">Content</Typography>
         </Popover.Content>
-        <Button onClick={handleClose}>OK</Button>
+        <Popover.Actions>
+          <Button onClick={handleClose}>OK</Button>
+        </Popover.Actions>
       </Popover>
     </StoryCenter>
   )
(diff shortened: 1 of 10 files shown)


laliq:
Need to create a description.

packages/eds-core-react/src/components/Popover/Popover.stories.tsx
```diff
@@ -10,23 +10,37 @@ import {
...
-const { Title, Content } = Popover
+const { Title, Content, Header } = Popover
 
 const StoryCenter = styled.div({
   display: 'flex',
   justifyContent: 'center',
   margin: '10rem',
 })
 
+const StyledCloseButton = styled(Button)(({ theme }) => {
```
mimarz, line 24: Our goal is for users to use just `Button` inside the `Popver.Header`. Looking at the styling in `Popover.Header` I don't think you need this?

packages/eds-core-react/src/components/Popover/PopoverHeader.tsx
```diff
@@ -0,0 +1,36 @@
...
+export type PopoverHeaderProps = HTMLAttributes<HTMLDivElement>
+
+const StyledPopoverHeader = styled.div`
+  display: grid;
+  grid-auto-flow: column;
+`
+
+const StyledDivider = styled(Divider)(({ theme }) => {
+  return css`
+    margin-left: -${theme.spacings.left};
```
mimarz, line 14: We want to avoid "hacks" with negative margins as it causes some unwanted consequences with different content. Maybe do the same refactor with spacings in `Popover` as we did with `Card`?

packages/eds-core-react/src/components/Popover/PopoverHeader.tsx
```diff
@@ -0,0 +1,36 @@
+import { forwardRef, HTMLAttributes } from 'react'
+import styled, { css } from 'styled-components'
+import { Divider } from '../Divider'
+
+export type PopoverHeaderProps = HTMLAttributes<HTMLDivElement>
+
+const StyledPopoverHeader = styled.div`
+  display: grid;
```
mimarz, line 8: What if we use `display:flex` and `justify-content: space-between`? 🤔

packages/eds-core-react/src/components/Popover/Popover.test.tsx
```diff
@@ -77,11 +78,14 @@ describe('Popover', () => {
 
     render(
       <TestPopover open onClose={handleOnClose}>
-        <div>some random content</div>
+        <Popover.Header>
```
mimarz, line 81: This test seems kinda redundant now as it only tests if the button was clicked. Suggest we just remove it.

mimarz (changes requested):
Looking good. Some tweaks can still be done.

Looks like we should do the same "conforming" we did on `Card` where the padding is moved from the container to the individual sub-components. This should make it easier to have the `Divider` for `Popover.Header` and better code :)

packages/eds-core-react/src/components/Popover/PopoverHeader.tsx
laliq: I added `position: absolute` with `left`&`right` 0. It does work.
What do you think about it?

packages/eds-core-react/src/components/Popover/PopoverHeader.tsx
laliq: Yes works nice 😊

packages/eds-core-react/src/components/Popover/Popover.stories.tsx
laliq: Now the close button is larger has 48px width and height. With the previous styling had 32px, shall i change this in style inline?

packages/eds-core-react/src/components/Popover/PopoverHeader.tsx
mimarz: I think it would be better to just fix this properly with doing the paddings the same way we did for `Card`. Its easier to maintain and more predicated for our users. Were you thinking you wanted a separate issue for that task?

packages/eds-core-react/src/components/Popover/Popover.test.tsx
mimarz: Just remove it :)

packages/eds-core-react/src/components/Popover/PopoverHeader.tsx
laliq: I see your point. 
Im not sure if i need a separate issue for it. The issue is a breaking change already. Would it be enough to make a clear description on the commits?
Also i found negative value in `Popover.Title`, i think it should be fixed to.

mimarz: Ok, you can do the changes in this PR as they are related to this refactor which as you say is already breaking. We should no longer need any of these negative margins if we fix the padding by moving them them to the sub-components :) 

What description we have on the commits doesn't matter in this case as they get squished on merge anyways. Suggest you just updated the description for this PR with the information instead :)

laliq: Thanks, I will do that. 👌

laliq:
In the example  `WithCloseButton` the size of the header is determined by `close btn`. How to implement `closeButton` spacing?

packages/eds-core-react/src/components/Popover/PopoverTitle.tsx
```diff
@@ -1,23 +1,11 @@
 import { forwardRef, HTMLAttributes } from 'react'
 import styled, { css } from 'styled-components'
-import { Divider } from '../Divider'
 
 const StyledPopoverTitle = styled.div(({ theme }) => {
   return css`
     margin-right: ${theme.entities.closeButton.width};
     max-width: calc(${theme.maxWidth} - ${theme.entities.closeButton.width});
```
oddvernes, line 7: It seems like the margin-right here was to take into account the hardcoded close button and can now be removed (and some margin-left can be added to the manually added closebutton instead.
Also the max-width calc can be replaced with just the max-width for same reasons?

laliq: Thanks @oddvernes.

packages/eds-core-react/src/components/Popover/Popover.stories.tsx
```diff
@@ -255,3 +276,57 @@ Compact.parameters = {
...
+
+export const WithCloseButton: Story<PopoverProps> = () => {
+  const [isOpen, setIsOpen] = useState<boolean>(false)
+  const anchorRef = useRef<HTMLButtonElement>(null)
+
+  const openPopover = () => setIsOpen(true)
+  const closePopover = () => setIsOpen(false)
+
+  return (
+    <EdsProvider>
```
oddvernes, line 288: The edsprovider is superflous here

packages/eds-core-react/src/components/Popover/Popover.stories.tsx
```diff
@@ -255,3 +276,57 @@ Compact.parameters = {
...
+          anchorEl={anchorRef.current}
+          onClose={closePopover}
+          open={isOpen}
+        >
+          <Popover.Header>
+            <Popover.Title>Title</Popover.Title>
+            <Button
+              style={{ height: '32px', width: '32px' }}
+              variant="ghost_icon"
+              title="Close popover"
```
oddvernes, line 311: aria-label is generally recommended instead of title for accessibility

packages/eds-core-react/src/components/Popover/Popover.stories.tsx
```diff
@@ -255,3 +276,57 @@ Compact.parameters = {
...
+        </Popover>
+      </StoryCenter>
+    </EdsProvider>
+  )
+}
+
+WithCloseButton.parameters = {
+  docs: {
+    description: {
+      story: 'Description comes here ',
```
oddvernes, line 329: I suppose this needs to be filled out still, but I am not sure myself in what scenarios it is recommended to use closebutton. @mimarz ? (Maybe as a security measure in a dense page full of interactive stuff you risk clicking on accidentally with clickoutside)

mimarz: Looking at the documentation for [Popover](https://eds.equinor.com/0b0c666ab/p/8809df-popover/b/49f4b2), its meant for interactive use-cases.

laliq: Oddbjørn's argument seems reasonable , but still user can use esc button to avoid collision with other clickable elements?
I still wander if the close button is needed, especially when there are two way to close it? I would see it as a informative for the user. 🤔

laliq: What if we say: Popover with close button can be use in a dense page, a page filled with many clickable elements to avoid collision with them. Close button can be useful when Popover do not contain any clickable elements.
Something along those lines?

oddvernes: Maybe we can avoid the reasoning for now and say something generic like, 'In some cases an explicit close button is recommended. Here is an example of how this can be implemented in the Header'

laliq: Nice @oddvernes

packages/eds-core-react/src/components/Popover/PopoverHeader.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { Divider } from '../Divider'
+
+export type PopoverHeaderProps = HTMLAttributes<HTMLDivElement>
+
+const StyledPopoverHeader = styled.div(({ theme }) => {
+  return css`
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    padding: calc(${theme.spacings.top} - ${theme.entities.title.spacings.top})
```
mimarz, line 12: Maybe just specify it to be `12px` in `theme.entities.title.spacings.top` for clarity. This seems to be an exception to normal spacing rules.  Would also verify with designers if this is intended as it seems weird with only this being 16px while the rest is 16px 😅

laliq: Just try to be sophisticated 😅
I will talk to the designers 👍

packages/eds-core-react/src/components/Popover/PopoverActions.tsx
```diff
@@ -0,0 +1,31 @@
...
+  ({ theme }) => {
+    return css`
+      display: grid;
+      grid-gap: 8px;
+      grid-auto-flow: column;
+      align-items: center;
+      padding: 0 ${theme.spacings.right} 0 ${theme.spacings.left};
+      :last-child {
+        padding-bottom: ${theme.spacings.bottom};
+      }
```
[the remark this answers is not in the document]
laliq: @mimarz I had a chat wit Bente about this, just to be sure if `Popover` can contain only `Popover.Actions` without title 🤔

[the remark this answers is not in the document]
laliq: The children of the `Popover` aren't `first-child`, the `ArrowWrapper` has `first-child` position.

mimarz (approved):
quality-quality-work

PR #1867, opened 2022-01-06
♿️  Remove fixed height on Button and Table

### LINKED ISSUE #1858: Remove fixed button height
EnergyVision has buttons with multiple lines of text, which is currently not possible in EDS due to the fixed height on the component. Remove the fixed height, but make sure the button still takes up the same place vertically by using padding and line-height, by using a `padding-top `and `padding-bottom` of `0.625em` for example. This should ideally be solved in Figma Broker by transforming the values from Figma to relative units.

image

This issue will solve [#653](https://github.com/equinor/energyvision/issues/653) in equinor/energyvision.

vnys:
My first thought was to add this to `@equinor/eds-labs` – but we already have a button component there with the CSS spike. We could potentially rename that to something else… or publish this to `@equinor/eds-core-react` to a `next` release so they can test it in Envis 🤔

How to test:
1. Spin up a Codespace from this pr
1. Wait for the dust to settle 
1. Run `pnpm storybook:core-react` in the terminal and open the preview in a browser 
1. Change the root font size in the browser* and observe how the button size changes

_*In Chrome, go to Settings → Appearance → Font size_

Resolves #1858
Resolves #1867


vnys:
I’ve left out the ghost icon button, since the icon size is specified in px – not sure how that fits with the needs in Envis 🤔 Also did the calculations manually for now to see how it works with first, which is why I’ve left some comments with the original values. So this is ready for review, but we will have to publish it to with a dist-tag to test it properly in Envis – so

packages/eds-core-react/src/components/Button/tokens/button.ts
```diff
@@ -98,7 +105,11 @@ export const button: ButtonToken = {
   },
   modes: {
     compact: {
-      height: compactButtonHeight,
+      minHeight: '1.7142em', // 24px
```
vnys, line 108: Perhaps this is superfluous 🤔  Just wanted to see if it had any effect on layout shift.

packages/eds-core-react/src/components/Button/tokens/button.ts
```diff
@@ -98,7 +105,11 @@ export const button: ButtonToken = {
...
+      spacings: {
+        top: 'calc(0.2857em - 1px)', // calc(4px - 1px) - 4/14
+        bottom: 'calc(0.2857em - 1px)',
+      },
       clickbound: {
```
vnys, line 113: …and perhaps this too. What good are clickbounds when compact mode breaks with UU in the first place?

vnys:
* What _doesn’t_ scale is the border-width. Maybe it should? 🤔 Any thoughts on that @equinor/eds-ux? We could use 1px as the minimum width for example when users decrease the font-size, but use a relative width when users increase the font-size.
* Border-radius is relative to the font-size, it’s 4px when the font-size is 14px as in Figma, but increases in parallell with the font-size.
* The icon doesn’t scale. Maybe it should? 🤔 When the button is scaled up, it looks a bit weird when the icon doesn’t scale.

SvSven:
Not everyone uses Chrome... 😬 Which is a good thing.

Anyway without diving into the philosophy of pixels vs relative units and all that, I had a look at storybook and it seems to solve our issues at least:

image

image

vnys:
> Not everyone uses Chrome... 😬  Which is a good thing.

Absolutely! I used Chrome as an example since that’s what most people use. I used Firefox Developer Edition too for a while, but on Chrome OS it’s pretty much just Chrome – I haven’t bothered with installing anything else.

wenche:
I think we should rename this PR to something that is more descriptive 👍  It goes far beyond a simple remove fixed height imho.

wenche:
People on the mighty Internet suggest to be explicit about the calculations for rem/em, as a result it might be easier to understand what is going on without changing the mental model? Leaving it here as an example, but if you change the font size on the root things will be more complicated
`font-size: calc(24 / 16 * 1rem);` <- In this example, 24 is the value we are aiming for 24px

wenche:
My last comment is about complexity. This obviously makes things a lot more complicated by using `em` and `rem` for all spacings, especially with a lot of decimals. I think it's hard to understand what is going on by just looking at the code, there's so much going on. This might or might not be what we want, but if we really want to go down this road we need to ensure that it makes sense in terms of a much better user experience. (because of the added complexity) People tend to have different opinions on which units to use for different use cases beyond typography. :)

vnys:
> It goes far beyond a simple remove fixed height

I think the other changes are side effects of removing the fixed height. Lets say you have a button with a border-radius of 5px and a font-size of 16px for example. If you increase the font-size to 32px then it looks like the border-radius is only 2.5px relative to the font-size. And the same goes for spacing. So the way the button behaves now, with the exception of border-width, is that when you increase the font-size, the button always looks the way it was designed in Figma, only larger. But if you take a screenshot of it, and scale that down, it should fit on top of the default one.

The reason I didn’t change the border-width, was because that would make the button look heavier as the font-size increases – which I don’t think is something we want necessarily. 

In EnergyVision all the margins and paddings are dynamic based on the size of the viewport, so then I think it would look strange if the button didn’t follow the same principles 🤷🏻‍♂️

vnys:
> be explicit about the calculations for rem/em

That’s a good idea, I’ll change that.

vnys:
> I think it's hard to understand what is going on by just looking at the code

I agree, so adding the calculations would make that easier. And using css custom properties makes it easier too. I revisited the calculations for the dynamic typography a couple of weeks ago in a [Codepen](https://codepen.io/vnys/pen/rNGvwWN) and did it with custom properties instead of using the Power Apps calculator I made a while back. It’s still relatively complex, but I would argue that using custom properties makes it more readable. The `--convertToRem` variable is only used as a multiplier to convert `px` to `rem`.

And there’s always a debate over which units to use, as you say. I think relative units make it more responsive and accessible, but I use px for some things, mostly borders. The discussion about rem vs. em is another topic, but I think that depends on what you want. If border-radius, padding and font-size is harmonious and everything falls apart if you change one of them I then that wouldn’t be a very robust design in my opinion.

In the example you mentioned on Slack from that CSS course, I think the author has some valid points. But in his first example, I wouldn’t use 2ems in both of those styles when they don’t mean the same, I would use an em value that would translate to the same px value. And I also agree with the point about having less space for text if the padding is increased, but that doesn’t take large screen layouts into considerations, where you actually _want_ more space – because the alternative would be narrow columns of content surrounded by white space which is precisely what we wanted to avoid. 

I think we should get @BirteThornquist to take a look at it too, I can publish the component to npm and you can try it out, perhaps it’s easier to evaluate when you use the component in a web page – and I’ll use the `calc()` function to make values easier to reason about and start working on the table component.

vnys:
…or alternatively I could make two quick examples in Codepen so we can compare the **font-size only** approach to the **all-in** approach.

benteljones:
From my point of view, it should definitely be considered whether the edge widths, the rounded corners, are also scaled when the buttons are increased. When the buttons get bigger and the rounded corners remain with a corner radius of 4, the shape will change character and the experience of proportions will be different.

This also applies to the line thickness of outline buttons, although it can withstand a larger scale before it changes expression. Icon scaling should also be considered.

In general, changing one single proportion in a molecule / template etc. will also give changes to the other "elements" if the goal is to give the same form experience.

vnys:
> From my point of view, it should definitely be considered whether the edge widths, the rounded corners, are also scaled when the buttons are increased. When the buttons get bigger and the rounded corners remain with a corner radius of 4, the shape will change character and the experience of proportions will be different.

That’s what I thought too, so the border-radius is relative to the font-size.
 
> This also applies to the line thickness of outline buttons, although it can withstand a larger scale before it changes expression. Icon scaling should also be considered.

Yes, I think you’re right about that too, that the icons should also scale. But then there’s the issue of pixel snapping, so we may have to investigate that in a separate issue 🤔 

> In general, changing one single proportion in a molecule / template etc. will also give changes to the other "elements" if the goal is to give the same form experience.

I agree.

vnys:
Note to self: 

Do a visual check on components that use button: 
- [ ] Singleselect
- [ ] Multiselect
- [ ] Search
- [ ] Popover
- [ ] Pagination
- [ ] Sidesheet

vnys:
Published to npm as `v0.18.0-envis.1`

packages/eds-core-react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@equinor/eds-core-react",
-  "version": "0.17.0",
+  "version": "0.18.0-envis.3",
```
oddvernes, line 3: I suppose this version number should go back to 0.17 before merging into develop

oddvernes (approved):
LGTM, just resolve conflicts 👍

packages/eds-core-react/package.json
vnys: Yes, good point!

PR #1870, opened 2022-01-07
🐛Fix argstable sync with useArgs

### LINKED ISSUE #1825: Enhance interactive Storybook stories with props-table
Our "Default" stories today have a props-table generated for them by Storybook based on the components api. This props-table is interactive and lets users test the component with different props values in real-time. 

Some of our components which are more complex and depends on external state (mostly via `useState`) may diverge and not reflect the actual state of the component example in the props-table.

We can use the [useArgs](https://storybook.js.org/docs/react/addons/addons-api#useargs) hook from Storybook to update the "args" state to make the props-table in sync with story/component state.

! Update the following list with stories we discover needs to be enchanced.

## Actions
- [x] Scrim diverges when toggling open/close state
- [x] Dialog diverges when toggling open/close state

oddvernes:
Resolves #1825

### DIFF
--- a/packages/eds-core-react/src/components/Dialog/Dialog.stories.tsx
+++ b/packages/eds-core-react/src/components/Dialog/Dialog.stories.tsx
@@ -56,9 +56,12 @@ export const Default: Story<DialogProps> = (args) => {
   const handleClose = () => {
     updateArgs({ open: false })
   }
+  const handleOpen = () => {
+    updateArgs({ open: true })
+  }
   return (
     <>
-      <Button aria-haspopup="dialog" onClick={() => updateArgs({ open: true })}>
+      <Button aria-haspopup="dialog" onClick={handleOpen}>
         Trigger Dialog
       </Button>
       <Dialog open={open} onClose={handleClose} isDismissable={isDismissable}>
@@ -79,12 +82,15 @@ export const Default: Story<DialogProps> = (args) => {
 
 export const Dismissable: Story<DialogProps> = () => {
   const [isOpen, setIsOpen] = useState(false)
+  const handleOpen = () => {
+    setIsOpen(true)
+  }
   const handleClose = () => {
     setIsOpen(false)
   }
   return (
     <>
-      <Button aria-haspopup="dialog" onClick={() => setIsOpen(true)}>
+      <Button aria-haspopup="dialog" onClick={handleOpen}>
         Trigger Dialog
       </Button>
       <Dialog open={isOpen} isDismissable onClose={handleClose}>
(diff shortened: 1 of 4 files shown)


oddvernes:
Table seems to be a separate issue from this to me. is density event supposed to be a prop for table? Is it not supposed to use edsprovider like everywhere else?

mimarz:
> Table seems to be a separate issue from this to me. is density event supposed to be a prop for table? Is it not supposed to use edsprovider like everywhere else?

Good point, its not. Maybe we can make a tailored storybook component for all our "compact" stories? Either way, a separate issue :)

packages/eds-core-react/src/components/Dialog/Dialog.stories.tsx
```diff
@@ -50,27 +51,25 @@ const Placeholder = styled.div`
...
-  const [isOpen, setIsOpen] = useState(open)
+  const [, updateArgs] = useArgs()
   const handleClose = () => {
-    setIsOpen(false)
+    updateArgs({ open: false })
   }
   return (
     <>
-      <Button aria-haspopup="dialog" onClick={() => setIsOpen(true)}>
+      <Button aria-haspopup="dialog" onClick={() => updateArgs({ open: true })}>
```
mimarz, line 61: Add a `handleOpen` function? To be concise with our other stories.

mimarz (changes requested):
Why did you use `@storybooj/client-api` instead of `@storybook/api`? The first seems to be [deprecated](https://www.npmjs.com/package/@storybook/client-api) and is not mentioned in the storybook documentation.

packages/eds-core-react/src/components/Dialog/Dialog.stories.tsx
oddvernes: Ok added `handleOpen `and `handleClose `to all the dialog stories

oddvernes:
> Why did you use `@storybooj/client-api` instead of `@storybook/api`? The first seems to be [deprecated](https://www.npmjs.com/package/@storybook/client-api) and is not mentioned in the storybook [documentation](https://www.npmjs.com/package/@storybook/client-api).

Something i spent much of friday trying to figure out. The /api one simply will not work (`TypeError: Cannot read properties of undefined (reading 'getCurrentStoryData')`) whereas the "deprecated" /client-api one works fine and is what everyone seems to use if you search for `import { useArgs } from '@storybook/client-api'` vs `import { useArgs } from '@storybook/api' `on github

mimarz:
> > Why did you use `@storybooj/client-api` instead of `@storybook/api`? The first seems to be [deprecated](https://www.npmjs.com/package/@storybook/client-api) and is not mentioned in the storybook [documentation](https://www.npmjs.com/package/@storybook/client-api).
> 
> 
> 
> Something i spent much of friday trying to figure out. The /api one simply will not work (`TypeError: Cannot read properties of undefined (reading 'getCurrentStoryData')`) whereas the "deprecated" /client-api one works fine and is what everyone seems to use if you search for `import { useArgs } from '@storybook/client-api'` vs `import { useArgs } from '@storybook/api' `on github

Fair enough. Maybe its fixed in the future, lets keep an eye out 👀

mimarz (approved):
LGTM 👍

PR #1888, opened 2022-01-12
✨ Icon compact support

### LINKED ISSUE #1827: Add compact density for `Icon`
With the addition of separate compact icons we need to update `Icon` to support this. How this will be implemented in code will depend on #1826 

- [ ] React

- [ ] Design

- [ ] Documentation

Edit:
Notes:
- Default size will need to be adjusted to 18x18 for `small` icons when density is compact, or else they become larger than `default` in a 24x24 size.

mimarz:
resolves #1827


oddvernes (approved):
LGTM 👍

PR #1891, opened 2022-01-13
✨ Disabled options in Combobox

### LINKED ISSUE #1430: Disable individual items in a MultiSelect
We are developing a tool, where we use a MultiSelect component to allow users to select one or more 'items' to be plotted in a figure. For various reasons, some of the items that may be selected in the MultiSelect do not have any data available for plotting. Thus, if the user selects such an item, nothing is plotted.

Now, we could just remove these "useless" items from the list that we pass to the MultiSelect component. However, for our use case, it is important for the user to be made aware that these items do exist, but that their data is missing. Therefore, we would instead like to keep the items in the dropdown list shown by the MultiSelect component, but make it so that they are not selectable. Perhaps these items could be shown as being "disabled" in some way (e.g., greyed out text).

To summarise, I would like to disable the selection of individual items in a MultiSelect component.

mimarz:
resolves #1430

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -17,6 +17,7 @@ scripts:
   test:lab: pnpm --filter @equinor/eds-lab-react run test
   test:utils: pnpm --filter @equinor/eds-utils run test
   test:watch:core-react: pnpm --filter @equinor/eds-core-react run test:watch
+  test:watch:lab: pnpm --filter @equinor/eds-lab-react run test:watch
   test:update-snapshots:core-react: pnpm --filter @equinor/eds-core-react run test:update-snapshots
   test:update-snapshots:lab: pnpm --filter @equinor/eds-lab-react run test:update-snapshots
   storybook:core-react: pnpm --filter @equinor/eds-core-react run storybook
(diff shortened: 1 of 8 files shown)


oddvernes:
Two comments that are not specific to disabled but the combobox in general: 

Perhaps we should have user-select: none on the item text since it looks a bit janky if you drag the mouse a bit as you are selecting something and text is highlighted.

Mouse down on menu item removes the focus border around the input until mouse up, I don't know if that is a design decision but it looks maybe a bit distracting to me.

mimarz:
> 



> Two comments that are not specific to disabled but the combobox in general:
> 
> Perhaps we should have user-select: none on the item text since it looks a bit janky if you drag the mouse a bit as you are selecting something and text is highlighted.

Didn't think of `use-selet:none`, added it now :)  Good suggestion! 👍 


> Mouse down on menu item removes the focus border around the input until mouse up, I don't know if that is a design decision but it looks maybe a bit distracting to me.

We have this behaviour with todays selects as well. Maybe a separate issue for it, as I see none of the `downshiftjs` examples have this...

PR #1897, opened 2022-01-17
 ✨Typed polymorphic Button

### LINKED ISSUE #1736: Casting elements using "as" and correct Typescript type
We are using the property `as` to mimick styled-components api for casting an element-type to one or the other. 

We have used `ElementType` from `react` for this but are uncertain now if its the correct type to use as its flagged as "any" value in reality.

Casting components from one element type to another also poses a problem as the props type is not changed/updated to reflect the new set of native attributes a component has. Most notable example of this is when `Button` is casted from `button` to `a` when `href` is defined. You are unable to set any of the `a` [properties](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attributes) as type is still set to `button.`

Affected components `Button`, `List`, `Typography`, `Breadcrumbs.Breadcrumb`

oddvernes:
Resolves #1736 

I guess this is more of a spike, so far it is only implemented on button
based on the code in [this](https://www.benmvp.com/blog/forwarding-refs-polymorphic-react-component-typescript/) article and [this](https://github.com/kripod/react-polymorphic-box) project

### DIFF
--- a/packages/eds-core-react/src/components/Button/Button.tsx
+++ b/packages/eds-core-react/src/components/Button/Button.tsx
@@ -1,4 +1,4 @@
-import { forwardRef, ElementType, ButtonHTMLAttributes } from 'react'
+import { forwardRef, ElementType, ReactElement } from 'react'
 import styled, { css, ThemeProvider } from 'styled-components'
 import { token as buttonToken } from './tokens'
 import { ButtonTokenSet, ButtonToken } from './Button.types'
@@ -7,6 +7,8 @@ import {
   bordersTemplate,
   outlineTemplate,
   spacingsTemplate,
+  PolymorphicComponentPropsWithRef,
+  PolymorphicRef,
   useToken,
 } from '@equinor/eds-utils'
 import { InnerFullWidth } from './InnerFullWidth'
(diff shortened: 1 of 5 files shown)


mimarz:
1. Maybe we could set these types in `Button.types`
2. Ideally we would want the type to change when `href` is defined. Currently set to be the type thats passed in through the `as` property, so maybe some more research on how to do this conditionally if a prop is set 🤔
3. We could add a separate types folder to `eds-utils` for the polymorphic type helpers.

oddvernes:
Strangely after rebase on develop, I now have to add tabIndex to ButtonProps otherwise i get
`Type 'number' is not assignable to type 'LibraryManagedAttributes<C, PropsWithoutRef<ComponentProps<C>>>[string]'.ts(2322)`
Not sure why this is

mimarz:
> Strangely after rebase on develop, I now have to add tabIndex to ButtonProps otherwise i get
> 
> `Type 'number' is not assignable to type 'LibraryManagedAttributes<C, PropsWithoutRef<ComponentProps<C>>>[string]'.ts(2322)`
> 
> error on tabIndex prop
> 
> Not sure why I get this now and not before

Try running our init script or restart vscode as different versions of typescript can cause weird behaviours

mimarz:
Based on findings here, and noted in #1736. I suggest closing this issue and solving this with a generic polymorphic type in `eds-utils` @oddvernes

PR #1899, opened 2022-01-17
Dialog Refactor

### LINKED ISSUE #1896: Conform on margin/padding for Dialog
For making our components easier to use we have decided to move padding from Dialog parent component, and add to child components.  As well implementing component structure based on:

- Header
- Cotent
- Actions

which will bring consistency and create better guidelines for developers.

Actions:

- Move padding to child components.
- Add `DialogHeader` component.
- Change names of the child components.  

This is a breaking change.

laliq:
Resolves #1896


oddvernes:
I tried to figure out why CustomContent have a minheight of 80px but the git history only goes back to the typescript merge so it is just always there

laliq:
> 

Same here.
Maybe designers should look in to it too?

oddvernes (approved):
LGTM 👍

PR #1908, opened 2022-01-24
Add keyboard navigation

### LINKED ISSUE #1901: No kb-navigation in Playground "settings" menu
**Describe the bug**

Keyboard navigation is not implemented/not working in storybook playground "settings" menu (last column in the table)

1. Open playground and navigate to of the menu buttons with keyboard
2. Press enter to open menu
3.  Menu is not focusable

**Expected behavior**

Should be able to navigate menu same way as in the menu stories, with arrow buttons

**Additional context**

I tried quickly implementing the focus logic from the menu stories but it did not readily work, perhaps it has something to do with the isOpen logic here defined by anchor-element which is set by an event which is sort of async

laliq:
Resolves #1901

### DIFF
--- a/packages/eds-core-react/stories/playground/DataTable.tsx
+++ b/packages/eds-core-react/stories/playground/DataTable.tsx
@@ -174,7 +174,7 @@ const MenuButton = ({ row }: { row: string[] }) => {
           focus={focus}
           onClose={closeMenu}
         >
-          <Menu.Item onKeyDown={onKeyPress}>
+          <Menu.Item onKeyDown={onKeyPress} onClick={closeMenu}>
             <Icon name="folder" size={16} />
             <Typography group="navigation" variant="menu_title">
               Open
@@ -188,7 +188,7 @@ const MenuButton = ({ row }: { row: string[] }) => {
               CTRL+O
             </Typography>
           </Menu.Item>
-          <Menu.Item onKeyDown={onKeyPress}>
+          <Menu.Item onKeyDown={onKeyPress} onClick={closeMenu}>
             <Icon name="copy" size={16} />
             <Typography group="navigation" variant="menu_title">
               Copy
--- a/packages/eds-core-react/stories/playground/DataTable.tsx
+++ b/packages/eds-core-react/stories/playground/DataTable.tsx
@@ -174,7 +174,7 @@ const MenuButton = ({ row }: { row: string[] }) => {
           focus={focus}
           onClose={closeMenu}
         >
-          <Menu.Item onKeyDown={onKeyPress} onClick={closeMenu}>
+          <Menu.Item onClick={closeMenu}>
             <Icon name="folder" size={16} />
             <Typography group="navigation" variant="menu_title">
               Open
@@ -188,7 +188,7 @@ const MenuButton = ({ row }: { row: string[] }) => {
               CTRL+O
             </Typography>
           </Menu.Item>
-          <Menu.Item onKeyDown={onKeyPress} onClick={closeMenu}>
+          <Menu.Item onClick={closeMenu}>
             <Icon name="copy" size={16} />
             <Typography group="navigation" variant="menu_title">
               Copy
--- a/packages/eds-core-react/stories/playground/DataTable.tsx
+++ b/packages/eds-core-react/stories/playground/DataTable.tsx
@@ -114,7 +114,6 @@ const toCellValues = (data: Data[], columns: Column[]) =>
 const MenuButton = ({ row }: { row: string[] }) => {
   const [anchorEl, setAnchorEl] = useState<HTMLElement>()
   const isElement = Boolean(anchorEl)
-
   const [isOpen, setIsOpen] = useState<boolean>(false)
   const [focus, setFocus] = useState<MenuProps['focus']>(null)
 
@@ -133,6 +132,7 @@ const MenuButton = ({ row }: { row: string[] }) => {
     e.preventDefault()
     e.stopPropagation()
     setAnchorEl(e.currentTarget)
+
     switch (key) {
       case 'Enter':
         isOpen ? closeMenu() : openMenu('first')
@@ -174,7 +174,7 @@ const MenuButton = ({ row }: { row: string[] }) => {
           focus={focus}
           onClose={closeMenu}
         >
-          <Menu.Item onKeyDown={onKeyPress} onClick={closeMenu}>
+          <Menu.Item onClick={closeMenu}>
             <Icon name="folder" size={16} />
             <Typography group="navigation" variant="menu_title">
               Open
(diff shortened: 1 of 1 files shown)


laliq:
Still i detect two 🐛 

1. When using arrows to focus  the `menu.item`,  the `menu` detaches itself from default position.
2. When clicking with mouse on the `menu.item` it won't close the `menu`.

packages/eds-core-react/stories/playground/DataTable.tsx
```diff
@@ -122,18 +161,20 @@ const MenuButton = ({ row }: { row: string[] }) => {
...
           aria-labelledby={`menu-button-${row.toString()}`}
-          open={isOpen}
+          open={isElement}
           anchorEl={anchorEl}
-          onClose={() => setAnchorEl(null)}
+          focus={focus}
+          onClose={closeMenu}
         >
-          <Menu.Item>
+          <Menu.Item onKeyDown={onKeyPress} onClick={closeMenu}>
```
oddvernes, line 177: Menu.item should not have onKeyDown, only onClick. It is a button element so pressing enter on it is same as onClick. This should fix the menu flying around the screen

oddvernes:
Actually, there is something quite funky going on, and I think pressing enter on the menu item does not count as a click after all. Going to the menu stories, onClick inside menuItem never registers anything when pressing enter (as opposed to clicking with mouse pinter), but `console.log(new Error())` in the closeMenu function claims it was triggered by onClick on the anchor element button. I am a little bit confused. Suggest @mimarz also takes a look

laliq:
> Actually, there is something quite funky going on...

Yes i agree @oddvernes, and i'm a bit confuse too...

oddvernes:
Maybe it should be its own ticket if the problem is deeper in the menu itself. And put this pr aside until it if fixed

mimarz:
> Actually, there is something quite funky going on, and I think pressing enter on the menu item does not count as a click after all. Going to the menu stories, onClick inside menuItem never registers anything when pressing enter (as opposed to clicking with mouse pinter), but `console.log(new Error())` in the closeMenu function claims it was triggered by onClick on the anchor element button. I am a little bit confused. Suggest @mimarz also takes a look

I am observing the same. 

* `MenuItem.onClick` is only triggered by pressing "space" key.
* Anchor element (button) `onClick` is triggered when `MenuItem` is focused and key press.
  * Adding `preventDefault` to `MenuList` `handleKeyPress` stops this bubbling
* This behaviour is probably a side-effect of trying to focus anchor element after `Menu` closes. Removing the [globalKeyPress for Enter](https://github.com/equinor/design-system/blob/94a1986d4bc48b0ff7eaa2a4bc0ae0d3eceb1c20/packages/eds-core-react/src/components/Menu/Menu.tsx#L71) "fixes" this behaviour.

As @oddvernes suggested, we need an issue to fix this bug in `Menu` :)

packages/eds-core-react/stories/playground/DataTable.tsx
```diff
@@ -112,7 +113,45 @@ const toCellValues = (data: Data[], columns: Column[]) =>
 
 const MenuButton = ({ row }: { row: string[] }) => {
   const [anchorEl, setAnchorEl] = useState<HTMLElement>()
-  const isOpen = Boolean(anchorEl)
+  const isElement = Boolean(anchorEl)
```
mimarz, line 116: Maybe keep this as `isOpen`? Looks confusing when you look at `Menu` and `open={isElement}`. Not sure If I understood the need for an additional `useState` 🤔

mimarz (changes requested):

packages/eds-core-react/stories/playground/DataTable.tsx
```diff
@@ -112,7 +113,45 @@ const toCellValues = (data: Data[], columns: Column[]) =>
...
+  const [isOpen, setIsOpen] = useState<boolean>(false)
```
mimarz, line 117: Maybe keep this (`isElement`) as `isOpen`? 

Looks confusing when you look at `Menu` and `open={isElement}`. Not sure If I understood the need for an additional `useState` 🤔

packages/eds-core-react/stories/playground/DataTable.tsx
```diff
@@ -112,7 +113,42 @@ const toCellValues = (data: Data[], columns: Column[]) =>
...
+  const openMenu = (focus: MenuProps['focus']) => {
+    setFocus(focus)
+  }
+  const closeMenu = () => {
+    setFocus(null)
+    setAnchorEl(null)
+  }
+
+  const onKeyPress = (e: React.KeyboardEvent<HTMLButtonElement>) => {
+    const { key } = e
```
oddvernes, line 128: I haven't checked but you might need a `if (key === 'tab') return` here as was needed in the stories?

laliq: Not sure what do you mean, maybe i'm missing something, but the tab key does work.

oddvernes: It does not for me until i add that line (but its 'Tab' with capital T) (I mean pressing tab should move focus to next element in table, not ne stuck on the menu button)

laliq: I see, let me add it then.

packages/eds-core-react/stories/playground/DataTable.tsx
```diff
@@ -112,7 +113,42 @@ const toCellValues = (data: Data[], columns: Column[]) =>
...
+    e.preventDefault()
+    e.stopPropagation()
+    setAnchorEl(e.currentTarget)
```
mimarz, line 131: Wouldn't this open `Menu` on any keypress on `Button`? 😅 

Maybe setting the `anchorEl` is best done inside `openMenu` as its what we are semantically doing when we set it.

mimarz (changes requested):

packages/eds-core-react/stories/playground/DataTable.tsx
laliq: You were absolutely right! I didin’t see it before, i was so focused on the menu element. Thanks a lot 👍

packages/eds-core-react/stories/playground/DataTable.tsx
laliq: Good call, thanks 👍

packages/eds-core-react/stories/playground/DataTable.tsx
```diff
@@ -112,7 +113,48 @@ const toCellValues = (data: Data[], columns: Column[]) =>
...
+    }
+    e.preventDefault()
+    e.stopPropagation()
+
+    switch (key) {
+      case 'Enter':
+        isOpen ? closeMenu() : openMenu('first', e)
+        break
+      case 'ArrowDown':
+        if (isOpen) {
```
oddvernes, line 144: The arrowDown and ArrowUp cases do nothing and can either be removed, or fixed so they open the menu on first or last item

oddvernes (approved):
LGTM 👍

mimarz (approved):
LGTM 👍

PR #1914, opened 2022-01-25
📝 Combobox documentation

### LINKED ISSUE #1912: Document Combobox
Create documentation for Combobox so its easier for our users to understand the intended use for this "new" component when releasing it to public testing.

mimarz:
resolves #1912

### DIFF
--- a/packages/eds-lab-react/.storybook/components/PropsTable.tsx
+++ b/packages/eds-lab-react/.storybook/components/PropsTable.tsx
@@ -1,0 +1,5 @@
+import { ArgsTable, PRIMARY_STORY } from '@storybook/addon-docs'
+
+export const PropsTable = ({ story = PRIMARY_STORY, ...props }) => (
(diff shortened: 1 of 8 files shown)


oddvernes:
Should maybe force height on a container so all the examples arent in a tiny window, like the wrapper used in the select stories? 
bilde

mimarz:
> Should maybe force height on a container so all the examples arent in a tiny window, like the wrapper used in the select stories? bilde

yeah, I removed the height as I intended to fix this #1913

oddvernes:
But the outer div owned by storybook have overflow hidden on it
bilde

mimarz:
> But the outer div owned by storybook have overflow hidden on it bilde

Was thinking we could use portals the same way we do for `Menu` & `Popover`? 🤔

oddvernes (approved):
Other than compact not being set with variable, LGTM 👍

PR #1918, opened 2022-01-26
Add keyboard selection

### LINKED ISSUE #1910: MenuItem keyboard selection
We currently don't have a clear example on how `MenuItem` is selected/triggered when pressing Enter key as it also triggers the `onClick` (due to event bubbling)  and causes some unwanted behaviours.

Notes: 
* Should we expect users to add key handlers on `MenuItem` like they do with `onClick`? 
  * Maybe we just need to add a shim for event bubbling for keystrokes (like we have for `MenuItem.onClick`)
    * This might break anchor element focus on close?  
* https://www.w3.org/TR/wai-aria-practices-1.1/examples/menu-button/menu-button-links.html

laliq:
Resolves #1910

### DIFF
--- a/packages/eds-core-react/src/components/Menu/MenuItem.tsx
+++ b/packages/eds-core-react/src/components/Menu/MenuItem.tsx
@@ -100,7 +100,7 @@ export type MenuItemProps = {
   /** Is disabled */
   disabled?: boolean
   /** onClick handler */
-  onClick?: (e: React.MouseEvent | React.KeyboardEvent) => void
+  onClick?: (e: React.MouseEvent) => void
 } & React.HTMLAttributes<HTMLButtonElement>
 
 export const MenuItem = memo(
@@ -117,15 +117,6 @@ export const MenuItem = memo(
     }
 
     const isFocused = index === focusedIndex
-
-    const handleKeyPress = (e: React.KeyboardEvent<HTMLButtonElement>) => {
-      const { key } = e
-      if (key === 'Enter') {
-        if (onClick) {
-          onClick(e)
-        }
-      }
-    }
 
     const props = {
       ...rest,
@@ -150,7 +141,6 @@ export const MenuItem = memo(
             }
           }
         }}
-        onKeyDown={handleKeyPress}
       >
         <Content>{children}</Content>
       </Item>
--- a/packages/eds-core-react/src/components/Menu/MenuItem.tsx
+++ b/packages/eds-core-react/src/components/Menu/MenuItem.tsx
@@ -101,6 +101,8 @@ export type MenuItemProps = {
   disabled?: boolean
   /** onClick handler */
   onClick?: (e: React.MouseEvent) => void
+  /** onClose handler */
+  onClose?: (e: React.KeyboardEvent) => void
 } & React.HTMLAttributes<HTMLButtonElement>
 
 export const MenuItem = memo(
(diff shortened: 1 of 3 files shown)


oddvernes (changes requested):
Currently the keypress only handles onclose, which only closes the menu without selecting anything. But it is onClick we need. This is what is used to call different functions. The problem is onClick expects a mouseevent . We need enter and click to behave exactly the same.
My suggestion was to just dispatch a click event when you press enter, which does seem to work perfectly
`      if (key === 'Enter') {
        e.target.dispatchEvent(new Event('click', { bubbles: true }))
      }`
But maybe @mimarz have some some reservations?:P

Also in menu stories something is trapping focus on the anchor button while tabbing, but this seems to be an existing bug, having to do with event.preventDefaut/stopPropagation in the stories maybe

packages/eds-core-react/src/components/Menu/MenuItem.tsx
```diff
@@ -101,6 +101,8 @@ export type MenuItemProps = {
   disabled?: boolean
   /** onClick handler */
   onClick?: (e: React.MouseEvent) => void
+  /** onClose handler */
+  onClose?: (e: React.KeyboardEvent) => void
```
mimarz, line 105: Suggest removing this as we already have an `onClose` passed on top level and it deviates from intended expectation with click vs. keydown.

mimarz (changes requested):
We need to make sure the expected behaviour is the same for when either click or keypress is used. That being that the `onClose` handler passed to `Menu` is called with the event. 

Introducing an `onClose` on `MenuItem` in addition to the on one `Menu` would cause confusion as on or the other would be called be it click or keypress.

mimarz:
> Currently the keypress only handles onclose, which only closes the menu without selecting anything. But it is onClick we need. This is what is used to call different functions. The problem is onClick expects a mouseevent . We need enter and click to behave exactly the same. My suggestion was to just dispatch a click event when you press enter, which does seem to work perfectly ` if (key === 'Enter') { e.target.dispatchEvent(new Event('click', { bubbles: true })) }` But maybe @mimarz have some some reservations?:P
> 
> Also in menu stories something is trapping focus on the anchor button while tabbing, but this seems to be an existing bug, having to do with event.preventDefaut/stopPropagation in the stories maybe

Not sure, mimicing a click event when a key is pressed feels very wrong :P Typing should not be a problem as we can just use "or" on the event type?

oddvernes:
> Not sure, mimicing a click event when a key is pressed feels very wrong :P Typing should not be a problem as we can just use "or" on the event type?

`onClick?: (e: React.MouseEvent | React.KeyboardEvent) => void `

`      if (key === 'Enter') {
        if (onClick) {
          onClick(e)
        }`

Yes this works if you also declare the event as mouse or keyboard downstream in the stories (I don't know how many people even use this event in their apps, so its only a small breaking change for those that do)

mimarz:
> 

I was thinking more for the `onClose` handler? Lets not mess expected native behaviours on handler 😅

laliq:
Thanks guys 🙏

laliq:
https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onclick

oddvernes (approved):
LGTM 👍 
I'll make a new issue for focus getting stuck on the anchor button in menu stories

mimarz (approved):
Quirky solution! LGTM 👍

Issue #1848, opened 2022-01-03
Update core-react & lab to use eds-utils


mimarz:
* Update all references to `utils` or `hooks` to use `@equinor/eds-utils` instead of local copies. 
* Delete all local copies once updated and tested

Issue #1004, opened 2021-01-11
Autowidth on Select


cbuv:
**Is your feature request related to a problem? Please describe.**
The Select must have width  equal to the the longest text in the items array to avoid having horizontal scroll in options. 
 The main friction is when filtering on big data set will cause column width to vary because we must have wide enough columns to avoid horizontal scrolling in a Select component within the table.  

**Describe the solution you'd like**
The options container should use necessary space horizontal.  (possible bigger the than the parent component)
A auto width property similar to material-ui.

**Describe alternatives you've considered**
Custom made select

**Additional context**
Material ui's Simple Select does solve this problem.  
https://material-ui.com/api/select/
![image](https://user-images.githubusercontent.com/48285335/104202669-81075380-542b-11eb-9d59-739aa78d6c78.png)

mimarz:
Thanks for your feedback! 

Do you have an example of the use-case where this is a problem for you (since you cant't make a fixed width that works)?
We have several options here which would also be discussed with our designer, such as may the text be truncated with ellipsis or similar if width is max of parent.

We'll have a look at it on our next planning meeting and get back to you on priority :)

cbuv:
I have a dropdown with labels of up to 36 characters in a table column. 
Being forced to set width on the column make the table too wide. 

Our tables have typically 10-15 columns. 
The designers could create worst-case design of a table with dropdown cell where options have up to 40 characters.

mimarz:
We had a look at this during planning today, and I think we need to look closer at this use-case with our designers @equinor/eds-ux to determine what would be the best solution or approach to this problem. @akvam showed us some use-cases from (one-of) the app you are working on, but maybe we could have a meeting with you and our designers to have a closer look at this?

cbuv:
Sure, lets arrange a meeting.

mimarz:
Notes for self: https://github.com/floating-ui/floating-ui/issues/794

PR #1919, opened 2022-01-27
🐛 Fix Combobox dropdown layering

### LINKED ISSUE #1913: Fix dropdown stacking layer in Combobox

mimarz:
fixes #1913 

* Introduced use of react portal & popper for placement. 
* Due to using popper, options width now scales based on content.
  * ~~This can be fixed in #1004 after #1848, as changes in `usePopper` hook are needed~~
    * Edit: We decided that this will be the default behaviour now and if requested add feature for toggling this later. Decided on PR-review meet 01.02.2022

### DIFF
--- a/packages/eds-lab-react/src/components/Combobox/Combobox.stories.tsx
+++ b/packages/eds-lab-react/src/components/Combobox/Combobox.stories.tsx
@@ -89,6 +89,7 @@ Introduction.args = {
   multiple: false,
   readOnly: false,
   disabled: false,
+  disablePortal: false,
 }
 
 export const Multiple: Story<ComboboxProps<MyOptionType>> = (args) => {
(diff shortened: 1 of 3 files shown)


oddvernes:
I do sometimes get the dropdown spawning in weird positions away from the input in the "with react hook form" example, if i fill it in, submit, reset, click on one input, close it again and then open another input it is somewhere far to the left of the screen until i scroll one tick and it repositions properly

mimarz:
> I do sometimes get the dropdown spawning in weird positions away from the input in the "with react hook form" example, if i fill it in, submit, reset, click on one input, close it again and then open another input it is somewhere far to the left on the screen until i scroll one tick and it repositions properly

What did you test on? I can't seem to reproduce the behaviour...

oddvernes:
Hm, I get it on both firefox and chrome every time. Also i get it on every field in hook form after reset (i had added disablePortal on one when i first wrote and that one was positioned normal)

PR #1922, opened 2022-01-27
👷‍♂️ Add Lab to Publish Storybook

### LINKED ISSUE #1921: Setup publish for eds-lab-react Storybook

mimarz:
resolves #1921

### DIFF
--- a/.github/workflows/publish_lab.yaml
+++ b/.github/workflows/publish_lab.yaml
@@ -6,6 +6,10 @@ on:
         description: 'Tag for npm package (next | latest)'
         required: true
         default: 'next'
+      storybook-slot:
+        description: 'Azure webapp slot for Storybook'
+        required: false
+        default: 'lab'
 jobs:
   setup:
     name: Setup base files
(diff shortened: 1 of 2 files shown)


nilsml (commented):
A couple of questions related to the solution (not the PR as such)

1. Why are we publishing all storybooks in the same script? Would all changes affect all storybooks? Or could we have separate actions for each and react on different triggers?
2. Why are we using Azure app instead of Radix?
3. Since we are using Azure web app, would it be an option to ditch Docker and deploy the app in a SaaS manner?

mimarz:
> A couple of questions related to the solution (not the PR as such)
> 
> 1. Why are we publishing all storybooks in the same script? Would all changes affect all storybooks? Or could we have separate actions for each and react on different triggers?
> 2. Why are we using Azure app instead of Radix?
> 3. Since we are using Azure web app, would it be an option to ditch Docker and deploy the app in a SaaS manner?

1. I started making a separate action but ended up using the same as we have used this for mainly manually deploying the Storybook as the other actions for the packages include automatic deployment of their respective storybooks to stay in sync with released package. That reminds me I need to update publish labs to also deploy this storybook :)

2. We started using azure before radix was a thing and haven't moved over yet.

3. What benefit would we get from using SaaS manner? I feel docker provides more flexibility, easier rollbacks and consistency.

nilsml:
Ok, fair enough. Since we are using docker, switching to Radix should be considered at some point since that is where we are hosting most other things. About ditching docker my point was that you take away some complexity from the solution and let someone else handle it. Also nice for scaling, but we don't need that here. I guess current solution is inline with Equinor's wish to containerize applications…

mimarz:
> Ok, fair enough. Since we are using docker, switching to Radix should be considered at some point since that is where we are hosting most other things. About ditching docker my point was that you take away some complexity from the solution and let someone else handle it. Also nice for scaling, but we don't need that here. I guess current solution is inline with Equinor's wish to containerize applications…

The plan has been to move to radix at one point, but its not being prioritised as we only host Storybook and a CDN now in Azure.

nilsml (approved):
Go for it!

PR #1931, opened 2022-02-02
🐛 Fixed trapped focus in menu stories

### LINKED ISSUE #1924: Menu stories: Focus trapped on anchor button
**Describe the bug**

While tabbing around in the menu stories in storybook, whenever focus lands on a menu anchor button, focus gets stuck on the button and you can not escape with tab or shift+tab

oddvernes:
resolves #1924

### DIFF
--- a/packages/eds-core-react/src/components/Menu/Menu.stories.tsx
+++ b/packages/eds-core-react/src/components/Menu/Menu.stories.tsx
@@ -68,6 +68,7 @@ export const Default: Story<MenuProps> = (args) => {
 
   const onKeyPress = (e: React.KeyboardEvent<HTMLButtonElement>) => {
     const { key } = e
+    if (key === 'Tab') return
     e.preventDefault()
     e.stopPropagation()
     switch (key) {
@@ -136,6 +137,7 @@ export const Complex: Story<MenuProps> = () => {
 
   const onKeyPress = (e: React.KeyboardEvent<HTMLButtonElement>) => {
     const { key } = e
+    if (key === 'Tab') return
     e.preventDefault()
     e.stopPropagation()
     switch (key) {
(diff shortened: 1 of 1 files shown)


mimarz (commented):
Is there now way to better fix interaction with `Menu` and keypress `Enter`? Maybe we need to step back and have a look at what we are trying to achieve. 

These "fixes" feels like we are adding lots more unnecessary complexion for our users (in keypress on anchor button). 

* Do we need a global keypress for `Enter`? 
* Can this be done in `MenuList` or `MenuItem` instead?

oddvernes:
> Is there now way to better fix interaction with `Menu` and keypress `Enter`? Maybe we need to step back and have a look at what we are trying to achieve.
> 
> These "fixes" feels like we are adding lots more unnecessary complexion for our users (in keypress on anchor button).
> 
>     * Do we need a global keypress for `Enter`?
> 
>     * Can this be done in `MenuList` or `MenuItem` instead?

As a user I would expect the component to handle focus on enter internally, but you mentioned last week there was some reason why we need to handle onkeydown outside of the component (I forgot what that reason was though 😅 )

mimarz:
> > Is there now way to better fix interaction with `Menu` and keypress `Enter`? Maybe we need to step back and have a look at what we are trying to achieve.
> > These "fixes" feels like we are adding lots more unnecessary complexion for our users (in keypress on anchor button).
> > ```
> > * Do we need a global keypress for `Enter`?
> > 
> > * Can this be done in `MenuList` or `MenuItem` instead?
> > ```
> 
> As a user I would expect the component to handle focus on enter internally, but you mentioned last week there was some reason why we need to handle onkeydown outside of the component (I forgot what that reason was though 😅 )

There was something that didn't work with attaching handlers on passed dom elements (anhchorEl). Maybe we should try to again to handle this internally in the Menu as the current approach feels cumbersome for our users. Especially since we've seen other component libraries do this internally aswell...

oddvernes:
Made a new issue about moving keydown handling of anchor into the component #1936

mimarz (approved):
LGTM 👍 Lets just hope we get to do #1936 in not long

PR #1933, opened 2022-02-02
🐛 menu: fix unwanted viewport scroll when kb-navigation

### LINKED ISSUE #1932: Menu: focus before popper init cause scroll to top
**Describe the bug**

Menu item recieves focus instantly when using keyboard navigation, one frame earlier than popper setting transform values to the menu container. If the user is scrolled down, the browser set scrolltop to 0 because at that time the menucontainer is positioned absolutely top left.

1. Go to menu stories in storybook and scroll down to second or third story
2. Click inside the story window and press tab so the anchor button is focused. Press enter to open menu
3. Observe as the browser instantly scrolls to the top and menu is out of sight

**Expected behavior**

viewport should not scroll when opening menu

oddvernes:
resolves #1932 

Use raf to delay focus by one frame so transform values from popper take hold

### DIFF
--- a/packages/eds-core-react/src/components/Menu/MenuItem.tsx
+++ b/packages/eds-core-react/src/components/Menu/MenuItem.tsx
@@ -128,8 +128,10 @@ export const MenuItem = memo(
       <Item
         {...props}
         ref={useCombinedRefs<HTMLButtonElement>(ref, (el) => {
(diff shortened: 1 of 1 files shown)


mimarz (approved):
LGTM 👍

PR #1942, opened 2022-02-07
✅ Add a11y test for Accordion

### LINKED ISSUE #1941: Add a11y test for Accordion

laliq:
Resolves #1941


laliq:
[Design patter for](https://www.w3.org/TR/wai-aria-practices-1.2/#accordion)`Accordion` shows that:

> The title of each accordion header is contained in an element with role button.
Each accordion header button is wrapped in an element with role heading that has a value set for [aria-level](https://www.w3.org/TR/wai-aria-1.2/#aria-level) that is appropriate for the information architecture of the page.

I have looked into [gov.uk](https://design-system.service.gov.uk/components/accordion/#known-issues-and-gaps) work and investigation that they made. And other DS, it look that the accordion `header` is a `button` element, while in EDS is a `h2` element.

Actions:
Wrap header in `button`.

mimarz:
> Test will not pass with `role = button` on `h` element.
> 
> [Design patter for](https://www.w3.org/TR/wai-aria-practices-1.2/#accordion)`Accordion` shows that:
> 
> > The title of each accordion header is contained in an element with role button.
> > Each accordion header button is wrapped in an element with role heading that has a value set for [aria-level](https://www.w3.org/TR/wai-aria-1.2/#aria-level) that is appropriate for the information architecture of the page.
> 
> I have looked into [gov.uk](https://design-system.service.gov.uk/components/accordion/#known-issues-and-gaps) work and investigation that they made. And other DS, it look that the accordion `header` is a `button` element, while in EDS is a `h2` element.
> 
> Actions: Wrap header in `button`.

Reading the articles you posted would it not make more sense to put `button` inside the header element?

oddvernes:
I guess this component needs to be restructured slightly so that StyledAccordionHeader continues to be the heading (with the as-attribute, but directly inside the should be a StyledAccordionHeaderButton where you move the various aria stuff. (Or maybe button outermost and the inner is the h2 etc)

I suppose screenreaders get confused if an element is both a heading-level and button at the same time since these are two different roles

laliq:
> I guess this component needs to be restructured slightly so that StyledAccordionHeader continues to be the heading (with the as-attribute, but directly inside the should be a StyledAccordionHeaderButton where you move the various aria stuff. (Or maybe button outermost and the inner is the h2 etc)

Yes, somewhere in this alley.

> I suppose screenreaders get confused if an element is both a heading-level and button at the same time since these are two different roles

Yes it is a bit off puzzle with screenreaders, as well as with key navigation.

oddvernes (approved):
It looks like accordion is still toggled if one presses the "action buttons" using keyboard (even though there is a stopPropagation). I'm wondering if there is a way to place these buttons outside of the header button while still have them on top of it visually. It should be doable with grid, but It looks tricky with regards to how children are handled so maybe this should be its own ticket. Otherwise this looks good to go for me 👍

PR #1947, opened 2022-02-08
✅ Add a11y test for Breadcrumbs

### LINKED ISSUE #1946: Add a11y test for Breadcrumbs

laliq:
Resolves  #1946


packages/eds-core-react/src/components/Breadcrumbs/Breadcrumbs.test.tsx
```diff
@@ -24,6 +25,16 @@ describe('Breadcrumbs', () => {
     )
     expect(asFragment()).toMatchSnapshot()
   })
+  it('Should pass a11y test', async () => {
```
mimarz, line 28: Could we also have the same test with `href="#"` on each `Breadcrumb`? They are then rendered as `a` elements and would be nice to try cover that in terms of a11y also.

mimarz (changes requested):

mimarz (approved):
LGTM 👍

PR #1960, opened 2022-02-08
✅ Fixed & improved Combobox tests

### LINKED ISSUE #1930: Fix act warnings in Combobox tests
Tests seems to seemingly work but throw a lot of errors with "act", while still being green. This is probably related to introducing popper. 

Updating the tests to be async and wait for options to pop should fix the issue.

Notes: https://stackoverflow.com/questions/60115885/how-to-solve-the-update-was-not-wrapped-in-act-warning-in-testing-library-re 

image

mimarz:
fixes #1930


oddvernes (approved):
LGTM 👍

PR #1961, opened 2022-02-08
✅  Add a11y test for Input

### LINKED ISSUE #1959: Add a11y test for Input

laliq:
Resolves #1959

### DIFF
--- a/packages/eds-core-react/src/components/Input/Input.test.tsx
+++ b/packages/eds-core-react/src/components/Input/Input.test.tsx
@@ -27,8 +27,18 @@ describe('Input', () => {
     const { asFragment } = render(<Input value="input value" readOnly />)
     expect(asFragment()).toMatchSnapshot()
   })
-  it('Should pass a11y test', async () => {
-    const { container } = render(<Input value="input value" readOnly />)
+  it('Should pass a11y test when using label', async () => {
+    const label = 'Input'
+    const { container } = render(
+      <label>
+        {label}
+        <Input />
+      </label>,
+    )
+    expect(await axe(container)).toHaveNoViolations()
+  })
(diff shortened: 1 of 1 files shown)


laliq:
Here is how [axe](https://dequeuniversity.com/rules/axe/4.2/label?application=axeAPI) describes the issue and solution.

mimarz:
> Here is how [axe](https://dequeuniversity.com/rules/axe/4.2/label?application=axeAPI) describes the issue and solution.

Jepp, its a good explanation for the a11y rules for `input` fields :) 

Since we are basically just wrapping an `input` field with our `Input` component I suggest making two a11y test that cover the use-cases in which we used this component. One for `aria-label` and the other using a `label`, either wrapped or by using the `for` property with `id`.

packages/eds-core-react/src/components/Input/Input.test.tsx
```diff
@@ -26,6 +27,20 @@ describe('Input', () => {
     const { asFragment } = render(<Input value="input value" readOnly />)
     expect(asFragment()).toMatchSnapshot()
   })
+  it('Should pass a11y test when using label', async () => {
+    const label = 'Input'
```
mimarz, line 31:
```suggestion
```

packages/eds-core-react/src/components/Input/Input.test.tsx
```diff
@@ -26,6 +27,20 @@ describe('Input', () => {
...
+    const { container } = render(
+      <label>
+        {label}
```
mimarz, line 34:
```suggestion
        Label text
```

mimarz (changes requested):
Suggest just typing the label text in the label, noe need for variable interpolation if its not needed.

PR #1964, opened 2022-02-09
👷‍♂️ Working published core-react with eds-utils

### LINKED ISSUE #1949: Test downstream packages with eds-utils
Now that labs and core-react are using eds-utils, we need to test and verify that everything is working for our users when they download these packages downstream (install from npmjs.org). 

Test
- [x] Next.js create react app
- [x] Planin create react app
- [x] Get third party to test new version doesn't break their application

mimarz:
resolves #1949 

Suggest keeping the -dev version names as its the latest versions that are published under the `next` tag which is what is in development

### DIFF
--- a/.github/workflows/publish_core_react.yaml
+++ b/.github/workflows/publish_core_react.yaml
@@ -162,7 +162,14 @@ jobs:
     runs-on: ubuntu-latest
     needs: [build-utils]
     steps:
-      - name: Get utils build
+      - uses: actions/cache@v2
+        name: Get base files
+        with:
+          path: |
+            ./*
+            ~/.pnpm-store
+          key: ${{ github.sha }}
+      - name: Get eds-utils build
         uses: actions/cache@v2
         with:
           path: ./packages/eds-utils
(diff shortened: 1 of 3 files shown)


vnys (changes requested):
Looks good 👍🏻 But we should have a discussion on package.json version numbers in the develop branch, you had some strong opinions about that a little while ago. Personally I don’t think it matters that much until the next release, and makes it easier to remember to bump to the right next version.

mimarz:
> Looks good 👍🏻 But we should have a discussion on package.json version numbers in the develop branch, you had some strong opinions about that a little while ago. Personally I don’t think it matters that much until the next release, and makes it easier to remember to bump to the right next version.

My opinion was more that we never knew which version incrementation (major, minor or patch) the next released version would have so it was easier to do that once we knew what the next release would be judging by the "summary" we've made in the changelog prior to a release. I agree now however that we can be more lenient on just increasing it to a patch, minor or even major base on what changes we do in a PR or even something like that.

mimarz:
I have made an issue for following up your request @vnys #1964

Now we need this PR approved as its blocking for our next release 😅

vnys (approved):
LGTM 👍🏻

PR #1971, opened 2022-02-10
✨ Combobox custom search filter

### LINKED ISSUE #1657: Support abbreviations search in Select components
It would be nice if we could search for the abbreviations/keys of the words. For instance in the picture below, if searching for FMRA it would display the "Fifteen Month Rolling Availabilty" option. 

image

Searching for this would be the same as the one over
image

mimarz:
resolves #1657


oddvernes (approved):
LGTM!

PR #1983, opened 2022-02-15
🐛 Remove window references breaking SSR

### LINKED ISSUE #1970: References to windows crashes on SSR
We have a reference to `window` in `Tooltip` which throws an error when used in a SSR setting (even though its not used until mounted).

We should also make sure we have no other `windows` references and make sure our `setTimeout` and `clearTimeout` are used correctly.

Actions
* Remove window references in `Tooltip` & `Snackbar`

Notes
* https://stackoverflow.com/questions/45802988/typescript-use-correct-version-of-settimeout-node-vs-window/56239226#56239226

oddvernes:
resolves #1970


mimarz (approved):
LGTM 👍

PR #1984, opened 2022-02-15
🐛 tooltip: merge popper.styles and style prop

### LINKED ISSUE #1982: Overriding z-index via style for Tooltip breaks its width
If the `zindex` is overriden using the `style` property it breaks the width of the tooltip. 

Might be some interference with popper or styled-components.

Example of bug here: https://codesandbox.io/s/hidden-shape-pei60?file=/src/App.tsx

oddvernes:
resolves #1982


mimarz (approved):
Had to be a simple bug 🤪 LGTM 👍

PR #1985, opened 2022-02-15
💄 Bump z-indexes inspired by MUI

### LINKED ISSUE #1981: Increase our z-indexs
Some users are still experiencing our components being hidden due to "low" z-index relative to what they are using. 

Suggest updating our values to match that of [mui's z-indexes](https://mui.com/customization/z-index/) as this is what most people are either coming from or using instead. These values should be no high enough to account for anything else people might use.

oddvernes:
resolves #1981

### DIFF
--- a/packages/eds-core-react/src/components/Chip/Chip.tsx
+++ b/packages/eds-core-react/src/components/Chip/Chip.tsx
@@ -40,7 +40,6 @@ const StyledChips = styled.div.attrs<StyleProps>(
   grid-auto-flow: column;
   grid-auto-columns: max-content;
   align-items: center;
-  z-index: 10;
 
   svg {
     fill: ${typography.color};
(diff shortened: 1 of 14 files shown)


oddvernes:
Btw should I also update Sidesheet and Topbar or no?

mimarz (approved):
Team review with @benteljones, @mimarz and @oddvernes approved 👍

mimarz:
> > Ok but if i follow mui (if sidesheet = drawer and topbar = app bar), then sidesheet will have a higher z-index than topbar, where before, topbar was above. Not sure if this can have any consequences if they are used in combination somehow
> 
> Can you give an answer here @equinor/eds-ux ?

After looking at [examples of use-cases ](https://www.figma.com/file/luNobpDIWTuQ9rhyovGLBI/Desktop?node-id=442%3A0)in Figma this is valid behaviour

PR #1987, opened 2022-02-15
✨ Combobox - Select all story

### LINKED ISSUE #961: MultiSelect, select all feature
As [mentioned in an issue in Downshift's repository](https://github.com/downshift-js/downshift/issues/935#issuecomment-738689141)  the `useMultiselect` hook will be improved. 

This will provide a better foundation for adding `Select all` option/action, to solve the issue of an easy overview on selected options and improve accessibility by making it easier to find the selected options.

mimarz:
resolves #961 

Suggest that this should just be an example (instead of built in) as it gives an indication on how to make other option types for quick selections in the combobox.

* Also found some bugs related to controlled that was fixed 🐛


oddvernes:
Would it be possible to change the state of the check all box based on manually checking all the other boxes, so it gets checked when everything is checked and indeterminate if at least one is checked, and empty when none are checked?

mimarz:
@oddvernes The checkboxes are controlled internally, making a possibility for intermediate state for an option externally could lead to other pitfalls or misuse. 

We could however solve this by doing the select all feature internally then...

oddvernes (approved):
In that case this is good enough for an external solution 👍

PR #1996, opened 2022-02-19
🐛 Autocomplete - Fix behaviour for placeholder declaration

### LINKED ISSUE #1651: If placeholder in multiselect is undefined, it should display the standard text
Now it is like this
image

We use this to have it say "all" when there is no items selected

mimarz:
resolves #1651

### DIFF
--- a/packages/eds-lab-react/src/components/Autocomplete/Autocomplete.tsx
+++ b/packages/eds-lab-react/src/components/Autocomplete/Autocomplete.tsx
@@ -220,6 +220,7 @@ function AutocompleteInner<T>(
     optionDisabled = () => false,
     optionsFilter,
     autoWidth,
+    placeholder,
     ...other
   } = props
   const anchorRef = useRef()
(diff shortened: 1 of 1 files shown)


oddvernes:
I guess it could be debated wether the placeholder should always show, or only show if 0 is selected. But then again there is label also where you in most cases you can put the important information. Feels a bit weird though if i put something as placeholder I wouldnt know if i accidentally clicked the clear all button, but if this is the feature requested 🤷

mimarz:
> I guess it could be debated wether the placeholder should always show, or only show if 0 is selected. But then again there is label also where you in most cases you can put the important information. Feels a bit weird though if i put something as placeholder I wouldnt know if i accidentally clicked the clear all button (unless the user makes placeholder text dynamic i suppose), but if this is the feature requested 🤷

You got me thinking. We also have the "meta" where we could have this information. Remains to be tested with screen-readers though... 🤔

oddvernes (approved):
LGTM!

PR #2006, opened 2022-02-22
✨ Icon files as svg sprites with small

### LINKED ISSUE #1885: Svg icons files format
With the addition for `small` & `default` icon sizes in Figma, we need to decide on how to structure the svg files we save in `assets/icons`. 

Suggestions
* Save with size as icon names in a folder name for icon?
   * `fullscreen/`
      * `small.svg`
      * `default.svg` 
* Save double up with an "annotation" in filename?
  * `fullscreen.svg` & `fullscreen_small.svg`
* Save as svg sprite?
   * Still `fullscreen.svg` file but users need to specifiy size with # when using like `fullscreen.svg#small`. How the default fallback here? 🤔

mimarz:
resolves #1885


vnys:
There are a couple of issues with using [the equinor logo](https://eds-static.equinor.com/logo/equinor-logo-primary.svg) as a template 🤔 In that file, the shape is the same – but the user can choose between three different colours (black, red, white) using fragment-identifiers, with black as the fallback in case no identifier is provided. 

While in the case of small and medium icon sizes, the shape is different – and we _don’t_ provide any means of choosing or adding colours. So perhaps we could use a different approach, while still having both shapes in the same file. 

The following suggestion is just from the perspective of using the svg-file as a sprite containing two icons, while we _could_ in fact create one gigantonourmous svg-sprite containing _all_ the icons in both sizes. But lets keep that for another discussion. 

So, if we rewrite the svg file like so: 

```svg
<svg xmlns="http://www.w3.org/2000/svg">
  <view id="medium" viewBox="0 0 24 24" />
  <view id="small" viewBox="0 24 18 18" />
  <path d="M7 10H5V5h5v2H7v3zm-2 4h2v3h3v2H5v-5zm12 3h-3v2h5v-5h-2v3zM14 7V5h5v5h-2V7h-3z" />
  <path d="M5 7H3V3h4v2H5v2zm-2 4h2v2h2v2H3v-4zm10 2h-2v2h4v-4h-2v2zm-2-8V3h4v4h-2V5h-2z" transform="translate(0 24)" />
</svg>
```

…then what we’re essensially doing is stacking the icons vertically, then creating two view elements as windows into the svg file that only show one icon at the time.

Then in order to use the icons in html, we would need some css, which could look something like this:

```css
  .icon.green {
    background:#007079;
  }
  .icon.blue {
    background: #243746;
  }
  .icon.medium {
    width: 24px;
    height: 24px;        
  }
  .icon.small {
    width: 18px;
    height: 18px;        
  }
  icon.medium.fullscreen {
    -webkit-mask-image: url(fullscreen.svg#medium);
    mask-image: url(fullscreen.svg#medium);
  }            
  icon.small.fullscreen {
    -webkit-mask-image: url(fullscreen.svg#small);
    mask-image: url(fullscreen.svg#small);        
  }
```

…and then the html. We can’t use the image element since we can’t use src for this, but we can use a span element, and add the role-attribute so it’s accessible for screen readers.

```html
  <span class="icon fullscreen medium green" role="img" alt="fullscreen"></span>
  <span class="icon fullscreen small blue" role="img" alt="fullscreen"></span>
```

I haven’t tested this with a gazillion different browsers, in fact I’ve only tested it in Chrome – so there may be some nasty surprises, but according to [caniuse.com](https://caniuse.com/css-masks) it should work in all modern browsers by adding the -webkit vendor prefix.

mimarz:
> The following suggestion is just from the perspective of using the svg-file as a sprite containing two icons, while we _could_ in fact create one gigantonourmous svg-sprite containing _all_ the icons in both sizes. But lets keep that for another discussion.

Lets make a separate issue for this. Only downside would be that all the icons are loaded even if your just using one :( 

> ```
> <svg xmlns="http://www.w3.org/2000/svg">
>   <view id="medium" viewBox="0 0 24 24" />
>   <view id="small" viewBox="0 24 18 18" />
>   <path d="M7 10H5V5h5v2H7v3zm-2 4h2v3h3v2H5v-5zm12 3h-3v2h5v-5h-2v3zM14 7V5h5v5h-2V7h-3z" />
>   <path d="M5 7H3V3h4v2H5v2zm-2 4h2v2h2v2H3v-4zm10 2h-2v2h4v-4h-2v2zm-2-8V3h4v4h-2V5h-2z" transform="translate(0 24)" />
> </svg>
> ```
> 
> ```
>   <span class="icon fullscreen medium green" role="img" alt="fullscreen"></span>
>   <span class="icon fullscreen small blue" role="img" alt="fullscreen"></span>
> ```
> 
> I haven’t tested this with a gazillion different browsers, in fact I’ve only tested it in Chrome – so there may be some nasty surprises, but according to [caniuse.com](https://caniuse.com/css-masks) it should work in all modern browsers by adding the -webkit vendor prefix.

I like this approach, but wouldn't this unfortunately be a breaking change for anyone using our icons they way they are now? 

We don't know exactly how people are using the svg icon files but I would guess people use them for basic html files as-well as using them in proprietary software.

vnys:
> Only downside would be that all the icons are loaded even if your just using one :(

Unless we build the sprite dynamically by using a query parameter where you send in a list if icons, similarly to how you use Google Fonts to fetch multiple fonts simultaneously 🤷🏻 Could be implemented as a simple Azure Function. 

> We don't know exactly how people are using the svg icon files but I would guess people use them for basic html files as-well as using them in proprietary software.

That’s true. Perhaps we should offer them as separate icon sets for comfortable and compact in addition to a combined file then, so devs can choose between using our logic or writing their own.

mimarz:
I agree. Having a separate icon files for small would probably be best at this point.

mimarz:
@vnys They the "small" icons are now added next to the existing ones with a `_small` sufix. This is the least intrusive change we can do for now.

vnys (changes requested):
All the files are missing new line at the end of the file, apart from that I agree with the conclusion. 

image

mimarz:
> All the files are missing new line at the end of the file, apart from that I agree with the conclusion.
> 
> image

Good catch! I've added a new issue for this #2076 as it will affect all the icons files (630+ files)

vnys (approved):
LGTM 👍🏻

PR #2012, opened 2022-02-22
✅ Add a11y test for Popover

### LINKED ISSUE #2011: Add a11y test for Popover

laliq:
Resolves #2011


oddvernes (approved):
LGTM! Altough maybe there is no need for individual tests for each subcomponents, just one with all? they are just divs, so not sure what axe even tests for here other than maybe contrast

laliq:
I did it as you suggested @oddvernes.

PR #2014, opened 2022-02-22
✅  Add a11y test for Progress

### LINKED ISSUE #2013: Add a11y test for Progress

laliq:
Resolves #2013


laliq:
Based on the [axe description](https://dequeuniversity.com/rules/axe/4.2/aria-progressbar-name?application=axeAPI) elements with `role="progressbar"` should use `aria-label` attr. to describe the progressbar. 
This error shows only in `LinearProgress`. While adding directly to the element `aria-label` will fix the error, but i think this description should be dynamic, based on the purpose of the progressbar, for expample `aria-label="File upload"` 🤔

oddvernes:
Don't forget to add aria-label to the stories as well 👍

oddvernes (approved):
LGTM!

PR #2016, opened 2022-02-23
💄 Focus ring visibility

### LINKED ISSUE #1733: Make focus-indicators more uniform across components
Both clickbounds and focus-indicators vary greatly across components, as can be seen in the screenshot. Most components have dotted outlines, but not all – and the spacing between the components and the focus-indicators are inconsistent. 

image

oddvernes:
resolves #1733

### DIFF
--- a/packages/eds-core-react/src/components/Button/tokens/icon.ts
+++ b/packages/eds-core-react/src/components/Button/tokens/icon.ts
@@ -22,6 +22,9 @@ const {
     compact__standard: compactClickboundHeight,
   },
   shape,
+  interactions: {
+    focused: { width: focusOutlineWidth },
+  },
 } = tokens
 
 export const primary: ButtonToken = mergeDeepRight(button, {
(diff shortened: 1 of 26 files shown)


packages/eds-core-react/src/components/Button/tokens/icon.ts
```diff
@@ -61,7 +64,7 @@ export const primary: ButtonToken = mergeDeepRight(button, {
     },
     focus: {
       outline: {
-        offset: '0px',
+        offset: `-${parseInt(focusOutlineWidth)}px`,
```
mimarz, line 67: This was the fix for having the focus frame "fit" when used inside table?

oddvernes: Yes the ghost icon button

mimarz: Its just weird if you have it in focus and hover over it, but who am I kidding. Like thats ever going to be a common use-case 🤪

mimarz (approved):
LGTM 👍

PR #2023, opened 2022-02-24
🐛 Adding all variants to quickvariants

### LINKED ISSUE #2022: Variant accordion_header is missing from Typography
**Describe the bug**

image
The initial fault was found in a Dialog, and seems to be reproduced anywhere as far as I can tell.
I did not try to render this inside an actual EDS Accordion, but I assume this is not related.

Use Typography component anywhere and send variant="accordion_header" as prop.

**Expected behavior**
Should output an accordion_header formatted text.

**Specifications**

- Version: 0.18.0
- Browser: Vivaldi 5.1 (Chromium 98)
- OS: Linux

oddvernes:
resolves #2022 

Is this correct though? previously only heading and paragraph variants would work, but all variants would be listed as available with typescript/in storybook. Now I have added all variants as possible which then matches the exposed types

### DIFF
--- a/packages/eds-core-react/src/components/Typography/Typography.tokens.ts
+++ b/packages/eds-core-react/src/components/Typography/Typography.tokens.ts
@@ -2,7 +2,7 @@ import { ComponentToken, tokens } from '@equinor/eds-tokens'
 import type { TypographyTokens } from '@equinor/eds-tokens'
 
 const { typography, colors: colorsToken } = tokens
-const { heading, paragraph } = typography
+const { heading, paragraph, ui, table, input, navigation } = typography
 
 const {
   interactive: {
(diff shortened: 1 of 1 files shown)


mimarz (commented):
The problem with spreading all variants as one is duplicate naming 🤔

oddvernes:
Well I have istead added a new story which gives an overview of variables in each group (excluding heading & paragraph since they dont require group)

packages/eds-core-react/src/components/Typography/Typography.stories.tsx
```diff
@@ -67,6 +67,152 @@ export const paragraphs = (): JSX.Element => (
...
+          </Typography>
+        </List.Item>
+      </List>
+    </List.Item>
+  </List>
+)
+GroupsOverview.parameters = {
+  docs: {
+    storyDescription:
+      'Variants ordered by their required group parameter. Example `<Typography group="input" variant="label"></Typography>`. Variants under `heading` and `paragraph` does not require the group parameter and are therefore not included here.',
```
vnys, line 212:
```suggestion
      'Variants ordered by their required group parameter. Example `<Typography group="input" variant="label"></Typography>`. Variants under `heading` and `paragraph` don’t require the group parameter and are therefore not included here.',
```

vnys (changes requested):
Just a minor typo, looks good apart from that.

vnys (approved):
LGTM 👍🏻

PR #2024, opened 2022-02-24
🐛 Fixes in Menu.Item & Menu.Section

### LINKED ISSUE #2015: Missing null check in MenuList
**Describe the bug**

MenuList crashes when a child is `null`, e.g., from a conditional child where the clause is false (`condition && <Menu.Item>`).

If the conditional child is outside of a `Menu.Section`, `MenuList` crashes on [L55](https://github.com/equinor/design-system/blob/25f768a035973b81906b71aee0d34d95e7b57ebb/packages/eds-core-react/src/components/Menu/MenuList.tsx#L55) with error `TypeError: can't access property "type", child is null`. Otherwise `MenuList` crashes on [L61]() with error `Error: React.cloneElement(...): The argument must be a React element, but you passed null.`. 

1. Add a conditional to the inner JSX of a <Menu> component.

**Expected behavior**

The `<Menu>` renders without crashing.

**Specifications**

- Version: 0.18.0
- Browser: Chrome v98.0.4758.102
- OS: macOS v12.2.1

**Additional context**

[Codesandbox](https://codesandbox.io/s/pedantic-glitter-h2zrmb?file=/src/App.js)

mimarz:
fixes #2015

* Fixed crash in `Menu` when using conditional rendering on `Menu.Item`
* Fixed divider not hiding in `Menu.Section` when first.

### DIFF
--- a/packages/eds-core-react/src/components/Menu/MenuList.tsx
+++ b/packages/eds-core-react/src/components/Menu/MenuList.tsx
@@ -54,6 +54,8 @@ export const MenuList = forwardRef<HTMLDivElement, MenuListProps>(
         ReactChildren.map(children, (child: MenuChild) => {
           if (!child) return child
           if (child.type === MenuSection) {
+            index++
+            const menuSectionIndex = index
             const updatedGrandChildren = ReactChildren.map(
               child.props.children,
               (grandChild: MenuChild) => {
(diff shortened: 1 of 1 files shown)


oddvernes (approved):
LGTM!

PR #2040, opened 2022-02-28
📝 Explain input validation better in Textfield story

### LINKED ISSUE #1972: Input validation does not work correctly on TextField component
The TextField component (in a number context) allows some basic input validation when using the arrows to increase/decrease the current value. However, you are still allowed to enter a random value outside the specified range.

**Side issue**: The component uses the browser tooltip (in the browser local language) and not the EDS tooltip.

|Chrome on Windows 10 (Norwegian)|Chrome on Windows 11 (English)|
|-----------------------|--------------------------|
|image|image|

Is it the responsibility of the frontend app that uses the component to handle this, or should the component itself sanitize the inputs as there is some input validation present? 

(Note that proper use of input sanitation should be done once more in the backend of the application that uses this, but this is outside the scope of this issue.)

Define a Textfield like this:
```
<TextField
    id={someId}
    max={10}
    min={1}
    onChange={(event: ChangeEvent<HTMLInputElement>) => setScore(event.target.valueAsNumber)}
    type="number"
    value={category.score?.toString() ?? ""}
/>
```

**Expected behavior**

When input validation is specified on the component with `min` and `max`, I expect the component to also handle user input and show an error if the value exceeds the range.

Regarding tooltip: I also expect the EDS tooltip to be used and not the browser native tooltip.

**Specifications**

- Version: 0.17.0
- Browser: Chrome
- OS:  Windows 10/11

oddvernes:
Resolves #1972

### DIFF
--- a/packages/eds-core-react/src/components/TextField/TextField.stories.tsx
+++ b/packages/eds-core-react/src/components/TextField/TextField.stories.tsx
@@ -504,7 +504,7 @@ Compact.parameters = {
   },
 }
 
-export const ExampleWithReactHookForm: Story<TextFieldProps> = () => {
+export const ValidationWithReactHookForm: Story<TextFieldProps> = () => {
   const { handleSubmit, control } = useForm({
     defaultValues: { data: '' },
   })
@@ -542,4 +542,12 @@ export const ExampleWithReactHookForm: Story<TextFieldProps> = () => {
       </Button>
     </form>
   )
+}
+ValidationWithReactHookForm.parameters = {
+  docs: {
+    description: {
+      story:
+        'How input validation is handled is up to the user. In this example we use <a href="https://react-hook-form.com/">React Hook Form</a>.',
+    },
+  },
 }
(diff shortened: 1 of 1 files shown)


packages/eds-core-react/src/components/TextField/TextField.stories.tsx
```diff
@@ -543,3 +543,11 @@ export const ExampleWithReactHookForm: Story<TextFieldProps> = () => {
     </form>
   )
 }
+ValidationWithReactHookForm.parameters = {
+  docs: {
+    description: {
+      story:
+        'How input validation is handled is up to the user. In this example we use <a href="https://react-hook-form.com/">React Hook Form</a>.',
```
vnys, line 550:
```suggestion
        'How input validation is handled is up to the developer. In this example we use <a href="https://react-hook-form.com/">React Hook Form</a>.',
```

vnys (changes requested):
Since “users” can mean both the end user and the developer, I would use the “developer” to avoid misunderstandings.

vnys (approved):
LGTM 👍🏻

PR #2062, opened 2022-03-04
🧑‍💻Menu anchor handle keypress internal

### LINKED ISSUE #1936: Menu: make anchor-element keyhandling internal to component
Due to increasing complexity of how to make the anchor button for menu behave properly with keyboard navigation, we should look into the feasability of moving this logic into the component, instead of offloading it to the users.

oddvernes:
Resolves #1936 

Make keyhandling of anchor internal (open and move focus to menu after pressing `enter`/`keyUp`/`keyDown`).
Adds eventlistener to the anchor for the keypresses, so a new requirement for developers now is to use `useState` to set the `anchorEl` ref. Previously in our examples we used useRef, but this will not trigger rerender to add listener when ref switches from null to the actual element

### DIFF
--- a/packages/eds-core-react/src/components/Menu/Menu.context.tsx
+++ b/packages/eds-core-react/src/components/Menu/Menu.context.tsx
@@ -5,19 +5,23 @@ import {
   MouseEvent,
   createContext,
 } from 'react'
+import type { FocusTarget } from './Menu.types'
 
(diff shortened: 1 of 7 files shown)


mimarz (approved):
LGTM 👍 

Just some suggestions on naming in the example code.

PR #2068, opened 2022-03-09
Fixes useOutsideClick type

cbuv:
this is not a bug, not a feature, but typescript type mismatch. 
eslint complains about wrong type


mimarz (approved):
LGTM 👍

PR #2072, opened 2022-03-09
🛑     Add extension Stylelint

### LINKED ISSUE #2071: Install Stylelint

laliq:
Resolves #2071


mimarz (changes requested):
Suggest also adding this extension to `.vscode/extensions.json` so that people not using codespaces also have it :)

.stylelint.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": [
+    "stylelint-config-recommended",
```
vnys, line 3:
```suggestion
```

The standard config [already extends recommended](https://github.com/stylelint/stylelint-config-standard/blob/main/index.js#L4) so we can remove this line.

vnys (changes requested):

mimarz (changes requested):
Stylelint out of the box only works for css or scss files I belive. Since most of our styling uses styled-components, we need to use a processor for the stylelint to work on _that_ styling.

https://styled-components.com/docs/tooling#stylelint

mimarz:
Closing this as its quite outdated by now 👴

PR #2093, opened 2022-03-24
📝  Divider documentation

### LINKED ISSUE #435: Create documentation for Divider
Write documentation for [Divider](https://eds.equinor.com/components/divider/)

**Sub tasks**

- [ ] Write documentation
- [x] Create examples
- [x] Add props-table
- [x] Add live preview

laliq:
Resolves #435


oddvernes (approved):
LGTM 👍 

Annoyingly the 3 dividers at the top have an apparent different thickness even though it's 3 of the same. I guess this is the consequence of having things 1px thick nowadays, similar to the problem with chip. if you add padding to the parent the lines alternates between which looks thin as you bump up and down the padding🤷.

PR #2094, opened 2022-03-24
🐛 Fix menu in scrim/Dialog

### LINKED ISSUE #2092: useOutsideClick breaks in Dialog component
Our hook `useOutsideClick` does not detect clicks when used inside the `Dialog` component. This means that `Menu` and `Popover` does not work as intended, as well as anyone else using this hook for other purposes inside `Dialog`. 

Notes:
* Might have something to do with stacking content as the document event listener for click is not triggered when click inside the `Dialog`
* Provided [codesandbox](https://codesandbox.io/s/staging-pine-rb42xv?file=/src/App.js) by user illustrating bug

oddvernes:
Resolves #2092 

I included a story (menu inside a dialog) in menu stories temporarily if anyone wants to test more, with the intent to remove it again before merging once approved

### DIFF
--- a/packages/eds-core-react/src/components/Menu/Menu.stories.tsx
+++ b/packages/eds-core-react/src/components/Menu/Menu.stories.tsx
@@ -4,6 +4,7 @@ import styled from 'styled-components'
 import {
   Menu,
   MenuProps,
+  Dialog,
   Typography,
   Button,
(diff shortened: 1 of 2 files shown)


mimarz (approved):
LGTM 👍

PR #2100, opened 2022-03-28
♻️ Autocomplete: Change id contraints for items

### LINKED ISSUE #2086: Change option id handling in Autocomplete
User feedback wants more freedom in regards to provided data-model to options. 

Currently we have `label` as being a required property on the provided objects model, as downshift needs a string value for each option to work properly with kb-navigation, a11y and other features.

We can however removed this need and instead make the `optionLabel` required. This will fulfil downshifts need for a option string value and removing the need for users to potentially map their data to fit the `Autocomplete`

mimarz:
resolves #2086 

* Added benefit is that we now support plain string arrays in addition to object arrays.

### DIFF
--- a/packages/eds-lab-react/src/components/Autocomplete/Autocomplete.docs.mdx
+++ b/packages/eds-lab-react/src/components/Autocomplete/Autocomplete.docs.mdx
@@ -29,7 +29,7 @@ const options = [
 />
 ```
 
-When passing in objects as options, make sure you define that `optionLabel` function. This function must return a string.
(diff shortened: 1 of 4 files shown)


oddvernes:
I suppose the documentation should be updated to reflect this change. (Label is no longer required for instance)

mimarz:
> I suppose the documentation should be updated to reflect this change. (Label is no longer required for instance)

yeah. I am debating wether we should have a console.warn if T is of type `object` and `optionLabel` is not provided. This is assuming `optionLabel` is now optional as its not required for string arrays.

oddvernes (approved):
LGTM!

PR #2107, opened 2022-03-30
Fix filename typos

berland:
Noted typos in filenames when browsing through the icon set.

I don't know how this affects downstream usage of EDS.

(`ispell` has been run through all filename, and no other typos was found)


mimarz (approved):
Thats a good bunch of typos 🙈 

These files are however auto-generated from our icon library in Figma, so we need to make sure to fix these typos there as well! #2109 

These changes will be breaking changes downstream for anyone using the static icon files and users of npm package `@equinor/eds-icons`

I've approved your changes nonetheless as we can use them to verify that changes in Figma was done correctly :) 

Thanks a lot for the fixes! 🙏

PR #2125, opened 2022-04-04
⬆️ Upgrade typescript

### LINKED ISSUE #2124: Update typescript
We are having some problems with type errors that are not present in new version of typescript. Since our current typescript version 4.2.4 is from june 2021 we want to update typescript. 

Type error issues #2108 #2118

mimarz:
fixes #2124


oddvernes (approved):
LGTM 👍

PR #2148, opened 2022-04-07
👷‍♂️  Simplified github actions checks & updated minor dependencies on core-react

### LINKED ISSUE #2130: Fix github actions deploy to correct Storybook slot
We used `inputs` under `workflow_dispatch` for deciding which storybook things were published to. We then turned on to deploy to development storybook on `on.push` to develop, but this trigger ignores `inputs` values which resulted in development storybook being deployed to production storybook...

mimarz:
resolves #2130 

* Storybook build is left out of checks for now as they are taking up lots of time already
* Updated patch updates on dependencies for `eds-core-react` package
* Only one `checks` script is run on PR/push now testing all packages as they are all dependant on eachother (instead of separate as earlier)
* Using the same scripts as local for building, testing, linting etc (was just replicated earlier)
* Adde type check scripts to packages and root
* Checks on this PR fail as using re-usable workflows has to target a branch in the repository, which is pointing to develop as is whats correct once merged.


mimarz:
Merging to continue work on in develop updating publish actions

PR #2158, opened 2022-04-13
📝 Tooltip documentation

### LINKED ISSUE #444: Create documentation for Tooltip
Write documentation for [Tooltip](https://eds.equinor.com/components/tooltip/)

**Sub tasks**

- [ ] Write documentation
- [ ] Create examples
- [ ] Add props-table
- [ ] Add live preview

laliq:
Resolves #444

### DIFF
--- a/packages/eds-core-react/src/components/Tooltip/Tooltip.docs.mdx
+++ b/packages/eds-core-react/src/components/Tooltip/Tooltip.docs.mdx
@@ -5,7 +5,7 @@ import { Tooltip } from '.'
 
 Shows when hovered, focused or long-pressed (touch), and display information such as a text label or a short description.
 
-<Links 
+<Links
     ariaUrl="https://www.w3.org/TR/wai-aria-practices/#tooltip"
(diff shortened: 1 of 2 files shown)


packages/eds-core-react/src/components/Tooltip/Tooltip.docs.mdx
```diff
@@ -0,0 +1,62 @@
+import { Links, PropsTable, Story } from './../../../.storybook/components'
+import { Tooltip } from '.'
+
+# Tooltip
+
+Shows when hovered, focused or long-pressed (touch), and display information such as a text label or a short description.
```
oddvernes, line 6: I am not sure if this claim about long-press is true, in our implementation at least? (on my android phone it activates on tap, and i think we need to implement long-press detection with a timeout or something if we want this)

oddvernes (changes requested):
I'm, unsure about the long press claim, but otherwise fine

packages/eds-core-react/src/components/Tooltip/Tooltip.docs.mdx
laliq: I will simplify the description.

oddvernes (approved):

PR #2160, opened 2022-04-19
💄 Search: Override Chrome autofill background

### LINKED ISSUE #2138: Search component AutoComplete blue color
**Describe the bug**

When you use autoComplete on the Search component it has this blue color added that does not look good.

1. Go to https://eds-storybook-react.azurewebsites.net/?path=/docs/inputs-search--default
2. Select an auto-complete value if you have any
3. Looks like the image below

image

**Expected behavior**

To not have this blue color

**Specifications**

- Version: 0.18.0
- Browser: Chrome
- OS: Windows

**Additional context**

Screenshot (19)

oddvernes:
resolves #2138

### DIFF
--- a/packages/eds-core-react/src/components/Search/Search.tsx
+++ b/packages/eds-core-react/src/components/Search/Search.tsx
@@ -117,6 +117,12 @@ const SearchInput = styled(Input)(({ theme, disabled }) => {
     &:focus {
       outline: none;
     }
+    &:-webkit-autofill {
+      box-shadow: 0 0 0px 1000px ${theme.background} inset;
+    }
+    &:autofill {
+      box-shadow: 0 0 0px 1000px ${theme.background} inset;
+    }
     ${disabled &&
     css`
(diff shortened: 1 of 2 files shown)


packages/eds-core-react/src/components/Search/Search.tsx
```diff
@@ -117,6 +117,12 @@ const SearchInput = styled(Input)(({ theme, disabled }) => {
     &:focus {
       outline: none;
     }
+    &:-webkit-autofill {
```
mimarz, line 120: Did it work without the vendor prefix? I thingk `styled-components` has built in autoprefixer 🤔

oddvernes: It does not. The non-prefixed alone does not work, and does not get autoprefixed, so this is more of a future-proofing (css-tricks recommmended including it)

mimarz (approved):
LGTM 👍

PR #2161, opened 2022-04-19
📝 Slider documentation

### LINKED ISSUE #433: Create documentation for Slider
Write documentation for [Slider](https://eds.equinor.com/components/slider/)

**Sub tasks**

- [ ] Write documentation
- [x] Create examples
- [x] Add props-table
- [x] Add live preview

laliq:
Resolves #433

### DIFF
--- a/packages/eds-core-react/src/components/Slider/Slider.stories.tsx
+++ b/packages/eds-core-react/src/components/Slider/Slider.stories.tsx
@@ -21,6 +21,47 @@ export const Introduction: Story<SliderProps> = (args) => {
   )
 }
 
+export const SimpleSlider: Story<SliderProps> = () => (
+  <>
+    <span id="simple-slider">Slide me</span>
+    <Slider value={4} min={0} max={10} ariaLabelledby="simple-slider" />
+  </>
+)
+SimpleSlider.storyName = 'Simple slider'
+SimpleSlider.decorators = [
+  (Story) => (
+    <div style={{ margin: '3rem' }}>
+      <Story />
+    </div>
+  ),
+]
+
+export const SimpleSliderWithSteps: Story<SliderProps> = () => {
+  return (
+    <>
+      <span id="even-simpler-slider">
+        Simple slider, no dots, no min or max values, steps of 10
+      </span>
+      <Slider
+        ariaLabelledby="even-simpler-slider"
+        value={50}
+        step={10}
+        minMaxDots={false}
+        minMaxValues={false}
+      />
+    </>
+  )
+}
+
+SimpleSliderWithSteps.storyName = 'Simple slider with steps'
+SimpleSliderWithSteps.decorators = [
+  (Story) => (
+    <div style={{ margin: '3rem' }}>
+      <Story />
+    </div>
+  ),
+]
+
 export const RangeSlider: Story<SliderProps> = () => {
   const [value, updateValue] = useState([30, 70])
   const changeHandler = (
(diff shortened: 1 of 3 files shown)


packages/eds-core-react/src/components/Slider/Slider.stories.tsx
```diff
@@ -1,29 +1,67 @@
...
   )
 }
 
+export const SimpleSlider: Story<SliderProps> = () => (
+  <>
+    <span id="simple-slider">Slide me</span>
+    <Slider value={4} min={0} max={10} ariaLabelledby="simple-slider" />
+  </>
+)
+SimpleSlider.storyName = 'Simple slider'
```
mimarz, line 30: The derived name from story should be the same as your are overriding here. So this should no longer be needed.

packages/eds-core-react/src/components/Slider/Slider.stories.tsx
```diff
@@ -1,29 +1,67 @@
...
+SimpleSlider.decorators = [
```
mimarz, line 31: Suggest using `<Stack>` component from our Storybook components instead of this.

mimarz (changes requested):

packages/eds-core-react/src/components/Slider/Slider.stories.tsx
laliq: I would have to overwrite display:
`const Stack = styled(SBStack)
  margin: 3rem;
  display: block;
`
Would it be right to do so @mimarz ?

packages/eds-core-react/src/components/Slider/Slider.stories.tsx
laliq: I kept it for the menu purpose. It keeps the consistency with the rest headers in the menu.

mimarz: If I am not mistaken Storybook with translate the story name `SimpleSlider` to `Simple slider` so having the this defined is redundant no?

packages/eds-core-react/src/components/Slider/Slider.stories.tsx
mimarz: Can't you just use `Stack` as it is?

laliq: Will look like that, will change span position:
Screenshot 2022-04-20 at 11 46 29

mimarz: Ah, this can be solved by changing the `flex-direction` to `row`?

packages/eds-core-react/src/components/Slider/Slider.stories.tsx
laliq: With `storyName`:
Screenshot 2022-04-20 at 11 50 38

Without `storyName`: 
Screenshot 2022-04-20 at 11 51 45

What do you prefer?
I kept in all the stories i worked on option with `storyName`, because sometimes there are longer  names: 
Screenshot 2022-04-20 at 11 53 55
Screenshot 2022-04-20 at 11 55 30

mimarz: Hmm, the "Simple slider" looks better I guess. Keep it as is then?  

I see we have a good mix of with and without higher/lower cases in our other stories. We need to change the others that are missing to match the style then :)

laliq: OK, yes i keep in mind the naming on each element i work on.

packages/eds-core-react/src/components/Slider/Slider.stories.tsx
laliq: > I bet there are other use-case we want different direction in our stories.

Yes indeed

laliq: What about the `gap` between the elements? Also elements are align `centre` this will move `span` text into the centre? 🤔

mimarz: The more I look at it the more I feel we need to change `Slider`'s props. I don't quite understand why `ariaLabelledby` is mandatory as using only `aria-label` is also valid a11y...  Reference: https://www.w3.org/TR/wai-aria-practices/#slider_roles_states_props

Ok, so maybe just leave it and then we can change it after we have fixed the properties? I'll make a follow up issue.

laliq: I see, ok i will do so then.

mimarz: Made an issue regarding this and some initial thoughts #2167 😅

oddvernes:
Is everything here resolved now?

laliq:
Wouldn’t be better to wait till #2167 issue is finished?

mimarz (changes requested):

mimarz (approved):
LGTM 👍

PR #2172, opened 2022-04-25
📝  SideSheet documentation

### LINKED ISSUE #448: Create documentation for Side sheet
Write documentation for [Side sheet](https://eds.equinor.com/components/side-sheet/)

**Sub tasks**

- [ ] Write documentation
- [ ] Create examples
- [ ] Add props-table
- [ ] Add live preview

laliq:
Resolves #448

### DIFF
--- a/packages/eds-core-react/src/components/SideSheet/SideSheet.docs.mdx
+++ b/packages/eds-core-react/src/components/SideSheet/SideSheet.docs.mdx
@@ -3,6 +3,18 @@ import { SideSheet } from '.'
 
 # SideSheet
 
+A toggleable area that is anchored to the edge of the viewport with supplementary content.
+
+<Links 
+    figmaUrl="https://www.figma.com/file/0bGXR2sCwMVSDNyyzu5BXrO5/UI%E2%80%94User-Interface?node-id=17790%3A0"
+    documentationUrl="https://eds.equinor.com/0b0c666ab/p/17d96d-side-sheet/b/282ac1"
+    sourceUrl="https://github.com/equinor/design-system/blob/develop/packages/eds-core-react/src/components/SideSheet/SideSheet.tsx"
+    npmUrl="https://www.npmjs.com/package/@equinor/eds-core-react"
+/>
+
+<Story id="surfaces-sidesheet--introduction" />
+<PropsTable />
+
 ## Usage
 
 ## Examples
--- a/packages/eds-core-react/src/components/SideSheet/SideSheet.docs.mdx
+++ b/packages/eds-core-react/src/components/SideSheet/SideSheet.docs.mdx
@@ -17,4 +17,20 @@ A toggleable area that is anchored to the edge of the viewport with supplementar
 
 ## Usage
 
+`SideSheet` is surfaces containing supplementary content.
+
+<ul>
+    <li>Side sheets contain content that supplements the screen’s primary UI region.</li>
+    <li>Side sheets display a wide variety of content and layouts.</li>
+    <li>Side sheets can be visible or hidden based on screen size or user need, and they often become bottom sheets on mobile.</li>
+</ul>
+
+```tsx
+import { SideSheet } from '@equinor/eds-core-react'
+
+<SideSheet>
+  <Child>Children</Child>
+</SideSheet>
+```
+
 ## Examples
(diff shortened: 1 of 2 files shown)


packages/eds-core-react/src/components/SideSheet/SideSheet.docs.mdx
```diff
@@ -0,0 +1,56 @@
...
+    <li>Side sheets contain content that supplements the screen’s primary UI region.</li>
+    <li>Side sheets display a wide variety of content and layouts.</li>
+    <li>Side sheets can be visible or hidden based on screen size or user need, and they often become bottom sheets on mobile.</li>
+</ul>
+
+```tsx
+import { SideSheet } from '@equinor/eds-core-react'
+
+<SideSheet>
+  <Child>Children</Child>
```
mimarz, line 32: `Child` is not a component we have in our libraries so this example might be confusing. Maybe just replace it with a comment something like; `// your content here`

packages/eds-core-react/src/components/SideSheet/SideSheet.stories.tsx
```diff
@@ -43,3 +78,200 @@ export const Default: Story<SideSheetProps> = (args) => {
...
+        onClose={() => setToggle(!toggle)}
+        style={{ overflowY: 'hidden' }}
+      >
+        <div style={{ padding: '6px', overflowY: 'auto', maxHeight: '80%' }}>
+          <Typography variant="h4">Details</Typography>
+          <TextField
+            id="storybook-details"
+            placeholder="Add description"
+            inputIcon={<Icon name="edit" title="Add description" />}
+            style={{ paddingTop: '16px' }}
```
mimarz, line 242: This looks like a typo?

packages/eds-core-react/src/components/SideSheet/SideSheet.stories.tsx
```diff
@@ -43,3 +78,200 @@ export const Default: Story<SideSheetProps> = (args) => {
...
+          ))}
+        </Table.Body>
+      </Table>
+      <SideSheet
+        open={toggle}
+        onClose={() => setToggle(!toggle)}
+        style={{ overflowY: 'hidden' }}
+      >
+        <div style={{ padding: '6px', overflowY: 'auto', maxHeight: '80%' }}>
+          <Typography variant="h4">Details</Typography>
```
mimarz, line 237: Suggest extend our `Stack` component with direction `row` and `left` alignment instead of just div using style attr.

packages/eds-core-react/src/components/SideSheet/SideSheet.stories.tsx
```diff
@@ -43,3 +78,200 @@ export const Default: Story<SideSheetProps> = (args) => {
...
+          </p>
+        </div>
+      </SideSheet>
+    </Container>
+  )
+}
+
+export const Scrolling: Story<SideSheetProps> = () => {
+  const cellValues = toCellValues(data, columns)
+  const [toggle, setToggle] = useState(false)
```
mimarz, line 181: Suggest default state to be open so its clearer to users that this is about side sheet

packages/eds-core-react/src/components/SideSheet/SideSheet.stories.tsx
```diff
@@ -43,3 +78,200 @@ export const Default: Story<SideSheetProps> = (args) => {
     </div>
   )
 }
+
+export const Placement: Story<SideSheetProps> = () => {
+  const cellValues = toCellValues(data, columns)
+  const [toggle, setToggle] = useState(false)
```
mimarz, line 84: Rename to "open" so its the same as in the other examples?
Suggest default state to be open so its clearer to users that this is about side sheet.

mimarz (changes requested):

packages/eds-core-react/src/components/SideSheet/SideSheet.stories.tsx
laliq: `Stack` makes things bit complicated in this layout; i would have to overwrite `align-items` & `flex-direction`. When acctually elements used in the component has proper attrs for spacing.
I try to make `Divider` full-bleed, but i can't find the way to overwrite padding 🤔

mimarz (changes requested):

mimarz (approved):
LGTM 👍

PR #2173, opened 2022-04-25
♿️ Slider a11y improvements

### LINKED ISSUE #2167: Fix Slider a11y properties
We currently have `ariaLabelledby` property which is mandatory on `Slider`. Not only does it differ from native prop name, but there are other valid combinations using `aria` attrs which are valid for `Slider`.

https://www.w3.org/TR/wai-aria-practices/#slider_roles_states_props

Actions
- Make `ariaLabelledby` optional and make note of removing it in future releases as native `aria-labelledby` should be used. 
- Are other a11y combinations with `aria-value/min/max/now` upheld?
- Accessible `aria-valuetext` story?

oddvernes:
resolves #2167 

- Added` aria-valuemin/max/now` (handled internally since these should be same as min/max/value anyway)
- Internally  set `aria-valuetext` with `getFormattedText`, which returns the return of `outputFunction`-prop, or raw value
- use `useId` instead of `ariaLabelledby` to construct internal ids
- Made `ariaLabelledby` optional with a deprecation warning log if used, and added support for native `aria-labelledby`

### DIFF
--- a/packages/eds-core-react/src/components/Slider/Slider.tsx
+++ b/packages/eds-core-react/src/components/Slider/Slider.tsx
@@ -13,7 +13,7 @@ import { slider as tokens } from './Slider.tokens'
 import { MinMax } from './MinMax'
 import { Output } from './Output'
 import { SliderInput } from './SliderInput'
-import { bordersTemplate } from '@equinor/eds-utils'
+import { bordersTemplate, useId } from '@equinor/eds-utils'
 
 const {
   entities: { track, handle, dot },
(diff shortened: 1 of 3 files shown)


packages/eds-core-react/src/components/Slider/Slider.tsx
```diff
@@ -296,9 +297,26 @@ export const Slider = forwardRef<HTMLDivElement, SliderProps>(function Slider(
...
   }
 
-  const inputIdA = `${ariaLabelledby}-thumb-a`
-  const inputIdB = `${ariaLabelledby}-thumb-b`
-  const inputId = `${ariaLabelledby}-thumb`
+  let inputIdA = useId(null, 'inputA')
+  let inputIdB = useId(null, 'inputB')
+  let inputId = useId(null, 'thumb')
+  const overrideId = rest['id'] ? rest['id'] : null
+  if (overrideId) {
```
mimarz, line 304: Maybe just check the `rest.id` here and create const inside "if" if true?

mimarz (approved):
Just a suggestion on slightly cleaner code, else it looks good! 👍

PR #2182, opened 2022-05-03
📝♿️ Progress a11y

### LINKED ISSUE #2020: Aria-label for Progress

oddvernes:
resolves #2020 

I added an invisible `output` element to `linear` and `circular` determinate progress so screenreaders should read out new values if it is inside a container with `aria-live` (I could not get it to read out `aria-valuenow`, I think maybe the element needs to be made focusable and be in focus for that to work). The output value is limited to 25/50/75/100% to limit readouts (its possible setting aria-live to assertive instead of polite also helps with this, its hard to tell with this behaving different in different browsers)

The other issues about aria-describedby/aria-label is already supported by `...rest` going to the element with role="progressbar" so I added stories to storybook showing this

### DIFF
--- a/packages/eds-core-react/src/components/Progress/Circular/Circular.docs.mdx
+++ b/packages/eds-core-react/src/components/Progress/Circular/Circular.docs.mdx
@@ -58,3 +58,9 @@ Determinate circular indicators fill the invisible, circular track with color, a
 ### Inside button
 
 <Story id="feedback-progress-indicators-circular--inside-button" />
+
+### Acessibility
+`Circular progress` have `role="progressbar"` and therefore either `aria-describedby` or `aria-label` should be included. 
+Click the button with keyboard and a screen reader activated to test.
+
+<Story id="feedback-progress-indicators-circular--accessibility" />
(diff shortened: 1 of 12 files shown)


oddvernes:
> When testing the accessibility story for linear and circular the voice over is stuck on "loading 0%" 🤔

Well first of all I am going to rebase on develop, stand by

oddvernes:
Hmm it looks like the mac voiceover reads out a bunch of stuff that is not read on windows. It is not supposed to read loading percentages for indeterminate loader. I need to investigate more on mac. The determinate loader demos on circular and linear works as it is supposed

oddvernes:
Fixed mac issue by not including output element for indeterminate

mimarz:
VoiceOver on Mac does not seem to read out loading states for linear and star accessibility stories... 🤔

mimarz (changes requested):
VoiceOver on Mac does not seem to read out loading states for linear and star accessibility stories... 🤔

oddvernes:
It was due to aria-busy. I removed it and made aria-live assertive as well so the loading readout interrupts the long winded readout of the button context

mimarz (approved):
Works now! ♿

PR #2206, opened 2022-05-16
✅ Fixed test warnings

### LINKED ISSUE #2185: Fix test warnings in packages
We have a few warnings coming from test even though they are green.

```shell
  console.error
    Error: Not implemented: window.computedStyle(elt, pseudoElt)
```
```shell
  console.error
    Warning: An update to ForwardRef(Tooltip) inside a test was not wrapped in act(...).

    When testing, code that causes React state updates should be wrapped into act(...):

    act(() => {
      /* fire events that update state */
    });
    /* assert on the output */

    This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
        at title (/Users/michael.marszalek/Documents/prosjekter/Equinor/code/design-system/packages/eds-core-react/src/components/Tooltip/Tooltip.tsx:105:7)
```

mimarz:
fixes #2185

### DIFF
--- a/packages/eds-core-react/src/components/Tooltip/Tooltip.test.tsx
+++ b/packages/eds-core-react/src/components/Tooltip/Tooltip.test.tsx
@@ -9,6 +9,7 @@ import { Tooltip, Button } from '../../'
 const StyledTooltip = styled(Tooltip)`
   background: red;
 `
+const openDelay = 100
 
 afterEach(cleanup)
 
@@ -23,6 +24,7 @@ describe('Tooltip', () => {
     const content = screen.getByText('Test')
 
     fireEvent.mouseOver(content)
+    await act(() => new Promise((r) => setTimeout(r, openDelay)))
 
     const tooltip = await screen.findByRole('tooltip')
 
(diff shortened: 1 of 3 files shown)


packages/eds-core-react/src/components/Tooltip/Tooltip.test.tsx
```diff
@@ -30,12 +30,12 @@ describe('Tooltip', () => {
   })
   it('Should pass a11y test', async () => {
     const { container } = render(
-      <Tooltip title="Tooltip" enterDelay={0}>
```
oddvernes, line 33: Do we not just go back to testing an empty document here and not the tooltip element? (Not that axe can really check anything useful for tooltip anyway)
bilde

mimarz: Added a 100ms after firevent in all tooltip tests. Validated by using screen debug to log dom elements.

oddvernes (approved):
Nice 👍

PR #2263, opened 2022-06-02
📝 Input documentation

### LINKED ISSUE #2262: Create documentation for Input

martalalik:
Resolves #2262

### DIFF
--- a/packages/eds-core-react/src/components/Input/Input.docs.mdx
+++ b/packages/eds-core-react/src/components/Input/Input.docs.mdx
@@ -26,6 +26,8 @@ import { Input } from '@equinor/eds-core-react'
 
 ### Types
 
+This is the native type of property and here is a small selection of examples of all the possible types.
+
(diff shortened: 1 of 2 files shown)


packages/eds-core-react/src/components/Input/Input.docs.mdx
```diff
@@ -0,0 +1,53 @@
...
+```tsx
+import { Input } from '@equinor/eds-core-react'
+
+<Input />
+```
+
+## Examples
+
+### Types
+
```
oddvernes, line 28: Should it be pointed out that this is the native type prop and here is a small selection of examples of all the possible types?

martalalik: Added

oddvernes: "Native `type` property" (not "native type of property")

PR #2264, opened 2022-06-02
🐛 Popover: check for onClose existence

### LINKED ISSUE #2236: Missmatch in typings in Popover
**Describe the bug**

Popover seems to check if onCLose is null, but type props state that type is either undefined or () => void
This makes it so that line this line: 
https://github.com/equinor/design-system/blob/7c4b75d5eded584c5f371078382c5ba4fd3e883b/packages/eds-core-react/src/components/Popover/Popover.tsx#L133

and  this line 
https://github.com/equinor/design-system/blob/7c4b75d5eded584c5f371078382c5ba4fd3e883b/packages/eds-core-react/src/components/Popover/Popover.tsx#L142

will always return oneclose !== null => true 

You get a error in the logs if you dont provide an onClose

**Expected behavior**
Suggest either opdating the input prop type to support undefined or updating the checks to support undefined

**Specifications**

- Version:
- Browser:
- OS:

**Additional context**

oddvernes:
resolved #2236


mimarz (approved):
LGTM 👍

PR #2267, opened 2022-06-02
📝 Textfield storybook: added datetime examples

### LINKED ISSUE #2198: Example of Textfield as datepicker
Add a story with using `type="date"` to use native datepicker. 

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date

oddvernes:
resolves #2198


mimarz (approved):
LGTM 👍

PR #2273, opened 2022-06-03
💄Card: changed to flex column, removed height, actions pushed down

### LINKED ISSUE #2248: Problems using Card inside grid/flex containers
There are some oddeties in `Card` which makes it hard to use inside flex/grid containers.

* Oddities: fit-content´, `align-items: center`,  `grid-auto-columns: auto;`

oddvernes:
resolves #2248 

Card can now be stretched when in a grid, this was not possible before due to height: fit-content. 
Card is now a flex column, with margin-top: auto on card.actions to push it down in the case where card is stretched. This means having multiple actions will kind of break the layout by evenly spacing the actions though, so users should have only one actions at the bottom (and just make a new styled div for rows inside content)


mimarz (approved):
LGTM 👍

PR #2276, opened 2022-06-07
📝  Paper documentation

### LINKED ISSUE #2266: Create documentation for Paper

martalalik:
Resolves #2266

### DIFF
--- a/packages/eds-core-react/src/components/Paper/Paper.stories.tsx
+++ b/packages/eds-core-react/src/components/Paper/Paper.stories.tsx
@@ -1,5 +1,5 @@
 import styled from 'styled-components'
-import { Paper, PaperProps } from '../..'
+import { Paper, PaperProps, Typography } from '../..'
 import { Story, ComponentMeta } from '@storybook/react'
 import { Stack as SBStack } from './../../../.storybook/components'
 import page from './Paper.docs.mdx'
@@ -23,6 +23,11 @@ const Wrapper = styled(Paper)`
   height: 150px;
   width: 150px;
 `
+const WrapperOverview = styled(Paper)`
+  height: 100px;
+  width: 50px;
+  padding: 10px;
+`
 
 export const Introduction: Story<PaperProps> = (args) => (
   <Stack>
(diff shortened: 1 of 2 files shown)


packages/eds-core-react/src/components/Paper/Paper.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+const Wrapper = styled(Paper)`
+  height: 150px;
+  width: 150px;
+`
+
+export const Introduction: Story<PaperProps> = (args) => (
+  <Stack>
+    <Wrapper {...args}>
+      <Paper {...args} />
```
oddvernes, line 30: Perhaps instead of/or in addition to argstable we make a grid of papers, one for each elevation, and write the name of the elevation in the center of the paper? I feel like that is more user friendly that having to switch with the dropdown

mimarz: We have the ordinal here: https://eds.equinor.com/0b0c666ab/p/568642-elevation/b/790b54

martalalik: Because i agree with both of you, i will make an overview, bear with me...

martalalik: I added `Elevation overview`, it gives nice compare of elevations. So developers don't have to jump between pages for to see only the elevation level.

oddvernes: Nice! Although the shadow values in elevation overview is a bit wrong, each shadow actually consists of 3 stacked shadows, but you only list the first 'part' of each shadow

martalalik: I update the view.

oddvernes: I don't know, this still loks a bit messy. Missing comma separation and color values
The box-shadows looks like this
`box-shadow: 0 11px 15px rgb(0 0 0 / 20%), 0 9px 46px rgb(0 0 0 / 12%), 0 24px 38px rgb(0 0 0 / 14%);`
which is a bit much for these tiny squares. Maybe either not listing out the box-shadow values at all or make the squares bigger and add the values in `<code>` blocks or somethingso it is clear it  is syntactically correct and could be copy pasted

mimarz: Maybe just remove the `box-shadow` values for now? I mean, they are using this component for elevation purpose, if they need the code for some reason they can inspect the component.

martalalik: I took away the values

mimarz (approved):
LGTM 👍

PR #2277, opened 2022-06-07
🐛 Fix textfield compact story

### LINKED ISSUE #2146: Textfield unit text flows over container when compact
**Describe the bug**

In compact mode the unit text expands past the input field instead of the field expanding 

1. See compact story in storybook, add/change the unit prop

**Expected behavior**

Should not overflow that way

oddvernes:
resolves #2146 

Turns out it wasn't a bug, per se, someone set style="200px" on the textfield, but style goes into  ...rest goes into the input and setting width on the input is wrong. It would have to be set on the outermost element wrapping both input and adornments to work. Fixed by making a div around the input in the story and set width on that. 
Also removed double edsprovider


mimarz (approved):
LGTM 👍

PR #2280, opened 2022-06-08
📝 Switch documentation 

### LINKED ISSUE #2279: Create documentation for Switch

martalalik:
Resolves #2279

### DIFF
--- a/packages/eds-core-react/src/components/Switch/Switch.docs.mdx
+++ b/packages/eds-core-react/src/components/Switch/Switch.docs.mdx
@@ -1,0 +1,6 @@
+import { Links, PropsTable, Story } from './../../../.storybook/components'
+import { Switch } from '.'
+
+# Switch
+## Usage
+## Examples
--- a/packages/eds-core-react/src/components/Switch/Switch.docs.mdx
+++ b/packages/eds-core-react/src/components/Switch/Switch.docs.mdx
@@ -41,3 +41,26 @@ import { Switch } from '@equinor/eds-core-react'
 ```
 
 ## Examples
+
+### Default states
+
+<Story id="inputs-selection-controls-switch--default-states" />
+
+### Alternative to label
+
+To comply with accessibility, a `label` is always required on inputs. In some cases though, a visual label is not desirable. <br />
+In such cases `aria-label` or `aria-labelledby` should be used.
+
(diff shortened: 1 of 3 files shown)


packages/eds-core-react/src/components/Switch/Switch.docs.mdx
```diff
@@ -0,0 +1,66 @@
...
+<ul>
+    <li>Toggle a single item on or off.</li>
+    <li>Immediately activate or deactivate something.</li>
+     <li>Switches make it easy to compare available options.</li>
+</ul>
+
+```tsx
+import { Switch } from '@equinor/eds-core-react'
+
+<Switch label="Label text" />
```
oddvernes, line 40: I am missing an example story to clue people in to use `checked` and `onChange` (maybe it should be here in the code example too) since it is not mentioned this is a checkbox and the attributes are not listed since they are native

oddvernes (changes requested):

packages/eds-core-react/src/components/Switch/Switch.docs.mdx
martalalik: I add an example in `DefaultStates` section.

packages/eds-core-react/src/components/Switch/Switch.stories.tsx
```diff
@@ -1,67 +1,65 @@
...
+      </li>
+      <li>
+        <Switch disabled label="You can't turn me on!" />
+      </li>
+      <li>
+        <Switch disabled defaultChecked label="You can't turn me off!" />
+      </li>
+      <li>
+        <Switch
+          label="I'm a controlled component"
```
oddvernes, line 47: Could perhaps the label be updated to say "I'm a controlled label: checked" when true and unchecked when false (or some other visual change)

martalalik: Updated

oddvernes: Getting closer, but you can simplify this a bit more by removing that outside label and do this instead, then you also don't need the extra useState:
`label={\`Output from slider is ${check ? 'checked' : 'unchecked'}\`}`

martalalik: Done

PR #2282, opened 2022-06-09
♻️  Reduce property deprecation warning intensity for Slider

mimarz:
Replaced `console.warn` with `@deprecated` annotation.

image


oddvernes (approved):
LGTM

PR #2296, opened 2022-06-10
♿️ Accordion header actions

### LINKED ISSUE #2091: Accordion Header design pattern
An example Introduced by EDS ➡️  **header with button** :
It is consider  as a bad practice, with the current [Accordion Design Pattern](https://www.w3.org/TR/wai-aria-practices-1.2/#accordion), and [Deque description](https://dequeuniversity.com/rules/axe/4.2/nested-interactive?application=axeAPI): 

> Interactive controls must not have focusable descendants .

Further more, this issue might need more attention because of the design pattern in EDS that is introduced today.

Lastly i would like to point some issues/questions that comes with EDS pattern:

- What is header and what is it's main function?
- What about creative approach  and design that engage users?
- What about accessibility?

oddvernes:
resolves #2091 

Now, buttons or whatever else you want to add to the right side in the header can be put into the `Accordion.HeaderActions` subcomponent. This is placed next to the button toggle element instead of inside it.

### DIFF
--- a/packages/eds-core-react/src/components/Accordion/Accordion.stories.tsx
+++ b/packages/eds-core-react/src/components/Accordion/Accordion.stories.tsx
@@ -111,14 +111,14 @@ export const Header: Story<AccordionProps> = () => {
             <Accordion.HeaderActions>
               <Icon
                 name="warning_outlined"
-                title="Attach file"
+                title="Warning"
                 size={16}
                 color="currentColor"
                 style={{ marginRight: '16px' }}
               />
               <Icon
                 name="error_outlined"
-                title="Notifications"
+                title="Error"
                 size={16}
                 color="currentColor"
               />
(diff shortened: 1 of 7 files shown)


packages/eds-core-react/src/components/Accordion/Accordion.stories.tsx
```diff
@@ -98,19 +108,21 @@ export const Header: Story<AccordionProps> = () => {
...
-            <Icon
-              name="notifications"
-              title="Notifications"
-              size={16}
-              color="currentColor"
-            />
+            <Accordion.HeaderActions>
+              <Icon
+                name="warning_outlined"
+                title="Attach file"
```
mimarz, line 114: The title doesn't seem to match the icon used here?

packages/eds-core-react/src/components/Accordion/Accordion.stories.tsx
```diff
@@ -98,19 +108,21 @@ export const Header: Story<AccordionProps> = () => {
...
+                size={16}
+                color="currentColor"
+                style={{ marginRight: '16px' }}
+              />
+              <Icon
+                name="error_outlined"
+                title="Notifications"
```
mimarz, line 121: The title doesn't seem to match the icon used here?

mimarz (approved):
LGTM 👍

PR #2301, opened 2022-06-13
Fix three issues with the lab DatePicker component

sebastianvitterso:
This PR solves/addresses three issues found in the DatePicker component in eds-lab-react. 

1. It is currently not possible to close the pop-over of the DatePicker by shift-tabbing away. Focus is moved away from the input field, but the pop-over stays. In forms with multiple DatePickers in a row, this might lead to multiple DatePickers being open at the same time. Fixed by using a ref to the 3rd-party-component, and running its `setOpen(false)`-function. Commit: ccacbf2
2. It is currently not possible to update the component's value from the outside of the component, as the component's value is currently initialized from the prop's value, then the prop is never read from again. This PR addresses this by adding a `useEffect` which updates the internal value whenever the external value is updated. Commit: c70a44d.
3. The color of the label and the calendar icon were too dark, compared to how they are designed in [EDS - Date/time picker](https://eds.equinor.com/0b0c666ab/p/07b531-datetime-picker/b/86826a). This is now updated, and passed directly to the component instead of unnecessarily overriding CSS (for the icon). Commits: 7f38785, 661ef2f, 2992a3d.

### DIFF
--- a/packages/eds-lab-react/src/components/DatePicker/DatePicker.tokens.ts
+++ b/packages/eds-lab-react/src/components/DatePicker/DatePicker.tokens.ts
@@ -10,6 +10,9 @@ const {
     navigation: { menu_title, button },
   },
   colors: {
+    text: {
+      static_icons__tertiary: { rgba: iconTertiary },
+    },
     interactive: {
       primary__resting: { rgba: focusColor },
     },
@@ -30,6 +33,7 @@ export interface DatePickerToken extends ComponentToken {
   colors: {
     green13: string
     green100: string
+    iconGray: string
   }
 }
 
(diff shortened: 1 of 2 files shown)


mimarz (changes requested):
Awesomesauce! 👏

We haven't looked much at the `Datepicker` code in a long time as its been a community contribution, but from what I can see it looks mostly good :) 

We use [Github keywords](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) for linking to issues that are fixed/resolved, so could you update the description with the other issue you also fixed?

PR #2327, opened 2022-06-20
🐛  Using Icon data should work properly if size is not defined

### LINKED ISSUE #2326: Icon size default value problems
Default value for `size` causes a bug if you you have defined width/height and not size, as default values for size overrides icon width/height in the component. Regardless.

mimarz:
fixes #2326


oddvernes (approved):
LGTM

PR #2331, opened 2022-06-21
💄 Decorators in stories

### LINKED ISSUE #2253: Normalize on story margins in Storybook
Before officially announcing new Storybook documentation we should normalize on how we apply more "air" to our stories.

We currently use one or the other apporach:

* Override margin/padding on `Stack` compnent
* Use decorator on individual story
* Just a div with style margin/padding

We should standarize on one approach for doing this to make it easier to maintain and provide clear code example for "inspect code" feature in SB. 

Notes:
* Decorators in SB still messes up "inspect code" for props table example but works fine for the other stories

With decorator:
image

Without decorator and some props set using table:
image

martalalik:
Resolves #2253


packages/eds-core-react/src/components/Accordion/Accordion.stories.tsx
```diff
@@ -52,26 +52,28 @@ export default {
...
+        <Accordion.Panel>Content 3</Accordion.Panel>
+      </Accordion.Item>
+    </Accordion>
+  )
+}
 
 export const Header: Story<AccordionProps> = () => {
   return (
-    <Wrapper>
+    <>
```
mimarz, line 76: Just a heads up, in React, if a single element is return you don't need fragments. Fragments is only used when you have multiple elements returned.

mimarz:
Since @martalalik is not here, we want to merge the changes done here as to not obstruct with other work done until she is back.

mimarz (approved):

PR #2335, opened 2022-06-21
🐛 Fixed Autocomplete onOptionsChange when not controlled & cleaned up react hooks form example

### LINKED ISSUE #2333: Autocomplete react form hook example is not working
When submitting form, error for values not being set it thrown.

mimarz:
resolves #2333

- Turns our there was also a small bug when using `onOptionsChange` and `selectedOptions` was NOT set which i fixed.
- Did a big clean up of react forms hook

### DIFF
--- a/packages/eds-core-react/src/components/Autocomplete/Autocomplete.docs.mdx
+++ b/packages/eds-core-react/src/components/Autocomplete/Autocomplete.docs.mdx
@@ -88,6 +88,8 @@ Compact `Autocomplete` using `EdsProvider`.
 <Story id="inputs-autocomplete--compact" />
 
 ### With `React Hook Form`
+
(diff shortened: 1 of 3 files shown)


martalalik (changes requested):
- Tested, and looks good.  👍 
- I guess we don't need `styled` import in story.
- Shall `Stack` be placed in decorator? 🤔

mimarz:
> * Tested, and looks good.  👍
> * I guess we don't need `styled` import in story.
`styled` is being used (for hooks-form example) or else we would have gotten linting warnings :) 
> * Shall `Stack` be placed in decorator? 🤔
I was thinking for doing this in a separate PR as not to have to many unrelated changes now

martalalik (approved):
LGTM

PR #2349, opened 2022-06-24
✨ Added `contained_icon` variant to Button

### LINKED ISSUE #2340: Add new contained_icon button in react
EDS design team has implemented the new variant of button Ghost Icon Contained:
https://www.figma.com/file/0bGXR2sCwMVSDNyyzu5BXrO5/UI%E2%80%94User-Interface?node-id=17394%3A47297
This variant needs to be implemented in React and Storybook

mimarz:
resolves #2340 

- Updated docs to reflect new `Button` icon variant
- Updated docs to streamline `a11y` to use `aria-label` for icon buttons instead of title on icon

The size for icon buttons ghost & contained are to big (48px vs 40px), this is due to wrong value in shape token. This will be fixed from Figma and new tokens will need to be exported to fix this.

### DIFF
--- a/packages/eds-core-react/src/components/Button/Button.docs.mdx
+++ b/packages/eds-core-react/src/components/Button/Button.docs.mdx
@@ -35,8 +35,9 @@ The `Button` come with four variants; `contained`, `outlined`, `ghost` and `ghos
 
 ### Icon button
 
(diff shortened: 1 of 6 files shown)


oddvernes (approved):
LGTM

Issue #1887, opened 2022-01-12
Update dependencies in figma-broker


mimarz:
Update all dependencies. 

svgo has a deprecation warning: 
![image](https://user-images.githubusercontent.com/1070981/149149758-f6f6a173-0f25-466c-92f7-52a3a809becf.png)

BirteThornquist:
Related: #2771

oddvernes:
SVGO has been updated by #2771 , the remaining dependencies are

![Image](https://user-images.githubusercontent.com/4499531/226330656-79435026-6660-4215-bffc-f3baf489baa0.png)

which I am slightly worried about updating if it breaks something

oddvernes:
This was has been covered by multiple pr's since, the latest being #3121

PR #2353, opened 2022-06-27
✨ Exported new icons

### LINKED ISSUE #2356: `wind_turbine_farm` icon not exporting correctly in code
After different attempt is seems that this icon is not exporting the right way in code:

Need to investigate why Figma is not drawing it properly

mimarz:
🚧 WIP - Pushing changes for investigation and highlight errors

There is a problem with icon export because `collapse` and `badge` is all black and `wind_turbine_farm` has some artifacts.

- **Removed**. Investigate possible fix #2356 - ~~`wind_turbine_farm` might be fixed by updating svgo #1887~~
- **Fixed** - ~~`collapse` & `badge` original svg from Figma has the same error (black box) so this might be something on Figma's end.~~

image

image


oddvernes (approved):
LGTM

PR #2378, opened 2022-07-08
feat(autocomplete): add disable clear button prop

### LINKED ISSUE #2383: autocomplete: add disable clear button prop

In some situations it would be preferable to hide the clear button, e.g. when a field has a default and can never be null.

a simple hideClearButton prop that is added onto the "showClearButton" computed bool.

**Additional context**
Documenting a pr already submitted by @Simrayz

Simrayz:
In some situations it would be preferable to hide the clear button, e.g. when a field has a default and can never be null.
This PR adds a simple hideClearButton prop that is added onto the "showClearButton" computed bool.

Resolves #2383


mimarz (approved):
LGTM 👍

PR #2381, opened 2022-07-08
♻️Replace useCombinedRefs with mergeRefs util

### LINKED ISSUE #2379: Replace useCombinedRefs
It turns out our useCombinedRefs implementation does not play well with the callback refs from floating-UI, due to our usage of useCallback inside.

Other projects that use floating-UI all use a mergeRef function that is pretty much the same as ours but without useCallback, and instead memoize the output in the components.

oddvernes:
resolves #2379


mimarz (approved):
LGTM 👍

PR #2395, opened 2022-08-04
✨ New internal component: `InputWrapper`

### LINKED ISSUE #2336: New "InputWrapper" component
Undecided on name, but I suggest we need a "inputWrapper" or "Field" like component which handles adornments, helpertext, label, meta and such for encompassing `Input`.

We currently have `Search`, `Autocomplete`, `TextField`, which all have different implementations of how an input is rendered along with its helpertext (or not), adornments, labels etc.

This makes it harder to maintain, ensure consistency and proper `a11y`. 

We also have several pending issues which could easily be solved by standardising with a `InputWrapper`;
 - #2294  
 - #1660 
 - #1209
 - #1003 (maybe)

Goals
- Investigate what features this component should have
  - `Textfield` has a similar component internally which might be extracted and used
- Check wether this can be used for `Radio`, `Checkbox` & `Switch`
- Make it easy to use our "form" components in forms

Notes
- Similar components in other libraries
  - Mantines `InputWrapper`: https://mantine.dev/core/input-wrapper/
  - Mui's `FormControl`: https://mui.com/material-ui/api/form-control/

mimarz:
resolves #2336 

Working on creating and converting this in a separate branch as to not pollute `develop` if re-writes need to be done.


oddvernes:
moving `height` to the wrapping div has caused this issue on textfield
bilde

a minor visual bug in windows, the scollbar is slightly overlapped by rightAdornments. Not sure we can do much about this
bilde

mimarz:
> moving `height` to the wrapping div has caused this issue on textfield bilde

Fixed the height so its the same as in production now. We still have some adjustments that will need to be done for textarea once its finished in EDS 2.0

> a minor visual bug in windows, the scollbar is slightly overlapped by rightAdornments. Not sure we can do much about this bilde

Didn't floating-ui some some kind of solution for this? If we could detect scrollbar and adjust padding accordingly, but again might be a very edge-case.

oddvernes:
> Didn't floating-ui some some kind of solution for this? If we could detect scrollbar and adjust padding accordingly, but again might be a very edge-case.

The way I did it on vlfk.no to detect scrollbar-width when open search was: 
```
var totalWidth = window.innerWidth
var widthMinusScrollBar = document.body.clientWidth
var scrollBarWidth = totalWidth - widthMinusScrollBar
```
Havent checked if there is a similar way for scrollbar in a box. Feels a bit overkill though

mimarz:
> > Didn't floating-ui some some kind of solution for this? If we could detect scrollbar and adjust padding accordingly, but again might be a very edge-case.
> 
> The way I did it on vlfk.no to detect scrollbar-width when open search was:
> 
> ```
> var totalWidth = window.innerWidth
> var widthMinusScrollBar = document.body.clientWidth
> var scrollBarWidth = totalWidth - widthMinusScrollBar
> ```
> 
> Havent checked if there is a similar way for scrollbar in a box. Feels a bit overkill though

hmm, yeah, maybe save it for if/when we need it. We already have quite a few big changes.

oddvernes (approved):
LGTM!

mimarz:
On second thought, should we maybe call the props on `Input`, `prefix` and `sufix` instead of `leftAdornments` and `rightAdornments`? @oddvernes

oddvernes:
> On second thought, should we maybe call the props on `Input`, `prefix` and `sufix` instead of `leftAdornments` and `rightAdornments`

I don't know i think pre/suffix is more specifically letters added to words while this is more anything added before/after the input. I see Mantine calls it `rightSection` and MUI calls it `startAdornment`, `endAdornment`. In cirquit though, its `renderPrefix`, `renderSuffix` 🤷‍♂️

PR #2399, opened 2022-08-09
✨ Add width to SideSheet

### LINKED ISSUE #2398: Add width to Sidesheet
After feedback we have decided to introduce a `width` property to `Sidesheet` so that people can define their own width if the variants dont fit the need.

`width` takes precedence over variants if defined.

martalalik:
Resolves #2398

### DIFF
--- a/packages/eds-core-react/src/components/SideSheet/SideSheet.tsx
+++ b/packages/eds-core-react/src/components/SideSheet/SideSheet.tsx
@@ -22,7 +22,7 @@ export type SideSheetProps = {
   onClose?: (Event) => void
   /** Open / close Side Sheet */
   open?: boolean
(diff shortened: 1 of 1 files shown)


packages/eds-core-react/src/components/SideSheet/SideSheet.tsx
```diff
@@ -22,6 +22,8 @@ export type SideSheetProps = {
   onClose?: (Event) => void
   /** Open / close Side Sheet */
   open?: boolean
+  /** Manualy controls width of Side Sheet */
```
mimarz, line 25:
```suggestion
  /** Override width of Side Sheet */
```

mimarz (approved):
Looks good, just a small suggestion on wording to be precise that it overrides width.

PR #2405, opened 2022-08-09
♻️ menu: replace popper with floating-UI

### LINKED ISSUE #2343: Refactor Menu to floating-ui
`Popperjs` (along with `react-popper`) has evolved to [floating-ui](https://floating-ui.com/) and we need to replace our `usePopper` with equivalent in `@floating-ui`

Actions
- Have a look at [examples in floating-ui ](https://floating-ui.com/docs/react-dom-interactions#examples) documentation for our respective component 
- Replace any of our `useIsMounted`, `useOutsideClick` and `useHideBodyScroll` hooks with `floating-ui` middleware.

oddvernes:
resolves #2343 

further improvements can be done by letting floating-ui handle keyboard navigation and add support for nested menus with `FloatingTree` but this is a larger undertaking to be handled by separate tickets 

Like with popover, it is unclear to me why the menu is hidden with visibility, but i am leaving it be

### DIFF
--- a/packages/eds-core-react/src/components/Menu/Menu.test.tsx
+++ b/packages/eds-core-react/src/components/Menu/Menu.test.tsx
@@ -73,15 +73,16 @@ describe('Menu', () => {
     })
   })
   it('Can extend the css for the component', async () => {
-    const { container } = render(
-      <StyledMenu open>
+    render(
+      <StyledMenu open data-testid="menu">
         <div>some random content</div>
       </StyledMenu>,
     )
+    const container = screen.getByTestId('menu')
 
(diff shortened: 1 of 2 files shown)


mimarz (approved):
I think you are the floating-ui expert, it looks fine to me and works like before :)

mimarz:
> Like with popover, it is unclear to me why the menu is hidden with visibility, but i am leaving it be

I think this was to avoid render on open or popper needed it anchor it correctly to the anchor. Might also be leftover before portal for when Menu was inline in the dom 🤔

PR #2410, opened 2022-08-11
✨Overrideable component spike

### LINKED ISSUE #2338: Polymorphic types for our "casted" components
Investigate and see if we can find a generic type we can use for our "casted" components in typescript.

Examples being a `Button` being casted to an `a` element, we want the type to dynamically change to show the type for `a` elements.

Copy of text from #1736 

> Based on the testing @oddvernes did in #1897, maybe we should add the polymorphic type in utils for use in other needed components. 
> 
> Based on [this reddit thread](https://www.reddit.com/r/reactjs/comments/tjwiak/what_is_the_best_way_of_writing_a_polymorphic/), I feel like this implementation was clean:
> 
> * Types: https://github.com/biswarup35/material-you/blob/v0.1.x/packages/material-meeu/src/types/PolymorphicWithRef.d.ts
> 
> * Usage in component: https://github.com/biswarup35/material-you/blob/v0.1.x/packages/material-meeu/src/Button/Button.tsx

Another example: https://github.com/mantinedev/mantine/blob/master/src/mantine-styles/src/theme/types/Polymorphic.ts

oddvernes:
resolves #2338 (perhaps)

This solution does seem to work with button, although you need to set `as="a"` explicitly, as just setting `href`, which converts it to an `a` internally, is not picked up by typescript so adding `target` attribute gets an error in that case.

This pr also fixes a bug i came across where height of button is 36px when as=button, 38px when other things due to different default box-sizing. So i added explicit box-sizing: border-box to force always 36px

### DIFF
--- a/packages/eds-core-react/src/components/Button/Button.docs.mdx
+++ b/packages/eds-core-react/src/components/Button/Button.docs.mdx
@@ -75,3 +75,15 @@ Example All button combinations.
 
 Compact `Button` using `EdsProvider`.
 <Story id="inputs-button--compact" />
+
+### Using the 'as' prop to render as another component
+
+The `as` prop allows you to render the button as another html element or even react component. 
+This is useful if you want to use the button as as a navigation component for a third party routing library such as rect-router.
+
+```tsx
+import { Link } from "react-router-dom"
+import { Button } from '@equinor/eds-core-react'
+
+<Button as={Link} to="/about">About</Button>
+```
--- a/packages/eds-core-react/src/components/Button/Button.docs.mdx
+++ b/packages/eds-core-react/src/components/Button/Button.docs.mdx
@@ -79,7 +79,11 @@ Compact `Button` using `EdsProvider`.
 ### Using the 'as' prop to render as another component
 
 The `as` prop allows you to render the `Button` as another html element or even react component. 
-This is useful if you want to use the `Button` as as a navigation component for a third party routing library such as react-router.
+
+```tsx
+<Button as="a" href="/about" target="_blank" rel="noreferrer">About</Button>
+```
+This is especially useful if you want to use the `Button` as as a navigation component for a third party routing library such as react-router.
 
 ```tsx
 import { Link } from "react-router-dom"
--- a/packages/eds-core-react/src/components/Button/Button.docs.mdx
+++ b/packages/eds-core-react/src/components/Button/Button.docs.mdx
@@ -78,7 +78,7 @@ Compact `Button` using `EdsProvider`.
 
 ### Using the 'as' prop to render as another component
 
-The `as` prop allows you to render the button as another html element or even react component. 
+The `as` prop allows you to render the `Button` as another html element or even react component. 
 This is useful if you want to use the `Button` as as a navigation component for a third party routing library such as react-router.
 
 ```tsx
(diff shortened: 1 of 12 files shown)


oddvernes:
Tested downstream on a create-react-app with `react-router`, doing this 
`<Button as={Link} to="/about">About</Button>`
Now works, whereas before you got typescript error, failed to compile

mimarz:
> Tested downstream on a create-react-app with `react-router`, doing this `<Button as={Link} to="/about">About</Button>` Now works, whereas before you got typescript error, failed to compile

Awesome! Did you mean `as="a"` in this example?

mimarz (changes requested):
Wicked! Suggest we add a story and test for testing that the element is casted? 

Should we still call the casting prop `as` or something else? The name was primarily used because its what styled-components use 🤔

oddvernes:
> Awesome! Did you mean `as="a"` in this example?

No, react router has its own `Link` component with its own attributes like `to`. It gets transformed into a normal `a` with `href` in the html output. But the point is typescript recognizes this and anything else you throw at it now

mimarz:
> > Awesome! Did you mean `as="a"` in this example?
> 
> No, react router has its own `Link` component with its own attributes like `to`. It gets transformed into a normal `a` with `href` in the html output. But the point is typescript recognizes this and anything else you throw at it now

høh, smart!

packages/eds-core-react/src/components/Button/Button.docs.mdx
```diff
@@ -75,3 +75,15 @@ Example All button combinations.
 
 Compact `Button` using `EdsProvider`.
 <Story id="inputs-button--compact" />
+
+### Using the 'as' prop to render as another component
+
+The `as` prop allows you to render the button as another html element or even react component. 
+This is useful if you want to use the button as as a navigation component for a third party routing library such as rect-router.
```
mimarz, line 82:
```suggestion
This is useful if you want to use the `Button` as as a navigation component for a third party routing library such as react-router.
```

packages/eds-core-react/src/components/Button/Button.docs.mdx
```diff
@@ -75,3 +75,15 @@ Example All button combinations.
...
+
+```tsx
+import { Link } from "react-router-dom"
+import { Button } from '@equinor/eds-core-react'
+
+<Button as={Link} to="/about">About</Button>
```
mimarz, line 88: Suggest adding a example with element also, `<Button as="a" to="/about">About</Button>`

packages/eds-core-react/src/components/Button/Button.docs.mdx
```diff
@@ -75,3 +75,15 @@ Example All button combinations.
 
 Compact `Button` using `EdsProvider`.
 <Story id="inputs-button--compact" />
+
+### Using the 'as' prop to render as another component
+
+The `as` prop allows you to render the button as another html element or even react component.
```
mimarz, line 81:
```suggestion
The `as` prop allows you to render the `Button` as another html element or even react component. 
```

mimarz (changes requested):

packages/eds-core-react/src/components/Button/Button.docs.mdx
oddvernes: This would not be possible ( 'to' does not exist on 'a') but i can add an example of 'a' with 'href' and 'target' and 'rel'

oddvernes: I expanded the story a little with a normal 'a'

mimarz: Nice!

mimarz (approved):
LGTM 👍

PR #2420, opened 2022-08-16
📝 Add ButtonGroup to Storybook

### LINKED ISSUE #2419: Button Group add to Storybook
More description in #394

martalalik:
Resolves #2419


martalalik:
I would still consider to create a separate story for `ButtonGroup`  🤔

packages/eds-core-react/src/components/Button/Button.docs.mdx
```diff
@@ -76,18 +92,39 @@ Example All button combinations.
 Compact `Button` using `EdsProvider`.
 <Story id="inputs-button--compact" />
 
-### Using the 'as' prop to render as another component
+## Gutton Group
```
oddvernes, line 95: Gutton Group

martalalik: 🙈

martalalik: done

packages/eds-core-react/src/components/Button/Button.docs.mdx
```diff
@@ -76,18 +92,39 @@ Example All button combinations.
...
+<Links 
+figmaUrl="https://www.figma.com/file/0bGXR2sCwMVSDNyyzu5BXrO5/UI%E2%80%94User-Interface?node-id=17394%3A48481" 
+documentationUrl="https://eds.equinor.com/0b0c666ab/p/22cd7f-toggle-button/b/86826a"
+sourceUrl="https://github.com/equinor/design-system/blob/develop/packages/eds-core-react/src/components/Group/Group.tsx"
+npmUrl="https://www.npmjs.com/package/@equinor/eds-core-react/"
+/>
+
+### Usage
+
+The buttons can be grouped by wrapping them with the `ButtonGroup` component.
```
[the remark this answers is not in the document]
martalalik: `Group` component is just a wrapper, that can be use for `ToggleButton`, and probably that is why i called it group. But yes story is for `ButtonGroup`.

martalalik:
Can someone review `Split Button Group`? I need some feedback, thanks.

martalalik:
To fix position of `menu button` added:
style={{ position: 'absolute', height: '36px' }}

martalalik:
Should `Group` include 

- [role="group"](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/group_role)
- `aria-label?:`

oddvernes:
> To fix position of `menu button` added: style={{ position: 'absolute', height: '36px' }}

Instead add `display: flex` to group and use same default button type for both buttons, but in the story, overwrite padding on the dropdown button to make it narrower and overwrite gap to 1px on group

martalalik:
Do we want `SplitButton` in `Compact` story?

oddvernes:
> Do we want `SplitButton` in `Compact` story?

Make sure it works at least, but i don't think there will be an issue we only override left/right padding to something smaller so it'll just get less tall

martalalik:
> Make sure it works at least, but i don't think there will be an issue we only override left/right padding to something smaller so it'll just get less tall

It works.

martalalik:
My proposition for naming is:

- if keeping `Group` & `Toggle` name i suggest to create folder `Buttons` in Storybook and put inside `Button`, `Group`, `Toggle`.
- if changing name to `ButtonGroup` & `ToggleButton` no need to create folder.

oddvernes:
> * `ToggleButton`

I would rename `Group` to `ButtonGroup`. And that is all for this pr. `ToggleButton` is the upcoming component we havent made yet which is going to be something along the lines of [ToggleButton](https://mui.com/material-ui/react-button-group/) from MUI / [SegmentedControl](https://mantine.dev/core/segmented-control/) from mantine / [SelectorGroup](https://circuit.sumup.com/?path=/docs/forms-selector-selectorgroup--base) from circuit

mimarz:
Should we maybe expose `Group` via a proxy component as `Button.Group` to keep inline with the style of our other components? 


Do we have other usages for this kind of grouping? Maybe `Chip` or `Avatar`? 

image

oddvernes:
> Should we maybe expose `Group` via a proxy component as `Button.Group` to keep inline with the style of our other components?
> 
> Do we have other usages for this kind of grouping? Maybe `Chip` or `Avatar`?
> 

But our pattern is Component.SubComponent whereas here it would be Child.Parent ? But I see what you mean, certain components could be wrapped in Component.Group if there is more

mimarz:
> > Should we maybe expose `Group` via a proxy component as `Button.Group` to keep inline with the style of our other components?
> > Do we have other usages for this kind of grouping? Maybe `Chip` or `Avatar`?
> 
> But our pattern is Component.SubComponent whereas here it would be Child.Parent ? But I see what you mean, certain components could be wrapped in Component.Group if there is more

Yeah, naming is tricky 🙈 We could also name it `Buttons` or just leave it as `Group` and have a prop for toggling of rounding borders and gap instead. Many ways to do this. 

I saw Mantine are using a combination for Component.SubComponent and Component.Parent, so its more of a way to show that some things are related to eachother 🤔

martalalik:
I was to quick 🙈

oddvernes:
If we one day decide to (or are asked) to make layout helper components like what mantine have, we should reserve very general names like "Group" for that i feel. But currently this is out of scope for eds as i understand (its too trivial and projects can quickly make them themselves anyway, it would make us responsible for everyone getting their layout correctly etc). ButtonGroup meanwhile only does one thing: its a wrapper that modifies the styling of buttons, nothing more, nothing less 🤷‍♂️

packages/eds-core-react/src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -0,0 +1,68 @@
...
+    @media (hover: hover) and (pointer: fine) {
+      &:hover {
+        border-radius: 0;
+      }
+    }
+  }
+  ${({ vertical }) =>
+    vertical
+      ? css`
+          display: grid;
```
oddvernes, line 26: Can we change it to always be display flex, and change `flex-direction` from` row` to` column` when verical?

oddvernes (changes requested):

packages/eds-core-react/src/components/ButtonGroup/ButtonGroup.tsx
martalalik: done

oddvernes: I tested some more without the "stack" wrapper, and realized it needs to be display: inline-flex actually (mui does the same) otherwise it will stretch 🤪

martalalik: done

oddvernes (approved):
LGTM!

mimarz:
After discussing on a call 24 august 2020, we decided the following:
- `Group` is renamed to `ButtonGroup`
- `ButtonGroup` is moved in under `Button` and exposed as a compound component (`Button.Group`)
- `ButtonGroup` stories are moved to `Button.docs.mdx` with the following stories
  - Split button
  - Orientation 
  - Existing compact story is updated with `ButtonGroup` example

martalalik:
Can anyone review it?

packages/eds-core-react/src/components/Button/Button.stories.tsx
```diff
@@ -308,6 +310,12 @@ export const Compact: Story<ButtonProps> = () => {
       <Button variant="ghost_icon">
         <Icon data={menu} title="Ghost icon menu"></Icon>
       </Button>
+      <Button.Group ria-label="button group compact">
```
oddvernes, line 313: missing a in aria

martalalik: done

oddvernes: Ok you misunderstood slightly; what i mean is above here is this (ria label):  <Button.Group ria-label="button group compact">
I don't think you need aria label on the buttons

martalalik: Oh i see a typo! 
Every button with icon should have aria-label. [Here](https://medium.com/100-days-in-kyoto-to-create-a-web-app-with-google/day-16-icon-buttons-should-be-labelled-with-aria-label-1056515d3707)

oddvernes: Oh true 😄

oddvernes (approved):
LGTM

PR #2422, opened 2022-08-17
✨♿️Sidebar link improvements

### LINKED ISSUE #2261: Sidebar: make Link component usable
A11y: Focus state design and make Link focusable (currently "a")

How is this component supposed to work? Probably should work with react routers which use Link attribute?

oddvernes:
Resolves #2261

Made links focusable and added focus outline

Made link overridable component so it can be used with routers


mimarz (approved):
LGTM 👍

PR #2424, opened 2022-08-17
 📝  Add custom label to selection controls

### LINKED ISSUE #2231: Example of using custom label with Selection controls
Feedback from a project was that EDS only supports label on right side of the selection controls. Maybe make a short story for our selection controls where we use `Label` or `Typography` with text on left side and in sidesheet or something similar 🤔

Maybe better description that is just a styled `input` with `type=radio/checkbox` 

Suggested [example](https://codesandbox.io/s/serene-golick-wfr72c?file=/src/App.tsx)

martalalik:
Resolve #2231

### DIFF
--- a/packages/eds-core-react/src/components/Radio/Radio.stories.tsx
+++ b/packages/eds-core-react/src/components/Radio/Radio.stories.tsx
@@ -1,6 +1,15 @@
-import { useState, useEffect } from 'react'
-import { Radio, RadioProps, Table, EdsProvider, Density, Label } from '../..'
+import { useState, useEffect, ChangeEvent } from 'react'
+import {
+  Radio,
+  RadioProps,
+  Table,
+  EdsProvider,
+  Density,
+  Label,
+  Typography,
+} from '../..'
 import styled from 'styled-components'
+import { tokens } from '@equinor/eds-tokens'
 import { ComponentMeta, Story } from '@storybook/react'
 import { data } from '../../stories/data'
 import { Stack } from './../../../.storybook/components'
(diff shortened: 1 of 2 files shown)


packages/eds-core-react/src/components/Radio/Radio.stories.tsx
```diff
@@ -116,6 +124,28 @@ export const AlternativeToLabel: Story<RadioProps> = () => (
 )
 AlternativeToLabel.storyName = 'Alternative to label'
 
+export const CustomLabel: Story<RadioProps> = () => {
+  return (
+    <div style={{ display: 'contents' }}>
```
oddvernes, line 129: I don't think this is the correct usage of display: contents and the end result is a bit weird witht the distance between label and radio. How about display flex , align-items center.

oddvernes (changes requested):

packages/eds-core-react/src/components/Radio/Radio.stories.tsx
martalalik: Pure provocation with `contents`, but yes you right.

oddvernes:
I would perhaps use the example linked in the parent issue, since that was the use case someone was asking about

packages/eds-core-react/src/components/Radio/Radio.stories.tsx
martalalik: done

packages/eds-core-react/src/components/Radio/Radio.stories.tsx
```diff
@@ -116,6 +124,28 @@ export const AlternativeToLabel: Story<RadioProps> = () => (
...
+      <Label label="Label" />
+      <Radio name="radio" />
```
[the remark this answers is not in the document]
martalalik: done

oddvernes (approved):
LGTM

PR #2432, opened 2022-08-18
✨ Make typography an overridableComponent

### LINKED ISSUE #2426: Update `Typography` with OverridableComponent type
Update `Typography` to use our `OverridableComponent` type to get the correct typings when its casted to a different html element using the `as` property. 

Look to `Button` for example of how `OverridableComponent` was implemented/used.

oddvernes:
Resolves #2426

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -9,7 +9,7 @@
     "docker-compose*": "yaml"
   },
   "editor.defaultFormatter": "esbenp.prettier-vscode",
-  "editor.formatOnSave": true,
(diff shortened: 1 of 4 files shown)


oddvernes:
Tested downstream with react-router. Also now you can do stuff like `<Typography as="time" dateTime="2022-08-19 20:00">2022-08-19 20:00</Typography>` and typescript won't complain about  dateTime prop, which was the case before

mimarz (approved):
LGTM 👍

PR #2436, opened 2022-08-19
🗑️ Card: removed cursor logic for onClick

### LINKED ISSUE #2435: Remove `onClick` from `Card`
Remove `onClick` from `Card` and its `pointer` changing.

Keep it simple.

oddvernes:
resolves #2435


mimarz (approved):
We haven't actually flagged this as deprecated and removed it in a later realease, but I hope we'll be fine as I don't think people are using this feature...

PR #2442, opened 2022-08-25
✨OverridableComponent breadcrumb

### LINKED ISSUE #2433: Update Breadcrumbs with OverridableComponent type
Mainly thinking of third part routers being supported

oddvernes:
resolves #2433

### DIFF
--- a/packages/eds-core-react/src/components/Breadcrumbs/Breadcrumb.tsx
+++ b/packages/eds-core-react/src/components/Breadcrumbs/Breadcrumb.tsx
@@ -1,4 +1,4 @@
-import { forwardRef, ElementType } from 'react'
+import { forwardRef, ElementType, useMemo } from 'react'
 import styled, { css } from 'styled-components'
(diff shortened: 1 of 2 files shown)


packages/eds-core-react/src/components/Breadcrumbs/Breadcrumb.tsx
```diff
@@ -24,30 +25,47 @@ const StyledTypography = styled(Typography)<StyledProps>`
...
     const props = {
       ...other,
       href,
       ref,
       maxWidth,
     }
+
     const showTooltip = maxWidth > 0
     const isHrefDefined = href !== undefined
+    let forwardedAs: ElementType
```
mimarz, line 56: Could this be rewritten as a ternarery operator? forwaredAs could be initiated with `as` value and then ternarery for if `isHrefDefined` or not. Maybe wrap it in a useCallback also?

mimarz (changes requested):

mimarz (approved):
LGTM 👍

PR #2446, opened 2022-08-29
✨ (feat) Add `createPolymorphicComponent`

### LINKED ISSUE #2443: Add Polymorphic component types

mimarz:
resolves #2443


mimarz:
Withdrawing this as using styled-components and extending styles causes problems for this approach. Changing our components to adapt is to big of a re-write at this moment

PR #2452, opened 2022-08-29
📸🐛Created missing snapshots

### LINKED ISSUE #2449: tests: Fix missing snapshots
Some tests have `expect(asFragment).toMatchSnapshot()` but the correct syntax is `expect(asFragment()).toMatchSnapshot()` otherwise the snapshot is empty

a quick search is looks like at least these component tests should be fixed:

Table
TableofContent
Tabs
TexField
Typography

oddvernes:
resolves #2449


mimarz (approved):
LGTM

PR #2458, opened 2022-09-01
✨ Create ToggleButton

### LINKED ISSUE #2455: ToggleButton: create new component

martalalik:
Resolves #2455

### DIFF
--- a/packages/eds-core-react/src/components/Button/Button.docs.mdx
+++ b/packages/eds-core-react/src/components/Button/Button.docs.mdx
@@ -101,7 +101,7 @@ The buttons can be grouped by wrapping them with the `Button.Group` component.
 ```tsx
 import { Button } from '@equinor/eds-core-react'
 
-<Button.Group aria-label="button group">
+<Button.Group aria-label="primary">
   <Button>Click me!</Button>
   <Button>Click me!</Button>
   <Button>Click me!</Button>
@@ -142,22 +142,27 @@ Is used to group related **options**.
 The `Button.Toggle`
 <ul>
   <li>emphasizes groups of related <code>Toggle</code> buttons,</li>
-  <li>controls the selected state of its child buttons when given its own `selected` prop.</li>
+  <li>controls the selected state of its child buttons when given its own <code>selected</code> prop.</li>
 </ul>
 
 ```tsx
 import { Button } from '@equinor/eds-core-react'
-import { save, edit, copy } from '@equinor/eds-icons'
+import { save } from '@equinor/eds-icons'
 
-<Button.Toggle aria-label="toggle button group">
+const [selectedButton, setSelectedButton] = useState([])
+const handleChange = (indexes: number[]) => {
+  setSelectedButton(indexes)
+}
+
+<Button.Toggle aria-label="actions" selected={selectedButton} onChange={handleChange}>
   <Button aria-label="save action">
     <Icon data={save} title="Ghost icon save"></Icon>
   </Button>
-  <Button aria-label="edit action">
-    <Icon data={edit} title="Ghost icon edit"></Icon>
+  <Button aria-label="save action">
+    <Icon data={save} title="Ghost icon save"></Icon>
   </Button>
-  <Button aria-label="copy action">
-    <Icon data={copy} title="Ghost icon copy"></Icon>
+  <Button aria-label="save action">
+    <Icon data={save} title="Ghost icon save"></Icon>
   </Button>
 </Button.Toggle>
 ```
@@ -172,5 +177,26 @@ import { save, edit, copy } from '@equinor/eds-icons'
 
 ## Examples
 
-### Toggle Button
-<Story id="inputs-button--toggle-button" />
+### Exclusive selection
+
+With exclusive selection, selecting one option deselects any other.
+
+<Story id="inputs-button--toggle-exclusive-selection" />
+
+### Multiple
+
+Multiple selection allows for logically-grouped options, to have multiple options selected, by using the `multiple` prop.
+
+<Story id="inputs-button--toggle-multiple" />
+
+### Enforced
+
+To enforce at least one button to be active, you can adapt your `handleChange` function.
+
+<Story id="inputs-button--toggle-enforced" />
+
+### Vertical
+
+The `Button.Toggle` can be displayed vertically using the `vertical` prop.
+
+<Story id="inputs-button--toggle-vertical" />
--- a/packages/eds-core-react/src/components/Button/Button.docs.mdx
+++ b/packages/eds-core-react/src/components/Button/Button.docs.mdx
@@ -134,5 +134,43 @@ The dropdown can change the button action (as in this example) or be used to imm
 
 <Story id="inputs-button--group-split" />
 
+## Button.Toggle
+Is used to group related **options**.
+
+### Usage
+
+The `Button.Toggle`
+<ul>
+  <li>emphasizes groups of related <code>Toggle</code> buttons,</li>
+  <li>controls the selected state of its child buttons when given its own `selected` prop.</li>
+</ul>
+
+```tsx
+import { Button } from '@equinor/eds-core-react'
+import { save, edit, copy } from '@equinor/eds-icons'
+
+<Button.Toggle aria-label="toggle button group">
+  <Button aria-label="save action">
+    <Icon data={save} title="Ghost icon save"></Icon>
+  </Button>
+  <Button aria-label="edit action">
+    <Icon data={edit} title="Ghost icon edit"></Icon>
+  </Button>
+  <Button aria-label="copy action">
+    <Icon data={copy} title="Ghost icon copy"></Icon>
+  </Button>
+</Button.Toggle>
+```
+
+### Accessibility
+
+#### ARIA
+<ul>
+  <li>Button.Toggle sets <code>aria-pressed="bool"</code> according to the button state.</li>
+  <li>You should provide an accessible label with <code>aria-label="label"</code>, <code>aria-labelledby="id"</code> or <code>&lt;label&gt;</code>.</li>
+</ul>
+
+## Examples
+
 ### Toggle Button
 <Story id="inputs-button--toggle-button" />
--- a/packages/eds-core-react/src/components/Button/Button.docs.mdx
+++ b/packages/eds-core-react/src/components/Button/Button.docs.mdx
@@ -133,3 +133,6 @@ The `Button.Group` can be displayed vertically using the `vertical` prop.
 The dropdown can change the button action (as in this example) or be used to immediately trigger a related action.
 
 <Story id="inputs-button--group-split" />
+
+### Toggle Button
+<Story id="inputs-button--toggle-button" />
--- a/packages/eds-core-react/src/components/Button/Button.docs.mdx
+++ b/packages/eds-core-react/src/components/Button/Button.docs.mdx
@@ -183,7 +183,7 @@ With exclusive selection, selecting one option deselects any other.
 
 <Story id="inputs-button--toggle-exclusive-selection" />
 
-### Multiple
+### Multiple selection
 
 Multiple selection allows for logically-grouped options, to have multiple options selected, by using the `multiple` prop.
 
--- a/packages/eds-core-react/src/components/Button/Button.docs.mdx
+++ b/packages/eds-core-react/src/components/Button/Button.docs.mdx
@@ -149,12 +149,7 @@ The `Button.Toggle`
 import { Button } from '@equinor/eds-core-react'
 import { save } from '@equinor/eds-icons'
 
-const [selectedButton, setSelectedButton] = useState([])
-const handleChange = (indexes: number[]) => {
-  setSelectedButton(indexes)
-}
-
-<Button.Toggle aria-label="actions" selected={selectedButton} onChange={handleChange}>
+<Button.Toggle aria-label="actions">
   <Button aria-label="save action">
     <Icon data={save} title="Ghost icon save"></Icon>
   </Button>
@@ -179,7 +174,7 @@ const handleChange = (indexes: number[]) => {
 
 ### Exclusive selection
 
-With exclusive selection, selecting one option deselects any other.
+Exclusive selection is a default sate, which means that by, selecting one option deselects any other.
 
 <Story id="inputs-button--toggle-exclusive-selection" />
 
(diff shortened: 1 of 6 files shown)


.vscode/settings.json
```diff
@@ -11,5 +11,8 @@
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.formatOnSave": true,
   "prettier.prettierPath": "./node_modules/prettier",
-  "typescript.tsdk": "node_modules/typescript/lib"
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "[typescriptreact]": {
+    "editor.defaultFormatter": "ms-vsliveshare.vsliveshare"
+  }
```
mimarz, line 17: Suggest reverting this change as it was probably added by the extension for some reason.

packages/eds-core-react/src/components/Button/Button.docs.mdx
```diff
@@ -101,7 +101,7 @@ The buttons can be grouped by wrapping them with the `Button.Group` component.
 ```tsx
 import { Button } from '@equinor/eds-core-react'
 
-<Button.Group aria-label="button group">
+<Button.Group aria-label="primary">
```
mimarz, line 104: Was this supposed to be changed? Doesn't make much sense... 🤔

packages/eds-core-react/src/components/Button/Button.docs.mdx
```diff
@@ -132,4 +132,71 @@ The `Button.Group` can be displayed vertically using the `vertical` prop.
...
+
+```tsx
+import { Button } from '@equinor/eds-core-react'
+import { save } from '@equinor/eds-icons'
+
+const [selectedButton, setSelectedButton] = useState([])
+const handleChange = (indexes: number[]) => {
+  setSelectedButton(indexes)
+}
+
```
mimarz, line 156: Suggest removing the "selected" and "onChange" in this example as they are not needed for this component to work. They are meant for more advance features if our users need it.

packages/eds-core-react/src/components/Button/Button.stories.tsx
```diff
@@ -310,7 +325,7 @@ export const Compact: Story<ButtonProps> = () => {
       <Button variant="ghost_icon" aria-label="menu action">
         <Icon data={menu} title="Ghost icon menu"></Icon>
       </Button>
-      <Button.Group aria-label="button group compact">
+      <Button.Group aria-label="compact">
```
mimarz, line 328: Maybe a more descriptive label? "compact button group" ?

packages/eds-core-react/src/components/Button/Button.stories.tsx
```diff
@@ -422,3 +437,139 @@ GroupSplit.decorators = [
     </Stack>
   ),
 ]
+
+export const ToggleExclusiveSelection: Story<ToggleButtonProps> = () => {
+  const [selectedButton, setSelectedButton] = useState([])
+  const handleChange = (indexes: number[]) => {
+    setSelectedButton(indexes)
+  }
```
mimarz, line 445: Suggest removing bloated change handler & useState

packages/eds-core-react/src/components/Button/Button.stories.tsx
```diff
@@ -422,3 +437,139 @@ GroupSplit.decorators = [
...
+    </Stack>
+  ),
+]
+
+export const ToggleMultiple: Story<ToggleButtonProps> = () => {
+  const [selectedButtons, setSelectedButtons] = useState([])
+  const handleChange = (indexes: number[]) => {
+    setSelectedButtons(indexes)
+  }
+
```
mimarz, line 479: Suggest removing bloated change handler & useState

packages/eds-core-react/src/components/Button/Button.docs.mdx
```diff
@@ -132,4 +132,71 @@ The `Button.Group` can be displayed vertically using the `vertical` prop.
...
+<ul>
+  <li>Button.Toggle sets <code>aria-pressed="bool"</code> according to the button state.</li>
+  <li>You should provide an accessible label with <code>aria-label="label"</code>, <code>aria-labelledby="id"</code> or <code>&lt;label&gt;</code>.</li>
+</ul>
+
+## Examples
+
+### Exclusive selection
+
+With exclusive selection, selecting one option deselects any other.
```
mimarz, line 182: Maybe specify this is the default state? 🤔

packages/eds-core-react/src/components/Button/Button.docs.mdx
```diff
@@ -132,4 +132,71 @@ The `Button.Group` can be displayed vertically using the `vertical` prop.
...
+
+<Story id="inputs-button--toggle-exclusive-selection" />
+
+### Multiple
```
mimarz, line 186:
```suggestion
### Multiple selection
```

packages/eds-core-react/src/components/Button/Button.stories.tsx
```diff
@@ -422,3 +437,139 @@ GroupSplit.decorators = [
...
+      <Button aria-label="calendar">
+        <Icon data={calendar} title="Ghost icon calendar"></Icon>
+      </Button>
+      <Button aria-label="time">
+        <Icon data={time} title="Ghost icon time"></Icon>
+      </Button>
+      <Button aria-label="alarm">
+        <Icon data={alarm} title="Ghost icon alarm"></Icon>
+      </Button>
+      <Button aria-label="timer" disabled>
```
mimarz, line 496:
```suggestion
      <Button aria-label="timer">
```

packages/eds-core-react/src/components/Button/ToggleButton/ToggleButton.tsx
```diff
@@ -0,0 +1,69 @@
...
+  useState,
+  useEffect,
+} from 'react'
+import { Button, ButtonProps, ButtonGroupProps } from '..'
+
+export type ToggleButtonProps = {
+  /** Multiple */
+  multiple?: boolean
+  /** Array of selected indexses. */
+  selected?: number[]
```
mimarz, line 16: Should we maybe call this `selectedIndexes` instead? What do you think @oddvernes ?

mimarz (changes requested):
Good work on the documentation! I would suggest some changes on how its structured as there is maybe a bit to much in the Button docs. We can split them up in sub-directories as you suggested, but lets save that for another issue afterwards.

I suggest removing "Enforced" and "Vertical" and replace them with a "controlled" story, which explains the use of `onChange` and `selected` properties. [Controlled vs uncontrolled](https://reactjs.org/docs/uncontrolled-components.html) is a common pattern for react components and it's what these properties are meant for.

packages/eds-core-react/src/components/Button/Button.docs.mdx
martalalik: `ButtonGroup` has `role=group`, which is read by VoiceOver. When adding `aria-label` with a word `group` it will read 2 times group. The VoiceOver is overloaded with unnecessary words. 
So i wanted to create a good pattern for naming `groups`. Conclusion was to get rid  of `group` word from `aria-label` because we have readable `grope` from `role`.

martalalik:
I'm concerned about component's name. Is it a `ToggleButton` or `ToggleButtonGroup`? 🤔

packages/eds-core-react/src/components/Button/Button.docs.mdx
martalalik: Isn't the `Exclusive selection` default state?

packages/eds-core-react/src/components/Button/Button.stories.tsx
martalalik: Same in previous comment ⬆️

packages/eds-core-react/src/components/Button/Button.docs.mdx
mimarz: Aha! I didn't test it with voiceover so that explains it. Maybe something like you did earlier with "example actions" or something like that then? Just thinking how a user might interpret this example and what kind of description `aria-label` should have

mimarz:
> I'm concerned about component's name. Is it a `ToggleButton` or `ToggleButtonGroup`? 🤔

Suggest just keeping it `ToggleButton`, its only internally and descriptive enough for us to understand.

packages/eds-core-react/src/components/Button/Button.docs.mdx
martalalik: In `ButtonGroup` docs describing a11y:

> Button.Group has role="group". 

Maybe adding additional description that this role is read by voiceover? 🤔  
And further more in: 

> You should provide an accessible label with aria-label="label", aria-labelledby="id" or <label>.

Advice to test with voiceover while naming the group 🤔

martalalik: In `ButtonGroup` i changed to something like that:
Screenshot 2022-09-21 at 12 10 23

martalalik:
> Is it ok that we send in regular unmoderated buttons? 

Separate issue?

mimarz:
> I would perhaps make it a separate story page like mui since it is its own thing, has its own props that should be in a propstable, and button story is already gigantic. I was confused trying to even find this.
> 

Agreed. We made a separate issue for this #2509

> Is it ok that we send in regular unmoderated buttons? Technically, I can break it by adding "illegal" (in this context) props such as `href`, while other props like` variant` won't do anything, obviously, because it is hijacked by the component. But on the other hand what developer would be so silly to do that. (I guess alternative would be the `<ToggleButtonGroup><ToggleButton>` where togglebutton is just a button with some props stipped away from the type)
> 

With us trying to be more open and flexible I suggest we allow unmoderated buttons, simpler to use and cleaner without bloating with to many sub-components which are just a shell of some other component. 

@oddvernes You do however highlight one use-case we didn't handle, and thats checking [isValidElement](https://reactjs.org/docs/react-api.html#isvalidelement) on `child` before cloning as well as calling `child.props.onClick` if its present. We need to fix this in this branch @martalalik

packages/eds-core-react/src/components/Button/Button.docs.mdx
```diff
@@ -112,7 +112,7 @@ import { Button } from '@equinor/eds-core-react'
 
 #### ARIA
 <ul>
-  <li>Button.Group has <code>role="group"</code>.</li>
+  <li>Button.Group has <code>role="group"</code>. The word <code>group</code> will be red by assistive technologies.</li>
```
oddvernes, line 115: Typo, should be: "Will be read"

packages/eds-core-react/src/components/Button/Button.docs.mdx
```diff
@@ -132,4 +132,60 @@ The `Button.Group` can be displayed vertically using the `vertical` prop.
...
+
+### Multiple selection
+
+Multiple selection allows for logically-grouped options, to have multiple options selected, by using the `multiple` prop.
+
+<Story id="inputs-button--toggle-multiple" />
+
+### Controlled
+
+To enforce at least one button to be active, you can adapt your `handleChange` function with `selectedIndexes` prop.
```
oddvernes, line 189: I think the text from the deleted "enfored" section has ended up as the text for the "controlled" section, but that has not been implemented here

packages/eds-core-react/src/components/Button/Button.stories.tsx
```diff
@@ -422,3 +437,86 @@ GroupSplit.decorators = [
     </Stack>
   ),
 ]
+
+export const ToggleExclusiveSelection: Story<ToggleButtonProps> = () => {
```
oddvernes, line 441: Still seems a bit weird that the story starts off with nothing selected, when the rule is "one should always be selected"

oddvernes (changes requested):

packages/eds-core-react/src/components/Button/ToggleButton/ToggleButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+              updatedSelection = pickedIndexes.includes(index)
+                ? pickedIndexes.filter((i) => i !== index)
+                : [...pickedIndexes, index]
+            }
+            setPickedIndexes(updatedSelection)
+            if (onChange) {
+              onChange(updatedSelection)
+            }
+          },
+        }
```
mimarz, line 64: All of this is related to overriding the child props with `cloneElement` which we only do if the child is a valid element. Suggest moving it inside the if statement as there is no need for this code to run if we aren't calling cloneElement?

mimarz (changes requested):

packages/eds-core-react/src/components/Button/Button.docs.mdx
martalalik: is this not enough as a dummy story?

packages/eds-core-react/src/components/Button/Button.docs.mdx
```diff
@@ -132,4 +132,60 @@ The `Button.Group` can be displayed vertically using the `vertical` prop.
...
+
+#### ARIA
+<ul>
+  <li>Button.Toggle sets <code>aria-pressed="bool"</code> according to the button state.</li>
+  <li>You should provide an accessible label with <code>aria-label="label"</code>, <code>aria-labelledby="id"</code> or <code>&lt;label&gt;</code>.</li>
+</ul>
+
+## Examples
+
+### Exclusive selection
```
mimarz, line 175: Maybe call it "single selection" instead of "exclusive selection" so the word is similar to our other components such as `Autocomplete`& `NativeSelect`? 🤔

packages/eds-core-react/src/components/Button/Button.stories.tsx
```diff
@@ -422,3 +437,95 @@ GroupSplit.decorators = [
...
+  const [selectedButtons, setSelectedButtons] = useState([0])
+  const handleChange = (indexes: number[]) => {
+    setSelectedButtons(indexes)
+  }
```
mimarz, line 445: Maybe remove this as its not really needed for this example.

mimarz (changes requested):

packages/eds-core-react/src/components/Button/Button.docs.mdx
martalalik: Definitely, especially when we have already `Multiple selection`.

mimarz (approved):
LGTM 👍 

Do you want to do #2509 next? Then we could get out a new release pretty fast!

oddvernes (approved):
LGTM

PR #2465, opened 2022-09-05
🔖Release eds-core-react@0.21.0

### LINKED ISSUE #2464: Release next version of eds-core-react

oddvernes:
resolves #2464

### DIFF
--- a/packages/eds-core-react/CHANGELOG.md
+++ b/packages/eds-core-react/CHANGELOG.md
@@ -8,25 +8,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [0.21.0] - 2022-09-05
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


oddvernes:
> Did we decide to deprecate `useCombinedRefs` also, or was there still some use for it? 🤔

From what I rememeber I asked if we should keep it and then you said get rid of it to avoid confusion 😅 
We at least don't need it, I replaced all usage with this

mimarz:
> > Did we decide to deprecate `useCombinedRefs` also, or was there still some use for it? 🤔
> 
> From what I rememeber I asked if we should keep it and then you said get rid of it to avoid confusion 😅 We at least don't need it, I replaced all usage with this

Aye, suggest marking it also as deprecated then :D

mimarz (approved):
LGTM 👍

I tested using Edge in Citrix and everything seemed to work fine :) 

Just remember to update the dates and removed `useCombinedRefs` before publishing! 👌

PR #2471, opened 2022-09-07
♿️ Focustrapping in Dialog with floating-ui

### LINKED ISSUE #2328: Convert dialog to native Dialog
For a11y and focus trapping, it might be sensible to use the native dialog element, which became evergreen back in march. Also the `inert` prop, albeit it is not supported in firefox  yet

oddvernes:
resolves #2328 

Instead of native dialog, I opted to use floating-ui's FloatingFocusManager to handle focus trapping and returning focus


mimarz (approved):
LGTM!
Maaan, floating-ui has it all 😍

PR #2491, opened 2022-09-15
📝 Draggable sidesheet story

### LINKED ISSUE #2490: Sidesheet: draggable example in storybook
Someone asked on slack if we supported draggable resize on sidesheet. We don't, but I said it should be possible to implement from the outside. I then tried to come up with something in the storybook and it worked quite well, so we might as well add it as another story for sidesheet?

oddvernes:
resolves #2490

### DIFF
--- a/packages/eds-core-react/src/components/SideSheet/SideSheet.docs.mdx
+++ b/packages/eds-core-react/src/components/SideSheet/SideSheet.docs.mdx
@@ -23,7 +23,7 @@ A toggleable area that is anchored to the edge of the viewport with supplementar
 <ul>
     <li>Side sheets contain content that supplements the screen’s primary UI region.</li>
     <li>Side sheets display a wide variety of content and layouts.</li>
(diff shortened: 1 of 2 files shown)


mimarz:
Neat feature! We have also had questions about something similar for `Table` columns, so this might ball up into a utils function or maybe use a third party lib. 

Has this been tested by @equinor/eds-ux ? This is a feature we want to promote? 

From a  a11y point of view, I felt like maybe the the "drag anchor" could have had a bigger clickbound for easier interaction?

oddvernes:
> Has this been tested by @equinor/eds-ux ? This is a feature we want to promote?
> 
> From a a11y point of view, I felt like maybe the the "drag anchor" could have had a bigger clickbound for easier interaction?

@lucasveil is going to take a look and we talked about it on friday standup also.
The drag anchor size is similar to slack and vs code, but yes of course those are mildly annoying too. It could be resized while keeping it visually similar but i suspect too big could be problematic also, you don't want to trigger it accidentally.

mimarz (approved):

PR #2494, opened 2022-09-16
🐛 hide closed menu with display: none

### LINKED ISSUE #2493: Menu: visibility hidden causes overflow scroll on body
**Describe the bug**

For some reason menu is only hidden with css visibility when closed. This causes horisontal scroll overflow on body if you resize the viewport, as floating-ui does not update positions when menu is closed.  We should change it to display none, or perhaps just conditionally render it

1. Open storybook on menu
2. Resize the window narrower until you start seeing horisontal scrollbar

**Expected behavior**
elements in floating-ui portals should not cause overflow

**Specifications**

- Version:
- Browser:
- OS:

**Additional context**

oddvernes:
resolves #2493 
Is this solution ok? I don't know why it was hidden with visibility. I'm guessing perhaps a concern for lag with conditional rendering large menus? Otherwise we can do that also instead of display none

Edit: fixed the same issue with popover although here overflow happens on the container element since popover is not portalized

### DIFF
--- a/packages/eds-core-react/src/components/Menu/Menu.tsx
+++ b/packages/eds-core-react/src/components/Menu/Menu.tsx
@@ -41,7 +41,7 @@ const MenuPaper = styled(Paper)<MenuPaperProps>`
   width: fit-content;
   min-width: fit-content;
   ${bordersTemplate(border)};
-  ${({ open }) => css({ visibility: open ? null : 'hidden' })}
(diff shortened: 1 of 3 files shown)


mimarz (approved):
I think it was done like this to avoid lag with conditional render and when menu was inline instead of portal.

A lot has changed since then, and it looks good still 👍

PR #2498, opened 2022-09-19
💚 Fix build errors node was higher than v16.x

### LINKED ISSUE #2376: Support node 18
Storybook is crashing whenever we run node higher than 16.

mimarz:
resolves #2376 

Turns out `rollup-plugin-size-snapshot` had dependencies to outdated webpack 4x which caused errors. This is just a fix so it works now. Added new issue to update remaining packages with new rollup config using babel for transpiling all code, #2499


oddvernes (approved):
LGTM

PR #2500, opened 2022-09-20
🐛Tabs: scroll with buttons story improvements

### LINKED ISSUE #2495: Overflow with next/prev buttons - disabling of the next button
**Describe the bug**

Using the EDS Tabs component, **Overflow with Next/Prev Buttons**, the next button does not become disabled when you reach the end of scrolling. It also does not take into account viewport resize.

_Link to storybook: https://eds-storybook-react.azurewebsites.net/?path=/docs/navigation-tabs_

1. Go to EDS Storybook
2. Go to Navigation -> Tabs -> **Overflow with next/previous buttons**
3. Click on the next button / right arrow until you reach the end
4. Check if the next button / right arrow becomes disabled

**Expected behavior**

* When you scroll to the end, the next button should be disabled (become greyed out)
Skjermbilde 2022-09-19 kl  09 59 57

**Specifications**

- Using browser Google Chrome on macOS

**Additional context**

**The screen recording shows that the next button does not become disabled when you reach the end of scrolling:**

https://user-images.githubusercontent.com/61127933/190972279-4e215fcc-5e85-4b21-aae4-d2a0d64dc757.mov

oddvernes:
resolves #2495 

fixes "next" button not disabling (it was a rounding error) and added code so it still works when resizing

### DIFF
--- a/packages/eds-core-react/src/components/Tabs/Tabs.stories.tsx
+++ b/packages/eds-core-react/src/components/Tabs/Tabs.stories.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect, useRef, useCallback, useLayoutEffect } from 'react'
+import { useState, useEffect, useRef, useCallback, useMemo } from 'react'
 import styled from 'styled-components'
 import {
   Tabs,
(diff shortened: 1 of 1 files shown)


mimarz (approved):
Lots of code for a feature in a story, but looks like it works! 👍

PR #2503, opened 2022-09-20
🐛autocomplete: same zindex as tooltip

### LINKED ISSUE #2502: `Autocomplete` dropdown when inside `Popover` is underneath
[from comment](https://github.com/equinor/design-system/issues/2501#issuecomment-1252121377) in #2501 

When `Autocomplete` is used inside our `Popover` the dropdown menu is displayed underneath the popover.

https://user-images.githubusercontent.com/24507952/191228539-5f5aadd0-e718-4964-951a-0e34b1e5717a.mov

Notes:
- This might be due to stacking content and how we use portals?

oddvernes:
resolves #2502 

Also tested menu inside popover but it works fine (they have same z-index)

### DIFF
--- a/packages/eds-core-react/src/components/Autocomplete/Autocomplete.tsx
+++ b/packages/eds-core-react/src/components/Autocomplete/Autocomplete.tsx
@@ -542,7 +542,7 @@ function AutocompleteInner<T>(
           position: strategy,
           top: y ?? 0,
           left: x ?? 0,
-          zIndex: 1500,
+          zIndex: 1400,
         },
       })}
     >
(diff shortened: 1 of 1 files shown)


packages/eds-core-react/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -542,6 +542,7 @@ function AutocompleteInner<T>(
           position: strategy,
           top: y ?? 0,
           left: x ?? 0,
+          zIndex: 1500,
```
mimarz, line 545: If it worked with `1400` should we not use that then? Autocomplete dropdown and Menu should have the same z-index

oddvernes: yes good point

mimarz (approved):
When tested with for example an `Autocomplete` inside Autocomplete flickers when closed as is in a portal. 

We created a separate issue to remedy this behaviour here #2505

PR #2507, opened 2022-09-21
🐛Popover: fix missing arrow 

### LINKED ISSUE #2506: Popover: Fix missing arrow

oddvernes:
resolves #2506
Also removed some redundant css


mimarz (approved):
LGTM 👍

Maybe change the title a little bit to "Popover: fix missing arrow" or something like that? Since we'll just use the auto-generate for release log and it uses PR title

PR #2508, opened 2022-09-21
✨ Add sticky prop to Topbar

### LINKED ISSUE #2504: Topbar: Conform sticky toggling
Suggest using a `sticky` property, same as with `Table.Head`.

Maybe set this to default `true` to avoid breaking changes or denoted in release notes as this will be changed.

martalalik:
Resolves #2504

### DIFF
--- a/packages/eds-core-react/src/components/TopBar/TopBar.tsx
+++ b/packages/eds-core-react/src/components/TopBar/TopBar.tsx
@@ -15,16 +15,15 @@ type AvailableElevations = keyof Pick<Elevations, 'none' | 'raised'>
 
 export type TopbarProps = {
   elevation?: AvailableElevations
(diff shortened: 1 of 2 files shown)


packages/eds-core-react/src/components/TopBar/TopBar.tsx
```diff
@@ -33,11 +32,19 @@ const StyledTopBar = styled(Paper)<TopbarProps>(({ theme }) => {
     ${bordersTemplate(theme.border)}
     ${spacingsTemplate(theme.spacings)};
     ${typographyTemplate(theme.typography)}
+
+    ${sticky
```
oddvernes, line 36: Could this not be `sticky &&` instead of ternary if else?

oddvernes (approved):
LGTM

PR #2510, opened 2022-09-21
Upgrade to React 18

### LINKED ISSUE #2474: Test our packages with React 18 upgrade

mimarz:
resolves #2474 

I have published these changes to the test Storybook for testing: https://eds-storybook-react-test.azurewebsites.net/

Tested with our storybook test and locally downstream with pack of eds-core-react. Everything seems to work fine, and I couldn't find any bugs (apart from the know with downshift). I guess it speaks for its self since we treat react as a peer-deps, a lot of our users are already using eds-core-react with React 18, I think we would have heard by now if there were any problems with that. 

The biggest change with this PR is then that our development environment is now using React 18. To begin with we should refrain from using new React 18 hooks/utils until its more adpoted. 

Only nuisance that still persists is that Storybook hasn't completely supported React 18 yet, so we get this error message in SB. 

image


mimarz:
Finally! Checks are green! 🙏

oddvernes (approved):
LGTM 😬

PR #2511, opened 2022-09-22
Add not dismissible for Popover

### LINKED ISSUE #2505: Popover: isDismissable support
Support for `isDismissable` on Portal as users might need to do several actions inside or using Autocomplete without closing Popover. 

This can also be done with ignoring onClose on Popover and just have a button setting open state to false. 

Decide which approach feels best

martalalik:
Resolves #2505

### DIFF
--- a/packages/eds-core-react/src/components/Popover/Popover.docs.mdx
+++ b/packages/eds-core-react/src/components/Popover/Popover.docs.mdx
@@ -107,6 +107,14 @@ const Demo = () => {
 
 <Story id="data-display-popover--with-close-button"/>
 
+### Not dismissable
+
+This story shows `Popover` with more complex content. <br />
+By ignoring `onClose` on Popover, and button setting `open` state to `false`, 
+this will give more control over closing Popover.
+
+<Story id="data-display-popover--not-dismissable"/>
+
 ### Compact
 
 Compact `Popover` using `EdsProvider`.
--- a/packages/eds-core-react/src/components/Popover/Popover.docs.mdx
+++ b/packages/eds-core-react/src/components/Popover/Popover.docs.mdx
@@ -107,13 +107,11 @@ const Demo = () => {
 
 <Story id="data-display-popover--with-close-button"/>
 
-### Not dismissable
+### Persistent Popover
 
-This story shows `Popover` with more complex content. <br />
-By ignoring `onClose` on Popover, and button setting `open` state to `false`, 
-this will give more control over closing Popover.
+To prevent dismissing `Popover` with click outside or escape key, simply leave out the `onClose` callback.
 
-<Story id="data-display-popover--not-dismissable"/>
+<Story id="data-display-popover--persistent-popover"/>
 
 ### Compact
 
--- a/packages/eds-core-react/src/components/Popover/Popover.stories.tsx
+++ b/packages/eds-core-react/src/components/Popover/Popover.stories.tsx
@@ -252,7 +252,9 @@ export const WithTooltip: Story<PopoverProps> = () => {
         <Popover.Header>
           <Popover.Title>Title</Popover.Title>
         </Popover.Header>
-        <Popover.Content>Content</Popover.Content>
+        <Popover.Content>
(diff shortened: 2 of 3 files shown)


packages/eds-core-react/src/components/Popover/Popover.stories.tsx
```diff
@@ -320,6 +321,79 @@ WithCloseButton.decorators = [
...
+        anchorEl={anchorRef.current}
+        open={isOpen}
+      >
+        <Popover.Header>
+          <Popover.Title>With Autocomplete</Popover.Title>
+        </Popover.Header>
+        <Popover.Content>
+          <Wrapper>
+            <Autocomplete label="Select a counties" options={counties} />
+            <Autocomplete label="Select a counties" options={counties} />
```
oddvernes, line 376: Typo: "Select a county"
Do we need two autocompletes in here?

packages/eds-core-react/src/components/Popover/Popover.docs.mdx
```diff
@@ -107,6 +107,14 @@ const Demo = () => {
 
 <Story id="data-display-popover--with-close-button"/>
 
+### Not dismissable
+
+This story shows `Popover` with more complex content. <br />
+By ignoring `onClose` on Popover, and button setting `open` state to `false`, 
+this will give more control over closing Popover.
```
oddvernes, line 114: Maybe something like: "To prevent dismissing popover with click outside or escape key, simply leave out the onClose callback"

packages/eds-core-react/src/components/Popover/Popover.stories.tsx
mimarz: Ditto, suggest just having the one `Autocomplete` and removing the `Wrapper` to avoid confusion.

martalalik:
Great suggestions, but i still not sure if `Not dismissible` title for the story is good 🤔

mimarz:
> Great suggestions, but still, I'm not sure if `Not dismissible` title for the story is good 🤔

I guess we are trying to convey and example of how you can work use interactive components inside the popover. Much like the "review PR" popup on github. So maybe something along the line of; Interactive content or form inside?

oddvernes:
> I guess we are trying to convey and example of how you can work use interactive components inside the popover. Much like the "review PR" popup on github. So maybe something along the line of; Interactive content or form inside?

Yes or maybe persistent popover

martalalik:
Something to look into Popover or just add style?
Screenshot 2022-09-23 at 10 02 37

mimarz:
> Something to look into Popover or just add style? Screenshot 2022-09-23 at 10 02 37
> 
> If adding `padding-bottom` to PopoverContetnt, that sorts the problem: Screenshot 2022-09-23 at 10 12 17
> 
> Wouldn’t that be accrue with Figma? 🤔 Screenshot 2022-09-23 at 10 18 17

how does the actual html and css look like? The `InnerWrapper` has a gap which should solve the spacing between content and actions sub-components.

martalalik:
> how does the actual html and css look like? The InnerWrapper has a gap which should solve the spacing between content and actions sub-components.

InnerWrapper has `grid-gap`, but it looks like the missing part of autocomplete is under the gap.

oddvernes:
It is because popover.content have overflow hidden, so the outline around the input gets clipped (because the gap counts as outside the container)

packages/eds-core-react/src/components/Popover/PopoverContent.tsx
```diff
@@ -3,7 +3,6 @@ import styled, { css } from 'styled-components'
 
 const ContentWrapper = styled.div(({ theme }) => {
   return css`
-    overflow: hidden;
```
oddvernes, line 6: I would revert this and instead fix it in #2517. Need to make sure popover works with a lot of content that fills the height (I assume header and actions should be fixed while content scrolls)

oddvernes (changes requested):

oddvernes (approved):
LGTM

PR #2513, opened 2022-09-22
🎨 Add FloatingOverlay to Scrim

### LINKED ISSUE #2470: Use FloatingOverlay from floation-ui in scrim?
It has a better scroll lock (definetly works in ios, adds padding to body on windows so we don't get content-shift) 
Investigate if this is better solution and implement if so

martalalik:
Resolves #2470


oddvernes (approved):
LGTM! This adds padding to body when there is a scrollbar so we don't get content jump as it opens, and should also work better at preventing scrolling on ios

PR #2519, opened 2022-09-26
✨Divider thickness option

### LINKED ISSUE #2515: Divider: add thickness options
Sidebar needs 2px light divider.

oddvernes:
resolves #2515 
There was some elaborate functional code to subtract divider height from margin bottom. I am not sure why it is important the total height of the divider + margins remain the same (perhaps so the total remains divisable by 8?). Regardless I reimplemented it in 1 line of css.

Should the prop name be `thickness` or `size`, like mantine? I feel like `thickness` is less ambigous since the` variant` changes margins and therefore total size as well

Also this component does not support compact yet but I imagine this might be needed at some point

### DIFF
--- a/packages/eds-core-react/src/components/Banner/__snapshots__/Banner.test.tsx.snap
+++ b/packages/eds-core-react/src/components/Banner/__snapshots__/Banner.test.tsx.snap
@@ -11,7 +11,7 @@ exports[`Banner Matches snapshot 1`] = `
   border: none;
   background-color: var(--eds_ui_background__light,rgba(247,247,247,1));
   margin-top: 16px;
-  margin-bottom: 15px;
+  margin-bottom: calc(16px - 1px);
   height: 1px;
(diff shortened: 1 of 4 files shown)


mimarz:
> resolves #2515 There was some elaborate functional code to subtract divider height from margin bottom. I am not sure why it is important the total height of the divider + margins remain the same (perhaps so the total remains divisable by 8?). Regardless I reimplemented it in 1 line of css.
> 
@vnys would have to answer on this 🤔

> Should the prop name be `thickness` or `size`, like mantine? I feel like `thickness` is less ambigous since the` variant` changes margins and therefore total size as well

I feel like `size` would be more in line with the rest our components, but **thickness** kinda fits the theme 😅
> Also this component does not support compact yet but I imagine this might be needed at some point

mimarz (approved):
LGTM 👍

PR #2524, opened 2022-09-27
🐛 Update spacing in Popover

### LINKED ISSUE #2517: Resolve Popover css selector bug
Discovered in PopoverContent
`    padding: 0 ${theme.spacings.right} 0 ${theme.spacings.left};
    :last-child {
      padding-bottom: ${theme.spacings.bottom};
    }
    :first-child {
      padding-top: ${theme.spacings.top};
    }`

this does not work, the correct syntax is ` > :last-child`. But then the spacing is doubled (16px padding + 16px gap) so this may not be correct either. Furthermore, padding can not be applied to all element types so it should probably be margin. Also it is unclear why not just have padding top and bottom on PopoverContent instead of children?

A quick search for :last-child or :first-child shows this is repeated multiple places so every case should be checked and corrected according to design

martalalik:
Resolves #2517

### DIFF
--- a/packages/eds-core-react/src/components/Popover/Popover.tsx
+++ b/packages/eds-core-react/src/components/Popover/Popover.tsx
@@ -58,6 +58,7 @@ const InnerWrapper = styled.div(({ theme }) => {
     width: max-content;
     max-width: ${theme.maxWidth};
     overflow: auto;
+    padding: ${theme.spacings.top} 0 ${theme.spacings.bottom} 0;
   `
 })
 
--- a/packages/eds-core-react/src/components/Popover/Popover.tsx
+++ b/packages/eds-core-react/src/components/Popover/Popover.tsx
@@ -58,7 +58,6 @@ const InnerWrapper = styled.div(({ theme }) => {
     width: max-content;
     max-width: ${theme.maxWidth};
     overflow: auto;
-    padding: ${theme.spacings.top} 0 ${theme.spacings.bottom} 0;
   `
 })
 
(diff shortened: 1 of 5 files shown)


oddvernes:
"A quick search for :last-child or :first-child shows this is repeated multiple places so every case should be checked and corrected according to design"

martalalik:
> "A quick search for :last-child or :first-child shows this is repeated multiple places so every case should be checked and corrected according to design"

New issue for it #2525

oddvernes:
That just looks like a duplicate of this parent issue but without any of the description

mimarz:
After testing this and old solution, I have to ask myself, was this css really wrong? When content/actions is last/first then it has the correct padding, which is maybe the intended behaviour.

oddvernes:
> After testing this and old solution, I have to ask myself, was this css really wrong? When content/actions is last/first then it has the correct padding, which is maybe the intended behaviour.

Actually you are correct 😂! I assumed first/last child was meant to refer to the children of the components, but it is actually meant to target the element _itself_ when it is last/first.

oddvernes:
@martalalik so essentially the only thing that needed to be fixed was overflow hidden, the last/first child stuff is there so padding only comes from the subcomponents in case people want to do their own thing inside popover, but I misunderstood what the selector was supposed to target. So you can either close this pr and make a new one to just fix overflow hidden,  or revert everything except overflow: hidden in this branch

mimarz (approved):
LGTM 👍

PR #2531, opened 2022-09-28
🐛Sidebar link active warning

### LINKED ISSUE #2529: SideBar: make active a transient prop?
errormessage when loading sidebar with links cast as react-router Link: 
```
Warning: Received `true` for a non-boolean attribute `active`.

If you want to write it to the DOM, pass a string instead: active="true" or active={value.toString()}.
```

oddvernes:
resolves #2529 
Turns out extending an overridable component (Button) inside another overridable component (SideBar.Link) does not work very well when overriding, and adds the props as html attrs onto the final component. But I realized it is pointless to extend Button here anyway, so I removed it and just use an `a` element as base.

### DIFF
--- a/packages/eds-lab-react/src/components/SideBar/SidebarLink/index.tsx
+++ b/packages/eds-lab-react/src/components/SideBar/SidebarLink/index.tsx
@@ -1,9 +1,4 @@
-import {
-  ButtonHTMLAttributes,
-  forwardRef,
-  AnchorHTMLAttributes,
-  ElementType,
-} from 'react'
+import { forwardRef, AnchorHTMLAttributes, ElementType } from 'react'
 import { sidebar as tokens } from '../SideBar.tokens'
 import {
   bordersTemplate,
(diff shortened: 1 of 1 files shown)


packages/eds-lab-react/src/components/SideBar/SidebarLink/index.tsx
```diff
@@ -35,15 +28,10 @@ const {
...
 
 type ContainerProps = {
-  active?: boolean
-} & StrippedButton
-
-type StrippedButton = Omit<
-  ButtonProps,
-  keyof ButtonHTMLAttributes<HTMLButtonElement>
->
+  $active?: boolean
```
mimarz, line 31: why did you choose `$active` as prop name with $?

oddvernes: I tested again with just `active` after the other change and nope, still got the warning, but with `$active` no warning so it seems to be needed here

mimarz (approved):

PR #2536, opened 2022-09-29
♿️  Add portal & focusManager to Popover

### LINKED ISSUE #2386: Popover in portal? (+ focusmanager)
Making this a separate issue to floating-ui, because I don't know if there was a good reason why the current implementation is not in a portal and not conditionally rendered but instead just hidden with css visibility, and everyone to discuss with is on vacation.

This shoud be very easy to implement with [FloatingPortal](https://floating-ui.com/docs/FloatingPortal) from floating-ui, as i did with tooltip. At the same time [FloatingFocusManager](https://floating-ui.com/docs/FloatingFocusManager) could be added for focus trapping (maybe as an optional thing, look at how mui/mantine does it)

martalalik:
Resolves #2386

### DIFF
--- a/packages/eds-core-react/src/components/Popover/Popover.tsx
+++ b/packages/eds-core-react/src/components/Popover/Popover.tsx
@@ -31,11 +31,8 @@ import {
   FloatingFocusManager,
 } from '@floating-ui/react-dom-interactions'
 
-type StyledPopoverProps = Pick<PopoverProps, 'open'>
-
-const PopoverPaper = styled(Paper)<StyledPopoverProps>(({ theme, open }) => {
+const PopoverPaper = styled(Paper)(({ theme }) => {
   return css`
-    ${{ display: open ? 'block' : 'none' }}
     ${typographyTemplate(theme.typography)}
     background: ${theme.background};
     ${bordersTemplate(theme.border)}
@@ -86,10 +83,10 @@ export type PopoverProps = {
   anchorEl?: HTMLElement | null
   /** Is Popover open */
   open: boolean
-  /** Disable use of react portal for dropdown */
-  disablePortal?: boolean
+  /** initializes react portal for dropdown, default to false. */
+  withinPortal?: boolean
   /** Determines whether focus should be trapped within dropdown,
-   * default to false */
+   * default to false. */
   trapFocus?: boolean
 } & HTMLAttributes<HTMLDivElement>
 
@@ -102,7 +99,7 @@ export const Popover = forwardRef<HTMLDivElement, PopoverProps>(
       style,
       open,
       onClose,
-      disablePortal,
+      withinPortal,
       trapFocus,
       ...rest
     },
@@ -215,10 +212,20 @@ export const Popover = forwardRef<HTMLDivElement, PopoverProps>(
 
     return (
       <>
-        {disablePortal ? (
-          popover
+        {withinPortal ? (
+          open && (
+            <FloatingPortal id="eds-popover-container">
+              {trapFocus
+                ? open && (
+                    <FloatingFocusManager context={context} modal={true}>
+                      {popover}
+                    </FloatingFocusManager>
+                  )
+                : open && popover}
+            </FloatingPortal>
+          )
         ) : (
-          <FloatingPortal id="eds-popover-container">
+          <>
             {trapFocus
               ? open && (
                   <FloatingFocusManager context={context} modal={true}>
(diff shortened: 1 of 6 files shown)


oddvernes:
I think the plan is to make portal optional via a prop similar to how we did it in autocomplete fyi

martalalik:
Putting this into the review to discuss documentation, that got bit of an update. There are some test failing too but this i will get sort it.

After adding portal and focus trapping, it brought some changes in the focus trapping behaviour. This also risen some questions about content of Popover.
[Figma](https://www.figma.com/file/0bGXR2sCwMVSDNyyzu5BXrO5/UI%E2%80%94User-Interface?node-id=17515%3A0) 

Screenshot 2022-09-30 at 16 09 54

I brought a suggestions which are reflected in docs.

**Proposition**
To standardise a Popover look in docs, i removed actions and I added closing button in the header instead.
Screenshot 2022-09-30 at 16 15 17

**Reasons:**

- Fits Figma design.
- Brings visible difference between Dialog & Popover.

oddvernes:
From what I remember close button was hardcoded in this component, but we removed it because there isn't much point of having it, except some specific cases, and then people can add it manually instead. 
I don't see the point of having a closebutton in all these samples. 

Also there should (probably?) not be focustrapping when there is no interactive content (see mantine documentation. MUI seems to force focustrap and scroll lock as if it was a dialog but I don't know if i agree on that). On that note, i see you still use `FloatingFocusManager` around `popover` when `trapFocus` is false, which then traps focus regardless?

If the paper is focusable (which it currently is set to be) we should add our focus style (`&:focus-visible` in `PopoverPaper` with `outlineTemplate` and states.focus.outline data in token)

On the hover example,  if you add `trapFocus` it freaks out and starts strobing, so this needs to be figured out.

I have a bug where the arrow disappears the second time you open a popover, this seems to affect all of the examples regardless if mouse or kb. I am wondering if this is related to the upgraded version of floating-ui. You should rebase this branch on develop and see if you also get this bug.

oddvernes:
I now see you introduced conditional rendering (open &&) but the component is written to specifically not do that, so this seems to be the root of some issues here, like arrow vanishing and just overall how it behaves. Looks like conditional rendering might be needed for trapping focus to work, not sure. If I change things around to allow conditional rendering (remove conditional dsiplay from css, add `whileElementsMounted: autoUpdate,`  to `useFloating`, remove the useEffect to call `autoUpdate` and set `modal={true}`) then it seems to be working better, although arrow is still bugged.

mimarz:
> I think the plan is to make portal optional via a prop similar to how we did it in autocomplete fyi

Maybe the toggle should be opposite of the toggle in `Autocomplete`. Not sure if we want the `Popover` to be in a portal by default as its "removed" from the stacking content. 

I see Mantine has the same, where you opt-in for a portal with `withinPortal` prop 🤔

packages/eds-core-react/src/components/Popover/Popover.tsx
```diff
@@ -199,5 +212,23 @@ export const Popover = forwardRef<HTMLDivElement, PopoverProps>(
         </PopoverPaper>
       </ThemeProvider>
     )
+
+    return (
+      <>
+        {disablePortal ? (
+          popover
```
oddvernes, line 219: This also needs to be conditionally rendered, and also I think you forgot to remove the css to hide when closed. Also FloatingFocusManager should be possible when no portal

packages/eds-core-react/src/components/Popover/Popover.tsx
```diff
@@ -125,47 +140,42 @@ export const Popover = forwardRef<HTMLDivElement, PopoverProps>(
       [floating, ref],
     )
 
-    useEffect(() => {
-      if (refs.reference.current && refs.floating.current && open) {
-        return autoUpdate(refs.reference.current, refs.floating.current, update)
-      }
-    }, [refs.reference, refs.floating, update, open])
```
oddvernes, line 132: Note that `refs` and `update` are no longer used and should be removed from useFloating (line 115-116)

oddvernes:
I found another issue. When we gave `Autocomplete` the same z-index as `Popover` it seemed to work fine, but when popover is in a portal, it appears later than autocomplete in the Dom, so now it is behind the popover again. So Autocomplete needs a higher z-index after all. I suppose the Tooltip z-index of 1500

oddvernes (approved):
LGTM!

PR #2540, opened 2022-09-30
🐛 Accordion: fixed shrinking icon

### LINKED ISSUE #2539: Accordion: fix shrinking icon
The left chevron icon shrinks when accordion header is squeezed and text starts overflowing. This is because it is in a flexbox and default is flex-shrink: 1 

This explains a screenshot i saw a while ago from the designers where this chevron was tiny

oddvernes:
resolves #2539


mimarz (approved):
LGTM 👍

PR #2541, opened 2022-09-30
✨ Add core typography tokens

vnys:
Core typography tokens in Figma. There are currently 4 scales, ranging from tight to relaxed, where the difference is the line-height. Line-heights are calculated based on font-size and rounded up to the nearest 4px to fit with the layout grid.

Known issues: Figma Tokens does not add new line at the end of the file, I will notify the developers about this


packages/eds-tokens/src/tokens/.gitkeep
```diff
@@ -0,0 +1 @@
+
```
mimarz, line 1: Don't think we need this file now?

mimarz (changes requested):
Not sure where these tokens best belong, in here or maybe `assets/tokens` ? We will either way need a script or something like styled dictionary to convert these to usable files for consumers of this package. Guess we can just move them somewhere else if needed. 

Sugggest maybe a README or something in `src/tokens` that states that these are files auto-generated related to figma-tokens plugin? How will this affect the plugin if some of these files are suddenly moved/edited by other?

packages/eds-tokens/src/tokens/.gitkeep
vnys: I don’t think so either, but it did a good job while it lasted. Just hanging there, waiting for some friends. And when they finally arrive it’s time to go. Bummer.

vnys:
> Not sure where these tokens best belong

I think putting them in `packages/eds-tokens/` makes sense. The tokens can be modified outside of Figma, as long as the format remains the same. The plugin checks if there are changes in the repo when it’s started, and asks the user if s/he wants to pull the latest changes.

I can add a README with an instruction on how to load the tokens in the Figma Tokens plugin as well.

vnys:
> Does it check against the whole repo or can it specify against a given path? 🤔

You can specify path and branch

mimarz (approved):
LGTM 👍 

I assume we'll use styled dictionary or the likes to transform these tokens later 🤞

PR #2542, opened 2022-10-04
🐛 Breadcrumb/Typography: use transient props

### LINKED ISSUE #2538: Breadcrumbs: fix polymorphic issue
Same problem as was discovered in SideBar.Link, where an overridable component  (now typography)  is used inside another overridable component (breadcrumb) and props end up getting sent all the way to the dom

oddvernes:
resolves #2538 

Now casting breadcrumb `as={Link}` from react router will not throw warnings about maxWidth being applied to a dom element


mimarz (approved):
LGTM 👍

PR #2556, opened 2022-10-05
✨Polymorphic tabs

### LINKED ISSUE #2434: Update Tabs Tab with OverridableComponent type
Investigate if this is possible so tabs can be used as nav for react router

oddvernes:
resolves #2434 

tested downstream with react-router-dom and nested routes like this:
bilde


mimarz (approved):
LGTM 👍

PR #2567, opened 2022-10-10
🐛 Accordion: prevent toggle from submitting form

### LINKED ISSUE #2566: Accordion trigger "submit" inside form
**Describe the bug**

Accordion header `StyledAccordionHeaderButton` triggers submit inside forms

**Expected behavior**

Accordion should not trigger submit. Set type="button" and probably preventdefault/stoppropagation for good measure

oddvernes:
resolves #2566 

tested by wrapping a form around one of the stories. Before: it submitted which reloads page. After: it does not submit


mimarz (approved):
LGTM 👍

PR #2570, opened 2022-10-11
📝 Update aria-label in Popover

### LINKED ISSUE #2569: Popover improve a11y in docs
Popover is lucking aria-labels in docs, makes is less accessible.
Story with Tooltip is questionable. Should Tooltip be used together with Popover 🤔 ?
Use `aria-label` instead.

martalalik:
Resolves #2569


oddvernes (changes requested):
Just fix the duplicated ids

martalalik:
https://www.w3.org/WAI/ARIA/apg/example-index/

oddvernes (approved):
LGTM

PR #2571, opened 2022-10-11
🔥 Remove commonjs path for in eds-icons & eds-tokens

### LINKED ISSUE #2559: Remove commonjs path in tokens & icons packages
We introduced a alternative path for our packages a while back when esm and commonjs modules support were abit shakey in the node world. 

We introduced a separate path for specifically targeting the cjs bundling packages.

Now build/transpilers/node have good support for esm, aswell as we remove this already for `core-react` a good while ago and we haven't had any feeback on people missing the alternative path.

Remove this alternative paths for `eds-tokens` and `eds-icson`

mimarz:
resolves #2559 

Remove the last of fallback support for commonjs when node had problems figuring out esm vs commonjs modules.

### DIFF
--- a/packages/eds-icons/commonjs/package.json
+++ b/packages/eds-icons/commonjs/package.json
@@ -1,4 +1,0 @@
-{
-  "type": "commonjs",
-  "main": "../dist/icons.cjs.js"
-}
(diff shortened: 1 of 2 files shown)


oddvernes (approved):
LGTM

PR #2586, opened 2022-10-13
🚧 Add virtual-scrolling story for Table

### LINKED ISSUE #2561: Example of virtual scrolling in Table

I would love to have a demonstration of best practice ™️ on how to implement virtual scrolling in the React table, as we often have vast amounts of data on the screen at any given time.

If I had the expertise in react to create a best practice ™️ example this would be viable
image

Alas I would probably make a hacky example that would be more like this
image

**Additional context**

martalalik:
Resolves #2561

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,5 +58,8 @@
     "jest-axe": "^6.0.0",
     "prettier": "2.7.1",
     "typescript": "^4.8.4"
+  },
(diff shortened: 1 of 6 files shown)


martalalik:
Need further investigation, based on description in issue https://github.com/equinor/design-system/issues/2561.

PR #2596, opened 2022-10-17
🐛  Improve a11y for screen readers for Tooltip

### LINKED ISSUE #2591: Tooltip: screen reader does not always read
With windows narrator the tooltip is read when on a button but not when on a link or a span.

Test on mac, does it read tooltip here?
Investigate if there is a way to entice narrator to read tooltip when opened (aria-live etc)

Apply Adam Argyles trick to have it announce the tooltip:
`  &::before {
    content: "; Has tooltip: ";
    clip: rect(1px, 1px, 1px, 1px);
    clip-path: inset(50%);
    height: 1px;
    width: 1px;
    margin: -1px;
    overflow: hidden;
    padding: 0;
    position: absolute;
  }`

martalalik:
Resolves #2591

### DIFF
--- a/packages/eds-core-react/src/components/Tooltip/Tooltip.stories.tsx
+++ b/packages/eds-core-react/src/components/Tooltip/Tooltip.stories.tsx
@@ -128,7 +128,7 @@ export const LongListWithTooltips: Story<TooltipProps> = () => {
       {items.map((_, i) => (
         // eslint-disable-next-line react/no-array-index-key
         <span key={i}>
-          <Tooltip title={`Icon ${i}`} placement="right" aria-live="polite">
+          <Tooltip title={`Icon ${i}`} placement="right">
             <Icon data={explore} />
           </Tooltip>
         </span>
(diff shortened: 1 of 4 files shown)


martalalik:
Unfortunately, `aria-live` attribute, causes doble reading of tooltip content.
It seems to be a know bug described [here](https://github.com/phetsims/a11y-research/issues/132).

packages/eds-core-react/src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -128,7 +128,7 @@ export const LongListWithTooltips: Story<TooltipProps> = () => {
       {items.map((_, i) => (
         // eslint-disable-next-line react/no-array-index-key
         <span key={i}>
-          <Tooltip title={`Icon ${i}`} placement="right">
+          <Tooltip title={`Icon ${i}`} placement="right" aria-live="polite">
```
oddvernes, line 131: You can remove these `aria-live`, they do nothing because they have to exist outside of and before the elements added (tooltip is conditionally rendered)

oddvernes (changes requested):

martalalik:
We are aware that Tooltip has anti a11y pattern, for today. Mainly because we can add Tooltip to any element, even too non-focusable. [W3C](https://www.w3.org/WAI/ARIA/apg/) is still working on [Tooltip pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/).

oddvernes (approved):
I think we can merge this now. We didn't solve the main point of he ticket (make it read tooltips with windows voiceover) but a lot of research was done that can be used if we try this again later!

PR #2597, opened 2022-10-18
👷 Updated actions using set-output to use environment files

### LINKED ISSUE #2580: Update github actions with environment files
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files

mimarz:
resolves #2580 

This won't run correctly as reusable workflows are used by branch name (We denote our reusable workflows with starting underscore in the name).

There is one warning for "set-output" still but I looks to be in the [Azure Web App deploy action ](https://github.com/Azure/webapps-deploy). This will probably need to be updated when a new version is released.


oddvernes (approved):
LGTM

PR #2624, opened 2022-10-22
:children_crossing: Visually disable components that have aria-disabled

### LINKED ISSUE #2544: Visually disable components that have aria-disabled?
In some cases, users would want to use `aria-disabled` instead of `disabled` on buttons and other disableable components, which ensures the elements is still keyboard-navigatable. But we do not apply the disabled style in these cases, which we should to make this usable.

denektenina:
resolves #2544


oddvernes (approved):
LGTM

PR #2643, opened 2022-11-07
🔥 Removed deprecated SingleSelect and MultiSelect

### LINKED ISSUE #2638: Remove singleselect & multiselect
These components have been marked as deprecated for quite a while and many releases now, so it is time to remove them

denektenina:
Resolves #2638


oddvernes (changes requested):
I think we can also remove commonStyles.ts and Select.tokens.ts, I do believe they were only used by select and multiselect while nativeSelect has its own token?

oddvernes:
I suppose also we no longer need the 'select' folder in storybook with only one select type left. The way to remove it is to go into NativeSelect.stories.tsx line 8 and change to `title: 'Inputs/NativeSelect',` and then in NativeSelect.docs.mdx change the id's to `inputs-nativeselect--introduction` etc.
**Note** that this will probably break the link in zeroheight so need to inform Rebecca to fix when we publish storybook

oddvernes (approved):
LGTM!

PR #2655, opened 2022-11-11
🔥Storybook: Disable compact stories where it is not implemented yet

### LINKED ISSUE #2653: Disable compact stories for components where compact has not been implemented
`Accordion` was reported in. I don't know if there are any others, but we should look over and comment out any that are not implemented yet since it is confusing (presumably the story were made with the expecation that compact was _imminent_ earlier in the year)

Edit, the list:

- Accordion
- Banner
- Dialog
- Popover
- TableOfContents
- Tabs
- TopBar
- Snackbar

oddvernes:
Resolves #2653 

Deleted snackbar story because I don't think this component will even have compact. The rest are commented out


denektenina (approved):
lgtm

PR #2674, opened 2022-11-25
Improve docs for Slider outputFunction

Strepto:
I spent a bit of time on finding this option, as it was not clear to me that the `outputFunction` processes the current value for display text.

I added a suggested change, but feel free to change it 👍


oddvernes (approved):
Good clarification! 👍

PR #2677, opened 2022-11-25
📝 Added missing changelog items

### LINKED ISSUE #2657: Button size for ghost_icon & contained_icon changed from 0.23 to 0.24
It seems that the default size for `ghost_icon` and `contained_icon` Button types changed from 48px to 40px in `comfortable` density while 0.24 was released.

Although I think the 40px size is generally better – so this change can be celebrated –, there are two issues here:

- the change was not documented, and
- other components, notably Checkbox & Radio – maybe others too – did not follow the change, which might create an unbalanced and misaligned situation in layouts.

[0.23 Codesandbox](https://codesandbox.io/s/button-default-size-eds-core-react-0-23-forked-qw3o1o?file=/src/App.tsx)
[0.24 Codesandbox](https://codesandbox.io/s/button-default-size-eds-core-react-0-24-fkj6xr?file=/src/App.tsx)

oddvernes:
resolves #2657 

I have already updated the release notes on github with the same entries

### DIFF
--- a/packages/eds-core-react/CHANGELOG.md
+++ b/packages/eds-core-react/CHANGELOG.md
@@ -9,19 +9,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-* 📝 `Textfield` Storybook: Added validation example by @oddvernes in https://github.com/equinor/design-system/pull/2640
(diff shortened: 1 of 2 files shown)


denektenina (approved):
lgtm

Issue #1909, opened 2022-01-24
Support large data sets in Autocomplete


mimarz:
Selects/Autocompletes/Comboboxes are in most cases used for selecting tags in Equinor. Tags are usually very large data sets so we should probably do some optimising for our upcoming Autocomplete to support large data sets. 

[Examples from downshift using react-window](https://codesandbox.io/s/github/kentcdodds/downshift-examples?file=/src/downshift/other-examples/react-window/index.js)

sindre-nistad:
We have a similar problem in a project, where ~5000 options is available in some cases.
This causes the component to be _very_ slow, and only render part of the items.

oddvernes:
Interesting! Well since we are currently looking at lazy-rendering table in #2561 there may be some lessons learned there. I am moving this ticket higher up from the debths of the backlog 👍

oddvernes:
@sindre-nistad We have a test release out now if you have the opportunity to check it out 👍  `@equinor/eds-core-react@next`. Currently virtualization is _always on_ just to avoid bloating this component with even more code but I may have to add a` virtualize` prop if this does not work in some cases like ssr (yet to be tested)

sindre-nistad:
Great! I'll check it out

sindre-nistad:
Initial testing shows great promise. I've tried it with a dropdown with ~5000 elements, and one with ~1800 elements.
Scrolling, and _really_ fast scrolling does not cause the app to slow to a crawl.
If I let the scroll wheel fly freely, the elements seem to render occasionally, but it still scrolls down. During _really fast_ scroll the elements may disappear completely.
Searching, and selecting works smoothly.

PR #2689, opened 2022-12-09
✨ Virtualized autocomplete

oddvernes:
resolves https://github.com/equinor/design-system/issues/1909

Remains to figure out a musterious error message in storybook and look into if i can make virtualization optional (its a bit decentralized with changes various placed that would need to be toggled)

### DIFF
--- a/packages/eds-core-react/package.json
+++ b/packages/eds-core-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@equinor/eds-core-react",
-  "version": "0.28.0-dev12052022",
+  "version": "0.28.0-dev12152022",
   "description": "The React implementation of the Equinor Design System",
   "sideEffects": [
     "**/*.css"
--- a/packages/eds-core-react/package.json
+++ b/packages/eds-core-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@equinor/eds-core-react",
-  "version": "0.27.0",
+  "version": "0.28.0-dev12052022",
   "description": "The React implementation of the Equinor Design System",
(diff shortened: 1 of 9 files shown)


oddvernes:
I added a `multiline` prop which is false by default. This is mainly relevant if someone have `autoWidth` set and the text overflows. By default text is cut off with ellipsis, but when you add `multiline` the text will wrap. This also changes virtualizer settings to measure height per item, so it adds a bit of performance overhead. The old behaviour was always wrap, but then we didnt have the performance penalty either.

I am uncertain if i should add a `virtualizerOverscan` prop or leave it hardcoded at 25. It sets how many elements are rendered above and below visible viewport. Lower numbers introduce glitchy appearance when scrolling fast as items struggle to render fast enough, while with bigger numbers the scrolling  is mostly fine but adds more up front delay on opening/searching/dragging scrollbar. I find 25 to be a good compromize but perhaps users would want more control over this?

oddvernes:
There was an issue where autocomplete items would not resize when switching between comfortable and compact, but I found a workaround for this. 

Finally the last outstanding issue, in the storybook Autocomplete stories, at least when running locally you'll probably see this in the console everytime opening the dropdown: 
bilde
This is some bug in storybook where window.onerror is captured but fails to send as xhr 
bilde
If I intercept this window.onerror myself, it turns out the error is `ResizeObserver loop limit exceeded`. It means ResizeObserver was not able to deliver all observations within a single animation frame. It is benign and can be ignored.

denektenina (approved):
LGTM

PR #2706, opened 2023-01-02
⬆️ Updated floating-ui to latest version

### LINKED ISSUE #2703: Update floating UI to latest version
It has changed name from `@floating-ui/react-dom-interactions` to `@floating-ui/react`. Since there have been many releases since the last time we updated, make sure to read changelog for all releases and take note of potential breaking changes that may be relevant to us and test every component that uses floating-ui

This might be relevant to #2693 and #2698 as well

oddvernes:
resolves #2703

### DIFF
--- a/packages/eds-core-react/package.json
+++ b/packages/eds-core-react/package.json
@@ -95,7 +95,7 @@
     "@equinor/eds-icons": "workspace:*",
     "@equinor/eds-tokens": "workspace:*",
     "@equinor/eds-utils": "workspace:*",
-    "@floating-ui/react-dom-interactions": "^0.10.1",
+    "@floating-ui/react": "^0.15.1",
     "@tanstack/react-virtual": "3.0.0-beta.30",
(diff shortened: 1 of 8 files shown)


oddvernes:
For some unexplained reason two sidebar tests fail after upgrading floating-ui. These tests does not fail in develop. Floating-ui is not used by sidebar and I can not see floating-ui installing any different versions of jest/testing-library in package-lock 🤔 

Continuing the investigation...

oddvernes:
I suspect the problem with sidebar was related to Tooltip and how the FloatingPortal is injected since v([0.13.0](https://github.com/floating-ui/floating-ui/releases/tag/%40floating-ui%2Freact-dom-interactions%400.13.0)) combined with the flimsy selector we had in the test (`expect(screen.getAllByRole('generic')[2]).toHaveStyle({`) So I think its fine 🤞

denektenina (approved):

Issue #2682, opened 2022-11-30
Button loading state design overhaul?


oddvernes:
As was mentioned in the eds slack channel, our design on eds.equinor.com for a "loading" button is a normal non-disabled button with a loader with neutral color. 
![bilde.png](https://images.zenhubusercontent.com/6143345153d23c1b5b6353fc/e600051b-fd86-4839-9696-5376a4136068)
But usually, you want a button in a loading state to be disabled, which then looks like this:
![bilde.png](https://images.zenhubusercontent.com/6143345153d23c1b5b6353fc/9bac501f-758d-4ef0-9c35-29b3e83df831)
[This](https://uxmovement.com/buttons/when-you-need-to-show-a-buttons-loading-state/) article has a good point that the button should be visually disabled but have a high contrast color for the indicator, which in our case would be this:
![bilde.png](https://images.zenhubusercontent.com/6143345153d23c1b5b6353fc/60cbc84f-a46b-4bd9-85be-c75de8916ee5)

BirteThornquist:
will be solved as a part of #2820

PR #2742, opened 2023-01-20
📝 Improved documentation on Button

### LINKED ISSUE #2724: Tooltip around disabled button is not shown in Chrome (works in Firefox)
Repro: https://codesandbox.io/s/dazzling-rosalind-28qeuh?file=/src/App.tsx

```
<Tooltip title="I'm disabled. Chrome users can't see this">
  <Button disabled={true}>I'm disabled. You can't click me</Button>
</Tooltip>
```

Tooltip on the above button won't be shown in Chrome. This seem to be caused by Chrome not firing events for disabled elements. See https://stackoverflow.com/a/18113961. Firefix does fire these events though.

A workaround can be to wrap the children of the Tooltip in a `<div>` as suggested in the linked Stackoverflow above. Although as you see in the repro link the positioning of the tooltip might need some work.

- Version:  "@equinor/eds-core-react": "0.28.0",
- Browser: Chrome 108.0.5359.125 (doesn't work) / Firefox 107.0 (works)
- OS: Windows 11

denektenina:
Resolves #2724

### DIFF
--- a/packages/eds-core-react/src/components/Button/Button.docs.mdx
+++ b/packages/eds-core-react/src/components/Button/Button.docs.mdx
@@ -86,14 +86,11 @@ Use the `Progress` component if you need progress indication on a button.
 
 <Story id="inputs-button-button--progress-button" />
 
-### With tooltip
-To display a tooltip on the `Button` you can wrap it in a `Tooltip`. If you need to disable the Button, the button should be disabled using
-`aria-disabled` instead of the `disabled` prop because Chrome doesn't expose the hover event on disabled buttons. 
-
-When using `aria-disabled`, the button must also be disabled in in the code, as described in the 
-[Accessibility-section](?path=/docs/inputs-button-button--accessibility).
-
-An alternative workaround is wrapping the `Button` inside a fitted `span`.
+### Disabled buttons and tooltip
+Due to inconsistent browser support for mouse events on disabled buttons, `Tooltip` 
+should not be used in combination with `disabled` on `Button`. Instead use `aria-disabled` 
+and disable the `onClick` in the code. This has the added benefit of making the 
+Tooltip accessible for people using keyboard navigation, as the button is still focusable.
 
 <Story id="inputs-button-button--with-tooltip" />
 
--- a/packages/eds-core-react/src/components/Button/Button.docs.mdx
+++ b/packages/eds-core-react/src/components/Button/Button.docs.mdx
@@ -41,16 +41,22 @@ import { Button } from '@equinor/eds-core-react'
 <Button as={Link} to="/about">About</Button>
 ```
 
+## Accessibility
+The `aria-disabled` attribute makes it possible to semantically disable the `Button` without hiding it from assistive technologies, such as screen readers. 
+In EDS, the `Button` will be visually styled as disabled, but it is up to the developer to actually disable the button.
+
+<Story id="inputs-button-button--accessibility" />
+
 ## Examples
 
 ### Basic
-The `Button` come with four variants; `contained`, `outlined`, `ghost` and `ghost_icon`.
+The `Button` come with five variants; `contained`,  `contained_icon`, `outlined`, `ghost` and `ghost_icon`.
 
 <Story id="inputs-button-button--basic" />
 
 ### Icon button
 The `ghost_icon` & `contained_icon`variant is meant for standalone icon buttons using the `<Icon>` component.  
-Make sure to defined `aria-label` with descriptive text of `Button` interaction.
+Make sure to define `aria-label` with descriptive text of `Button` interaction.
 
 <Story id="inputs-button-button--icon-button" />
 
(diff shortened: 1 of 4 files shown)


packages/eds-core-react/src/components/Button/Button.docs.mdx
```diff
@@ -68,32 +74,35 @@ Use variants to differentiate the hierarchy.
...
 ### Progress button
 Use the `Progress` component if you need progress indication on a button.
 
 <Story id="inputs-button-button--progress-button" />
 
-### All
-Example All button combinations. 
+### With tooltip
+To display a tooltip on the `Button` you can wrap it in a `Tooltip`. If you need to disable the Button, the button should be disabled using
+`aria-disabled` instead of the `disabled` prop because Chrome doesn't expose the hover event on disabled buttons.
```
oddvernes, line 91: This is a bit too specific, and therefore not entirely correct. Tooltip also does not work in `edge`, and in `safari` there is an onmouseover event, but no onmouseleave event, so the tooltip gets stuck always on. Instead i suggest: 

Title: "Disabled buttons and tooltip"
text: "Due to inconsistent browser support for mouse events on disabled buttons, `Tooltip` should not be used in combination with `disabled` on `Button`. Instead use `aria-disabled` and manually block the button `onclick` action. This has the added benefit of making the `Tooltip` accessible for people using keyboard navigation, as the button is still focusable."

packages/eds-core-react/src/components/Button/Button.stories.tsx
```diff
@@ -139,46 +178,126 @@ Hierarchy.decorators = [
...
+    setIsSubmitting(true)
+  }
+
+  return (
+    <>
+      <Button
+        aria-disabled={isSubmitting ? true : false}
+        onClick={!isSubmitting ? onSubmit : undefined}
+      >
+        {isSubmitting ? <Progress.Dots /> : 'Fetch data'}
```
oddvernes, line 225: The progress indicator should have color="primary" as discussed in #2682 (It has been tentatively agreed with Birte this is the better solution)

packages/eds-core-react/src/components/Button/Button.stories.tsx
```diff
@@ -139,46 +178,126 @@ Hierarchy.decorators = [
...
+
+  const onSubmit = () => {
+    setIsSubmitting(true)
+  }
+
+  return (
+    <>
+      <Button
+        aria-disabled={isSubmitting ? true : false}
+        onClick={!isSubmitting ? onSubmit : undefined}
```
oddvernes, line 223: Should add `aria-label={isSubmitting ? 'loading data' : null}` whenever there is icon only indicating the state

packages/eds-core-react/src/components/Button/Button.stories.tsx
```diff
@@ -139,46 +178,126 @@ Hierarchy.decorators = [
...
-    <Button>
-      <Progress.Circular size={16} color="neutral" />
-    </Button>
-    <Button variant="ghost_icon">
-      <Progress.Circular size={16} color="primary" />
-    </Button>
+    <Tooltip title="This is what a tooltip looks like">
+      <Button>Hover me</Button>
+    </Tooltip>
+    <Tooltip title="Tooltip doesn't show in Chrome">
```
oddvernes, line 265: can we change text to: "This tooltip only shows for people using firefox and using mouse. Don't do this!"

packages/eds-core-react/src/components/Button/Button.stories.tsx
```diff
@@ -139,46 +178,126 @@ Hierarchy.decorators = [
...
+      <Button disabled>Disabled button</Button>
+    </Tooltip>
+    <Tooltip title="Tooltip shows in Chrome with aria-disabled">
```
oddvernes, line 268: Can we change text to: "Tooltip works in all browsers and with keyboard navigation when using aria-disabled"

packages/eds-core-react/src/components/Button/Button.stories.tsx
```diff
@@ -139,46 +178,126 @@ Hierarchy.decorators = [
...
+      <Button aria-disabled>Aria-disabled button</Button>
+    </Tooltip>
+    <Tooltip title="Tooltip shows in Chrome because Button is wrapped in span">
```
oddvernes, line 271: Delete the example using span (for reasons explained above)

packages/eds-core-react/src/components/Tooltip/Tooltip.docs.mdx
```diff
@@ -71,3 +71,14 @@ Long list of elements that have a tooltip. Used for testing if tooltip "lag" whe
 Inputs of type radio, checkbox and switch will have the tooltip applied to the input element. A good way to get the tooltip to also apply to the label is to simply wrap the input in a span.
 
 <Story id="data-display-tooltip--radio-and-checkboxes" />
+
+### Tooltip on Button
+To display a tooltip on the `Button` you can wrap it in the `Tooltip`. If you need to disable the `Button`, the button should be
```
oddvernes, line 76: Do the same changes here as I suggested in button docs

packages/eds-core-react/src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -152,3 +152,20 @@ export const RadioAndCheckboxes: Story<TooltipProps> = () => (
   </>
 )
 RadioAndCheckboxes.storyName = 'Radio and checkboxes'
+
+export const TooltipOnButton: Story<TooltipProps> = () => (
+  <>
+    <Tooltip title="This is what a tooltip looks like">
```
oddvernes, line 158: Do the same changes here as I suggested in the button stories

oddvernes (approved):
LGTM!

PR #2751, opened 2023-01-31
✨ Add tooltip support for ToggleButton

### LINKED ISSUE #2720: Support Tooltip in Button.Toggle buttons
As reported [on slack](https://equinor.slack.com/archives/CJT20H1B9/p1673532166076649) by Pål Mathias Klungresater Nødseth

Currently, adding a tooltip around a button inside `Button.Toggle` will discard the tooltip and the button inside the tooltip because children are checked for type and then cloned with additional attributes. Removing the check and cloning the tooltip will add back the button and tooltip but breaks the toggle logic so we can't do that straight forward. Perhaps check if type is Tooltip and handle that case especially will work? Otherwise maybe we could find a simple enough workaround and add it to the storybook

oddvernes:
resolves #2720


denektenina (approved):
lgtm

Issue #2763, opened 2023-02-02
The Paper component adds elevation as a dom attribute


bjartebore:
**Describe the bug**

Setting elevation on a Paper components adds an irregular `elevation="..."` attribute to the rendered html component. This is not valid html and clutters the dom. 

**Steps to reproduce the bug**

1. Add a Paper component or other component as a Card
2. Add the elevation argument to the component
3. Inspect the dom

**Expected behavior**

The element should only show classnames used render the element

**Specifications**

- Version: ^0.28.0
- Browser: Chrome (109.0.5414.119)
- OS: MacOS

**Additional context**

<img width="276" alt="image" src="https://user-images.githubusercontent.com/1978938/216274027-b29cb035-2442-46de-924e-c03ecb5e6f18.png">

bjartebore:
Documentation for transient props can be found here:
https://styled-components.com/docs/api#transient-props

oddvernes:
Good catch! I have already gotten rid of some attributes by using transient props here and there. We'll add it here too (there's probably more that should be transient too, its something wee need to look at)

oddvernes:
resolved bu #2764

PR #2764, opened 2023-02-02
Fixed issue #2763 by making the elevation props transient

bjartebore:
The elevation prop of Paper was beeing rendered to the dom, this is not valid html and should be transient


oddvernes (approved):
LGTM!

PR #2776, opened 2023-02-09
🐛 Autocomplete: updated deps & refactor floating-ui refs

### LINKED ISSUE #2773: Make Autocomplete dropdown render in portal
Reported by Vegard Haugland on [slack](https://equinor.slack.com/archives/CJT20H1B9/p1675782785371459).

copied from slack:

We encountered a fun bug in <Autocomplete/> when we upgraded from v0.28.0 to v0.29.1.
Long story short:

- Some (but not all) of our <Autocomplete/> instances didn't render the dropdown values when the onClick event was fired.
- We solved the issue by setting disablePortal to true for the troublesome instances.

Can anyone elaborate on how this solved the issue? Full disclosure: I don't know anything about the downshift library other than it is something that is being used by <Autocomplete/>.

Here the bug is solved (by setting the` disablePortal` attribute to true on the two right-most instances of Autocomplete)

oddvernes:
Does not resolve any issues

I have not been able to recreate this bug (where dropdown sometimes does not render when using portal). So, I have updated downshift, floating-ui and tanstack-virtual and changed to using new ref setters with floating ui.

### DIFF
--- a/packages/eds-core-react/package.json
+++ b/packages/eds-core-react/package.json
@@ -95,9 +95,9 @@
     "@equinor/eds-icons": "workspace:*",
     "@equinor/eds-tokens": "workspace:*",
     "@equinor/eds-utils": "workspace:*",
-    "@floating-ui/react": "^0.16.0",
+    "@floating-ui/react": "^0.19.1",
     "@tanstack/react-virtual": "3.0.0-beta.30",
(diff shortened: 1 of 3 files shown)


oddvernes:
Although it didn't fix the reported issue, I'll merge this pr as it updates dependencies and includes an important refactor.

PR #2786, opened 2023-02-20
Fix icons preview link

jsolaas:


oddvernes (changes requested):
Thank you for the fix! Actually we are in the process of migrating over to a new base url for storybook which is live now: `https://s478stedsstorybookdev.z16.web.core.windows.net/?path=/docs/icons--preview` could you change it to that? (We might as well start here hehe)

jsolaas:
> Thank you for the fix! Actually we are in the process of migrating over to a new base url for storybook which is live now: `https://s478stedsstorybookdev.z16.web.core.windows.net/?path=/docs/icons--preview` could you change it to that? (We might as well start here hehe)

Sure, updated!

oddvernes (approved):
LGTM!

PR #2788, opened 2023-02-21
✨ Autocomplete async

### LINKED ISSUE #2090: In order to use EDS components in my application, as a developer, I need Autocomplete to support async
User requested feature from `Autocomplete` testing. 

A lot of data is fetched from an api based on user input. A loading state for `Autocomplete` when options are loading based on user input would help with this use-case for our users.

Quick thoughts on this feature request:
* Loading state design?
* Is this even doable/good a11y wise with dynamic content?
* Disabled for user interaction component when "loading"?
* Error loading message goes where? Helpertext?

oddvernes:
resolves #2090

Introduces `onInputChange` and `loading` props and a new **Async search autocomplete** story


denektenina (approved):
Looks good, great example in the documentation!

PR #2797, opened 2023-03-02
Started UI for new Accordion inside SideBar

### LINKED ISSUE #2608: Add nested links subcomponent to the SideBar
The amplify sidebar has a flat list of links. We can still create an alternate nested list of collapsible links as a subcomponent of SideBar, as per the original EDS design for sidebar.

denektenina:
Resolves #2608


oddvernes:
Need to update snapshot to pass tests
`pnpm test:update-snapshots:core-react sidebar`

packages/eds-core-react/src/components/SideBar/SideBar.tsx
```diff
@@ -46,7 +46,7 @@ const GridContainer = styled.div<ContainerProps>(({ theme, open }) => {
       'content'
       'footer';
     background-color: ${theme.background};
-    overflow: hidden;
+    overflow: scroll;
```
oddvernes, line 49: This needs to be `overflow-x: auto;` or Sidebar will look like this for 90% of our users:  
bilde
Additionally something is causing overflow in the Y direction, this should also be identified and fixed

packages/eds-core-react/src/components/SideBar/SideBarAccordion/index.tsx
```diff
@@ -0,0 +1,320 @@
...
+import styled, { css } from 'styled-components'
+import {
+  bordersTemplate,
+  outlineTemplate,
+  OverridableComponent,
+  mergeRefs,
+} from '@equinor/eds-utils'
+import { arrow_drop_down, IconData } from '@equinor/eds-icons'
+import { chevron_down, chevron_up } from '@equinor/eds-icons'
+import { useId } from '@equinor/eds-utils'
```
oddvernes, line 18: multiple import statements for same source (did your linter not pick up this?)

packages/eds-core-react/src/components/SideBar/SideBarAccordionItem/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+          background: menuActiveBackground,
+          typography: { color: typographyActiveColor },
+        },
+        hover: { background: menuHoverBackground },
+      },
+    },
+  },
+} = tokens
+
+const Container = styled.div<{ $active?: boolean }>(({ $active }) => {
```
oddvernes, line 26: Should this not be a styled.a?

packages/eds-core-react/src/components/SideBar/SideBarAccordionItem/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+  return css`
+    display: grid;
```
oddvernes, line 28: Need to add some css to reset some potential default styles since it is polymorphic. 
`text-decoration: none;` to prevent this:
bilde
And then `border: 0; width: 100%; padding: 0; ` background-color transparent instead of none to prevent this (as="button"):
bilde

We are also missing the focus outline for eds (copy it from sidebarLink)

packages/eds-core-react/src/components/SideBar/SideBarAccordion/index.tsx
```diff
@@ -0,0 +1,320 @@
...
+        placement={'right-start'}
+        id="menu-default"
+        aria-labelledby="anchor-default"
+        onClose={closeMenu}
+        anchorEl={anchorEl}
+      >
+        {Children.map(children, (child) => {
+          const item = child as ReactElement<AccordionItemProps>
+          return (
+            <Menu.Item onClick={item.props.onClick}>
```
oddvernes, line 310: Need to forward the `as` prop otherwise menu item is always a button when we want it to be an `a` or router `Link` etc

packages/eds-core-react/src/components/SideBar/SideBarAccordion/index.tsx
```diff
@@ -0,0 +1,315 @@
...
+                color={getIconColor()}
+              />
+            </ClosedSidebarButton>
+          </AccordionHeader>
+        </MenuItem>
+      </Tooltip>
+      <Menu
+        open={menuIsOpen}
+        placement={'right-start'}
+        id="menu-default"
```
oddvernes, line 297: This should probably not have an id? otherwise there could be several divs with the same id

oddvernes (approved):
LGTM

PR #2799, opened 2023-03-03
⬆️ Update svgo

### LINKED ISSUE #2771: Update SVGO
This plugin to optimize svg used when creating icons has not been updated for 2 years and has gone through major changes.
We should update this at some point when there are no new icons in the pipeline, as it is likely all svg paths will change sligthy for every icon.
This also requires some rewrites as the api has changed significantly.

[SVGO repo](https://github.com/svg/svgo)

oddvernes:
Resolves #2771


oddvernes:
I configured it so the output is the same as before, the only difference is the lowercase z changed to uppercase Z in svg paths (z is the "closepath" character and can be either uppercase or lowercase. The new version just uses uppercase)

denektenina (approved):

PR #2801, opened 2023-03-06
🐛 DialogHeader: fix css syntax errors

### LINKED ISSUE #2798: Styling not working for Dialog due to typo
**Describe the bug**

Typo in the two following lines in the component DialogHeader (packages\eds-core-react\src\components\Dialog\DialogHeader.tsx)
    justify-content: space-betweene;
    algin-items: center;

Should be:
    justify-content: space-between;
    align-items: center;

Just look at the code

**Expected behavior**
The applied styling to work 

**Specifications**

- Version:
- 0.29.1
- Browser: N/A
- OS: N/A

oddvernes:
resolves #2798 

Not sure if moving `...rest` into the inner div is the right call. The problem the reporter of this bug have is setting style on DialogHeader, but it has a surrounding div to contain both the header and `Divider` to prevent an extra gap due to dialog being a grid. 
Normally we: 
1. set ...rest to the outermost element
2. make sure `style` and `className` goes to the same place

But here we would want style to go to the styled component. So either seperate it from classname or be consistent

### DIFF
--- a/packages/eds-core-react/src/components/Dialog/DialogHeader.tsx
+++ b/packages/eds-core-react/src/components/Dialog/DialogHeader.tsx
@@ -27,8 +27,8 @@ export const DialogHeader = forwardRef<HTMLDivElement, DialogHeaderProps>(
     }
 
     return (
-      <div {...props}>
-        <StyledDialogHeader>{children}</StyledDialogHeader>
+      <div>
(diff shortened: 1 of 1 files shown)


oddvernes:
Update: we decided to not move props due to the potential breaking change for people overwriting styles here with classes or styled components

PR #2814, opened 2023-03-14
:bookmark: Release eds-icons@0.19.1

### LINKED ISSUE #2812: 🔖Release eds-icons@0.19.1
I think we need to release this before next core react due to the updated paths possibly breaking snapshots? (no new icons just updated svg path for every icon, only change is letter z changed from lower caser to upper case)

denektenina:
Resolves #2812


oddvernes (approved):
LGTM

PR #2826, opened 2023-03-22
react-datepicker: added selects for changing view

### LINKED ISSUE #2830: DatePicker custom navigation
The pull request connected to this issue does the following: 

### Description
The DatePicker component has a cumbersome navigation when you have to navigate several years.

- To simplify navigation i've added support for adding a custom header component.
- Forwarded disabled prop to allow disabling of DatePicker
- added min-width 128px to avoid icon overlapping label/text
- added disableAfterDate (considered overriding using a filterDate prop, but kept it in line with existing code instead)
- reworked filterDate to be able to use both disableBeforeDate and disableAfterDate at the same time
- removed the custom label and imported EDS Label
- Added disabled styling
- Forwarded minDate & maxDate

### minDate

When using disableBeforeDate for a future date, minDate seems like the only way to start picking dates in future without setting a date value

Also use this in our project to feed the component:

```
const minDate = disableBeforeDate;
const maxDate = disableAfterDate;
```

Otherwise if you want the user to start picking dates from 2024, they will start on current month with everything disabled. This way the user gets to start at the beginning of the valid time range

### DatePickerSelectHeader
There is an example DatePickerSelectHeadercomponent in this PR that's put in a Story for testing, but I'm not sure if it's wanted as it hasn't gone through the official channels and requires some more work

bovan:
### Description
The DatePicker component has a cumbersome navigation when you have to navigate several years.

- To simplify navigation i've added support for adding a custom header component.
- Forwarded disabled prop to allow disabling of DatePicker
- added min-width 128px to avoid icon overlapping label/text
- added disableAfterDate (considered overriding using a filterDate prop, but kept it in line with existing code instead)
- reworked filterDate to be able to use both disableBeforeDate and disableAfterDate at the same time
- removed the custom label and imported EDS Label
- Added disabled styling
- Forwarded minDate & maxDate

### minDate

When using disableBeforeDate for a future date, minDate seems like the only way to start picking dates in future without setting a date value

Also use this in our project to feed the component:

```
const minDate = disableBeforeDate;
const maxDate = disableAfterDate;
```

Otherwise if you want the user to start picking dates from 2024, they will start on current month with everything disabled. This way the user gets to start at the beginning of the valid time range

### DatePickerSelectHeader
There is an example DatePickerSelectHeadercomponent in this PR that's put in a Story for testing, but I'm not sure if it's wanted as it hasn't gone through the official channels and requires some more work

Resolves #2830

### DIFF
--- a/packages/eds-lab-react/src/components/DatePicker/SelectHeader.tsx
+++ b/packages/eds-lab-react/src/components/DatePicker/SelectHeader.tsx
@@ -1,18 +1,13 @@
 import styled from 'styled-components'
-import { getMonth, getYear } from 'date-fns'
 import { datePicker as tokens } from './DatePicker.tokens'
 import { arrow_back, arrow_forward } from '@equinor/eds-icons'
-import { typographyTemplate } from '@equinor/eds-utils'
-import { Typography, Icon, Button } from '@equinor/eds-core-react'
+import { Icon, Button, NativeSelect } from '@equinor/eds-core-react'
 import { ReactDatePickerCustomHeaderProps } from 'react-datepicker'
 
-type PopupHeaderProps = ReactDatePickerCustomHeaderProps & {
-  changeDate?: (date: Date) => void
-}
-
-const PopupHeader: React.FC<PopupHeaderProps> = ({
+const SelectHeader: React.FC<ReactDatePickerCustomHeaderProps> = ({
   date,
-  changeDate,
+  changeYear,
+  changeMonth,
   decreaseMonth,
   increaseMonth,
   prevMonthButtonDisabled,
@@ -33,6 +28,11 @@ const PopupHeader: React.FC<PopupHeaderProps> = ({
     'December',
   ]
 
+  const years: number[] = []
+  for (let i = 1970; i < 2060; i++) {
+    years.push(i)
+  }
+
   return (
     <Header>
       <IconButton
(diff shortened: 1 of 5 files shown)


packages/eds-lab-react/src/components/DatePicker/SelectHeader.tsx
```diff
@@ -0,0 +1,136 @@
+import styled from 'styled-components'
+import { datePicker as tokens } from './DatePicker.tokens'
+import { arrow_back, arrow_forward } from '@equinor/eds-icons'
+import { Icon, Button, NativeSelect } from '@equinor/eds-core-react'
+import { ReactDatePickerCustomHeaderProps } from 'react-datepicker'
+
+const SelectHeader: React.FC<ReactDatePickerCustomHeaderProps> = ({
```
oddvernes, line 7: I am thinking maybe this could have a bit more descriptive name since it is something that can be imported tied to the datePicker. "DatePickerSelectHeader" maybe?

bovan: if it's wanted as part of this component yes.

but keep in mind I just added this as a test example, as f.ex. "September" becomes too big of a word to fit in the select, and decreasing the font-size and margins to make it fit will probably make it non-WCAG-friendly :)
A more detailed look at it would be required to see how we can fit in all the elements, rather than me just hacking away at it. I'm not using this in our project as I've created my own custom version with specific features for us

the changes in DatePicker.tsx & Token is the important ones that enables a more flexible usage of the component

bovan:
> BTW if you add
> 
> ```
> source: {
>         type: 'code',
>       },
> ```
> 
> Inside parameters.docs in the export detfault at around line 242 in Datepicker.stories.tsx like so: bilde
> 
> The show code button will start working properly because I see this is broken at the moment. And also in the custom header story, a little text description about the custom header (how to import the one you made and perhaps a link to react-datepicker documentation about this (https://reactdatepicker.com/#example-custom-header), so it is easier to understand would be great 👍

I've added this, but the storybook in general for this part (date picker or just labs?) are broken, and I don't have time available to go too much outside of the scope of fixing a broken component :)
I'd rather just rip out the Select component and leave its usage undocumented than spend lots of time with Storybook

oddvernes (approved):
LGTM!

PR #2832, opened 2023-03-28
DatePicker - icon is selectable and has hover effect

bovan:
### Why

In previous PR I made DatePicker wasn't clickable anymore as it was no longer child of `<Label />`, it seems like it should be.

### How

Added button wrapper around SVG to get hover effect and added an onClick to open calendar

Expanded button to 32px to get a larger "hit zone" without making the icon larger. 
The icon looks a bit shifted upwards, but I think that's because of the shape of the icon


oddvernes:
Good catch! But I think this can be solved much simpler by just adding `pointer-events: none` to the icon, which is purely decorative anyway. Clicking anywhere on the input should open the datepicker regardless. I can fix this while doing dependency upgrades (which is coming along but I am having some issue with typescript in downstream testing at the moment) 🙂

oddvernes:
Closing this as it has been solved seperately

Issue #2839, opened 2023-03-30
css variables for typography colors


bjartebore:
I would like to globally override typography colors

I'm getting really frustrated to the point where I get no sleep by the fact that the colors of the typography are not possible to globally override

I would like the typography to use css variables with a fallback to the current color

writing a lot of boilerplate code

PR #2840, opened 2023-03-30
Added css variables for typography colors

bjartebore:
Related issue #2839


oddvernes:
This was certainly missing 😅! I will test that it works and then we'll have to make a new release of eds-tokens.

Be aware though, that the process that will _eventually_ lead to this "figma-broker" (along with the current eds-tokens) being deprecated has started, **but** it is likely to take a good while to figure out and implement the new token system. That will also cause the naming of css-variables coming from tokens to change, so will be a large breaking change when that time comes for people using these 😬.

oddvernes (approved):
LGTM!

PR #2852, opened 2023-04-18
✨ Autocomplete optionComponent

### LINKED ISSUE #2487: Customise individual items in <Autocomplete>

When using the `Autocomplete` selector / dropdown, I'd like to be able to customise the different items.
In my use case, I have a list of projects, some of which the user have read access to, while other they can write to.
I'd like to convey this information with an icon, and tooltip (or different) for each element in the list.

I'd like write something like this

```typescript tsx
<Autocomplete
    multiple
>
    {projects.map((project: string) => <Autocomplete.Item
        text={project}
    >
        <Icon variant={useHasAccess(project, user) ? 'writable' : 'readable' />
<Autocomplete.Item />
<Autocomplete />
```

It could be possible to simulate this behaviour with `useRef`, and some CSS, or simply give an indication after the user has made their selection.

**Additional context**

oddvernes:
resolves #2487 

`optionComponent` allows users to provide their own template `(option: T) => ReactNode`
I made a new story and updated the async story with flag pictures in the options

### DIFF
--- a/packages/eds-core-react/src/components/Autocomplete/Autocomplete.docs.mdx
+++ b/packages/eds-core-react/src/components/Autocomplete/Autocomplete.docs.mdx
@@ -67,6 +67,11 @@ how the label should be rendered and searched for. This function must return a s
 
 <Story id="inputs-autocomplete--option-label" />
 
+### Customized option component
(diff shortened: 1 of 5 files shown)


denektenina (approved):
LGTM

PR #2863, opened 2023-04-26
💄Sidebar: Remove visible scrollbar when closed

### LINKED ISSUE #2853: Handle overloaded Sidebar
When the Sidebar has so many items that it needs a vertical scroll, the closed view (when only the icon is showing) also gets a horizontal scrollbar. This is because the vertical scrollbar steals space that the content of the sidebar was using, so horizontal scroll becomes necessary to show the whole content. We should get rid of the horizontal scrollbar somehow. Ideally without making the component less accessible or the styling more "hacky"

Image

oddvernes:
resolves #2853 

I propose we visibly hide the scrollbar when sidebar is collapsed. The scrolling is still accessible with mouse scrollwheel or dragging finger on touch screen or tabbing through items on keyboard. this is the solution they have in the slack sidebar (where you click between organizations). Also considering this is a pretty unlikely edge case 🤷‍♂️


denektenina (approved):
Seems like a good solution

PR #2866, opened 2023-05-05
⬆️ Storybook v7

### LINKED ISSUE #2514: ⬆️ Upgrade storybook 7.0 (previously mdx 2)
MDX has been updated to 2.0 with big changes: https://mdxjs.com/blog/v2/

Storybook will eventually move to mdx-2, migration guide for experimental support: https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#opt-in-mdx2-support

oddvernes:
resolves #2514 

I had to remove the cjs output from the build and use "exports" field (+ type: "module") in package.json (instead of main and module) to make this work. if this is an issue we can possibly as an alternative keep cjs and change file extensions on the module outputs to .mjs

I switched to vite from webpack5 as the storybook builder as a result of the problems above (thats when the error message got informative enough to realize what was wrong) 

I used automigrate to csf3 in labs but core-react is still csf2. It is a lot of work currently to fix the csf3 autogenerated code (the render function in StoryObj is an arrow function not recognized as a react function so hooks in the story make it fail. Have to move all functions out of object, or change from arrow to named function)


nilsml (changes requested):
Truly a big PR!!
I tried starting the Storybook both on my regular Codespace and also on a brand new Codespace set up only for this PR. I was unable to build and start the Storybook.
Failing with messages like:

```
ERR! Error: Failed to resolve entry for package "@equinor/eds-tokens". The package may have incorrect main/module/exports specified in its package.json.
ERR!     at packageEntryFailure (file:///workspaces/design-system/packages/eds-core-react/node_modules/.pnpm/vite@4.3.5/node_modules/vite/dist/node/chunks/dep-934dbc7c.js:23384:11)
```

PR #2891, opened 2023-06-02
📝 Storybook: Add null as a possible value for useState hooks that receive HTML elements to avoid type error

FredrikMWold:
Adds null as possible type for anchor Elements in storybooks documentation.
This does not produce and error in the eds project itself as typescript is not set to strict, but does produce an error in most projects.

The default for most popular react bootstrapping tools like: create-react-app, create-vite and create-next-app default to typescript in strict mode.


oddvernes (approved):
LGTM! We should look into turning on ts strict mode as well so we are more aware of these things

Issue #2895, opened 2023-06-05
Implement data-grid enhancements


vnys:
## Features

These are the prelimenary features @yusijs has in mind:

- [x] Pagination
- [x] Sorting
- [x] Filtering
- [x] Grouping (mostly done)
- [x] Virtualization

And then at a later stage followed by

- [ ] Row expansion (tre-struktur)
- [ ] Column-sorting
 
## Documentation

I suggest adding a separate Storybook local to the component, and then we’ll include that into the main storybook. I think that’s the cleanest way to do it since the current storybook lives inside the EDS Core React package. Initially we had the storybook in a separate folder outside packages, but since that’s not how it’s intended to use, it caused all kinds of problems.

PR #2931, opened 2023-06-28
Implement eds-data-grid package. Completes #2895

yusijs:


oddvernes (dismissed):
If you could rebase on develop and then add the various scripts (lint, build etc) to the main package.json in root and make sure they work. I just ran lint which returned 31 errors, 38 warnings. Currently the checks don't run in the CI but i've added this package to the workflow now. And make sure to run `pnpm run init` after rebase as there is a bunch of dependency updates 👍

oddvernes (approved):
While there is probably some more work to be done, and nits that can be picked, this pr is mainly about getting everything up and running. All checks are green and the package builds and runs downstream as I have tested in a create-react-app. LGTM! 👍

Issue #2930, opened 2023-06-27
Typography: add conditional types to props


oddvernes:
variants not in the groups `heading` and `paragraph ` requires the correct `group` prop which is confusing to users. It is possible to set up conditional types so that typescript catches it

oddvernes:
The first attempt at this failed unfortunately (https://github.com/equinor/design-system/pull/2932) but lets keep this issue in the backlog. It would be a great benefit to get this to work

vnys:
Worth looking into, or better to replace the Typography component?

PR #2932, opened 2023-06-28
Refactored Typograhy to have conditional types for group and variants

bjartebore:
This will ensure that people only can select the correct variant within the group selected

#2930


bjartebore:
This does not work due to forwardRef and I see no way to fix it without removing forwardRef

PR #2950, opened 2023-07-06
♻️🚸Dialog: refactor to use native dialog

### LINKED ISSUE #2949: ♻️🚸Dialog: rewrite to use native dialog
We have a lot of issues with dialog (z-index, focus trapping, autofocus etc) so I am rewriting and trying out using native `<dialog>` element instead of Scrim/Floating-ui. It looks like i will be able to reproduce the old behaviour and api without any breaking changes

oddvernes:
resolves #2949 

removes `Scrim` and `floating-ui` as dependencies and instead uses a native `<dialog>` element in modal mode (uses `DialogElement.showModal()`)
I kept the functionality the same as the old Dialog to not introduce breaking changes (which means i did block pressing escape to close unless `isDismissable` is true, maybe something to discuss


oddvernes:
Note to self: check dialog inside dialog case and hidebodyscroll

PR #2995, opened 2023-08-09
Chips should only have hover on onClick events

### LINKED ISSUE #2994: Chips has a hover effect when no click action exists
**Describe the bug**

When there is no onClick event on the Chip there is still is a hover effect indicating there should/could be an action

1. Add a Chip without onClick behaviour
2. Hover the mouse over the element
3. nod

**Expected behavior**
Nothing should happen on hover when no click behaviour exists

bjartebore:
resolves #2994

### DIFF
--- a/packages/eds-core-react/src/components/Chip/Chip.tsx
+++ b/packages/eds-core-react/src/components/Chip/Chip.tsx
@@ -45,18 +45,6 @@ const StyledChips = styled.div.attrs<StyleProps>(
     fill: ${typography.color};
   }
 
-  ${({ clickable }) => clickable && css`
-    @media (hover: hover) and (pointer: fine) {
-      &:hover {
-        color: ${states.hover.typography.color};
-
-        svg {
-          fill: ${states.hover.typography.color};
-        }
-      }
-    }
-  `}
-  
   &:focus {
     outline: none;
   }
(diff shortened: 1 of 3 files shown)


packages/eds-core-react/src/components/Chip/Chip.tsx
```diff
@@ -45,16 +45,18 @@ const StyledChips = styled.div.attrs<StyleProps>(
     fill: ${typography.color};
   }
 
-  @media (hover: hover) and (pointer: fine) {
-    &:hover {
-      color: ${states.hover.typography.color};
+  ${({ clickable }) => clickable && css`
```
oddvernes, line 48: I see there is already a clickable check a little further down on line 70-something that only adds corsor, would it be possible to add the color and fill there instead so its in one place? also the error state has its own hover color even further down that is not checked for clickable

oddvernes (changes requested):

packages/eds-core-react/src/components/Chip/Chip.tsx
bjartebore: ofc, and I am a little embarrassed that I did not see it myself

oddvernes (approved):
LGTM

PR #2997, opened 2023-08-10
Added script for expanding variant modes

### LINKED ISSUE #2989: Expand variant modes in code
We use modes for variants in Figma. What that means is that in code we have tokens such as these, taken from the file `semantic.tokens.json5` in `design-tokens/src/color`:

```json5
{
  icon: {
    input: {
      '<notification>-focused': {
        $type: 'color',
        $value: '{notification.em}'
      }
    }
  }
}
```

This means that `notification` in this case is to be considered a variable for the notification modes you’ll find in `design-tokens/src/color/modes`:

- notification.error.tokens.json5
- notification.info.tokens.json5
- notification.success.tokens.json5
- notification.warning.tokens.json5

The same method is used for button variants. There are no other similar modes at this point. So _before_ we transform the tokens with style dictionary, we need an intermediate step where we merge the variant modes into the semantic tokens file. If we look at the content of `notification.error.tokens.json5`, we see that `notification.em` resolves to `error.em`, so what we want then is to transform the first json block into this one:

```json5
{
  icon: {
    input: {
      'error-focused': {
        $type: 'color',
        $value: '{error.em}'
      },
      'info-focused': {
        $type: 'color',
        $value: '{info.em}'
      },
      'success-focused': {
        $type: 'color',
        $value: '{success.em}'
      },
      'warning-focused': {
        $type: 'color',
        $value: '{warning.em}'
      }
    }
  }
}
```

denektenina:
resolves #2989

### DIFF
--- a/design-tokens/merge.js
+++ b/design-tokens/merge.js
@@ -1,4 +1,5 @@
-const fs = require('fs').promises
+const fsPromises = require('fs').promises
+const fs = require('fs')
 const JSON5 = require('json5')
 
 async function execute() {
@@ -40,7 +41,7 @@ async function execute() {
           })
         } else {
           console.error(
-            `MERGE ERROR: No available files for mode category ${modeCategory}`,
+            `🧐 No available mode-files for mode-category '${modeCategory}' - this tree was not modified`,
           )
           res[key] = value // TODO consider if this should be handled differently
         }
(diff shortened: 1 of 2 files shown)


oddvernes (approved):
As discussed we can merge this into the other branch first and let Victor have a look over before merging that into develop

PR #3036, opened 2023-08-28
✨Tabs: controlled mode

### LINKED ISSUE #3022: <Tabs.Tab active> attribute not respected
**Describe the bug**

According to the documentation, the `active` attribute on the `<Tabs.Tab />` component can be set to `true` to make that tab the active one. This does not work, though. The `active` attribute is not respected in any way.

```` 
<Tabs>
  <Tabs.List>
   <Tabs.Tab>First tab</Tabs.Tab>
    <Tabs.Tab active={true}>Second tab</Tabs.Tab>
  </Tabs.List>
</Tabs>
```` 
**Expected behavior**

In the example, the second tab should be marked as 'active' by having a highlighted underline.
Instead, the first tab is marked as active.

**Specifications**

- Version:  "@equinor/eds-core-react" @ "0.32.4",
- Browser:  Microsoft Edge Versjon 115.0.1901.203 (Offisiell build) (64-biters)
- OS: Microsoft Windows 10 Enterprise

**Additional context**

Additional discussion on slack here: https://equinor.slack.com/archives/CJT20H1B9/p1692611170152419

Using `activeTab` and `onChange` on the `Tabs` component is an alternative, but requires extra logic for mapping to indexes and detecting url changes. When using in combination with routing, it's easier/better/more robust to have the logic on each tab instead since you then no longer need to relate to any positional index.

oddvernes:
resolves #3022 

The feature request was to fix the `active` prop on `<Tabs.Tab>` so it could be used to manually set active tab when using tabs as router links, however this proved difficult due to how the internal logic controls the active state. So instead I opted to implement it the way MUI does it by having a `value` prop on `<Tabs.Tab>` that sets the active tab if it matches the value sent into `activeTab`.
I also created a storybook story with a practical example using `react-router-dom`

### DIFF
--- a/packages/eds-core-react/package.json
+++ b/packages/eds-core-react/package.json
@@ -76,6 +76,7 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-hook-form": "^7.45.0",
+    "react-router-dom": "^6.15.0",
     "remark-gfm": "^3.0.1",
     "rollup": "^3.25.3",
     "rollup-plugin-delete": "^2.0.0",
--- a/packages/eds-core-react/pnpm-lock.yaml
+++ b/packages/eds-core-react/pnpm-lock.yaml
@@ -133,6 +133,9 @@ devDependencies:
   react-hook-form:
     specifier: ^7.45.0
     version: 7.45.0(react@18.2.0)
+  react-router-dom:
+    specifier: ^6.15.0
+    version: 6.15.0(react-dom@18.2.0)(react@18.2.0)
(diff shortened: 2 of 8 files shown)


vegar:
It's an improvement for sure. 

I see that the `active` property still exist on `Tab` component. What stops the next person for trying to use it the way I did?
Should it be removed? Is it possibel to remove it?

oddvernes:
I put an `/** @ignore */` on it so at least it doesnt show up in the documentation. It is used internally, controlled by TabList

PR #3050, opened 2023-09-13
⬆️ Styled components v6

### LINKED ISSUE #2884: ⬆️ Update styled-components to v6
Another major update is around the corner as `styled-components` is currently making release candidates for v6.0.0. It may be the reason for this extremely strange bug reported in #2879. Something to be weary of when upgrading, in addition to other breaking changes.

We should wait until after they are out of rc and released 6.x.x officially before doing this

oddvernes:
Resolves #2884 

This was a big one. 
~~styled changed from default to named import~~ edit: this was not compatible with v5, so changing back and turning off the warning instead
They added types internally, deprecating `@types/styled-components`, but the types are not 1:1 and a bunch of types are missing. At least some of these will be added back in the next release of styled components though 🤷‍♂️.
Had to refactor how types are assigned to props on some components
Automatic filtering of props has been removed, so now every prop used in a styled component must be transient (have `$` added to the front of it) otherwise it gets forwarded to the DOM as invalid props. I had to go through every component and detect all the props that needed to be transient.
All snapshots have changed since sc no longer adds vendor prefixes + some other changes due to new version of stylis formats css differently
Had to change many other tests too due to stylis not trimming spaces any more


denektenina (approved):
LGTM

PR #3066, opened 2023-09-21
fix(sidebar): add aria-label to toggle button

ingeridhellen:
The toggle button causes an accessibility warning for missing text label. Also, without the label the button is hard to find in playwright tests and the like.


oddvernes (approved):
Good catch!

ingeridhellen:
@oddvernes When do you plan to release the package?

PR #3069, opened 2023-09-25
🐛 Autocomplete fix crash caused by merging refs

### LINKED ISSUE #3059: Error when using Autocomplete with multiple enabled
**Describe the bug**

We have seen an issue with Autocomplete when the multiple property is enabled. When selecting an option in the Autocomplete with the Enter-button and then immediately removing the alternative with the clear button by using the mouse, we get the error: "Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops."

This happens when using basic properties of the Autocomplete. We also get the error "Error: Minified React error 185" when performing the same steps in the first example called "Multiple" in the EDS storybook. https://storybook.eds.equinor.com/?path=/story/inputs-autocomplete--multiple

1. Create an Autocomplete component and enable the multiple property such that you are able select multiple options from the list of alternatives. For example use the Autocomplete component in the EDS storybook example "Multiple"
2. Move to an option in the list with the up/down keys and hover your mouse over where you expect the clear button to be when you have selected an option.
3. Then select an option with the enter-key on the keyboard. After the option is selected, quickly remove the option selected by clicking the clear button "x". Then an error shows with "Maximum update depth exceeded"/ "Error: Minified React error 185".

**Expected behavior**

We expected the Autocomplete-component to remove the selections without giving an error. The component does not give the same error when multiple is disabled.

**Specifications**

- Browser: Edge
- OS: Windows 10
- Version eds-core-react: 0.31.1

**Additional context**

Error when running fusion application locally:
image

Error when producing the bug in EDS storybook example:
image

oddvernes:
resolves #3059 and hopefully all the other ways of triggering this infinite loop

### DIFF
--- a/packages/eds-core-react/package.json
+++ b/packages/eds-core-react/package.json
@@ -97,7 +97,7 @@
     "@equinor/eds-utils": "workspace:*",
     "@floating-ui/react": "^0.25.1",
     "@tanstack/react-virtual": "3.0.0-beta.54",
-    "downshift": "^8.2.0"
+    "downshift": "^8.2.1"
   },
   "engines": {
     "pnpm": ">=4",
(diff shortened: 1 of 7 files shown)


denektenina (approved):
lgtm!

PR #3139, opened 2023-11-02
✨Added variant and helper-text to Autocomplete

### LINKED ISSUE #2836: autocomplete: invalid state
Autocomplete should have a visual invalid state in line with other form elements (in the case of `required` for example)

Currently this does not exist in figma either

denektenina:
resolves #2836


denektenina:
The reason we are not using the InputWrapper here is because Autocomplete and InputWrapper both have their own theme provider, which collides with each other. This should not be a problem once the new token system is in place, so it should be fixed then. Now we are instead using HelperText from InputWrapper to get the same styling.

oddvernes (approved):
LGTM

PR #3140, opened 2023-11-06
feat: allow React Node type for label props to allow it to be more customizable

### LINKED ISSUE #3141: ✨Label: allow html children
Requested by @FredrikMWold
Change Label prop type from String to ReactNode to allow html as children for label. His particular use case was to add a red "required star" (which will be added natively with eds 2.0), but there could be other use cases as well such as highligthing specific words etc.

FredrikMWold:
This allow reactNode to be assigned to the label prop for the Label, TextField and Autocomplete component. This makes it easier to customize the label for a field.

An example of a use case this solves is adding a "required star" to a field in a form

resolves #3141


oddvernes (approved):
LGTM!

PR #3142, opened 2023-11-08
✨ Breadcrumb: custom separator

### LINKED ISSUE #3001: Custom separators for Breadcrumbs
Especially when using Breadcrumbs for routing (e. g., in a header navbar), it is intuitive to display a ">" as a separator to indicate consecutive routes. 

Would be nice to have something along the lines of [Material-UIs solution](https://mui.com/material-ui/react-breadcrumbs/#custom-separator).

oddvernes:
resolves #3001 

the regular separator have an inline padding of 16px. If the seperator contains an icon, It gets an inline margin of -8px to make the spacing roughly similar. 
bilde
   

pending design review/figma update


oddvernes:
Agreed with @BirteThornquist to set `inline-margin` of icon to -9px to most closely match spacing as designed (when a character is used)

PR #3161, opened 2023-11-28
🧑‍💻SideSheet: open required, conditional render header, title, button

### LINKED ISSUE #3157: Optional props on Sidesheet component not really optional

Currently all props on the Sidesheet component are set as optional. But in reality they are not really optional.

I would want the props to be true optional. If i don't want the title, and don't send it in I don't want the h2 element to render.
If I don't send a close function in I don't want the close button to render. And I don't understand why the open prop is a optional parameter as well. 

In short I want the props that are optional to have conditional rendering, and the ones that isn't optional to be set as that.

Currently I am making a wrapper with styled components to set display none on the elements that I don't want to show. And it feels unecessary. Which I have filled with this..
`
    #discussion > div:first-child {
        display: none;
    }
`

"discussion" is the id i have sent to ID prop, which probably shouldn't be a optional prop either.

oddvernes:
resolves #3157 

changes:

- made `open` a required prop
- conditionally render `header` if `title` and/or `onClose` exists
- conditionally render `title`
- conditionally render `close button` if `onClose` is defined

### DIFF
--- a/packages/eds-core-react/src/components/SideSheet/SideSheet.test.tsx
+++ b/packages/eds-core-react/src/components/SideSheet/SideSheet.test.tsx
@@ -11,28 +11,32 @@ const StyledSidesheet = styled(SideSheet)`
 
 describe('SideSheet', () => {
   it('Matches snapshot', () => {
-    const { asFragment } = render(<SideSheet variant="large" title="Title" />)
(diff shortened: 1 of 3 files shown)


AgeAKristiansen:
Thank you!

Issue #3070, opened 2023-09-25
🚸 Autocomplete: improve ux of singleselect


oddvernes:
[Discussion on slack](https://equinor.slack.com/archives/CJT20H1B9/p1695366581731959). There is no visual distinction between selected value and search string in the input field, which can lead to people thinking they have selected a value when the real selected value is something completely different. Vegar Vikan suggested the dropdown is always open with a "no match found" text when the text is search input. I then suggested the text should revert to selected value on blur as well

PR #3208, opened 2024-01-05
feat: clear input field on blur if the user has not selected any option

FredrikMWold:


oddvernes:
Hello and sorry for the late reply! You are on the right track here but you are using onBlur on the element instead of downshift's stateReducer which is more proper here I think. Then it would be 

```
case useCombobox.stateChangeTypes.InputBlur:
            return {
              ...changes,
              inputValue: changes.selectedItem
                ? getLabel(changes.selectedItem)
                : '',
            }
```

However while testing this I came across some pre-existing bugs that needed to be fixed at the same time, plus we also needed to add a new "no options" ui-element (see original issue #3070) requiring some designer input, so I decided to handle this task myself. I will however use your test though!

PR #3231, opened 2024-01-23
:sparkles: Improve data grid

### LINKED ISSUE #3234: ✨ Data-grid improvements
Copied from the pull request:

- ✨ Reexport `@tanstack/react-table` types to ease typing in apps using the data grid
- 📌 Move eds-core-react to peer dependencies
  - This is neccessary because EDS uses React Context and the grid and the project should
have the same React instance running. This makes it possible to set EDS Density of the
table above the component.
- ♻️ Move text truncating into default cell to enable overwriting cell content
  - This enables custom cells like popover, autocomplete or other cells that overflows the cell itself.
- 🐛 Inherit row background color for pinned cells
  - This ensures hover color on the whole row when columns are pinned
- 🐛 Support 100% width
  - Support string `width` and `height`
- ✨ Allow setting minWidth of table
- ✨ Expose getRowId callback from react-table
- ✨ Expose virtualizer ref
  - This is needed to be able to run "scroll to" functionality in apps.
- 🐛 Hide virtualizer rows top and bottom rows when not needed

In order to implement this in our project, we need to be able to use custom cells, set 100% width and use the EdsProvider with the table. Also, there are some minor bugs discovered during implementation.

These improvements were carried out by @magnh

magnh:
resolves #3234 

This pull request introduces the following improvements to the data-grid:

Most of these changes have been discussed with @yusijs.

- ✨ Reexport `@tanstack/react-table` types to ease typing in apps using the data grid
- 📌 Move eds-core-react to peer dependencies
  - This is neccessary because EDS uses React Context and the grid and the project should
have the same React instance running. This makes it possible to set EDS Density of the
table above the component.
- ♻️ Move text truncating into default cell to enable overwriting cell content
  - This enables custom cells like popover, autocomplete or other cells that overflows the cell itself.
- 🐛 Inherit row background color for pinned cells
  - This ensures hover color on the whole row when columns are pinned
- 🐛 Support 100% width
  - Support string `width` and `height`
- ✨ Allow setting minWidth of table
- ✨ Expose getRowId callback from react-table
- ✨ Expose virtualizer ref
  - This is needed to be able to run "scroll to" functionality in apps.
- 🐛 Hide virtualizer rows top and bottom rows when not needed

In order to implement this in our project, we need to be able to use custom cells, set 100% width and use the EdsProvider with the table. Also, there are some minor bugs discovered during implementation.

### DIFF
--- a/packages/eds-data-grid-react/src/EdsDataGrid.tsx
+++ b/packages/eds-data-grid-react/src/EdsDataGrid.tsx
@@ -323,7 +323,7 @@ export function EdsDataGrid<T>({
           ...parentRefStyle,
           width: scrollbarHorizontal ? width ?? '100%' : 'auto',
           overflow: 'auto',
-          contain: 'strict',
+          contain: 'layout paint style',
         }}
         ref={parentRef}
       >
--- a/packages/eds-data-grid-react/src/EdsDataGrid.tsx
+++ b/packages/eds-data-grid-react/src/EdsDataGrid.tsx
@@ -65,6 +65,7 @@ export function EdsDataGrid<T>({
   columnPinState,
   scrollbarHorizontal,
   width,
+  minWidth,
   height,
 }: EdsDataGridProps<T>) {
   const [sorting, setSorting] = useState<SortingState>(sortingState ?? [])
@@ -331,6 +332,7 @@ export function EdsDataGrid<T>({
             style: {
               tableLayout: scrollbarHorizontal ? 'fixed' : 'auto',
               width: table.getTotalSize(),
+              minWidth: scrollbarHorizontal ? width : 'auto',
             },
           }}
         >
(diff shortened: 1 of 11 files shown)


oddvernes:
> I cannot edit reviewers, but I would like a review from @yusijs 😄
> 
> It seems like the checks are failing due to a missing secret, but I'm not sure whether it's on my side or yours?

 I have added @yusijs as a reviewer 🙂 
The checks failing from forked pull requests are on our end. We need to fix the github action, but in the meantime we  can run the tests locally 👍

packages/eds-data-grid-react/src/EdsDataGrid.tsx
```diff
@@ -298,9 +321,9 @@ export function EdsDataGrid<T>({
         style={{
           height: height ?? 'auto',
           ...parentRefStyle,
-          width: scrollbarHorizontal ? width : 'auto',
-          tableLayout: scrollbarHorizontal ? 'fixed' : 'auto',
+          width: scrollbarHorizontal ? width ?? '100%' : 'auto',
           overflow: 'auto',
+          contain: 'strict',
```
yusijs, line 326: Change this to `contain: layout paint style` to ensure auto height works

yusijs (changes requested):

oddvernes:
✅ Lint
✅ Tests
✅ Types

oddvernes:
lint/test/types all seems to be in order still 👍

yusijs (approved):
LGTM!

oddvernes (approved):
LGTM!

oddvernes:
Yes I will merge 👍 Was just trying the build downstream in an app for the final test  but looks good

PR #3268, opened 2024-02-12
Data-grid: Allow controlling size

yusijs:
Closes https://github.com/equinor/design-system/issues/3269

As it is implemented today, the column size is only controlled within the datagrid. This should be possible to control externally from the consuming application, to allow preserving the previous settings.


oddvernes (approved):
LGTM

PR #3273, opened 2024-02-13
🐛 Enable contain strict when width and height is provided

magnh:
Enables 100% width without requiring any custom code

@yusijs


yusijs (approved):
LGTM!

PR #3339, opened 2024-03-06
Feature/thalse 3052 update pnpm

torleifhalseth:
related issue #3339


package.json
```diff
@@ -10,7 +10,7 @@
     "lint:data-grid-react": "pnpm run lint ./packages/eds-data-grid-react",
     "lint:lab": "pnpm run lint ./packages/eds-lab-react",
     "lint:utils": "pnpm run lint ./packages/eds-utils",
-    "init": "pnpm multi install --force && pnpm run build",
```
oddvernes, line 13: I think maybe we should keep --force, from the documentation:
_Force reinstall dependencies: refetch packages modified in store, recreate a lockfile and/or modules directory created by a non-compatible version of pnpm. Install all optionalDependencies even they don't satisfy the current environment(cpu, os, arch)._
So this should be similar to deleting node_modules

oddvernes:
This works on windows with latest pnpm 👍 (which also cut build time in half on my slow machine from almost 5 minutes to 2,5 minutes)
I will update github actions to use  pnpm@"<9.0.0" in develop so we can test if it works in gh actions

oddvernes (approved):
LGTM

package.json
torleifhalseth: I am not sure how cache is preserved in the environments running our GitHub actions, but it looks like we could setup the cache of the store using config from [this example](https://pnpm.io/continuous-integration#github-actions). 

Not sure if this is related, but I was hoping --force would not be needed.

torleifhalseth: It looks like the example creates a shared cache "pnpm" and then I guess it would be listed in [our caches.](https://github.com/equinor/design-system/actions/caches)

PR #3355, opened 2024-03-19
Allow enabling/disabling multi row selection in eds-data-grid

mhwaage:
Forwards multiRowSelection options to tanstack, allowing users to enable/disable multi row selection.


yusijs (approved):
LGTM!

packages/eds-data-grid-react/src/EdsDataGrid.tsx
```diff
@@ -227,6 +228,7 @@ export function EdsDataGrid<T>({
     debugHeaders: debug,
     debugColumns: debug,
     enableRowSelection: rowSelection ?? false,
+    enableMultiRowSelection: multiRowSelection,
```
torleifhalseth, line 231: I would recommend using the same property name as the library.

yusijs: @torleifhalseth The `enable` prefix was added to make it clearer what the prop does. I'd argue that prefixing `enableMultiRowSelection` as well makes sense, given this.

oddvernes: have we concluded on this? Is there any further work needed or can i merge this?

yusijs: Imo we can leave it as is, since it matches the pattern for enableRowselection already in place.

PR #3359, opened 2024-03-20
Datagrid: Add possibility for custom filters

### LINKED ISSUE #3353: Allow more custom filtering
In my project, I would like to allow users to use filters to search for files. Specifically, I would like to filter based on:
File names - each file name is unique, so the default filter that adds a multiselect is not helpful, instead I would want a simple search field.
Dates - again the default filter for strings is a multi-select which is unhelpful, instead (e.g.) a datepicker would be handy.

The ability to pass in my own components for defining a filter function. Something along the line of (rough draft):

```javascript
function MyFilter() {
  const {setColumnFilterFn} = useColumnContext(); // <- Hypothetical column context for enabling the custom component to set/update the filter function, default filter undefined
  
  return (
    <input
      onChange={(e) => setColumnFilterFn((s) => s.includes(e.target.value))}
     />
  )
}
```

i.e. the custom component can update the column filter function via a context hook, and returns a component that will be displayed in the dialog

* Developers add custom filters outside of the data table. This is a bit inelegant for filters that are only intended for the data table.
* EDS adds more filters/explicit 'options' for which filter to use. This will likely become too bloated, although even with the current suggestion it would make sense to expose the 'default' functions (e.g. if users want to coerce a filter)

yusijs:
Closes #3353

TODO:

- [x] Tests
- [x] Examples
- [x] Documentation in storybook

### DIFF
--- a/packages/eds-data-grid-react/src/EdsDataGrid.stories.tsx
+++ b/packages/eds-data-grid-react/src/EdsDataGrid.stories.tsx
@@ -4,6 +4,7 @@ import {
   columns,
   expandColumns,
   groupedColumns,
+  helper,
   Photo,
   PostComment,
 } from './stories/columns'
@@ -30,6 +31,7 @@ import { tokens } from '@equinor/eds-tokens'
 import { action } from '@storybook/addon-actions'
 import { EdsDataGridProps } from './EdsDataGridProps'
 import { Virtualizer } from './types'
+import { FilterWrapper } from './components/FilterWrapper'
 
 const meta: Meta<typeof EdsDataGrid<Photo>> = {
   title: 'EDS Data grid',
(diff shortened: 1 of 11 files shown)


packages/eds-data-grid-react/src/components/TableHeaderCell.tsx
```diff
@@ -137,11 +137,12 @@ export function TableHeaderCell<T>({ header, columnResizeMode }: Props<T>) {
           desc: <Icon data={arrow_down} />,
         }[header.column.getIsSorted() as string] ?? null}
 
-        {header.column.getCanFilter() ? (
+        {header.column.getCanFilter() &&
+        !header.column.columnDef.meta?.customFilterInput ? (
           // Supressing this warning - div is not interactive, but prevents propagation of events to avoid unintended sorting
           // eslint-disable-next-line jsx-a11y/click-events-have-key-events,jsx-a11y/no-static-element-interactions
           <div onClick={(e) => e.stopPropagation()}>
```
torleifhalseth, line 144: Is this propagation challenge handled in custom filter input?

packages/eds-data-grid-react/src/EdsDataGrid.stories.tsx
torleifhalseth, line 1: Would it make sense to add a story where we use both custom filters and sorting?

torleifhalseth (approved):
Looks like a great feature @yusijs 👍

packages/eds-data-grid-react/src/EdsDataGrid.stories.tsx
yusijs: I've added two custom filters to the Filtering story, think that should cover it? Or anything in particular you were thinking about that I'm missing?

packages/eds-data-grid-react/src/EdsDataGrid.stories.tsx
```diff
@@ -56,7 +58,132 @@ export const Introduction: StoryFn<EdsDataGridProps<Photo>> = (args) => {
...
+  }: {
+    onChange: (value: { start: string; end: string }) => void
+    value: { start: string; end: string }
+  }) => {
+    const [start, setStart] = useState<string>(value?.start ?? '')
+    const [end, setEnd] = useState<string>(value?.end ?? '')
+
+    useEffect(() => {
+      onChange({ start, end })
+    }, [start, end])
```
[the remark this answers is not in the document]
yusijs: I added the dep to the array instead, and wrapped onChange in useCallback to avoid rerendering to often.

packages/eds-data-grid-react/src/EdsDataGrid.stories.tsx
torleifhalseth: Sorry, my comment is vague. I can't remember why I asked, but I believe I had some worry that the combination of the new custom filters and supporting sorting could cause issues.  But maybe you already tested this?

torleifhalseth: I tested adding `enableSorting` to the story locally and it seems to be working as expected.
Screenshot 2024-04-02 at 08 38 31

torleifhalseth (approved):
LGTM 🚢

PR #3377, opened 2024-04-09
Update node in Codespaces

### LINKED ISSUE #3376: Update node version in Codespaces

vnys:
closes #3376


vnys:
The latest version of Debian is bookworm, but I don’t see any point in updating that at the moment, so I propose we stick with bullseye. Also, since the devs in the team use node v18.x, I did not update to v20. 

> The 18.x release line now moves into "Active LTS" and will remain so until October 2023. After that time, it will move into "Maintenance" until end of life in April 2025.

oddvernes (approved):
lgtm

PR #3379, opened 2024-04-09
Update issue template

vnys:
closes equinor/design-system-internal#78

### DIFF
--- a/.github/ISSUE_TEMPLATE/generic-issue.md
+++ b/.github/ISSUE_TEMPLATE/generic-issue.md
@@ -1,10 +1,26 @@
 ---
 name: Generic issue
 about: Use this template to create issues where none of the other templates apply
-title: ''
+title: 'In order to <achieve some value>, as a <type of user>, I want <some functionality>.'
 labels: ''
 assignees: ''
 
 ---
+**Description**
+
+A clear and concise description of what this issue is about.
+
+**Definition of done**
+
+1.
+2.
+3.
+4.
+
+**Owner**
+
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/generic-issue.md
```diff
@@ -1,10 +1,27 @@
 ---
 name: Generic issue
 about: Use this template to create issues where none of the other templates apply
-title: ''
+title: 'In order to <achieve some value>, as a <type of user>, I want <some functionality>.'
```
torleifhalseth, line 4: In my experience, the order of the different segments is defined like this:
As a [user role], I want [goal/desire] so that [benefit]

Something we could consider?

User Role: This specifies who the user is or the role they are performing in the context of the story. It helps to identify the target audience or specific persona that the user story is designed to address.

Goal/Desire: This part of the user story explains what the user wants to achieve or the task they wish to perform. It focuses on the action or the feature they are interested in.

Benefit: This explains why the user wants to achieve this goal or perform this action, highlighting the value or benefit they expect to gain from it.

vnys: The difference is this format puts [emphasis on the business value](https://www.infoq.com/news/2008/06/new-user-story-format/)

torleifhalseth: Ok. Not sure how common this is and if it adds more confusion than value. I guess we can use this and get feedback from the team. 🚀

torleifhalseth (approved):
LGTM :shipit:

PR #3380, opened 2024-04-09
🐛 TextField: dynamically update padding for icon

### LINKED ISSUE #3354: 🐛 Input: make left/right-adornments reactive
Currently if an icon is added to a textfield dynamically such as adding an error icon when invalid, the input padding is not updated causing the icon to potentially overlap other content such as the calendar button when type="date".
One workaround is to force refrech the component using key
```
key={canSubmit.toString() + '1'}
inputIcon={canSubmit && <Icon data={error} />}

but we should make this work automatically as adding and removing warning icon is probably a common use case

### PR DESCRIPTION
**Pull request by oddvernes**
resolves #3354 

I could not find any other way to make this work then forcing a re-render in `TextField` if `rightAdornments` change

Updated the TextField "with icons" story with a button to toggle icon to demonstrate that it works
bilde

### DIFF
--- a/packages/eds-core-react/src/components/TextField/TextField.tsx
+++ b/packages/eds-core-react/src/components/TextField/TextField.tsx
@@ -4,6 +4,9 @@ import {
   TextareaHTMLAttributes,
   forwardRef,
   ForwardedRef,
+  useState,
+  useCallback,
+  useEffect,
 } from 'react'
 import { useId } from '@equinor/eds-utils'
 import { InputWrapper } from '../InputWrapper'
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: packages/eds-core-react/src/components/TextField/TextField.tsx [hunk 1]
```diff
@@ -82,6 +85,12 @@ export const TextField = forwardRef<HTMLDivElement, TextFieldProps>(
     ref,
   ) {
     const helperTextId = useId(null, 'helpertext')
+    const [, updateState] = useState<object>({})
+    const forceUpdate = useCallback(() => updateState({}), [])
+    //if rightAdornments are updated dynamically, we need to force re-render the component so the Input can set correct padding
+    useEffect(() => {
+      forceUpdate()
+    }, [unit, inputIcon, forceUpdate])
```
torleifhalseth, line 93: It would be nice if you could show me where we lose reactivity and why this is needed. I am hoping we can solve this without this complexity. 

In my mind, the component should re-render when the prop changes, and then we recalculate the padding based on the new props.

oddvernes: It does re-render when the prop changes and it does recalculate the padding. The problem is this all happens before the dom updates, so it is the old values. In `Input`, this works correctly, but in `TextField`, which wraps `Input` we get the delayed dom update 🤷

torleifhalseth: I understand now. The reason why it was not working as expected for the `TextField` component was that we always passed the `rightAdornments` property. 

I added the `hasRightAdornments` check so that we only pass it when `unit` or `inputIcon` are defined in the props.

torleifhalseth (approved):
I think this is a good solution, but you should probably test and confirm @oddvernes

PR #3396, opened 2024-04-22
💄Menu: focus-ring from :focus to :focus-visible

### LINKED ISSUE #3394: Focus state on click in menu
**Describe the bug**

For dropdown menu, the focus border apper on click. my thoughts is that hover effect, and active sign gives enough signal. So i am suggesting that Focus border is only shown for keybord navigation. 

image

**Extra**
Posted about this in the slack channel and a answer was: 

The focus state is important for keyboard navigation. In Menu for some reason it is also visible on click. I am not sure if that was a conscious design/ux decision back in the day or just an oversight in the implementation, because it uses :focus instead of :focus-visible which is how it is done everywhere else (only show up during keyboard navigation). I recommend filing a bug ticket so we can decide if we should change this :+1:

oddvernes:
resolves #3394 

I believe this was an oversight when making the component and not intended interaction design

### DIFF
--- a/packages/eds-core-react/src/components/Menu/MenuItem.tsx
+++ b/packages/eds-core-react/src/components/Menu/MenuItem.tsx
@@ -70,7 +70,6 @@ const Item = styled.button.attrs<StyleAttrsProps>(({ $isFocused }) => ({
       : css`
           @media (hover: hover) and (pointer: fine) {
             &:hover {
-              z-index: 1;
               cursor: pointer;
               background: ${hover.background};
             }
--- a/packages/eds-core-react/src/components/Menu/MenuItem.tsx
+++ b/packages/eds-core-react/src/components/Menu/MenuItem.tsx
@@ -61,9 +61,6 @@ const Item = styled.button.attrs<StyleAttrsProps>(({ $isFocused }) => ({
           svg {
             fill: ${icon.states.disabled.typography.color};
           }
-          &:focus {
-            outline: none;
-          }
           @media (hover: hover) and (pointer: fine) {
             &:hover {
               cursor: not-allowed;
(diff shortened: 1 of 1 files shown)


packages/eds-core-react/src/components/Menu/MenuItem.tsx
```diff
@@ -61,9 +61,6 @@ const Item = styled.button.attrs<StyleAttrsProps>(({ $isFocused }) => ({
           svg {
             fill: ${icon.states.disabled.typography.color};
           }
-          &:focus {
```
torleifhalseth, line 64: Will this get the default browser focus now?

packages/eds-core-react/src/components/Menu/MenuItem.tsx
```diff
@@ -78,7 +75,7 @@ const Item = styled.button.attrs<StyleAttrsProps>(({ $isFocused }) => ({
               background: ${hover.background};
             }
           }
-          &:focus {
+          &:focus-visible {
             z-index: 3;
```
torleifhalseth, line 79: Do you know why we need this?

packages/eds-core-react/src/components/Menu/MenuItem.tsx
oddvernes: nothing happens on focus now, only when focusing using keyboard (`focus-visible`) where it gets the same moss-green dashed focus as our other components.

packages/eds-core-react/src/components/Menu/MenuItem.tsx
oddvernes: This is needed so the focus.ring does not clip below the next item. However the z-index: 1 above in hover is no longer needed so i removed that one

torleifhalseth (approved):
LGTM 💅

PR #3401, opened 2024-04-23
🔥 Tabs: Remove invalid props from TabListProps

### LINKED ISSUE #3400: 🐛 Tabs: remove props from tabList
`TabListProps` contains `variant` and `scrollable` but these values are fetched from `TabsContext` and creates a styled-components warning if actually applied to the component, so the type should be updated to just `HTMLAttributes<HTMLDivElement>`

oddvernes:
resolves #3400


torleifhalseth (approved):
LGTM 🚢

Issue #3007, opened 2023-08-14
Improve typing of Autocomplete


olavis:
@oddvernes How about this?
```typescript
type Options<T extends Array<unknown>> = T extends Array<string>
  ? Array<string>
  : { objects: T; label: (option: T[number]) => string }
  ```
 Defining options in Autocomplete: `options: Options<T[]>`

Example, using `string[]` and `MyOptionType` as object type:
```typescript
type StringType = Options<string[]> // string[]
type ObjectType = Options<
  {
    label: string
    symbol?: string | undefined
    trend?: string | undefined
  }[]
> /*
  {
    objects: {
        label: string;
        symbol?: string | undefined;
        trend?: string | undefined;
    }[];
    label: (option: {
        label: string;
        symbol?: string | undefined;
        trend?: string | undefined;
    }) => string;
  }
 */
```
Then you'd define options like this when you're using objects as options (it type errors if you don't include label and informs about the missing properties) `options={{ objects: options, label: optionLabel }}`, and yet no complaints from TypeScript if you define options as string array: `options={['Option1', 'Option2']}`.

_Originally posted by @olavis in https://github.com/equinor/design-system/issues/2976#issuecomment-1677448486_

oddvernes:
resolved by #3408

PR #3408, opened 2024-04-29
fix: improves type safty for autocomplete when optionLabel is required

FredrikMWold:
This pull request improves autocomplete type safety.
The way it works now is that if the options are different from an array of strings then `optionLabel` is required. Typescript does not inform the user of this and the only feedback you get is a thrown error when you try and open the dropdown.

This PR makes it so that the `optionLabel` is optional if the options is an array of string, if not `optionLabel` becomes required.


packages/eds-core-react/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -195,6 +195,28 @@ const findPrevIndex: IndexFinderType = ({
   return prevIndex
 }
 
+type OptionLabelProps<T> = T extends string
```
oddvernes, line 198: Just thinking of one case, we do support options also being raw numbers without needing optionsLabel, is that possible to incorporate here? (to avoid creating a breaking change)

oddvernes (changes requested):

oddvernes:
Linting and tests are passing fine locally so disregard the failing github actions 👍

packages/eds-core-react/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -195,6 +195,28 @@ const findPrevIndex: IndexFinderType = ({
   return prevIndex
 }
 
+type OptionLabelProps<T> = T extends string | number
```
oddvernes, line 198: with this solution i am getting this error 
bilde

oddvernes (changes requested):

packages/eds-core-react/src/components/Autocomplete/Autocomplete.tsx
FredrikMWold: Can you try and restart your typescript server after checking out the changes? I am not getting the same error

image

oddvernes: Ah yes you are quite right! Everything seems to be in order now

oddvernes (approved):
LGTM!

oddvernes:
This resolves #3007

Issue #3405, opened 2024-04-25
Override date format on DatePicker component


v3gard:
**Is your feature request related to a problem? Please describe.**

Here in the civilized part of the world, we use date formats like `dd/mm/yy`. To avoid confusing users that prefers `mm/dd/yy`  (i.e. those that doesn't follow the obviously preferred standard), we should consider providing a universally accepted solution to a problem like `03/08/24`

Depending on where you are in the world, this could mean
* August 3rd 2024
* March 8th 2024

**Describe the solution you'd like**

1. A callback function that provides the selected date according to the ISO standard (e.g. `yyyy-mm-dd` / `2024-03-08`, which allows me to fully customize the output value that is shown to the end user in the date picker.

2. An example in the Story Book (to make it obvious to developers on how to apply this).

**Describe alternatives you've considered**

Creating a local variant that exists in the app we're responsible for. This does not bring any community value though....

**Additional context**

![image](https://github.com/equinor/design-system/assets/343078/47591804-3b5d-4d5c-899b-9862bfdd1602)

yusijs:
Spitballing a few potential implementations:

- Disallow input by text if a format-prop is passed in, only allow picking from the calendar (easiest to implement obviously)
- Show the formatted values when outside the field, but when focused we show the segmented inputs as today (ok-ish middleground?)
- Manually override the segments from `react-aria`'s `useDateFieldState` hook (seems like a bad road to go down)

oddvernes:
My first thought was showing the formatted value when input is blurred and then switching to showing the actual inputs on focus (which differs depending on what language you have set in the browser settings, not sure if you was aware of that @v3gard )


![Image](https://github.com/equinor/design-system/assets/4499531/56baf141-4df7-4976-b423-9c7eac913718)

v3gard:
> My first thought was showing the formatted value when input is blurred and then switching to showing the actual inputs on focus

I agree, this sounds like a good solution.

> which differs depending on what language you have set in the browser settings

@oddvernes Yes, I am aware of that. What confused me what this also overrides my local regional settings. 

Note that I prefer English in the Chrome UI (and not as locale)

![image](https://github.com/equinor/design-system/assets/343078/d2f2172c-96de-4b8b-b68d-96301f65b8d1)

I have set my locale (Windows regional settings) set to Norwegian

![image](https://github.com/equinor/design-system/assets/343078/3ec294f7-e57c-4c67-a31e-d9b5d9bd9ff4)

[...]

PR #3415, opened 2024-05-02
feat(datepicker): Enable custom date-formats

yusijs:


oddvernes (approved):
LGTM!

packages/eds-core-react/src/components/Datepicker/DatePicker.tsx
```diff
@@ -133,8 +146,12 @@ export const DatePicker = forwardRef<HTMLDivElement, DatePickerProps>(
         }
       : undefined
 
+    useEffect(() => {
+      if (!defaultValue && !value) setInnerValue(null)
+    }, [defaultValue, value])
+
```
torleifhalseth, line 152: Would be nice with a description of this side-effect so that we remember why it is there in the future.

torleifhalseth (approved):
Not sure how the formatting of the input works from a UX perspective, but other than that I like that it is based on Intl.DateTimeFormat 👍 

Added a small comment about the `useEffect` that we could describe in code.

yusijs:
@torleifhalseth there is a short video in the issue here -> https://github.com/equinor/design-system/issues/3405#issuecomment-2089910513

Agree on adding a note on the effect 👍

PR #3416, opened 2024-05-02
🚸 Autocomplete: implement native popover

### LINKED ISSUE #3413: 🚸Autocomplete etc: Implement popover attribute
Now that `popover `was added in the [last firefox update](https://caniuse.com/?search=popover), we should investigate if popover can be implemented in our floating components. This needs to play nice with floating-ui which still needs to handle positioning.
This will solve some issues arising from using components such as `Autocomplete` inside` Dialog` which requires it to have a minimum height to fit the autocomplete dropdown as it currently needs to be rendered inline

oddvernes:
resolves #3413

### DIFF
--- a/packages/eds-core-react/src/components/Autocomplete/Autocomplete.tsx
+++ b/packages/eds-core-react/src/components/Autocomplete/Autocomplete.tsx
@@ -38,18 +38,15 @@ import {
   useToken,
   bordersTemplate,
   useIsomorphicLayoutEffect,
-  useIsInDialog,
 } from '@equinor/eds-utils'
 import { AutocompleteOption } from './Option'
 import {
   offset,
   flip,
-  shift,
   size,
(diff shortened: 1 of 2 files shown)


packages/eds-core-react/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -335,6 +344,12 @@ function AutocompleteInner<T>(
     ...other
   } = props
 
+  if (disablePortal) {
+    console.warn(
+      'Autocomplete "disablePortal" prop has been deprecated. Autocomplete now uses the native popover api',
+    )
+  }
+
```
torleifhalseth, line 352: This is a nice developer experience 🎖️

torleifhalseth (approved):
LGTM

PR #3425, opened 2024-05-06
⬆️ Updated prod dependencies

### LINKED ISSUE #3423: ⬆️core-react: Update prod dependencies
Planning on making a release soon but should updtate prod dependencies first

oddvernes:
resolves #3423 

there is a bug in styled-components v6.1.9 that is [already reported](https://github.com/styled-components/styled-components/issues/4299) so I installed v6.1.8 instead. It has another issue where the conditional type that got added to Autocomplete (89dd0df678fefd5de9654281b4c6616be560755e) does not work properly if you overwrite Autocomplete styles by doing `const StyledAutocomplete = styled(Autocomplete)`, then `optionlabel` becomes **required** regardless and you have to do `optionLabel={(label: string) => label}`. Could be related to [this](https://github.com/styled-components/styled-components/issues/4303) or the multitude of other issues related to typescript & attrs currently so something to keep an eye on


packages/eds-core-react/src/components/Autocomplete/Autocomplete.test.tsx
```diff
@@ -175,6 +175,7 @@ describe('Autocomplete', () => {
     const onChange = jest.fn()
     render(
       <StyledAutocomplete
+        optionLabel={(label: string) => label}
```
torleifhalseth, line 178: Do we need to add this to the tests?

torleifhalseth (approved):
LGTM 🚢

packages/eds-core-react/src/components/Autocomplete/Autocomplete.test.tsx
oddvernes: after updating styled-components yes, that's what I am saying above. It is probably a bug in sc

Issue #3426, opened 2024-05-06
When clicking Autocomplete's Select All really fast, everything crashes.


mhwaage:
**Describe the bug**

When clicking Autocomplete's Select All really fast, everything crashes.

**Steps to reproduce the bug**

1. Have a convoluted enough enironment that there is lag when selecting options in Autocomplete
2. Set up a controlled Autocomplete component
3.  Click Select All twice, faster than the component can update between
4. Crash

**Expected behavior**
No crash

**Specifications**

- Version: 0.36.1
- Browser: Edge/firefox
- OS: Windows/Mac

**Additional context**
![image](https://github.com/equinor/design-system/assets/57612883/91eff74f-6a28-42b5-896d-5de34f46ff86)
Crash is reported as happening in:
![image](https://github.com/equinor/design-system/assets/57612883/74fd6871-2731-4874-a45d-ea2d84dc7a72)

mhwaage:
I'll try to experiment with resolving locally, since it has proven difficult to set up an MRE.

oddvernes:
I found the [commit](https://github.com/equinor/design-system/pull/1987/commits/d41466737a678fa10c5292f1cd87a9fabf60ff27) where this useEffect "if (isControlled)" was added from [this](https://github.com/equinor/design-system/pull/1987) pull request. Unfortunately it doesn't us anything about why it was added, just "`Also found some bugs related to controlled that was fixed 🐛`". Im going to test more if removing this doesn't have any side effects. Unfortunately havent been able to reproduce the crash yet

oddvernes:
reolved by #3428

PR #3428, opened 2024-05-07
Bugfix/select all crashes

mhwaage:
Removes a seemingly unnecessary useEffect, which resolves an issue where choosing select all would crash when clicked twice in rapid succession. (#3426)

### DIFF
--- a/packages/eds-core-react/src/components/Autocomplete/Autocomplete.tsx
+++ b/packages/eds-core-react/src/components/Autocomplete/Autocomplete.tsx
@@ -726,12 +726,6 @@ function AutocompleteInner<T>(
     reset: resetCombobox,
   } = useCombobox(comboBoxProps)
 
-  useEffect(() => {
-    if (isControlled) {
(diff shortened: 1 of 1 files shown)


mhwaage:
Some important caveats:
1. I was not able to create a test case that reproduces the issue, hence no tests are added for preventing regressions.
2. The useEffect being removed is the one where the code was crashing. There is no obvious reason why the useEffect should be in place, and I couldn't see any rationale for it via git blame.
3. The removal does not cause any test failures, but I don't know how good the test coverage is.
4. The removal resolves the crash reported in #3426 when I install the package locally into that same code base.

oddvernes (approved):
While it is a bit unnerving to not know what problem this particular bit of code was supposed to solve (guessing in the past, `selectedOptions` sent into the Autocomplete on load did not get set initially, or something like that, perhaps a bug with downshift 🤷‍♂️), I am unable to find any problems with removing it, so lets do it 👍

PR #3429, opened 2024-05-08
🐛 Autocomplete select all: consider disabled items

### LINKED ISSUE #3427: When clicking select all in a controlled component, disabled options are also selected
**Describe the bug**

In a controlled autocomplete, where some of the options are disabled, select all ignores that the options are disables and selects them anyway. (possibly

1. Create a controlled Autocomplete with some disabled options and Select All enabled
2. Click Select all
3. Disabled options are selected
4.

**Expected behavior**

Select All should not select disabled options
**Specifications**

- Version: 0.36.1
- Browser: Edge
- OS: Windows

**Additional context**
image

oddvernes:
resolves #3427 

changes
- moved defaultValue of `optionDisabled` to an external const so it does not break useMemo's where it is a dependency
- "select all" now only selects/unselects items that are not disabled
- "clear" button only unselects all selected items that are not disabled
- "x/y selected" text - changed y from "all not disabled items length" to "all items length" (due to the fact that disabled items can still have selected state)

### DIFF
--- a/packages/eds-core-react/src/components/Autocomplete/Autocomplete.tsx
+++ b/packages/eds-core-react/src/components/Autocomplete/Autocomplete.tsx
@@ -425,7 +425,6 @@ function AutocompleteInner<T>(
     addSelectedItem,
     removeSelectedItem,
     selectedItems,
-    reset: resetSelection,
     setSelectedItems,
   } = useMultipleSelection(multipleSelectionProps)
 
@@ -779,7 +778,8 @@ function AutocompleteInner<T>(
 
   const clear = () => {
     resetCombobox()
-    resetSelection()
+    //dont clear items if they are selected and disabled
+    setSelectedItems([...selectedDisabledItems])
     setTypedInputValue('')
   }
   const showClearButton =
--- a/packages/eds-core-react/src/components/Autocomplete/Autocomplete.tsx
+++ b/packages/eds-core-react/src/components/Autocomplete/Autocomplete.tsx
@@ -428,38 +428,37 @@ function AutocompleteInner<T>(
     setSelectedItems,
   } = useMultipleSelection(multipleSelectionProps)
 
-  const notDisabledItems = useMemo(() => {
+  const enabledItems = useMemo(() => {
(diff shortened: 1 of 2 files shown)


torleifhalseth (approved):
LGTM :shipit:

packages/eds-core-react/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -431,18 +434,21 @@ function AutocompleteInner<T>(
...
-  }, [inputOptions, selectedItems])
+  }, [notDisabledItems, selectedItems])
 
   const toggleAllSelected = () => {
-    if (selectedItems.length === inputOptions.length) {
+    if (selectedItems.length === notDisabledItems.length) {
       setSelectedItems([])
     } else {
-      setSelectedItems(inputOptions)
+      setSelectedItems(notDisabledItems)
```
mhwaage, line 451: Sorry for drive-by reviewing and also sorry for potentially adding more work, but:

Consider a situation where one of the disabled options has been selected¹. In this scenario, I would expect the value of the disabled options to be unaffected by toggling select all; i.e. the setSelectedItems needs to take those into account. In addition, the comparison `selectedItems.length === notDisabledItems.length` would be wrong in this scenario.

¹This could e.g. happen in a scenario where users are allowed to apply filters to which end results should be selected, without those filters removing selected values.

oddvernes: Ah you are quite right, of course a disabled item can also be selected. So I need a "selectableItems.length" as well basically

oddvernes:
I have now taken into account that a disabled item can be selected, and the "select all" toggle will only select/deselect items that are not disabled. I also made a change to the "x / y selected" text, where y used to be `all items length - disabled items length`, is now changed to` all items length`
But this also made me realize that our "clear all" button, the "x" in the input, clears disabled items as well if they are selected. So I have made a change there as well so "clear" only clears non-disabled items.

PR #3447, opened 2024-05-16
🎉  Create eds-tokens-sync, eds-tokens-build and add tokens from Figma to eds-tokens 

### LINKED ISSUE #3442: In order to use new tokens, as a developer, I want these available using the eds-tokens package
#### General objective

Add new tokens output to the eds-tokens package

#### Specific objectives
- Add existing code from [tokens-sync](https://github.com/equinor/design-system-tokens/tree/dev/packages/tokens-sync)
- Test the build and output using NPM pack locally and test the token input internally in React components.
- Publish a beta version of the eds-tokens package with new colour tokens.

#### Outcome

When developers can consume the new tokens from CSS, JS, and JSON using the new tokens-package beta.

#### Owner

@torleifhalseth

torleifhalseth:
resolves #3442


.github/workflows/sync-figma-to-tokens.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Install Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '18.20.2'
+
+      - name: Clone repo
+        uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: npm ci
```
oddvernes, line 21: shouldn't this be pnpm? I don't know what kind of consequences it can have to use npm in a pnpm monorepo

.github/workflows/sync-figma-to-tokens.yml
```diff
@@ -0,0 +1,36 @@
...
+        working-directory: packages/eds-tokens
+
+      - name: Sync variables from Figma file to canonical tokens in GitHub
+        run: npm run update-tokens && npm run build
```
oddvernes, line 25: also npm ?

packages/eds-core-react/src/components/Button/tokens/contained.ts
```diff
@@ -18,55 +18,55 @@ const {
...
   },
-
   shape: {
     button: { borderRadius: buttonBorderRadius },
   },
 } = tokens
 
 export const primary = mergeDeepRight(button, {
-  background: primaryColor,
+  background: `var(--eds-color-surface-button-primary-filled, ${primaryColor})`,
```
oddvernes, line 27: These button experiments should be reverted before we can merge. We'll try and do this the other way around first using the override variables in the existing tokens to inject new tokens, and then figure out the proper pattern for implementing new tokens directly later

.github/workflows/sync-tokens-to-figma.yml
```diff
@@ -0,0 +1,25 @@
+name: Sync tokens to Figma
+on:
```
oddvernes, line 2: Is this safe to do as an action? Will we be able to see the console output if something went wrong?

packages/eds-core-react/src/components/Button/tokens/contained.ts
torleifhalseth: Reverted ⏮️

.github/workflows/sync-tokens-to-figma.yml
torleifhalseth: Updating the variables in Figma could break; if it does, we must revert changes using the history view in Figma.  If something unexpected occurs in the script, this should be logged in the execution of the action.

oddvernes (approved):

PR #3450, opened 2024-05-21
fix: Datetime-input is formatted with / instead of :

yusijs:
I forgot to account for literals/placeholder-values, so it acts a bit... off now.

In latest release:
image

This fix:
image


oddvernes (approved):
LGTM

PR #3460, opened 2024-05-23
🏷️ Allow ReactNode as title in Tooltip

magnh:

Add `ReactNode` to the `title` prop in the `Tooltip` component.

The title is used directly in the JSX and passing custom components as title seems to work.

Not sure this is intended, up to the EDS team.

### DIFF
--- a/packages/eds-core-react/src/components/Tooltip/Tooltip.tsx
+++ b/packages/eds-core-react/src/components/Tooltip/Tooltip.tsx
@@ -7,6 +7,7 @@ import {
   cloneElement,
   useMemo,
   useEffect,
+  ReactNode,
 } from 'react'
 import { createPortal } from 'react-dom'
 import styled from 'styled-components'
@@ -81,7 +82,7 @@ export type TooltipProps = {
   /** Tooltip placement relative to anchor */
   placement?: Placement
   /** Tooltip title */
-  title?: string
+  title?: ReactNode
   /** Tooltip anchor element */
   children: React.ReactElement & React.RefAttributes<HTMLElement>
   /** Delay in ms, default 100 */
@@ -90,7 +91,7 @@ export type TooltipProps = {
    * @default document.body
    * */
   portalContainer?: HTMLElement
-} & HTMLAttributes<HTMLDivElement>
+} & Omit<HTMLAttributes<HTMLDivElement>, 'title'>
 
 export const Tooltip = forwardRef<HTMLDivElement, TooltipProps>(
   function Tooltip(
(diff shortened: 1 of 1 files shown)


packages/eds-core-react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -76,7 +77,7 @@ export type TooltipProps = {
   /** Tooltip placement relative to anchor */
   placement?: Placement
   /** Tooltip title */
-  title?: string
+  title?: string | ReactNode
```
oddvernes, line 80: string is already included in ReactNode type so `string | ` can be removed here. Another thing is, `title` is also a valid html attribute (which is string), so for this to work, you need to replace` & HTMLAttributes<HTMLDivElement>` at the bottom of the type here with  `& Omit<HTMLAttributes<HTMLDivElement>, 'title'>`

oddvernes (changes requested):

oddvernes:
I have to check a bit more if this should be allowed. This might get vetoed by our designers since tooltip should be strictly for short feedback etc and to use Popover for more complex formatted output

yusijs:
Just to chime in:
We have often also wanted to have a bit more comples values in the tooltip. Supersimple example is just being able to make part of the text cursive/bold, e.g "You should **not** do xyz". Another example is just to have an extra line to improve readability by splitting the text naturally. We often fall back to the native title-attribute here, since it adheres to \n for linebreaks (in some browser/os-combos at least)

This has been requested by our designers several times, where the popover is far to intrusive but the tooltip is to limited.

oddvernes:
Our [guidelines](https://eds.equinor.com/0b0c666ab/p/66a3cf-tooltip/b/49f4b2) HARD disagree on this 😂 
But who wrote this and why? we will probably never know 🤷 
bilde

yusijs:
We actually discussed this with a former EDS designer (Lucas) who said that it should be treated as guidelines and not rules. In this case it seems like it's expected as a rule. Imo, the result of such restrictions doesnt lead to a better / more consistent design across Equinor, but rather the opposite, since those who have the requirement will just roll their own variants. 

I would be interested getting some perspectives from the EDS designers (@BirteThornquist / @benteljones?) on why this rule is in place, and their thoughts on the points I outlined above.

benteljones:
We're going to go through all the old guidelines and create new examples. Several of them are outdated. Although we want the tooltip to be primarily for short, one-line information, we need to facilitate a slightly more adapted context. As you point out @yusijs , it will not lead to a more consistent use, but rather the opposite if the restrictions don't cover what the users need. 
We will take these needs with us into the work on an updated 2.0 component. Thank you for the important feedback. 🌸

oddvernes (approved):

PR #3487, opened 2024-06-05
:label: Make Autocomplete options prop readonly

magnh:

Allow `options` prop in `Autocomplete` to be readonly. This way you can pass readonly arrays such as the one below without getting any TypeScript errors.

```
const READONLY_OPTIONS = [
  { id: 1, label: "One" },
  { id: 2, label: "Two" },
] as const
```

The autocomplete should not change the options array anyways, so I cannot see any reason why not to support readonly options prop.

(became a fork by accident)

### DIFF
--- a/packages/eds-core-react/src/components/Autocomplete/Autocomplete.tsx
+++ b/packages/eds-core-react/src/components/Autocomplete/Autocomplete.tsx
@@ -250,7 +250,7 @@ export type AutocompleteChanges<T> = { selectedItems: T[] }
 
 export type AutocompleteProps<T> = {
   /** List of options in dropdown */
-  options: T[]
+  options: readonly T[]
   /** Label for the select element */
   label: ReactNode
   /** Array of initial selected items


packages/eds-core-react/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -250,7 +250,7 @@ export type AutocompleteChanges<T> = { selectedItems: T[] }
 
 export type AutocompleteProps<T> = {
   /** List of options in dropdown */
-  options: T[]
+  options: readonly T[]
```
oddvernes, line 253: looks like this needs some more work for typescript to be happy
bilde

oddvernes (changes requested):

packages/eds-core-react/src/components/Autocomplete/Autocomplete.tsx
oddvernes: I will fix this (adding readonly) in another pr for #3514. By the way looks like `downshift `is happy again if I add `as T[]` behind `availableItems`

oddvernes:
Will be resolved by #3515 instead

PR #3518, opened 2024-06-13
✨💄 Legacy colors override

### LINKED ISSUE #3510: ✨Create eds-2-overrides.css
In eds-tokens, create a new `eds-2-overrides.css` file (maybe with a better name). Paste in the color variables from the old `tokens.css` in the root of eds-tokens (but skip the infographic ones)  and make sure to rename the ones that are different in the js output overrides.
example
```
    static_icons__default: {
      hex: '#3d3d3d',
      hsla: 'hsla(0, 0%, 23.9%, 1)',
      rgba: 'var(--eds_text__static_icons__default, rgba(61, 61, 61, 1))'
    },
```
looks like 
 ` --eds_text_static_icons__default: rgba(61, 61, 61, 1);`
in `tokens.css`, so there is an underscore missing between text and static that needs to be added. Not sure if this is the case in all or just some variables.

next each of the variables must be given a value from the new eds 2.0 semantic variables. 
I think only
```
  --eds_interactive_secondary__resting: rgba(36, 55, 70, 1);
  --eds_interactive_secondary__link_hover: rgba(23, 36, 47, 1);
```
are missing on account of secondary button being removed, so in this case i guess we'll leave it, or use the values that had been intended for 2.0

Once tokens are published, people can import this file after variables.css and they should have 2.0 colors + dark theme out of the box in the eds-core-react components. We also need to make sure if there are hardcoded colors anywhere in the components that these are overridden too.

oddvernes:
resolves #3510 

Work in progress. I have extracted all the color override variables from the old tokens into an new css file. Now we just need to hook them up to the relevant new 2.0 variables

TODO: This pr should also install a theme switcher for storybook. Perhaps [this one](https://storybook.js.org/addons/storybook-dark-mode). Ideally we should be able to use our own tokens for the storybook ui as well


torleifhalseth:
Look at the global CSS variables if we want to replace the default value of these globally in the future.

PR #3519, opened 2024-06-14
🎉 support shorthand for spacing variables when outputReferences is false (trimmed)

torleifhalseth:
✨ Include spacing variables in the output


oddvernes (approved):
Im sure there is stuff we have to change later but lets merge it for now

PR #3523, opened 2024-06-14
🔨 Cyclical dependency toggle spacing

oddvernes:
This seems to fully work as of right now, but only without the  `'eds/css/spacing/shorthand'` transformer.
can be tested with import ` '@equinor/eds-tokens/css/variables.css'` or ` '@equinor/eds-tokens/css/variables-verbose.css'` in `eds-core-react\.storybook\preview.mjs`


packages/eds-tokens-build/src/index.ts
```diff
@@ -68,6 +71,48 @@ const lightDarkTransform: StyleDictionary.Transform = {
   },
 }
 
+//MARK: SpaceToggleTransform
+const densitySpaceToggleTransform: StyleDictionary.Transform = {
```
torleifhalseth, line 75: Great work. Could we a create separate file for this transform and maybe also for the other complex transforms we have? I think that would clean up this file and make it more readable.

packages/eds-tokens-build/src/index.ts
```diff
@@ -228,7 +278,7 @@ StyleDictionary.registerTransform({
 
 const cssTransforms = [
   'name/cti/kebab',
-  'eds/css/spacing/shorthand',
+  //'eds/css/spacing/shorthand',
```
torleifhalseth, line 281: What remains to make this work?

packages/eds-tokens/build/css/index-trimmed.css
```diff
@@ -1,3 +1,3 @@
 @import 'color/theme-trimmed.css';
-@import 'spacing/comfortable-trimmed.css';
-@import 'spacing/spacious-trimmed.css';
+@import 'cyclical-dependency-setup.css';
```
torleifhalseth, line 2: Could we name this file something like "density-setup"?

torleifhalseth (changes requested):
Looks great. Added some comments.

packages/eds-tokens-build/src/index.ts
oddvernes: We'll do this in a sepereate branch later so we keep the git history for these changes, plus there is no time

packages/eds-tokens-build/src/index.ts
oddvernes: As discussed, we'll find a solution for this later as seperate tokens  (if we do it at all)

PR #3528, opened 2024-06-17
Release eds tokens 1.0.0 beta 1

oddvernes:
This pr must not be merged, it is only for inspection


packages/eds-tokens-build/src/index.ts
```diff
@@ -329,7 +332,7 @@ const _extend = ({
           },
         ],
       },
-      ts: {
+      /*       ts: {
```
torleifhalseth, line 335: Why are we excluding the ts and JSON transforms?

packages/eds-tokens/package.json
```diff
@@ -55,7 +42,7 @@
     "elements.css"
   ],
   "scripts": {
-    "build": "rollup -c --bundleConfigAsCjs && pnpm run types && pnpm --filter @equinor/eds-tokens-sync run build && pnpm --filter @equinor/eds-tokens-build run build",
+    "build": "rollup -c --bundleConfigAsCjs && pnpm run types",
```
torleifhalseth, line 45: I guess we can assume that the token-sync and token-build are pre-build, but I think it makes sense to add `tokens-in-code:build` here. Then the build of all the things are isolated in the package, and we don't have to specify both the old and new build in the root package.json.

package.json
```diff
@@ -8,12 +8,14 @@
     "lint:lab": "pnpm run lint ./packages/eds-lab-react",
     "lint:utils": "pnpm run lint ./packages/eds-utils",
     "init": "pnpm install --force && pnpm run build",
-    "build": "pnpm run build:icons && pnpm run build:tokens && pnpm run build:utils && pnpm run build:core-react && pnpm run build:data-grid-react && pnpm run build:lab",
+    "build": "pnpm run build:icons && pnpm run build:tokens && pnpm run build:utils && pnpm run build:core-react && pnpm run build:data-grid-react && pnpm run build:lab && pnpm run build:token-build && pnpm run build:token-sync",
```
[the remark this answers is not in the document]
oddvernes: It should not be run as part of build, it should be manually ran by a human and comitted same as we do with the old tokens

packages/eds-tokens-build/src/index.ts
oddvernes: they were misnamed (-verbose) and we didn't have time to fix it

packages/eds-tokens/package.json
oddvernes: No they absoluetly should be seperate

oddvernes: `tokens-in-code:build` should be considered as part of the transform step. It is only ran when we do `transform:tokens`. The reason for this is that `lightningcss` is an external dependency that will get updated over time which could affect the output. Changes to output must always be verified to make sure nothing unexpectely changes. Unlike 1.0 tokens, 2.0 tokens does not have a build step, only sync and transform

PR #3542, opened 2024-06-25
:bug: Do not attach document.body if SSR

magnh:

This pull request improves support for SSR for `Autocomplete` by checking if `typeof document === "undefined"` before attaching `document.body` to `floating-ui`.

This should not change any existing functionality since the code is already checking that document is defined using `?`. `typeof` can be used on the server.

Get this error when our `Next.js` application tries to render an Autocomplete.

image

Did not manage to reproduce it in CodeSandbox, so it was just as fast creating a PR for it 😄


oddvernes (approved):
This is great, thanks for the help! I had mistakedly thought "?." would be enough

PR #3544, opened 2024-06-28
✨ update colors

torleifhalseth:
- Remove the theme suffix from the color files and use light and dark.
- Output colors as hex
- Delete old token collections
- Update color tokens and transform new code


oddvernes (approved):
I guess this is all how it is supposed to be (however preposterous those 100 or so table interactive tokens looks) 🤷‍♂️

oddvernes:
I really feel like we should reconsider oklch as the output color though. It doesnt add any value, it just makes the values less readable and file size larger. It doesn't really matter what format it is as long as its srgb space anyway so it might as well be hex 🤔

PR #3547, opened 2024-07-02
fix(Icon): improve icon type definition by making it a descriminated union on name and data

FredrikMWold:
This pull request improves the type defintion of the `Icon` component by making it a discrimited union on the props `name` and `data`. This makes it so that you have to supplie either as if you the `Icon` ends up crashing and throwing an error


oddvernes (approved):
LGTM!

PR #3555, opened 2024-07-05
⬆️ Update dev dependencies

### LINKED ISSUE #3551: ⬆️ Update dev dependencies july 2024
Some dependencies with major version changes may need their own tickets but this ticket should cover all the minor/patch updates at least

oddvernes:
resolves #3551


oddvernes:
Everything is up to date except styled-dictionary and eslint which has gotten their own tickets as these are larger undertakings. SVGO update expectedly caused icon paths to slightly change so this is why all the icons appear to be updated (there are no visual changes)

PR #3565, opened 2024-07-29
Bugfix/add missing bounds check for findIndex

mhwaage:
* Adds a bounds check for Autocomplete's findIndex, to prevent infinite recursion once the index has gone beyond bounds.
* Adds test to prevent regression

### DIFF
--- a/packages/eds-core-react/src/components/Autocomplete/Autocomplete.test.tsx
+++ b/packages/eds-core-react/src/components/Autocomplete/Autocomplete.test.tsx
@@ -336,7 +336,8 @@ describe('Autocomplete', () => {
       <Autocomplete
         options={items}
         label={labelText}
-        optionDisabled={() => true}
+        // Somewhat contrived condition to emulate a scenario where an undefined item would return true for the 'option' being disabled
+        optionDisabled={(item) => item !== items[0]}
       />,
     )
     const labeledNodes = await screen.findAllByLabelText(labelText)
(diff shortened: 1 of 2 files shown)


oddvernes:
I thought this had been fixed by https://github.com/equinor/design-system/pull/3534 but I guess I missed something? I'll take a look when i return from vacation next week!

mhwaage:
> I thought this had been fixed by #3534 but I guess I missed something? I'll take a look when i return from vacation next week!

Possible that I implemented this fix on an outdated version, I will verify!

mhwaage:
> > I thought this had been fixed by #3534 but I guess I missed something? I'll take a look when i return from vacation next week!
> 
> Possible that I implemented this fix on an outdated version, I will verify!

#3534 fixes some of the issues but still misses edge cases. I will update the tests to cover those edge cases.

oddvernes:
I was trying to reproduce the edge case but think i found another one instead. 

1. Run storybook and go to the disabled option story
2. type am so only the disabled amazon option is in the list
3. press down arrow
Another crash in findindex

mhwaage:
> I was trying to reproduce the edge case but think i found another one instead.
> 
> 1. Run storybook and go to the disabled option story
> 2. type "am" so only the disabled amazon option is in the list
> 3. press down arrow
>    Another crash in findindex

Yeah, this is the same edge case that this PR _should_ be fixing. Is it not fixed by this pr?

oddvernes:
> Yeah, this is the same edge case that this PR _should_ be fixing. Is it not fixed by this pr?

My mistake, the error was in the story 🤦‍♂️ `item.trend === '📉'` should be `item?.trend === '📉'` on line 354

oddvernes (approved):
LGTM

PR #3578, opened 2024-08-07
🩹 Fix error in parsing value for scrollbar-width

Benjamin-Lyon:
This PR fixes a minor issue in the CSS value for the ⁠`scrollbar-width` property in the ⁠`TabList.tsx` component, which resulted in errors being logged in the console of the Firefox Developer Tools. Changing the value to `none` fixes this.


oddvernes (approved):
lgtm!

PR #3610, opened 2024-09-02
🐛 blind fix for "disconnected" popover

### LINKED ISSUE #3609: Hovering over a cell in our AGGrid in UDD causes an error
**Describe the bug**

image

1.
2.
3.
4.

**Expected behavior**
image

**Specifications**

- Version: n/a
- Browser: n/a
- OS: n/a

**Additional context**
- eds-core-react@0.41.3 is the version of eds which causes the error. 
- eds-core-react@0.40.1 works fine!
- 

oddvernes:
resolves #3609

### DIFF
--- a/packages/eds-core-react/package.json
+++ b/packages/eds-core-react/package.json
@@ -1,10 +1,14 @@
 {
   "name": "@equinor/eds-core-react",
 <<<<<<< HEAD
+<<<<<<< HEAD
   "version": "0.41.5",
 =======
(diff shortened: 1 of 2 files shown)


oddvernes:
After making a test release, the report back is that this did not work

oddvernes:
ioc feedback: adding isConnected check and putting it in a 1ms settimeout makes the component render but in the top left corner, so floating-ui is not triggered in time

PR #3611, opened 2024-09-02
🐛 Remove dropdown svg icon when rendering multi-select

mhwaage:
image


mhwaage:
I swear this is a well-known issue, but maybe it's just that I've seen it before and not raised an issue? Let me know if I should make a formal bugreport first

PR #3626, opened 2024-09-13
feat(datepickers): Enable setting locale from prop. 

### LINKED ISSUE #3596: Data pickers show random dateformats
**Describe the bug**

We are trying to use the DatePicker and DateRangePicker but we struggling to let them consistently 

For some users they get the format mm/dd/yyyy instead of dd/mm/yyyy

Wrong format
Screenshot 2024-08-26 at 14 00 49

Correct format
Screenshot 2024-08-26 at 14 01 06

Also the component uses random separator in different browsers on my own machine

Chrome
Screenshot 2024-08-26 at 14 04 34

Brave
Screenshot 2024-08-26 at 14 01 06

We have tried setting the locale to en-GB for the entire react application 

```tsx 
<I18nProvider locale="en-GB">
...
</I18nProvider>
```

1.
2.
3.
4.

**Expected behavior**

A consistent date format across users and browsers to avoid confusion on what the date since showing the date `02/06/2025`  as `06/02/2025` can have big consequences if the the users misunderstands the date

There should also be an ability to set the format manually, at least the separator. E.g `dd.mm.yyyy` instead of `dd/mm/yyyy`

**Specifications**

- Version:
- Browser:
- OS:

**Additional context**

yusijs:
Resolves #3596

### DIFF
--- a/packages/eds-core-react/src/components/Datepicker/utils/useGetLocale.ts
+++ b/packages/eds-core-react/src/components/Datepicker/utils/useGetLocale.ts
@@ -3,7 +3,8 @@ import { useLocale } from 'react-aria'
 export const useGetLocale = (locale?: string) => {
   const { locale: externalLocale } = useLocale()
   // react-aria defaults to navigator.language if no locale is provided. If these are equal, we override by using the system default locale
-  const defaultLocale = navigator.language
+  const defaultLocale =
+    (typeof navigator !== 'undefined' && navigator.language) || 'en-US'
   const fallbackLocale = new Intl.DateTimeFormat().resolvedOptions().locale
   return (
     locale ??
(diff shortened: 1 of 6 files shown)


packages/eds-core-react/src/components/Datepicker/utils/useGetLocale.ts
```diff
@@ -0,0 +1,13 @@
+import { useLocale } from 'react-aria'
+
+export const useGetLocale = (locale?: string) => {
+  const { locale: externalLocale } = useLocale()
+  // react-aria defaults to navigator.language if no locale is provided. If these are equal, we override by using the system default locale
+  const defaultLocale = navigator.language
```
oddvernes, line 6: I think there needs to be an existence check for navigator since it will be undefined in nextjs

oddvernes (changes requested):

oddvernes (approved):
LGTM

PR #3648, opened 2024-10-07
🎨 Pass className to root

arkadiy93:
Like in EDS, we are also using styled components. 
It would be nice to be able to pass the className prop to the root to have more control of the styles.

https://styled-components.com/docs/basics#styling-any-component


oddvernes:
I see that we are not passing `{...rest}` props either which is where className and style is normally inlcuded. I'm not sure why there is a className there already since it is unused, but perhaps we need to keep it now that it has been published and possibly used by people. 
I think we should pass {...rest} as well (+ need to extend type with `& HTMLAttributes<HTMLDivElement>`). Note that we pass some styles required for `enableVirtual` that should not be interferred with so maybe style should be left as is 🤷

oddvernes (approved):
LGTM!

PR #3658, opened 2024-10-18
EDS Data Grid Feature: onRowContextMenu

tlastad:
I have added a `onRowContextMenu` feature to allow for right-click on table rows.
We need this to be able to use context menus on tables generated using Eds Data Grid.

`onRowContextMenu` works the same as the already existing `onRowClick` functionality. It just uses the other mouse button.

Tests and storybook is updated, and runs.
If there's anything I've missed, pleas let me know.

### DIFF
--- a/packages/eds-data-grid-react/src/EdsDataGrid.stories.tsx
+++ b/packages/eds-data-grid-react/src/EdsDataGrid.stories.tsx
@@ -286,6 +286,76 @@ RowSelection.args = {
   columnResizeMode: 'onChange',
 } satisfies Partial<EdsDataGridProps<Photo>>
 
+const BasicModal = ({
+  text,
+  top,
+  left,
+  onHide,
+}: {
+  text: string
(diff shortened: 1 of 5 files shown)


packages/eds-data-grid-react/src/EdsDataGrid.stories.tsx
```diff
@@ -286,6 +286,76 @@ RowSelection.args = {
   columnResizeMode: 'onChange',
 } satisfies Partial<EdsDataGridProps<Photo>>
 
+const BasicModal = ({
```
yusijs, line 289: Could you change this to use the Popover from eds-core-react instead?

tlastad:
I changed it to use a menu instead, as that is more in line with what it probably will be used for.

oddvernes (approved):
Looks good! Approved!

tlastad:
Thank you :)

PR #3660, opened 2024-10-22
fix: when using menu in a form, the form is submitted

FredrikMWold:
When using `Menu` in a form, the form is submitted when clicking on a menu item if `closeMenuOnClcik` is set to false

The problem is a missing `type="button"`.
Looks like this bug has been hiding behind the fact that having it in a form and having `closeMenuOnClick` set to `false` is rare.
The `onClick` callback being called twice when you set `closeMenuOnClick` set to `false` and dont give a `onClose` callback seems to have some have resolved for that instance


oddvernes (approved):
Nice catch!

Issue #3667, opened 2024-10-30
Exposing FilteringState


OdaKNilsen:
**Is your feature request related to a problem? Please describe.**

My users want default filters applied when opening table view in the Post Well Analytics application

**Describe the solution you'd like**

I want EdsDataGrid to expose properties: filteringState and possibly also onFilteringChange

**Describe alternatives you've considered**

Re-implementing our own table and a separate filtering solution. 

**Additional context**

On request I can push a branch I made solving the issue, it's not tested or anything, just for demo.

OdaKNilsen:
Patiently waiting for a release of @equinor/eds-data-grid-react so I can test https://github.com/equinor/design-system/pull/3670 :-)

oddvernes:
Ask and ye shall recieve 🙂  https://github.com/equinor/design-system/releases/tag/eds-data-grid-react%400.7.3

PR #3670, opened 2024-11-04
feat(data-grid): Expose table-instance via ref

### LINKED ISSUE #3586: Add access to current table state in eds-data-grid

I would like to be able to access the current table state, such as when it is filtered, in order to have the ability to export the table data as it is displayed at any given time.

Tanstack table seems to have an `onStateChange` prop, that perhaps could be exposed through the EdsDataGrid component.

I couldn't think of any reasonable alternatives, given I'd like to keep using EdsDataGrid. If there happens to be a hack or workaround for accessing current table state with the existing props please let me know!

**Additional context**

My goal is to maintain the functionality of the old table solution, which included a bunch of imperative and explicit handling of filtering the table data, so it was pretty straightforward to provide the current table data to an export-to-excel method.

yusijs:
There are a few instances where people want / need to access the `table` instance returned from `useReactTable`, or where certain features that we can't prioritize immediately (for example #3667) is required. By exposing the instance, we allow the consumers to bypass these limitations until we can either address their request, or we still allow them to do whatever they need to do without us having to officially ™️ support it.

Resolves #3586

### DIFF
--- a/packages/eds-data-grid-react/src/EdsDataGrid.tsx
+++ b/packages/eds-data-grid-react/src/EdsDataGrid.tsx
@@ -85,6 +85,7 @@ export function EdsDataGrid<T>({
   height,
   getRowId,
   rowVirtualizerInstanceRef,
+  tableInstanceRef,
   columnSizing,
   onColumnResize,
   expansionState,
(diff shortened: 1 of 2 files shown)


oddvernes (approved):
LGTM

PR #3672, opened 2024-11-08
fix: fixes issue where sidebar default state always starts as closed

FredrikMWold:
This PR addresses a bug in the sidebar component where setting the `open` state to `true` by default results in an initial layout shift. The component briefly renders as closed before opening, causing an unwanted animation.

### Issue

The issue stems from using a `useEffect` to sync the internal state with the `open` prop, leading to a mismatch on initial render.

### Fix

The internal state now defaults directly to the `open` prop, ensuring the sidebar renders correctly without unnecessary transitions.


oddvernes (approved):
Lgtm!

PR #3689, opened 2024-11-20
🐛 Autocomplete: undefined param in optionDiabled

### LINKED ISSUE #3677: Autocomplete 'option' parameter becomes 'undefined' in 'optionDisabled' callback
**Describe the bug**

The `optionDisabled` callback will be called with the 'option' argument as `undefined` when the option is removed from the available `options`.

The type definition of the `optionDisabled` function suggest that the `option` argument will not be undefined, but always contain the actual `option` to  evaluate. This easily (caused a bug for us) will cause an error where we access the `option` argument as if it's there (it can be undefined).

https://codesandbox.io/p/sandbox/react-18-playground-forked-dts5gr

**Expected behavior**

It should only call `optionDisabled` with actual options, and never  `undefined`.

**Specifications**

- Version: "@equinor/eds-core-react": "^0.42.2
- Browser: Chrome Version 130.0.6723.116 (Official Build) (64-bit)
- OS: Windows 11

oddvernes:
resolves #3677 

For some reason `selectedItems` (returned from `useMultipleSelection` from downshift) contains an array with null (`[null]`) if you have a controlled singleselect and update `selectedOptions` with an empty array from the outside. This then causes the component to crash if `optionDisabled` is defined and tries to read a property from null. I have fixed this by adding an extra null check when creating the `selectedDisabledItemsSet`

### DIFF
--- a/packages/eds-core-react/src/components/Autocomplete/Autocomplete.tsx
+++ b/packages/eds-core-react/src/components/Autocomplete/Autocomplete.tsx
@@ -467,7 +467,7 @@ function AutocompleteInner<T>(
   const allDisabled = enabledItems.length === 0
 
   const selectedDisabledItemsSet = useMemo(
-    () => new Set(selectedItems.filter(optionDisabled)),
+    () => new Set(selectedItems.filter((x) => x !== null && optionDisabled(x))),
     [selectedItems, optionDisabled],
   )
 


torleifhalseth (approved):
LGTM 🚢

PR #3690, opened 2024-11-20
feat(eds-core-react): :sparkles: use locale prop for text and formats when it is defined

### LINKED ISSUE #3638: Datepicker min/max value validation errors do not follow the selected locale
We want to use the DatePicker's min and max value to help the user select an appropriate date, but have encountered a blocker. It turns out the DatePicker produces Norwegian text in our English language application.

Regardless of the locale setting, the DatePicker displays the validation error in Norwegian.
image

Steps to reproduce for min-value, but max-value behaves the same
1. Create a DatePicker with min value and locale set to anything other than Norwegian, e.g. "en"
2. Use the keyboard to select a date earlier than the min value

To quickly reproduce it, go to 
https://storybook.eds.equinor.com/?path=/story/inputs-dates-datepicker--with-min-and-max-value&locale:en
(notice that "en" is selected), and follow the steps. You'll see
image
(American date format and Norwegian validation error)

**Expected behavior**

It's expected that the validation error is displayed in the selected language, just as the calendar's months and days use the selected locale's language.

Instead, the validation error is in Norwegian. I don't know why Norwegian is selected in particular, but it's not unreasonable in my case.

**Specifications**

- Version: "@equinor/eds-core-react": "^0.42.0"
- Browser: Firefox set to display pages in US English. I have tested it in Edge too, and it's stuck showing English despite the calendar showing German.
- OS: Windows using US English as the language and Norwegian date format.

** Other info **
I think I've traced the problem back to Adobe's code in https://github.com/adobe/react-spectrum/blob/main/packages/%40react-stately/form/src/useFormValidationState.ts, so I don't think your code is to blame directly. In any case, something isn't using the specified locale.

torleifhalseth:
Fixes #3638

### DIFF
--- a/packages/eds-core-react/src/components/Datepicker/DatePicker.stories.tsx
+++ b/packages/eds-core-react/src/components/Datepicker/DatePicker.stories.tsx
@@ -106,6 +106,7 @@ export const CustomLocale: StoryFn = () => {
   const [locale, setLocale] = useState('en-US')
   const locales = [
     { value: 'en-US', label: 'English' },
+    { value: 'nb-NO', label: 'Norwegian' },
     { value: 'uk', label: 'Ukrainian' },
     { value: 'sv-SE', label: 'Swedish' },
     { value: 'zh-Hans', label: 'Chinese (Simplified)' },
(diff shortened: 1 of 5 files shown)


oddvernes (approved):
LGTM

PR #3695, opened 2024-11-21
🐛 Tooltip: attempt to fix flicker issue

### LINKED ISSUE #3694: Tooltip flicker when hovering button or chip
**Describe the bug**

After upgrading from eds-core-react 0.40.1 to 0.42.4, we noticed that our tooltips have started flickering.
This happens when I hover the button, but the flickering stops when I hover the Icon on the button.
Also, tooltips on chips flicker all the time when the chip is hovered.

1. Add a button with an icon or a chip and wrap the button/chip in a tooltip
```html
<Tooltip title={'test'} placement="bottom">
  <Button variant={'ghost'}>
    This is a test <Icon data={comment} />
  </Button>
</Tooltip>
``` 

2. Hover the button text then hover the icon
3. The tooltip flickers every time you move your mouse or when hovering over a chip

**Expected behavior**

Expect the tooltip to stay still

**Specifications**

- Version: 0.42.4 / 0.42.5 (tried both)
- Browser: chrome
- OS: Windows

**Additional context**

oddvernes:
resolves #3694 

I will make a `next` release from this branch first so the reporter of the bug can test if the issue is resolved

### DIFF
--- a/packages/eds-core-react/package.json
+++ b/packages/eds-core-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@equinor/eds-core-react",
-  "version": "0.42.6-dev-11212024",
+  "version": "0.42.5",
   "description": "The React implementation of the Equinor Design System",
   "sideEffects": [
     "**/*.css"
--- a/packages/eds-core-react/package.json
+++ b/packages/eds-core-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@equinor/eds-core-react",
-  "version": "0.42.5",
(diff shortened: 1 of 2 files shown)


oddvernes:
The flicker issue, as it turned out, arose from style isolation in a micro frontend architecture. It was solved by adding an optional `portalContainer` for the tooltip in #3696.  I'll still merge this pr as it aligns with the popover implementation which is a bit more robust when it comes to edge cases when triggering `showPopover()`

PR #3706, opened 2024-12-12
fix: icon id for next.js

### LINKED ISSUE #3705: `<Icon />` component uses custom logic for getting unique ids, that doesn't produce equal results on server and client (in next.js)
### Describe the bug

The `Icon` component uses custom logic for getting unique ids to match `aria-described-by` tags to a `<title>`-tag with an `id`. This custom id logic doesn't produce equal results on server and client, since it's based on `Math.random()`.

To track down the bug, let's trek through the source code of the `Icon` component:
Assumptions: 
```typescript
import { external_link } from '@equinor/eds-icons'
// the component usage
<Icon data={external_link} title="Open in new tab"/>
```

The title tag is rendered on [Icon.tsx:150](https://github.com/sebastianvitterso/design-system/blob/69b53d2cc7eb67b4d2a3d462da0f966facdc5ee5/packages/eds-core-react/src/components/Icon/Icon.tsx#L150):
```typescript
{title && <title id={titleId}>{title}</title>}

```
Let's look for `titleId` on [Icon.tsx:138](https://github.com/sebastianvitterso/design-system/blob/69b53d2cc7eb67b4d2a3d462da0f966facdc5ee5/packages/eds-core-react/src/components/Icon/Icon.tsx#L138):
```typescript
titleId = `${icon.prefix}-${icon.name}-${count}`
```

Now, `count` comes from the call to `findIcon` on [Icon.tsx:108](https://github.com/sebastianvitterso/design-system/blob/69b53d2cc7eb67b4d2a3d462da0f966facdc5ee5/packages/eds-core-react/src/components/Icon/Icon.tsx#L108):
```typescript
const { icon, count } = findIcon(name, data, size)
```

`findIcon`, when the `data` prop isn't empty, returns count from `customIcon`, as we see on [Icon.tsx:57](https://github.com/sebastianvitterso/design-system/blob/69b53d2cc7eb67b4d2a3d462da0f966facdc5ee5/packages/eds-core-react/src/components/Icon/Icon.tsx#L57):
```typescript
let { icon, count } = data ? customIcon(data) : get(name)
```

`customIcon` simply returns a random number as its `count` value, which is where the problem comes from:
[Icon.tsx:50-53](https://github.com/sebastianvitterso/design-system/blob/69b53d2cc7eb67b4d2a3d462da0f966facdc5ee5/packages/eds-core-react/src/components/Icon/Icon.tsx#L50-L53)
```typescript
const customIcon = (icon: IconData) => ({
  icon,
  count: Math.floor(Math.random() * 1000),
})
```

### Proposed solution

Use React's `useId` hook (or implement it yourselves), as that is stable between server-side and client-side rendering, allowing us to simply do the following in the `Icon` component: 
```typescript
const id = useId()
const { icon } = findIcon(name, data, size)

// ...

if (title) {
  titleId = `${icon.prefix}-${icon.name}-${id}`
// ...
```

Now I do realize that you have to support `react < 18`, so I propose you build a stable `useId` replacement in that case. 

How about testing something like the following (based on your current `@equinor/eds-utils` `hooks/useId.ts`): 

```typescript
let counter = 0

export const useId = (idOverride: string, type?: string): string => {
  const [defaultId, setDefaultId] = useState(idOverride)
  const id = idOverride || defaultId

  useEffect(() => {
    if (defaultId == null) {
      setDefaultId(`eds-${type ? type + `-` : ''}${counter++}`)
    }
  }, [defaultId, type])
  return id
}
```

### Steps to reproduce the bug

1. Build a next.js app
2. Have an icon in the app, give it an `IconData` object as the `icon` prop and a string for the `title` attribute.
3. See error

The bug results in a warning/error message from next.js in dev mode: 
<a href="https://github.com/user-attachments/assets/783dfbd8-988b-4520-88ca-7e44e85528ce" target="_blank" >
  screenshot
</a>

### Expected behavior

No error message, as the id's should be equal on server and client.

### Specifications

- Version: `"@equinor/eds-core-react": "^0.42.5",`
- Browser: Chrome Version 131.0.6778.109 (Official Build) (arm64)
- OS: MacOS Sequoia 15.1.1 (24B91)

### Additional context

None

sebastianvitterso:
Does two things: 

1. Updates custom `useId` hook to be stable (actually just return an increasing number).
2. Make use of the updated `useId` hook in `Icon.tsx` for title id.

Closes #3705, see that for more details on "why".

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,5 +67,6 @@
     "prettier": "3.3.3",
     "typescript": "^5.5.4"
   },
-  "browserslist": "last 2 Chrome versions, last 2 firefox versions, last 2 safari versions, last 2 edge versions, not dead"
+  "browserslist": "last 2 Chrome versions, last 2 firefox versions, last 2 safari versions, last 2 edge versions, not dead",
+  "packageManager": "pnpm@9.15.0+sha1.8bfdb6d72b4d5fdf87d21d27f2bfbe2b21dd2629"
 }
--- a/packages/eds-core-react/src/components/Accordion/__snapshots__/Accordion.test.tsx.snap
+++ b/packages/eds-core-react/src/components/Accordion/__snapshots__/Accordion.test.tsx.snap
@@ -101,7 +101,7 @@ exports[`Accordion Matches snapshot 1`] = `
         class="c0"
       >
         <button
-          aria-controls="eds-accordion-12345-panel-1"
+          aria-controls="eds-accordion-0-panel-1"
           aria-expanded="true"
           class="c1"
           data-testid="header1"
@@ -133,9 +133,9 @@ exports[`Accordion Matches snapshot 1`] = `
         </button>
       </h2>
       <div
-        aria-labelledby="eds-accordion-12345-header-1"
+        aria-labelledby="eds-accordion-0-header-1"
         class="c5"
-        id="eds-accordion-12345-panel-1"
+        id="eds-accordion-0-panel-1"
         role="region"
       >
         Details 1
--- a/packages/eds-core-react/src/components/Accordion/__snapshots__/Accordion.test.tsx.snap
+++ b/packages/eds-core-react/src/components/Accordion/__snapshots__/Accordion.test.tsx.snap
@@ -101,7 +101,7 @@ exports[`Accordion Matches snapshot 1`] = `
         class="c0"
       >
         <button
-          aria-controls="eds-accordion-0-panel-1"
+          aria-controls="eds-accordion-1-panel-1"
           aria-expanded="true"
           class="c1"
           data-testid="header1"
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -67,5 +67,6 @@
     "prettier": "3.3.3",
     "typescript": "^5.5.4"
   },
-  "browserslist": "last 2 Chrome versions, last 2 firefox versions, last 2 safari versions, last 2 edge versions, not dead"
+  "browserslist": "last 2 Chrome versions, last 2 firefox versions, last 2 safari versions, last 2 edge versions, not dead",
+  "packageManager": "pnpm@9.15.0+sha1.8bfdb6d72b4d5fdf87d21d27f2bfbe2b21dd2629"
```
oddvernes, line 71: Should remove this packageManager field, i guess this was autogenerated by you IDE or something

oddvernes (changes requested):
Looks good, just fix the thw things i commented on 👍

oddvernes (approved):

PR #3716, opened 2025-01-03
🧑‍💻add 'use client' to relevant index files

### LINKED ISSUE #3702: feat: support next.js out-of-the-box
### Is your feature request related to a problem? Please describe.

Currently, using the library in a Next.js-application is a bit of a hassle. In short, it simply does not allow them. There are workarounds, and generally they still require users to put the styled components inside `"use client"` components. 

A dear friend of ours, however, recommended a solution like this:

```typescript
"use client"
import { Button as EdsButton } from '@equinor/eds-core-react'

export Button as styled(EdsButton)``
```

In their project, that worked well.

We experimented further with this, and found that this works just as well: 

```typescript
"use client"
export { Button } from '@equinor/eds-core-react'
```

Further experimentation found that we could also just do: 

```typescript
"use client"
export * from '@equinor/eds-core-react'
```

Which let us just import EDS components from this file instead of directly from `@equinor/eds-core-react`.

I've made PR #3701 to solve this exact issue. However, because `npm run build` runs rollup before this is published as an npm package, all directives are stripped away, so the `"use client"` addition is removed in the js bundle. 

A simple fix to this is to use [Ephem/rollup-plugin-preserve-directives](https://github.com/Ephem/rollup-plugin-preserve-directives), (found [here](https://github.com/rollup/rollup/issues/4699#issuecomment-1465302665)), which seems to be an alright rollup-plugin for keeping directives in the bundles.

### Describe the solution you'd like

`"use client"` at the top of `index.js` in the bundle.

This requires:
1. `"use client"` at the top of `index.ts` in the repo.
2. Rollup-plugin for keeping directives in bundled code.

### Describe alternatives you've considered

No others, I suppose. 

### Additional context

[A reddit thread shortly describing nextjs, SSR and styled-components](https://www.reddit.com/r/nextjs/comments/15bvkx2/comment/jtsojqm/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button).

oddvernes:
resolves #3702 

it was reported that 'use client' on the main index file did not "inherit" down to sub-components. This pr adds 'use client' to the index of components with sub-components. I don't have a test setup to check if this works so I'll make a dev release for  @sebastianvitterso to check

### DIFF
--- a/packages/eds-core-react/package.json
+++ b/packages/eds-core-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@equinor/eds-core-react",
-  "version": "0.42.5",
+  "version": "0.43.0-dev-01062025",
   "description": "The React implementation of the Equinor Design System",
   "sideEffects": [
     "**/*.css"
--- a/packages/eds-core-react/src/components/Accordion/index.ts
+++ b/packages/eds-core-react/src/components/Accordion/index.ts
@@ -1,3 +1,4 @@
+'use client'
 import { Accordion as BaseAccordion } from './Accordion'
 import { AccordionItem, AccordionItemProps } from './AccordionItem'
 import { AccordionHeader, AccordionHeaderProps } from './AccordionHeader'
--- a/packages/eds-core-react/src/components/Banner/index.ts
+++ b/packages/eds-core-react/src/components/Banner/index.ts
@@ -1,3 +1,4 @@
+'use client'
 import { Banner as BaseBanner, BannerProps } from './Banner'
 import { BannerIcon, BannerIconProps } from './BannerIcon'
 import { BannerMessage, BannerMessageProps } from './BannerMessage'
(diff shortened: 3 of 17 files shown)


torleifhalseth (approved):
LGTM

sebastianvitterso:
I opened an issue in the Nextjs repo https://github.com/vercel/next.js/issues/74585 so they can help. Seems to be a very strange edge-case we're hitting here.

oddvernes:
For reference, @sebastianvitterso found [this](https://ivicabatinic.from.hr/posts/multipart-namespace-components-addressing-rsc-and-dot-notation-issues) article explaining the issue at hand. So to fully support ssr we would need to either change to the pattern descibed in this article (big breaking change) or make the subcomponents available for import individually (maybe ideally imported from a sub-path and not root directly to avoid chaos)

PR #3718, opened 2025-01-07
EDS Data Grid Feature: +onRowDoubleClick event

oleksiishv:
I've added a new event to the EDS Data Grid: `onRowDoubleClick`
It propagates `onDoubleClick` event on the row. Works the same way as `onRowClick` event


oddvernes (approved):
LGTM!

PR #3725, opened 2025-01-29
🧱 Export subcomponents individually for ssr support

### LINKED ISSUE #3723: ✨Add support for composite components in nextjs
Ref this issue https://github.com/equinor/design-system/issues/3702, our initial attempt to add 'use client' would not work in next.js, instead we discovered that sub components (such as Table.Header) are undefined. 

@sebastianvitterso started [this](https://github.com/vercel/next.js/issues/74585) discussion at vercel and then discovered [this](https://ivicabatinic.from.hr/posts/multipart-namespace-components-addressing-rsc-and-dot-notation-issues) article which gets to the bottom of the problem and proposes a couple of solutions. 

I think most realistically, exporting the subcomponents individually is how we should go about it. This is also how they've done it over at Designsystemet ([example](https://github.com/digdir/designsystemet/blob/next/packages/react/src/components/Details/index.ts)). 

Note also that individual exports have been [requested](https://github.com/equinor/design-system/issues/3038) in the past, but then the reasoning was to easier create jest tests downstream, so we denied the request for being not worth the effort. However, supporting ssr is perhaps a bit more important, so we should reconsider.

oddvernes:
resolves #3723

### DIFF
--- a/packages/eds-core-react/package.json
+++ b/packages/eds-core-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@equinor/eds-core-react",
-  "version": "0.44.0-dev-01292025",
+  "version": "0.43.0",
   "description": "The React implementation of the Equinor Design System",
   "sideEffects": [
     "**/*.css"
--- a/packages/eds-core-react/package.json
+++ b/packages/eds-core-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@equinor/eds-core-react",
-  "version": "0.43.0",
+  "version": "0.44.0-dev-01292025",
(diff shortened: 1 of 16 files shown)


oddvernes:
A note: we have discovered that while this works for static components such as Card, it does not work with some components with internal post-processing of children (ReactChildren.map/cloneElement), such as Menu and Accordion

sebastianvitterso (approved):
Looks good, and after testing, it works well too.

sebastianvitterso:
Nice! Now we're just eagerly awaiting a new release 😄

PR #3728, opened 2025-02-05
🐛 Sidesheet: fix shrinking close button

### LINKED ISSUE #3727: Sidesheet close-button shrinks when a long title is given
**Describe the bug**

The circular icon-button shrinks horizontally, making it oblong. Doesn't look very good.

Image

1. Sidesheet
2. Long title
3. Tadaa

**Expected behavior**

Circular button

**Specifications**
N/A

**Additional context**
This is quickly fixed by adding `flex-shrink: 0` to the button element, but we don't have access to that through the SideSheet.

oddvernes:
resolves #3727


sebastianvitterso (approved):
Doesn't take much!

PR #3730, opened 2025-02-07
feat: make TextField id optional

mhwaage:
* Allows users to omit creating their own Id for Textfield by falling back to a generated id

### DIFF
--- a/packages/eds-core-react/src/components/TextField/TextField.tsx
+++ b/packages/eds-core-react/src/components/TextField/TextField.tsx
@@ -81,7 +81,7 @@ export const TextField = forwardRef<HTMLDivElement, TextFieldProps>(
     },
     ref,
   ) {
-    const id = useId(_id, "input");
+    const id = useId(_id, 'input')
     const helperTextId = useId(null, 'helpertext')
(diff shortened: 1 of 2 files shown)


packages/eds-core-react/src/components/TextField/TextField.tsx
```diff
@@ -29,7 +29,7 @@ type SharedTextFieldProps = {
   /** Variants */
   variant?: Variants
   /** Input unique id. This is required to ensure accesibility */
```
oddvernes, line 31: Should update this comment which is used by storybook propstable. Perhaps just remove the entire comment

torleifhalseth (approved):
Thank you for your contribution, @mhwaage  — We appreciate you taking the time to improve this! I’ve been hesitant about adding this feature when the maintenance cost might outweigh the value. Since we already use and maintain `useId` in our library, we’ve encountered some challenges with this hook, especially in server-side rendering.  

After discussing this with @oddvernes , we agreed that supporting this feature makes sense. It would also be great to move toward using the `useId` provided directly by React. Thanks again for your work on this — it’s very much appreciated!

PR #3734, opened 2025-02-14
feat: adds loading state to Button controlled by loading prop

FredrikMWold:
This PR enhances the Button component by adding a new `loading` prop.  

When `loading` is set to `true`, the button displays a `Progress.Circular` indicator, making it easier to show a loading state.

### DIFF
--- a/packages/eds-core-react/src/components/Button/Button.stories.tsx
+++ b/packages/eds-core-react/src/components/Button/Button.stories.tsx
@@ -1,16 +1,16 @@
-import { useState, ChangeEvent } from 'react'
+import { add, menu, refresh, save } from '@equinor/eds-icons'
+import { Meta, StoryFn } from '@storybook/react'
+import { ChangeEvent, useState } from 'react'
 import {
   Button,
   ButtonProps,
-  Icon,
-  EdsProvider,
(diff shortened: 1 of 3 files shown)


FredrikMWold:
Could make the button keep its size, when showing the loading state. 
Did not do that because in the examples they have. The button size does change.

torleifhalseth (changes requested):
Hey @FredrikMWold,  

Thank you for the contribution. We appreciate the effort and thought you’ve put into this! Simplifying the loading state for buttons is helpful, and we see the appeal of making it easier to implement.  

That said, we aim to keep our components as composable and flexible as possible, and adding a specific `loading` prop would tie the Button to one particular implementation. Since we already have Storybook examples showing how to build a loading button with both dots and circular/spinner progress components, we feel that it keeps the right balance between flexibility and ease of use.  

@magnh made a great point about maintaining the button’s width to prevent layout shifts — that’s a nice touch. It could be a great addition to our existing examples! Defining the color of the progress component could also make those examples more functional.  

If you’re up for it, we’d love a contribution focused on improving the examples with these ideas! Thanks again for your work  🚀

PR #3749, opened 2025-03-10
🐛 useHideBodyScroll: Refactor for better reliability

### LINKED ISSUE #3746: Body styles not resetting correctly for multiple Dialogs
**Describe the bug**

The styling in the body is not cleared correctly when closing a Dialog, if multiple Dialogs have been open simultaneously. This means that the body will have overflow: "hidden", so all scroll is disabled. 

1. Open a Dialog.
2. Open another Dialog (they don't need to be nested in the HTML).
3. Close one Dialog.
4. Close the other dialog.

The body now still has style="overflow: hidden; padding-right: 15px;".

**Expected behavior**

I would expect the body's style to have returned to normal. 

**Specifications**

- Version: 0.42.5
- Browser: Microsoft Edge
- OS: Windows 11

**Additional context**

To try to get around this problem, I originally tryed to reset the styles manually in the onClose-handler of the last dialog i closed. The styles where then reset, but then set back to "overflow: hidden; padding-right: 15px;".

pomfrida:
Resolves #3746 
This PR refactors our scroll locking hook to improve reliability and fix edge cases related to multiple overlapping dialogs.

## Key Changes:

- **Improved state tracking**: Added a `wasActive` ref to track the active state of each hook instance
- **Centralized state management**: Moved from instance-specific state to a shared dialog state object
- **Fixed cleanup logic**: Ensured cleanup function properly decrements the counter and restores scroll state
- **Added protection against race conditions**: Added Math.max to prevent negative counter values
- **Better handling of transitions**: Added explicit handling of active-to-inactive transitions while component is mounted

## Technical Implementation:

- Used a shared module-scoped object to track dialog state across instances
- Added proper active state tracking per hook instance to prevent double-counting
- Fixed edge cases with overlapping dialogs

The code has been manually tested with multiple dialogs and scrims to verify proper behavior in all scenarios.


oddvernes (approved):
LGTM!

PR #3754, opened 2025-03-20
🐛 DatePicker & DateRangePicker: Ensure consistent styling on state changes

### LINKED ISSUE #3747: Datepicker - When enabling from a disabled state, the calender icon is placed under the date
**Describe the bug**

1. Set the datepicker to disabled by default
2. Have logic that enables the datepicker by action
3. The calendar icon will now be below the date

**Expected behavior**

The calendar icon should always be at the same spot on the right of the date.

**Specifications**

- Version: 133.0.3065.92
- Browser: Edge 
- OS: Windows 10

Also reproduced in Firefox

**Additional context**

|State|Picture|Comment|
|-|-|-|
|From disabled to enabled|Image|The icon is moved underneath the date|
|Never disabled|Image|The icon is at its rightful spot|
|Temp fix|Image|Added custom style to move the element to the right. The text and icon is still not the same|

Code: https://github.com/equinor/surveymgmt-app/pull/529/files#diff-7f81251e760a8c8e35b0a39c48601805a73959f1f1e234252158f8b823cb5f99

A bit of code

```typescript

   import { DatePicker } from '@equinor/eds-core-react';
   import styled from '@emotion/styled';


    const onChangeExpiryDate = (setDate: Date | null, dirty: boolean | null) => {
        // adjust for timezones so the timer will not move the selected date (returns the negative value for +01:00 ex)
        let todaysDateUtc = new Date();
        let maxDateUtc = addDays(todaysDate,6);

        // todaysDateUtc must be set to 0 hours, 0 minutes etc because of the if condition below. 
        // Otherwise the incoming date is lower than todays date because the time of day.
        todaysDateUtc.setHours(0,0,0,0);
        todaysDateUtc.setUTCMinutes(todaysDateUtc.getTimezoneOffset() * -1);
        maxDateUtc.setUTCMinutes(maxDateUtc.getTimezoneOffset() * -1);
        setDate?.setUTCMinutes(setDate.getTimezoneOffset() * -1);

        if (setDate != null && setDate <= maxDateUtc && setDate >= todaysDateUtc) {
            const stringDate = setDate?.toISOString();
            if (stringDate != null) {
                if(dirty === null || dirty === true){
                    setValue('expiryDate', stringDate, { shouldDirty: true });
                }
                else {
                    setValue('expiryDate', stringDate, { shouldDirty: false });
                }
            }
        }
    };

  <DatePickerStyled 
                id='expiry-date'
                label='Expiry date'
                value={initalValueDate()}
                minValue={new Date(todaysDate.toISOString().split('T')[0])}
                maxValue={new Date(maxDate.toISOString().split('T')[0])}
                style={{ width: '40%' }}
                onChange={(e) =>onChangeExpiryDate(e, true)}
                hideClearButton={true}
                timezone='UTC'
                disabled={values.contactPersons?.length === 0}
            />
    
    // temp fix:
    const DatePickerStyled = styled(DatePicker)({
    '& > div': { display: 'flex' },
});



```

pomfrida:
Resolves #3747 
This fixes the misalignment of the calendar icon in `DatePicker` when toggling between disabled and enabled states and ensures the correct font is applied consistently.

## Key Changes:
- Fixed icon offset: The calendar icon no longer shifts position when enabling `DatePicker` from a disabled state.
- Ensured consistent font-family: The date input now retains the correct font-family after state changes.

## Technical Implementation:
- Preserved `Toggle` in the DOM by applying `visibility: hidden` instead of removing it when `disabled={true}`.
- Explicitly set font-family in `DateFieldSegments` to prevent incorrect font inheritance after enabling `DatePicker`.
- Ensured layout stability by maintaining space for `Toggle`, preventing flexbox or grid shifts when re-enabling the component.

The fix has been manually tested in Storybook to verify that the icon remains correctly aligned and the font remains consistent across state transitions.

### DIFF
--- a/packages/eds-core-react/src/components/Datepicker/fields/DateFieldSegments.tsx
+++ b/packages/eds-core-react/src/components/Datepicker/fields/DateFieldSegments.tsx
@@ -33,7 +33,14 @@ export const DateFieldSegments = forwardRef(
     )
 
     return (
-      <div {...fieldProps} style={{ display: 'flex' }} ref={ref}>
+      <div
+        {...fieldProps}
+        style={{
+          display: 'flex',
+          fontFamily: 'Equinor, Arial, sans-serif',
+        }}
+        ref={ref}
+      >
         {state.segments.map((segment, i) => (
           <DateSegment key={i} segment={segment} state={state} />
         ))}
--- a/packages/eds-core-react/src/components/Datepicker/fields/Toggle.tsx
+++ b/packages/eds-core-react/src/components/Datepicker/fields/Toggle.tsx
@@ -35,8 +35,8 @@ export const Toggle = ({
   valueString: string
   showClearButton: boolean
 }) => {
-  return readonly || disabled ? null : (
-    <>
+  return (
+    <div style={{ visibility: readonly || disabled ? 'hidden' : 'visible' }}>
       {showClearButton && (
         <StyledButton
           disabled={disabled}
@@ -69,6 +69,6 @@ export const Toggle = ({
       >
         <Icon data={icon} />
       </StyledButton>
-    </>
+    </div>
   )
 }


oddvernes (approved):
LGTM!

Issue #3259, opened 2024-02-06
Autocomplete: Duplicate entries in selectedItems for controlled Autocomplete


tomktjemsland:
**Describe the bug**

I am trying to replace an old custom drop-down solution with Autocomplete. The Autocomplete component is _Controlled_ with `multiple` selections enabled. Options are a list of objects on the format: `{ id: 'Aker Barents', selected: true, title: 'Aker Barents' }`.  Enabling options works as intended, but when attempting to disable options `selectedItems` given `onOptionsChange` has duplicate entries instead!

**Steps to reproduce the bug**

1. Create a controlled Autocomplete component with `multiple` enabled.
2. Set the `options` property to a list of objects with a `selected` flag available.
3. Filter objects on `selected === true` and use this as data for the  `selectedOptions` property.
4. Log `selectedItems` available in `onOptionsChange`.

**Expected behavior**

I expect objects to disappear from `selectedItems` when I try to disable them in the drop-down.

**Specifications**

- Version: 0.35.1
- Browser: Chrome
- OS: Windows

**Additional context**

![image](https://github.com/equinor/design-system/assets/48794877/3fc2b809-f1b3-495d-8b40-27badcd65401)

oddvernes:
Thanks for the report! We will investigate this issue further

BirteThornquist:
I understand you are already looking into this @yusijs

yusijs:
> I understand you are already looking into this @yusijs 

Perhaps - if my thinking is not to optimistic. :)

oddvernes:
A possible clue https://github.com/downshift-js/downshift/blob/master/src/hooks/useCombobox/README.md#itemtokey

oddvernes:
This should be fixed by #3455, by using the new `itemCompare` prop. Please try it out in the new release (0.39.0) and let us know if it works!

mhwaage:
This is still broken on 0.41.0

yusijs:
> The intended functionality is still broken in 0.41.0.

Just want to point to the title of my commit/pr 😅:
[Possible solution for object-checking](https://github.com/equinor/design-system/commit/e8c79fc16c38e8df0d546168bcaf3cdf3b241923)


> e: and yet, I am not able to reproduce that when trying to make a unit test 😢
I know the feeling, it was a pain to nail the behaviour down..

mhwaage:
OK, I am able to reproduce this in a development situation, but not usre the best way of exposing it for a unit test.
[...]

PR #3600, opened 2024-08-28
Bugfix/deselect multiple does not work with complex items

mhwaage:
Blind attempt at fixing edge case of #3259

PR #3763, opened 2025-04-04
expose itemToKey, add deprecation notice to itemCompare

mhwaage:
this is yet-another blind fix for issue #3259 - this time, intended to fix an issue where sometimes, downshift doesn't even trigger the onSelectedItemsChange function

### DIFF
--- a/packages/eds-core-react/src/components/Autocomplete/Autocomplete.stories.tsx
+++ b/packages/eds-core-react/src/components/Autocomplete/Autocomplete.stories.tsx
@@ -281,8 +281,8 @@ export const OptionComponent: StoryFn<AutocompleteProps<MyOptionType>> = (
         initialSelectedOptions={
           [JSON.parse(JSON.stringify(options[1]))] as MyOptionType[]
         }
-        itemCompare={(item, compare) => {
-          return item.label === compare.label
+        itemToKey={(item) => {
+          return item.label
         }}
         multiline
       />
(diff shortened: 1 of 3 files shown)


mhwaage:
Frustratingly, this is exactly the kind of stuff the unit test i added in #3600 was supposed to catch, and yet here we are. I suppose that the somewhat simplistic objects results in the reuse of object references?

mhwaage:
regarding deprecating `itemCompare`:
A [naive search ](https://github.com/search?q=org%3Aequinor+itemCompare&type=code)across the equinor org, only reveals 48 occurrences of the _term_. Inspecting those, the most problematic-looking ones (according to me) are:
https://github.com/equinor/petec-well-experience/blob/2ae3011f2b9cb6af08d6e0332605f32e7c96df50/src/pages/admin/Wellbores/WellTable/Autocomplete.tsx#L62
https://github.com/equinor/slots-transparency/blob/0875692620a6ac641538593a3026475c32239cd3/frontend/src/components/intervals/search/SearchCriteriaInput.tsx#L81
https://github.com/equinor/prix/blob/08dae6a3636d6de30561f468615e6e0702977a5e/src/Ras.Ui/ClientApp/src/components/features/Filter/components/SingleSelectFilter.tsx#L26

all of which should be possible to convert to use `itemToKey` as far as I can tell.

pomfrida:
Hi @mhwaage, thanks for your contribution – and extra kudos for following up on your own bug report! 🙌

One thing to make sure of: we should update the Storybook examples to reflect the new `itemToKey` prop to help guide others in the transition. Would you be up for adding that to this PR? 

Looks good overall – thanks again! 🙏

pomfrida (approved):
LGTM!

PR #3764, opened 2025-04-08
feat(datagrid): Enable external filters, hide filter-icon when not ac…

yusijs:
Hiding the icon was discussed w/ @BirteThornquist a while back by email. Note that this only applies to the built-in filters, and not custom filters.

image


yusijs:
I also experienced some issues w/ running the storybook locally due to version-discrepancies in storybook. Something broke between 8.4.7 and 8.6.12 (addons were at 8.6.12 locally, while storybook was on 8.4.7, which broke due to some internal path resolutions)

pomfrida:
Thanks for your contribution @yusijs ! 🙌

Just a couple of accessibility observations after testing in Storybook:
- When the filter icon is hidden it’s still included in the tab order. This results in a keyboard focus on a visually hidden element which is not compliant with [WCAG 2.4.7 (Focus Visible)](https://www.w3.org/WAI/WCAG22/Understanding/focus-visible.html).
- If you press Enter on the invisible element without applying any filter, the icon remains visible afterwards even though no filtering has taken place. This might feel inconsistent from a user perspective.

To improve accessibility and UX, I would suggest:
1. making the filter icon visible when it receives keyboard focus so the user can see what’s focused before interacting.
2. ensuring the icon stays hidden unless a filter is actually applied and not just opened.

Let us know what you think, and thanks again for the nice work on this! 👏

PR #3777, opened 2025-04-23
fix(Autocomplete, TextField): moves refs from containers to input element to fix scroll to field error with react hook form

FredrikMWold:
## Summary

This pull request addresses an issue with integrating the `Autocomplete` and `TextField` components with `react-hook-form`.

## Problem

When using these components, `react-hook-form`'s scroll-to-error functionality does not behave correctly. This is because `react-hook-form` expects refs to be attached directly to the underlying `<input>` elements. In the current implementation, the refs are attached to container elements instead, which breaks this behavior.

## Solution

Refs are now correctly forwarded to the input elements themselves. This ensures that `react-hook-form` can properly detect and scroll to fields with validation errors.

## Additional Notes

A quick demo has been added to the `Autocomplete` + `react-hook-form` story to help you test and verify the fix.


FredrikMWold:
@torleifhalseth 
I removed some more of the pr demo code from the autocomplete story.

Do you also want to keep the max-height that is on the react-hook-form example?

torleifhalseth (approved):
Yes, that makes sense to test the scroll behaviour 👍

PR #3782, opened 2025-04-25
fix(eds-core-react, eds-data-grid-react): :bug: prevent triggering sorting on resize

torleifhalseth:
update resizer component to prevent mouse events from triggering sorting


pomfrida (approved):
LGTM 👍

PR #3788, opened 2025-05-06
fix: removes 'item is:' console log from autocomplete

FredrikMWold:
This pr removes the "item is: " console.log from autocomplete as it is causing a lot of logging in out applications
image


pomfrida (approved):
Thanks a lot for cleaning up our mess 🙌 🥇

PR #3799, opened 2025-05-09
feat(eds-core-react): :sparkles: Export Autocomplete Option Label

### LINKED ISSUE #3665: Expose the AutoComplete Option Label

My primary goal to disable certain options, and also apply a Popover (https://storybook.eds.equinor.com/?path=/docs/data-display-popover--docs) to the disabled elements. We want the Popover to explain to the end user _why_ the option is disabled.

The disabling part works fine using the AutoComplete.optionDisabled-prop. On the other hand, to apply the Popover on the disabled options we were thinking to make use of the AutoComplete.optionComponent-prop to provide a custom option that uses the Popover. Naively this works fine, but the problem is that the default option [Label component](https://github.com/equinor/design-system/blob/develop/packages/eds-core-react/src/components/Autocomplete/Option.tsx#L44) has a bunch of styling that we would really want to retain on our slighty modified Option-component. The mentioned Label is not exposed, so we cannot wrap it in our `optionComponent`-code.

We would like the Autocomplete Option Label component (https://github.com/equinor/design-system/blob/develop/packages/eds-core-react/src/components/Autocomplete/Option.tsx#L44) to be exported, such that we can wrap it in our Popover, but still retain the default Autocomplete styling (that you maintain).

Alternatively you could consider providing other ways we can achieve our described goals of explaining to the end user _why_ an option is disabled.

An alternative approach for us to get the behavior we want is also to duplicate the code here https://github.com/equinor/design-system/blob/develop/packages/eds-core-react/src/components/Autocomplete/Option.tsx#L44 over into our own app, but we would rather try to avoid that.

pomfrida:
Resolves #3665

### DIFF
--- a/packages/eds-core-react/src/components/Autocomplete/Option.tsx
+++ b/packages/eds-core-react/src/components/Autocomplete/Option.tsx
@@ -41,19 +41,21 @@ const StyledListItem = styled.li<StyledListItemType>(
   },
 )
 
-const Label = styled.span<{ $multiline: boolean }>(({ theme, $multiline }) => {
-  return css`
-    ${spacingsTemplate(theme.entities.label.spacings)}
-    text-overflow: ellipsis;
-    white-space: ${$multiline ? 'normal' : 'nowrap'};
-    overflow: ${$multiline ? 'initial' : 'hidden'};
-    overflow-wrap: anywhere;
-    /* hack to fix clipping issue in firefox (#3170) */
-    @supports (-moz-appearance: none) {
-      overflow: ${$multiline ? 'initial' : 'clip'};
-    }
-  `
-})
+export const AutocompleteOptionLabel = styled.span<{ $multiline: boolean }>(
+  ({ theme, $multiline }) => {
+    return css`
+      ${spacingsTemplate(theme.entities.label.spacings)}
+      text-overflow: ellipsis;
+      white-space: ${$multiline ? 'normal' : 'nowrap'};
+      overflow: ${$multiline ? 'initial' : 'hidden'};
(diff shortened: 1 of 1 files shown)


packages/eds-core-react/src/components/Autocomplete/Option.tsx
```diff
@@ -41,19 +41,21 @@ const StyledListItem = styled.li<StyledListItemType>(
...
-    white-space: ${$multiline ? 'normal' : 'nowrap'};
-    overflow: ${$multiline ? 'initial' : 'hidden'};
-    overflow-wrap: anywhere;
-    /* hack to fix clipping issue in firefox (#3170) */
-    @supports (-moz-appearance: none) {
-      overflow: ${$multiline ? 'initial' : 'clip'};
-    }
-  `
-})
+export const AutocompleteOptionLabel = styled.span<{ $multiline: boolean }>(
```
torleifhalseth, line 44: Would it be simpler to name it Label? What is the downside of having a generic name?

torleifhalseth (approved):
LGTM 🚢  

Added a small comment, but you can decide if it makes sense to adjust.

packages/eds-core-react/src/components/Autocomplete/Option.tsx
pomfrida: It would be simpler, and Label was the original name when it wasn't exported. But I wanted to follow the naming of the other exported variables in the file - like  "AutocompleteOption". And since we also have a different Label component I thought it was best to differentiate the two.

PR #3804, opened 2025-05-14
chore(eds-tokens-build, eds-tokens-sync): ⬆️ Upgrade dependencies from dependabot

pomfrida:
resolves #3798 

I did not downgrade @babel/runtime as dependabot suggested because it did not seem like it knew the lates version we were on in its description anyway. It thought we were on 7.24.0 but actually we were already on 7.27.1


torleifhalseth (approved):
King kong 🦍

PR #3805, opened 2025-05-14
feat(eds-core-react): ✨ Banner.Message - Allow ReactNode as children

### LINKED ISSUE #3801: Allow children of Banner.Message to be ReactNode

Currently the component `Banner.Message` only accept 'string' as children. We have the need to express a richer `Banner.Message` content. Example of our usage:

Image

Today we forcefully override the 'string' limitation of the `Banner.Message` children to pass in `ReactNode` as children, instead of just string.

Change type of `Banner.Message` `children` prop to be `ReactNode`.

We currently override the `string` limitation by casting our `ReactNode` to `string`:
```
<Banner>
  <Banner.Message>
    {<div>my custom stuff</div> as unknown as string} 
  </Banner.Message>
</Banner>
```

This "solution" is a bit dangerous, since this "happen to work" now, but might break in the future if `Banner.Message` start doing some `string` specific operations on the `children`.

pomfrida:
Resolves #3801

### DIFF
--- a/packages/eds-core-react/src/components/Banner/Banner.docs.mdx
+++ b/packages/eds-core-react/src/components/Banner/Banner.docs.mdx
@@ -67,3 +67,19 @@ Banners can supplement their message using a supporting icon.
 
 <Canvas of={ComponentStories.TextAndIconAndAction} />
 
+### Complex Banner Message (special cases)
+
+While using string content for Banner.Message is the recommended approach for most use cases, there are situations where more complex content might be necessary. For these special cases, Banner.Message now supports ReactNode children.
+
+<ul>
+  <li>Use string content for Banner.Message whenever possible to maintain consistency and proper styling.</li>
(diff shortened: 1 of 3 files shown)


millus (approved):
Looks great and solves the issue! Just a thought, would it be useful to also have a story showcasing/informing about this new feature allowing other content than just string as BannerMessage?

torleifhalseth (approved):
LGTM ✨

pomfrida:
> Looks great and solves the issue! Just a thought, would it be useful to also have a story showcasing/informing about this new feature allowing other content than just string as BannerMessage?

Thanks @millus  You make a good point! 🙌 
While this change was primarily made to support a specific use case for one of our users, you're right that we should document this capability. I'll add a story that demonstrates using ReactNode as children, and explain that string content remains the recommended approach for most cases, as we added ReactNode support for special cases where more complex content is needed.

This way, we can maintain our existing best practices while still documenting the new flexibility for those who need it.

PR #3807, opened 2025-05-15
feat(eds-data-grid-react): 🔥 remove redundant placeholder in DebouncedInput

### LINKED ISSUE #3756: Prefill filter values with column's min/max in EDS Data Grid

Our users were confused that filtering values have placeholder value 0 in the EDS data grid. They wondered if the filtering values were defined as from 0 to 0, which shouldn't give any results.

I would suggest leaving the spaces empty or have the min/max values in the column as placeholders (The same as in parenthesis above filtering input fields, these are quite useful for us!)

**Additional context**

I thought of just fixing it in our app, but perhaps this is something useful for more than just our project?
Image

pomfrida:
Resolves #3756


torleifhalseth (approved):
LGTM 🔥

Issue #3825, opened 2025-06-03
Datagrid - dynamic row height scrolling issue


arkadiy93:
**Describe the bug**

Scrolling issues with dynamic row height.

**Steps to reproduce the bug**

1. Insert content where rows have dynamic height
2.Scroll

**Expected behavior**

Expect no flickering

https://github.com/user-attachments/assets/82a8929f-b968-45da-a6e0-460dbebcc148

PR #3840, opened 2025-06-13
datagrid issue - virtual table and rows with dynamic height

arkadiy93:
https://github.com/equinor/design-system/issues/3825

### DIFF
--- a/packages/eds-data-grid-react/src/EdsDataGrid.tsx
+++ b/packages/eds-data-grid-react/src/EdsDataGrid.tsx
@@ -382,10 +382,16 @@ function EdsDataGridInner<T>(
     return density === 'compact' ? 32 : 48
   }, [density])
 
-  const virtualizer = useVirtualizer({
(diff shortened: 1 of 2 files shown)


pomfrida (changes requested):
Great work! 🙏 And apologies for the late review.. 
This solves an important problem and we really appreciate you tackling this.

As a design system, we need to maintain high quality since many teams depend on these components. Could you add some unit tests for:

- `isFirefox()` detection
- Dynamic height measurement (happy path) 
- Edge case where element has no height (returns 0)

Example of what we're looking for:
```typescript
describe('Dynamic row height', () => {
  it('measures element height when not Firefox', () => {
    // Mock isFirefox to return false
    // Create element with known height
    // Verify measurement is correct
  });
  
  it('skips measurement in Firefox', () => {
    // Mock isFirefox to return true  
    // Verify measureElement is undefined
  });
 
  
  it('returns 0 for hidden elements', () => {
    // Create hidden element
    // Verify returns 0 (expected behaviour)
  });
});

pomfrida (approved):
Excellent! 👏

PR #3841, opened 2025-06-13
feat(eds-core-react): ✨ enhance Autocomplete onAddNewOption - stories, tests, and accessibility

### LINKED ISSUE #3794: Add support for user-defined options in Autocomplete

We currently use the Autocomplete component for predefined select options. However, there is no way to allow users to input and select values that are not part of the predefined list. This becomes a limitation when a user needs to add their own value on the fly, similar to the [freeSolo](https://mui.com/material-ui/react-autocomplete/#free-solo) behaviour in Material-UI. 

**Describe the solution you’d like**

Autocomplete component should support a “free solo” mode where users can input custom values that are not part of the current options. Ideally, when the user types a new value and selects it (e.g. by pressing Enter), it should be treated as a valid selection and optionally added to the list.

**Describe alternatives you’ve considered**

Users could consider adding a separate text input field, but it breaks the unified experience of using a single component for both predefined and custom input. Another option would be maintaining a local state that merges new user input into the options array, but that adds significant complexity on the implementer’s side.

**Additional context**

The request originated from a user (Anand Prakash Shukla) via Slack, who asked for a feature similar to Material-UI’s freeSolo prop.

pomfrida:
Resolves #3794 

Adding to #3833

### DIFF
--- a/packages/eds-core-react/src/components/Autocomplete/Autocomplete.test.tsx
+++ b/packages/eds-core-react/src/components/Autocomplete/Autocomplete.test.tsx
@@ -81,6 +81,10 @@ const waitForVirtualizedOptions = async (
   return validOptions.slice(0, expectedCount)
 }
 
+const StyledAutocomplete = styled(Autocomplete)`
+  clip-path: unset;
+`
+
 describe('Autocomplete', () => {
   it('Matches snapshot', async () => {
     render(<Autocomplete options={items} label={labelText} />)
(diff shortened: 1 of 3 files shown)


packages/eds-core-react/src/components/Autocomplete/Autocomplete.test.tsx
```diff
@@ -81,6 +81,10 @@ const waitForVirtualizedOptions = async (
   return validOptions.slice(0, expectedCount)
 }
 
+const StyledAutocomplete = styled(Autocomplete)`
+  clip-path: unset;
+`
```
torleifhalseth, line 86: Is this ensuring that the component is not visually clipped and behaves as expected in the test environment? If so maybe we should add a small comment on it.

packages/eds-core-react/src/components/Autocomplete/Autocomplete.test.tsx
```diff
@@ -571,3 +537,236 @@ describe('Autocomplete', () => {
     expect(autocomplete.nodeName).toBe('INPUT')
   })
 })
+describe('Autocomplete: Add new options feature', () => {
+  it('Can add new options', async () => {
```
torleifhalseth, line 541: Great work with the test 👍

packages/eds-core-react/src/components/Autocomplete/Autocomplete.test.tsx
```diff
@@ -571,3 +537,236 @@ describe('Autocomplete', () => {
...
+    const buttonNode = await screen.findByLabelText('toggle options', {
+      selector: 'button',
+    })
+    fireEvent.click(buttonNode)
+
+    const addOption = screen.queryByTestId('add-item')
+    expect(addOption).not.toBeInTheDocument()
+  })
+
+  it('Shows add option only when no matches are found', async () => {
```
torleifhalseth, line 594: From a UX perspective, I worry that it could be confusing when the “ add new option” feature is not available. But we can iterate and improve this in the future!

torleifhalseth (approved):
Great work on the tests 🎖️

packages/eds-core-react/src/components/Autocomplete/Autocomplete.test.tsx
pomfrida: Tbh I have no idea, I just had to move it out to be able to use it in my dedicated test block 😅

PR #3868, opened 2025-06-27
feat(eds-data-grid-react): ✨ New `ClickableCell` component

### LINKED ISSUE #3839: Data grid: `onCellClick` defined but not exposed

It is currently not possible to use the cell click handler `onCellClick()` in EDS Data grid.

E.g., the following code in the EdsDataGrid-component does not log to the console:

```javascript
onCellClick={() => console.log("cell clicked!")}
```
It is apparently passed to `TableBodyCell`, but is ignored from there.

`onCellClick` would work similarly to `onRowClick`

For the time being, I have created a custom cell component with an `onClick` handler:

```javascript

cell: ({ row }: { row: Row<{ [key: string]: string }> }) => (
  <div onClick={() =>handleCellClick(Number(col.id))}>
    <Typography>{row.original[`${col.id}`]}</Typography>
  </div>
),
```

but this is suboptimal in my use case.

**Additional context**

Originally mentioned on slack here: https://equinor.slack.com/archives/CJT20H1B9/p1748436933075109

pomfrida:
## ✨ New ClickableCell Component
This PR introduces a new ClickableCell component that enables interactive cells within the EDS DataGrid, following the design specifications from [Figma Desktop Design](https://www.figma.com/design/luNobpDIWTuQ9rhyovGLBI/Desktop?node-id=5112-91759&t=RrxLyUDIlm1EqiMe-4).

### 🎯 What this adds:

- Accessible clickable cells with proper ARIA labels (required) and button semantics
- Full keyboard navigation support (Enter, Space, Tab)
- Visual feedback states using EDS design tokens
- Seamless DataGrid integration with proper styling coordination

### 💻 Usage example:

```jsx
const clickableColumns = [
    helper.accessor('id', {
      header: 'ID',
      cell: ({ getValue }) => {
        const cellId = `id-${getValue()}`
        return (
          <ClickableCell
            onClick={() => handleCellClick(cellId)}
            isSelected={selectedCell === cellId}
            ariaLabel={`View details for ID ${getValue()}`}
          >
            {getValue()}
          </ClickableCell>
        )
      },
      id: 'clickable_id',
      size: 100,
    }),
    ....continue];
    
    const cellStyle = (row: Row<Photo>, columnId: string): CSSProperties => {
    if (
      ['clickable_id', 'clickable_title', 'clickable_actions'].includes(
        columnId,
      )
    ) {
      return { padding: 0 }
    }
    return {}
  }

    <EdsDataGrid {...args} columns={clickableColumns} cellStyle={cellStyle} />
    
```

### 📋 Integration notes:
- Should be used with column cell renderers as shown above (normal pattern for EdsDataGrid)
- Required: Parent cells must use padding: 0 in cellStyle to avoid double padding
- Component provides hover and focus styling automatically through the browser
- Selection state (`isSelected`) must be managed by parent component as shown in story

This component addresses the need for interactive cells while maintaining EDS design consistency and accessibility standards.

resolves #3839

### DIFF
--- a/packages/eds-data-grid-react/src/components/ClickableCell/ClickableCell.tokens.tsx
+++ b/packages/eds-data-grid-react/src/components/ClickableCell/ClickableCell.tokens.tsx
@@ -31,26 +31,9 @@ export const clickableCell: ClickableCellToken = {
     vertical: 'inherit',
     horizontal: 'inherit',
   },
-  //  states: {
-  //   hover: {
-  //     border: {
-  //       type: 'border',
-  //       width: '1px',
-  //       color: 'transparent',
-  //       style: 'solid',
-  //     },
-  //   },
-  //   focus: {
-  //     outline: {
-  //       type: 'outline',
-  //       offset: '3px',
-  //       style: 'dashed',
-  //       color: outline.color,
-  //       width: outline.width,
-  //     },
-  //   },
   states: {
     hover: {
+      background: 'rgba(220, 220, 220, 0.4)',
       outline: {
         type: 'outline',
         offset: '-1px',
@@ -67,6 +50,9 @@ export const clickableCell: ClickableCellToken = {
         color: outline.color,
         width: outline.width,
       },
+    },
+    active: {
+      background: 'rgba(200, 200, 200, 0.5)',
     },
   },
 }
--- a/packages/eds-data-grid-react/src/components/ClickableCell/ClickableCell.tokens.tsx
+++ b/packages/eds-data-grid-react/src/components/ClickableCell/ClickableCell.tokens.tsx
@@ -11,7 +11,7 @@ const {
     },
     interactive: {
       table__cell__fill_hover: { rgba: hoverBackgroundColor },
-      focus: { rgba: focusColor },
+      table__cell__fill_activated: { rgba: activeBackgroundColor },
     },
   },
   interactions: { focused: outline },
@@ -37,7 +37,7 @@ export const clickableCell: ClickableCellToken = {
       outline: {
         type: 'outline',
         offset: '-1px',
-        style: 'dashed',
+        style: 'solid',
         color: outline.color,
         width: outline.width,
       },
@@ -46,13 +46,20 @@ export const clickableCell: ClickableCellToken = {
       outline: {
         type: 'outline',
         offset: '-1px',
-        style: 'dashed',
+        style: 'solid',
         color: outline.color,
         width: outline.width,
       },
     },
     active: {
-      background: focusColor,
+      background: activeBackgroundColor,
+      outline: {
+        type: 'outline',
+        offset: '-1px',
+        style: 'solid',
+        color: outline.color,
+        width: outline.width,
+      },
     },
   },
 }
--- a/packages/eds-data-grid-react/src/components/ClickableCell/ClickableCell.tokens.tsx
+++ b/packages/eds-data-grid-react/src/components/ClickableCell/ClickableCell.tokens.tsx
@@ -11,6 +11,7 @@ const {
     },
     interactive: {
       table__cell__fill_hover: { rgba: hoverBackgroundColor },
+      table__cell__fill_hover: { rgba: focusBackgroundColor },
       table__cell__fill_activated: { rgba: activeBackgroundColor },
     },
   },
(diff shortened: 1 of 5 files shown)


packages/eds-data-grid-react/src/components/TableBodyCell.tsx
```diff
@@ -37,8 +38,16 @@ export function TableBodyCell<T>({ cell }: Props<T>) {
...
 
+  const handleClick = useCallback(
+    (event: React.MouseEvent<HTMLTableCellElement>) => {
+      onClick?.(event)
+    },
+    [onClick],
+  )
+
   return (
     <StyledCell
```
torleifhalseth: Adding an onClick event to a table cell makes it inaccessible to keyboard and screen reader users by default. To improve accessibility, we can add tabIndex={0}, handle keyboard events (Enter/Space), and set role="button" if the cell should act as a button. We could/should also include aria-label unless the content already describes the action. 

Alternatively, consider placing a <button> inside the cell for interactive content.

packages/eds-data-grid-react/src/components/TableBodyCell.tsx
```diff
@@ -37,8 +38,16 @@ export function TableBodyCell<T>({ cell }: Props<T>) {
       : table.getTotalSize() - header.getStart() - cell.column.getSize()
   }, [pinned, cell.column, table])
 
+  const handleClick = useCallback(
```
torleifhalseth: Using useCallback in this context is not harmful, but it is only beneficial if you are optimizing for referential stability—such as when TableBodyCell is memoized or when you pass handleClick to memoized children. Otherwise, its benefit is marginal, and you could omit it for simplicity.

packages/eds-data-grid-react/src/components/TableRow.tsx
```diff
@@ -20,6 +20,14 @@ export function TableRow<T>({
 }: Props<T>) {
   const { rowClass, rowStyle } = useTableContext()
 
+  const handleCellClick = useCallback(
```
torleifhalseth: I think the same applies for the use of useCallback in this context as mentioned in table cell.

pomfrida:
Discussed this with @torleifhalseth & @emirgens today and realised clickable cells in a table might be an anti-pattern, and not something we want in a design system. We have to discuss more with the designers to see if we can find better alternatives.

**Background**: This feature was requested by a user who needed interactive cells in their table. While the implementation is technically sound, we've identified several UX and accessibility concerns that make this pattern problematic for a design system. See referenced links above in this PR for use case examples. 

#### Concerns identified:
- No visual indication that cells are clickable, creating confusion for users
- Accessibility issues - screen readers can't communicate interactivity without additional markup
- Inconsistent UX pattern - tables are typically expected to display information, not trigger actions
- Missing hover states and other interactive feedback that users expect from clickable elements

#### Better alternatives to explore:
- Action buttons within cells 
- Links in cells - for navigation use cases, use proper <a> elements with clear styling
- Separate action column - dedicated column for row-level actions with proper button components

#### Next steps: 
- Collaborate with design team to understand the original user need
- Explore and document recommended patterns for interactive table content
- Consider creating specific components for common table interaction patterns

**Decision**: Putting this PR on hold until we can align on the right approach. The technical implementation can be preserved if we decide to move forward, but we should first ensure we're solving the user's problem in the most accessible and intuitive way.

**Additional context**: It's worth noting that the current data grid implementation originated from community contributions rather than the core EDS team, and we have identified several accessibility concerns throughout the component. This reinforces the importance of carefully evaluating new interactive patterns before adding them to the design system.

packages/eds-data-grid-react/src/components/ClickableCell/ClickableCell.tokens.tsx
```diff
@@ -0,0 +1,49 @@
...
+  typography: {
+    table: { cell_text: cellTypography },
+  },
+  colors: {
+    text: {
+      static_icons__default: { rgba: typographyColor },
+    },
+    interactive: {
+      table__cell__fill_hover: { rgba: hoverBackgroundColor },
+      table__cell__fill_hover: { rgba: focusBackgroundColor },
```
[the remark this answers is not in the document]
pomfrida: Doesn't exist

millus (approved):
LGTM ✨

PR #3889, opened 2025-08-01
fix(table): export table row with pascal case

magnh:

This pull request changes the exported name of the table row component from `tableRow` to `TableRow` as it was in earlier versions.


torleifhalseth (approved):
LGTM. Thanks @magnh!

Issue #3896, opened 2025-08-12
Add expandable detail panel to Data Grid


magnh:
**Is your feature request related to a problem? Please describe.**

I would like to render a detail panel similar to [the implementation in Mantine React Table](https://www.mantine-react-table.com/docs/guides/detail-panel).

**Describe the solution you'd like**

Add an optional `renderDetailPanel` prop to the `EdsDataGrid` of type `(context: Row<T>) => ReactNode`. The returned JSX would render below the row (as in [Mantine React Table](https://www.mantine-react-table.com/docs/guides/detail-panel#enable-detail-panel-demo)) when the row is expanded (e.g. using `row.toggleExpanded()`).

**Describe alternatives you've considered**

This is currently not possible with the current implementation of `EdsDataGrid`.

I would suggest that we keep the cell implementation per project (i.e. not in the package).

Suggested solution in pull request #3898 

**Additional context**

N/A

PR #3898, opened 2025-08-13
feat(eds-data-grid-react): add detail panel to data grid

magnh:

This pull request expands the `EdsDataGrid` props API by adding a `renderDetailPanel` prop.

This is a suggested implementation of #3896


yusijs (changes requested):
There are issues with combining this with virtual scroll. 

We need to either disable this (or limit the size / expanded count?) when virtualization is turned on, or pass in props to control the row height when expanded by using estimateSize in the virtualizer.

https://github.com/user-attachments/assets/949c05ff-70bc-40e5-ae18-6dde299a02ae

pomfrida:
Doing a clean-up and closing this due to inactivity after review and we don't have the resources to focus on this component at time being. Please feel free to resubmit after updating according to review.

PR #3908, opened 2025-08-19
feat(eds-core-react): add disabled prop to Tooltip

FredrikMWold:
## Summary
This PR adds a `disabled` prop to the **Tooltip** component, making it possible to turn off tooltips programmatically.  
Without this, developers need to wrap tooltips in conditional logic, which can add extra noise especially in larger JSX blocks:

```jsx
{isDisabled ? (
  <button>Save</button>
) : (
  <Tooltip content="Save changes">
    <button>Save</button>
  </Tooltip>
)}
```

This approach works, but it makes the code a bit harder to scan at a glance. With the new prop, the intent is clearer and the code stays more concise:

```jsx
<Tooltip content="Save changes" disabled={isDisabled}>
  <button>Save</button>
</Tooltip>
```

### DIFF
--- a/packages/eds-core-react/src/components/Tooltip/Tooltip.test.tsx
+++ b/packages/eds-core-react/src/components/Tooltip/Tooltip.test.tsx
@@ -93,6 +93,21 @@ describe('Tooltip', () => {
     const tooltip = await screen.findByRole('tooltip')
     expect(tooltip).toBeInTheDocument()
   })
+  it('is not rendered when diabled', async () => {
+    render(
+      <Tooltip title="Tooltip" placement="right-start" disabled>
+        <div>Test</div>
+      </Tooltip>,
+    )
+
+    const content = screen.getByText('Test')
+
+    fireEvent.mouseEnter(content)
+    await act(() => new Promise((r) => setTimeout(r, openDelay)))
+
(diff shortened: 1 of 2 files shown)


torleifhalseth (commented):
Would it be a possible solution to pass the the initial state of `open`?

torleifhalseth:
It could be that I misunderstood the context and need.

FredrikMWold:
> It could be that I misunderstood the context and need.

The change is not about programmatically opening or closing the tooltip. It’s about enabling or disabling the tooltip’s hover behavior.  

For example, imagine a form with a field that can be disabled under certain conditions. When the field is disabled, you want a tooltip to explain why. When the field is enabled, you don’t want the tooltip to appear at all. In this case, you want to disable the tooltip.  

Currently, the only way to achieve this is to conditionally wrap the field with a tooltip, as shown in the PR description:

```jsx
{isDisabled ? (
  <Field />
) : (
  <Tooltip content="Some explanation">
    <Field />
  </Tooltip>
)}
```

In real applications, the JSX is often more complex than this simplified example. Using this pattern repeatedly makes the code harder to read and maintain. Adding a `disabled` prop allows you to keep the tooltip in place while controlling whether it activates, making the code much cleaner and easier to follow.

PR #3924, opened 2025-08-27
feat(eds-core-react): always show "add new option" in Autocomplete when onAddNewOption is provided

FredrikMWold:
## Problem

While observing some users, I noticed that many don’t realize the autocomplete input is searchable. Instead of typing, they scroll through the list manually.  

This behavior makes the **“Add New Option”** feature effectively invisible, since it only appears when a user searches for something that doesn’t exist.  

## Solution

When an `onAddNewOption` callback is provided, the **first option in the list** will always be the “Add New Option.”  

This makes the feature discoverable, even for users who don’t search.


FredrikMWold:
- The add new option should now not be clickable if empty or only white-space.
- Added aria-live polite 
- Fixed aria-disabled not being applied
- It already had aria-label ```aria-label={`Add new option: ${value}`}``` should this be different?
- It already had the role `option`
- Set the aria-selected to false on the add new option since the option itself is not "selectable"

pomfrida (approved):
Very nice work, thanks a lot @FredrikMWold  🙌

PR #3930, opened 2025-08-28
chore: 🔥 Remove individual package lockfiles

pomfrida:
- Remove tracked lockfiles from packages and scripts directories 
- These files were being tracked despite .gitignore rules   
- Use workspace lockfile only for consistent dependency resolution


torleifhalseth (approved):
LGTM 🚢

PR #3932, opened 2025-08-29
Don't call onOptionsChange when clicking "Add new"

FredrikMWold:
This PR fixes a bug where the `onOptionsChange` callback was incorrectly triggered in single-select mode when the "Add new option" button was clicked.


pomfrida (approved):
LGTM 👍

PR #3959, opened 2025-09-04
3952 implement figma design

eddiman:


Chibuzor-Nwemambu (changes requested):
Great work 💪🏾 

There are some errors caught by github actions. I think we should get that fixed before I review the code.
https://github.com/equinor/design-system/actions/runs/17465197795

apps/design-system-docs/src/css/eds_static.css
torleifhalseth, line 1: You can import the static variables directly from the eds-tokens package. I did this in the generator here: https://github.com/equinor/design-system/blob/develop/packages/eds-color-palette-generator/src/app/globals.css#L1

eddiman: I'll do this in a new PR (Y)

apps/design-system-docs/src/theme/Footer/Links/Simple/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+  return <span className="footer__link-separator">·</span>
+}
+
+function SimpleLinkItem({ item }: { item: Props['links'][number] }) {
+  return item.html ? (
+    <span
+      className={clsx('footer__link-item', item.className)}
+      // Developer provided the HTML, so assume it's safe.
+      // eslint-disable-next-line react/no-danger
+      dangerouslySetInnerHTML={{ __html: item.html }}
```
Chibuzor-Nwemambu, line 16: A bit unsure about this `dangerouslySetInnerHTML` which shouldn't be used except if the content is properly sanitized and from a trusted source.

Leaving this comment here as a trail, if we need it.

Chibuzor-Nwemambu (approved):
Nicely done @eddiman 💪🏾 🚚 

This contains a lot of changes so we will merge. However, as we discussed, the footer needs to be redone(preferably without ejecting) in order to avoid changes made directly to the DOM.

PR #4009, opened 2025-09-29
style(eds-demo): 💄 Typography System & Typography component

pomfrida:


torleifhalseth (changes requested):
The styles.module.css in the static and dynamic exercise folder is used for demo/tutorial purpose and should be removed.

packages/eds-demo/src/app/_components/Typography.tsx
```diff
@@ -0,0 +1,235 @@
...
+  as: Component = "p",
+  className = "",
+  weight,
+  style,
+  ...props
+}: TypographyProps) {
+  // Build data attributes based on props
+  const dataAttributes = {
+    "data-font-size": size,
+    ...(lineHeight === "squished" && { "data-line-height": "squished" }),
```
torleifhalseth: Could you pass the value directly like you do with size?

packages/eds-demo/src/app/_components/Typography.tsx
```diff
@@ -0,0 +1,235 @@
...
+ *   as="h2"
+ * >
+ *   Heading
+ * </Typography>
+ */
+export function Typography({
+  children,
+  size = "md",
+  lineHeight = "normal",
+  baselineAligned = true,
```
torleifhalseth: I like the convention where we prefix booleans with is or has. What do you think @pomfrida?

packages/eds-demo/src/app/_components/Typography.tsx
```diff
@@ -0,0 +1,235 @@
...
+      {...dataAttributes}
+      {...props}
+    >
+      {children}
+    </Component>
+  );
+}
+
+// ============================================
+// Preset components for common use cases
```
torleifhalseth, line 110: Awesome with all these preset components. This is what our users want. 🎖️

packages/eds-demo/src/app/_components/Typography.tsx
```diff
@@ -0,0 +1,235 @@
...
+// ============================================
+
+interface HeadingProps extends Omit<TypographyProps, "as"> {
```
torleifhalseth, line 113: Really nice interface/type using Omit from the base type.

packages/eds-demo/src/app/_components/Typography.tsx
```diff
@@ -0,0 +1,235 @@
...
+      as={`h${level}` as As}
+      size={size}
+      lineHeight={lineHeight}
+      weight={weight}
+      {...props}
+    >
+      {children}
+    </Typography>
+  );
+}
```
torleifhalseth, line 143: I think we can remove the children from the prop destructure and pass it through rest props. This pattern works well for wrapper components that just forward all props (including children) to a child component.

```suggestion
export function Heading({
  level = 1,
  size = "lg",
  lineHeight = "squished",
  weight = "lighter",
  ...props
}: HeadingProps) {
  return (
    <Typography
      as={`h${level}` as As}
      size={size}
      lineHeight={lineHeight}
      weight={weight}
      {...props}
    />
  );
}
```

packages/eds-demo/src/app/globals.css
```diff
@@ -22,11 +22,27 @@ h2,
...
+
+.app {
+  background-color: var(--eds-color-bg-canvas);
+  color: var(--eds-color-text-strong);
+}
+
+.top-bar,
+.sidebar {
+  background-color: var(--eds-color-bg-surface);
+}
```
torleifhalseth: This global.css is shared in the layout and used on all pages. I believe this is a problem for the static and dynamic exercise because one of the tasks is to define these colours.

package.json
```diff
@@ -25,6 +25,7 @@
     "build:storybook": "pnpm --filter @equinor/eds-core-react run build:storybook",
     "build:storybook:lab": "pnpm --filter @equinor/eds-lab-react run build:storybook",
     "build:storybook:data-grid": "pnpm --filter ./packages/eds-data-grid-react run build:storybook",
+    "demo": "pnpm --filter eds-demo run dev",
```
torleifhalseth:
```suggestion
    "demo:dev": "pnpm --filter eds-demo run dev",
```

packages/eds-demo/src/app/_components/Typography.tsx
```diff
@@ -0,0 +1,232 @@
...
+      weight={weight}
+      {...props}
+    />
+  );
+}
+
+/**
+ * Body component for body text
+ *
+ * Uses medium size (14px) by default with normal line-height
```
vnys: The default font-size in the EDS is 16px, following the browser default.

packages/eds-demo/src/app/_components/Typography.tsx
vnys: …at least Fusion and Amplify 😉 They have asked specifically for a more opinionated typography component, but I assume others would chime in as well if we ask them. We need to ensure we have the same in Figma and Code though, so I suggest you and @eddiman or @alexlvrs start looking at that @pomfrida once we get the colours out the door.

packages/eds-demo/src/app/_components/Typography.tsx
```diff
@@ -0,0 +1,232 @@
...
+    <Typography as="span" size={size} {...props}>
+      {children}
+    </Typography>
+  );
+}
+
+/**
+ * Label component for form labels and tags
+ *
+ * Uses extra small size (10.5px) by default with medium weight
```
vnys: The default font-size for label is 12px (it used to be 10.5px).

packages/eds-demo/src/app/globals.css
```diff
@@ -22,8 +22,14 @@ h2,
...
-h6 {
-  font-family: Equinor, sans-serif;
+h6,
+h1.type,
+h2.type,
+h3.type,
+h4.type,
+h5.type,
+h6.type {
+  font-family: Equinor, sans-serif !important;
```
vnys: Why do we use the !important directive for font-family?

packages/eds-demo/src/app/layout.tsx
```diff
@@ -1,3 +1,4 @@
+import { Inter } from "next/font/google";
```
vnys, line 1: Can we use the CDN, or is this something that’s needed by Tailwind? 

<https://cdn.eds.equinor.com/font/eds-uprights-vf.css>

vnys (changes requested):
Couple of minor changes required. I see that @millus has already implemented the CDN for the fonts, so no need to do anything about that.

packages/eds-demo/src/app/layout.tsx
pomfrida: I needed to use Next.js font optimisation to avoid FOUT, but I can combine it with the CDN for the Equinor font since that's only used for headers.

packages/eds-demo/src/app/globals.css
pomfrida: I fought with Tailwind 😅 But I increased CSS specificity instead of using !important

This ensures Equinor font applies to headings without needing !important, since the body selector increases specificity enough to override Tailwind's font-sans on the body element.

packages/eds-demo/src/app/_components/Typography.tsx
```diff
@@ -0,0 +1,233 @@
...
+    <Typography as="span" size={size} {...props}>
+      {children}
+    </Typography>
+  );
+}
+
+/**
+ * Label component for form labels and tags
+ *
+ * Uses extra small size (12px) by default with bolder weight
```
vnys:
```suggestion
 * Uses small size (12px) by default with bolder weight
```

vnys (approved):
Just one minor typo in a comment, but apart from that it LGTM 👍

PR #4015, opened 2025-09-30
feat: Redesign the Meet the team page

Chibuzor-Nwemambu:


apps/design-system-docs/docs/about/team/team.css
```diff
@@ -0,0 +1,15 @@
+.img-row {
```
eddiman, line 1: At smaller viewports the grid elements should wrap. Now they overflow their container

apps/design-system-docs/docs/about/team/team.css
```diff
@@ -0,0 +1,15 @@
...
+}
+.img-row img {
+  width: 100%;
+  aspect-ratio: 1 / 1;
+  object-fit: cover;
+  border-radius: 50%;
+}
+
+p {
+  margin: 0 0 3px;
```
eddiman: I think we also have to agree at some point to whether use rem or actual spacings from the tokens. Not a big fan of magic pixel values (maybe im just too nitpicky here)

PR #4022, opened 2025-10-01
feat(eds-color-palette-generator): order colors and use text input for hex

torleifhalseth:


packages/eds-color-palette-generator/src/components/ColorManagement.tsx
```diff
@@ -46,15 +46,18 @@ export const ColorManagement = ({
                 />
               </div>
 
-              {/* Color Picker */}
+              {/* Color Hex Input */}
```
vnys, line 49: Can we input colours in OKLCH as well? Since we’re promoting the use of OKLCH, can we use that as default, with the additional option of using HEX values?

torleifhalseth: I used HEX for now because that is what most people (we) have available, but I totally agree that it should be possible to use other formats to define the base colour. I will create an issue describing this.

torleifhalseth: Created this: https://github.com/equinor/design-system/issues/4024

PR #4059, opened 2025-10-08
docs(design-system-docs): Add storybook iframe to all components

eddiman:
- Adds iframe from storybook to all components
- Puts components.md on top of sidebar
- Renamed directories in /components to be capitalized


emirgens (approved):
Let see IRL

PR #4129, opened 2025-10-21
chore: Rename follow-up

Chibuzor-Nwemambu:
Update all references from `develop` to `main`

This task involves more than just a PR approval, but verification that all relevant changes have been made.
(please see the parent issue for the list of verifiable tasks)

- [x] Update branching documentation
- [x] Update CI/CD pipelines
- [x] Configure branch protection
- [x] Communicate to the team
- [x] Change radix config branch

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -3,14 +3,14 @@ on:
   workflow_dispatch:
   pull_request:
     branches:
-      - develop
+      - main
     paths:
-      - 'apps/**'
-      - 'packages/**'
-      - '.github/workflows/**'
+      - "apps/**"
+      - "packages/**"
+      - ".github/workflows/**"
   push:
     branches:
-      - develop
+      - main
 env:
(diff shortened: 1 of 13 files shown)


.github/workflows/checks.yaml
```diff
@@ -3,14 +3,14 @@ on:
   workflow_dispatch:
   pull_request:
     branches:
-      - develop
+      - main
     paths:
-      - 'apps/**'
-      - 'packages/**'
-      - '.github/workflows/**'
+      - "apps/**"
```
torleifhalseth, line 8: What is our practise single our double?

Chibuzor-Nwemambu: That was an automatic change, not instigated by me 😂 . Not entirely sure how that happened as well. But good question, what is our practice? Is it single or double?

torleifhalseth: single quote is default and it is defined in the .prettierrc.yaml

PR #4174, opened 2025-10-30
chore: Documentation updates and devcontainer improvements 

Hjaf:
Added links and EDS assets for Power BI.
**Note**: These assets were provided by @KarenHelgoey
 
 
Updated devcontainer configuration to permit doc development using devcontainers. 
Some fixes remains:
- Warnings are still present on build, updates are suggested.
- node version should be synced automatically with the version defined in the deployment image, see [DockerfileEdsDocs](https://github.com/equinor/design-system/blob/6c6a4346f954cccdc313d54fb4f7690d8ba6e05f/DockerfileEdsDocs#L2)
- adding `next-line` linting exceptions not possible because the linting extension automatically adds newline on save: example `<!-- markdownlint-disable-next-line MD033 -->` 
- locally run devcontainer may not work unless --host is added to yarn start (docs updated)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,7 @@ node_modules
 pnpm-debug.log
 .DS_Store
 # use pnpm
-/.pnpm-store/*
+/.pnpm-store
 package-lock.json
 packages/*/package-lock.yaml
 scripts/*/package-lock.yaml
--- a/apps/design-system-docs/README.md
+++ b/apps/design-system-docs/README.md
@@ -15,6 +15,7 @@ yarn start
 ```
 
 This command starts a local development server and opens up a browser window. Most changes are reflected live without having to restart the server.
+> Note: If you encounter any issues in devcontainers, try running the command with the following flag: `--host 127.0.0.1`
 
 ## Build
 
--- a/apps/design-system-docs/docs/about/getting-started/develop/citizen_developers.md
+++ b/apps/design-system-docs/docs/about/getting-started/develop/citizen_developers.md
@@ -16,7 +16,6 @@ We've made it straightforward to begin building with EDS. You can be up and runn
 ### Power Apps
 
 <!-- markdownlint-disable-next-line MD033 -->
-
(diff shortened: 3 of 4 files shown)


.devcontainer/devcontainer.json
```diff
@@ -70,15 +62,10 @@
       "onAutoForward": "notify"
     }
   },
-  "onCreateCommand": "pnpm install && pnpm recursive install --filter ./packages",
+  "onCreateCommand": "npm install -g pnpm gitmoji-cli && pnpm install && pnpm recursive install --filter ./packages",
```
torleifhalseth, line 65: When is this onCreateCommand triggered? Is it when initialising a new container? Maybe we could use the `pnpm run init` command we have in the main package.json?

torleifhalseth, line 65: gitmoji-cli? 🤯 Is Elsa involved?

.devcontainer/devcontainer.json
```diff
@@ -70,15 +62,10 @@
...
   "features": {
-    "docker-in-docker": "20.10",
-    "docker-from-docker": "20.10",
-    "git": "latest",
-    "github-cli": "latest",
-    "ghcr.io/devcontainers/features/sshd:1": {
-      "version": "latest"
-    }
+    "ghcr.io/eitsupi/devcontainer-features/jq-likes:2": {},
+    "ghcr.io/devcontainers/features/node:1": {}
```
torleifhalseth, line 69: I am guessing this includes all we need @Hjaf ?

.devcontainer/devcontainer.json
Hjaf: https://containers.dev/implementors/json_reference/#lifecycle-scripts 

First time a new container is created. 
As this container is fairly general purpose I'd suggest not doing any building or installing container creation at all 😄

.devcontainer/devcontainer.json
Hjaf: I can't really see the usecase for these anymore. AFAIK the [docker extension](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-docker) should cover these needs. Maybe @vnys can shed some light on why docker-in-docker and docker-from-docker were added. The same applies ot the git github-cli and sshd which i believe is present by default in the base image

PR #4158, opened 2025-10-27
chore: ⬆️ Update dependencies

pomfrida:
Resolves #4132

PR #4200, opened 2025-11-04
feat(eds-tokens): support spacing levels

torleifhalseth:
This pull request introduces improvements and fixes to the spacing utilities and demos in the design system. The main changes include the addition of new selectable space utility demos, the introduction of new CSS utility classes for spacing, and corrections to variable naming in the token CSS. These updates enhance the flexibility and accuracy of spacing utilities and their documentation.

**Spacing utility improvements and bug fixes:**

* Fixed incorrect variable names for spacing tokens in `foundation.css` to use the correct format (e.g., `--eds-spacing-inline-4xs` instead of `--eds-spacing-inline-4-xs`). This ensures the spacing utilities reference the correct CSS variables. [[1]](diffhunk://#diff-0fac253047089fc5ec411cd5d7518d085f670895e744a02f2da01e6e3236a8bbL2-R12) [[2]](diffhunk://#diff-0fac253047089fc5ec411cd5d7518d085f670895e744a02f2da01e6e3236a8bbL42-R62) [[3]](diffhunk://#diff-0fac253047089fc5ec411cd5d7518d085f670895e744a02f2da01e6e3236a8bbL92-R97)

**New spacing utility classes:**

* Added new utility classes for selectable padding and gap (`.selectable-p`, `.selectable-px`, `.selectable-py`, `.selectable-gap`, `.container-gap`, `.page-gap`, etc.) to `foundation.css`, allowing for more granular and standardized spacing control in layouts.

**Storybook/demo enhancements:**

* Added two new Storybook stories, `SelectableSpace` and `SelectableSpaceDensityComparison`, to `Spacing.stories.tsx` to demonstrate how selectable space utilities work with different sizes, proportions, and density modes.
* Introduced new CSS styles in `Spacing.stories.css` to support the selectable space demos, providing better visual structure and clarity in the documentation.

**Tooling configuration:**

* Updated `.eslintrc.js` to include `vite.*.config.ts` files in the ESLint configuration, ensuring consistent linting for all Vite config files.

### DIFF
--- a/packages/eds-core-react/src/components/Spacing/Spacing.stories.tsx
+++ b/packages/eds-core-react/src/components/Spacing/Spacing.stories.tsx
@@ -214,6 +214,96 @@ export const SelectablePadding: Story = {
   },
 }
 
+export const ContainerPadding: Story = {
+  render: () => {
+    const proportions = ['squished', 'squared', 'stretched'] as const
+
+    return (
+      <DemoSection
+        title="Container Padding"
+        description="The .container-p utility provides padding for container elements that responds to space proportions. Requires the data-space-proportions attribute."
+      >
+        <div className="flex flex-row flex-wrap items-start container-gap-x">
+          {proportions.map((proportion) => (
+            <div
+              key={proportion}
+              className="flex flex-col items-start selectable-gap-x"
+            >
+              <Paragraph className="mb-2">
+                <strong>
+                  {proportion.charAt(0).toUpperCase() + proportion.slice(1)}
+                </strong>
+                <br />
+                <code>data-space-proportions=&quot;{proportion}&quot;</code>
+              </Paragraph>
+              <div
+                data-space-proportions={proportion}
+                className="container-p spacing-demo-box"
+              >
+                <Typography
+                  family="ui"
+                  size="md"
+                  baseline="grid"
+                  lineHeight="default"
+                  weight="normal"
+                  tracking="normal"
+                >
+                  Container
+                </Typography>
+              </div>
+            </div>
+          ))}
+        </div>
+      </DemoSection>
+    )
+  },
+}
+
+export const PagePadding: Story = {
+  render: () => {
+    const proportions = ['squished', 'squared', 'stretched'] as const
+
+    return (
+      <DemoSection
+        title="Page Padding"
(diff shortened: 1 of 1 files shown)


packages/eds-tokens/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "@equinor/eds-tokens",
   "version": "1.1.4",
   "description": "Design tokens for the Equinor Design System",
+  "type": "module",
```
pomfrida, line 5: I found out last week that this is not needed as the package exports already define both "import" and "require" conditions for dual ESM/CJS support. Adding "type": "module" forces the package to be treated as ESM-only, which can break CommonJS consumers and conflicts with our dual-package strategy. (I did this in https://github.com/equinor/design-system/pull/4158 after going back and forth so it was probably not communicated clearly, I am sorry for that)

packages/eds-tokens/package.json
```diff
@@ -15,9 +16,6 @@
     "./css/variables.css": "./build/css/variables.min.css",
     "./css/foundation": "./build/css/foundation.min.css",
     "./css/foundation.css": "./build/css/foundation.min.css",
-    "./css/spacing/primitives.css": "./build/css/spacing/primitives-trimmed.css",
-    "./css/spacing/comfortable.css": "./build/css/spacing/comfortable-verbose.css",
-    "./css/spacing/spacious.css": "./build/css/spacing/spacious-verbose.css",
```
pomfrida, line 20: The removed exports still have their corresponding files in the build folder (primitives-trimmed.css, comfortable-verbose.css, spacious-verbose.css), but removing them from package.json exports means consumers can no longer import them by name. This is a breaking change for anyone using these imports. Should add:

- Migration note to use `@equinor/eds-tokens/css/variables.css` instead
- Breaking change indicator in the commit (use `!` in commit msg for release please to pick up on the breaking change so it can update the changelog accordingly)

packages/eds-tokens/package.json
torleifhalseth: I added it back because we need it when executing vite config files that uses esm imports. The exports field is for consumers of your package (determines which file to load), while "type": "module" is for your package's own tooling and config files.

packages/eds-tokens/package.json
torleifhalseth: Lets do that!

PR #4202, opened 2025-11-04
docs(design-system-docs): Update Design System Docs README file

### LINKED ISSUE #4193: Update Design System Docs README file
## Description

The Docusaurus documentation README currently instructs users to use `yarn`, but the monorepo uses `pnpm` as specified in the root `package.json`. This causes dependency conflicts when running the docs locally from the devcontainer.

## Problem

* Users following the docs README get errors due to package manager mismatch
* The `packageManager` field in root `package.json` specifies `pnpm`, but docs say to use `yarn`
* Mixing package managers (`yarn` and `pnpm`) causes inheritance issues from root package.json

## Expected Behavior

* All apps in the monorepo should be run using `pnpm` from the root directory
* Documentation should consistently reflect this across all READMEs

## Proposed Solution

Update the Docusaurus documentation README to:

* Remove all references to `yarn`
* Add instructions to run the docs using `pnpm` from the monorepo root
* Include commands like `pnpm --filter docs dev` (or similar, depending on workspace setup)
* Add a note that the monorepo uses `pnpm` exclusively

## Additional Context

This was discovered when trying to run the Docusaurus instance locally from the devcontainer. The issue was temporarily fixed by adding a `packageManager` field locally, but the proper solution is to standardize on `pnpm` throughout.

## Checklist

* [ ] Update docs README
* [ ] Verify pnpm commands work correctly
* [ ] Check for any other READMEs that might reference yarn

Chibuzor-Nwemambu:
Fixes #4193

### DIFF
--- a/apps/design-system-docs/README.md
+++ b/apps/design-system-docs/README.md
@@ -14,14 +14,8 @@ Before you begin, ensure you have the following installed:
 This documentation site is part of the EDS monorepo. All commands should be run from the **root directory** of the monorepo, not from this directory.
 
 If you haven't already, start by setting up the monorepo:
-
-```bash
-# Navigate to the root directory
-cd ../..
-
-# Install dependencies and build packages
-pnpm run init
-```
+- First navigate to the root directory
+- Then install dependencies and build packages by running `pnpm run init`
 
 ## Available Commands
 


apps/design-system-docs/README.md
```diff
@@ -1,41 +1,151 @@
...
+
+If you haven't already, start by setting up the monorepo:
+
+```bash
+# Navigate to the root directory
+cd ../..
+
+# Install dependencies and build packages
+pnpm run init
+```
```
pomfrida: The `cd ../..` instruction is relative and confusing since it assumes the user is currently in the design-system-docs directory. Since all commands should be run from the root directory anyway, we can simplify this section.

```suggestion
This documentation site is part of the EDS monorepo. All commands should be run from the **root directory** of the monorepo (`/design-system`), not from this directory.

If you haven't already, start by setting up the monorepo from the root:

```bash
# Install dependencies and build packages
pnpm run init
```
```


This makes it clearer without relying on relative paths.. What do you think?

pomfrida: Ok don't commit my suggestion because its not including the ``` on the end here and it won't add it here either. It's conflicting with the suggestion``` thingy

pomfrida: Screenshot 2025-11-04 at 14 21 59

pomfrida (approved):
Great 👍

PR #4206, opened 2025-11-04
refactor(eds-color-palette-generator): 4178 move color palette generator app to apps folder

Chibuzor-Nwemambu:


apps/eds-color-palette-generator/tsconfig.json
```diff
@@ -11,17 +15,27 @@
     "moduleResolution": "bundler",
     "resolveJsonModule": true,
     "isolatedModules": true,
-    "jsx": "preserve",
+    "jsx": "react-jsx",
```
torleifhalseth, line 18: Do we need to change this?

Chibuzor-Nwemambu: ~Absolutely not, we don't 😅 . I'll revert that now.~ 
Scratch that. Actually, we **do** need to change this. This happened automatically when I ran the `build` command, and the change was made by Next.js itself with the message:
```
   We detected TypeScript in your project and reconfigured your tsconfig.json file for you.
   The following suggested values were added to your tsconfig.json. These values can be changed to fit your project's needs:

        - include was updated to add '.next/dev/types/**/*.ts'

   The following mandatory changes were made to your tsconfig.json:

        - jsx was set to react-jsx (next.js uses the React automatic runtime)
```

So I think we need to keep this most especially since it is required by **Next 16**.

Good thing is that, it means we don't have to import react in every file.

torleifhalseth: Perfect.

apps/eds-color-palette-generator/src/hooks/useMounted.ts
```diff
@@ -0,0 +1,34 @@
...
+ * ```
+ */
+export function useMounted(): boolean {
+  const [isMounted, setMounted] = useState(false)
+
+  useEffect(() => {
+    // Use setTimeout to avoid synchronous setState in useEffect
+    const timeoutId = setTimeout(() => {
+      setMounted(true)
+    }, 0)
```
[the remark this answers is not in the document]
torleifhalseth: I will look at this @Chibuzor-Nwemambu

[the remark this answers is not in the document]
torleifhalseth: The linter initially flagged calling setState in an effect as potentially problematic, but for a mounted hook, this is the intended and correct behavior. Added a disable comment for this legitimate use case and removed the setState.

PR #4216, opened 2025-11-06
docs: update pages in design tokens

emirgens:
Update design tokens documentation for elevation, grid, shape, spacing, and typography
Cleanup


vnys (approved):
This documentation is seriously outdated, but I think we just have to go with what we have for now and make an issue to update it.

PR #4219, opened 2025-11-10
feat: ✨ Input and TextField - new version for EDS 2.0. using new tokens

millus:
This pull request introduces the new EDS 2.0 `Input` component as a low-level, flexible form input for the next generation of the Equinor Design System. It includes the component implementation, types, documentation, comprehensive tests, and updates to package exports and Storybook configuration to support the new features.

**New EDS 2.0 Input component:**

- Added the new `Input` component in `src/components/next/Input/Input.tsx`, supporting flexible props such as adornments, text, invalid/disabled/readOnly states, and rendering as either `input` or `textarea`. The component uses EDS 2.0 token system data attributes and is fully accessible.
- Defined the `InputProps` type in `Input.types.ts` to document and enforce the supported props for the new component.
- Added an index file to export the new component and its types.
- Created a snapshot for the new component to support regression testing.

**Documentation and Storybook:**

- Added a new MDX documentation file for the `Input` component, covering usage, examples, and guidance for composition and accessibility.
- Updated the Storybook preview configuration to include the "Next (EDS 2.0)" section for the new components.

**Testing:**

- Added a comprehensive test suite for the new `Input` component, covering accessibility, states, adornments, EDS 2.0 token system attributes, and CSS classes.

**Package and build updates:**

- Updated `package.json` to include the build output for EDS 2.0 CSS files and ensure they are exported and included in the published package. [[1]](diffhunk://#diff-e88202ef1b9055378f94f509f148df36c5dbb72354fe302636dd96f53d958e03L26-R27) [[2]](diffhunk://#diff-e88202ef1b9055378f94f509f148df36c5dbb72354fe302636dd96f53d958e03L38-R41)

Also added TextField component that uses both the new Field wrapper and Input component.

### DIFF
--- a/packages/eds-core-react/src/components/next/Input/Input.tsx
+++ b/packages/eds-core-react/src/components/next/Input/Input.tsx
@@ -22,7 +22,7 @@ export const Input: OverridableComponent<InputProps, HTMLInputElement> =
     },
     ref,
   ) {
-    const colorAppearance = invalid && !disabled ? 'danger' : 'neutral'
+    const tone = invalid && !disabled ? 'danger' : 'neutral'
 
     const showErrorIcon = invalid && !disabled
     const hasStartAdornment = startText || startAdornment
@@ -35,7 +35,7 @@ export const Input: OverridableComponent<InputProps, HTMLInputElement> =
     return (
       <div
         className={containerClasses}
-        data-color-appearance={colorAppearance}
+        data-color-appearance={tone}
         data-font-size="md"
         data-selectable-space="sm"
         data-space-proportions="squished"


packages/eds-core-react/src/styles.css
torleifhalseth, line 1: Not 100% sure, but maybe we could name this `index.css`? It is not super important, but it feels more aligned with conventions.

millus: Yes, changed to index.css, thank you! Also fixed it so there is a bundled and minified version using rollup now 👍

packages/eds-core-react/build/index.css
```diff
@@ -0,0 +1,122 @@
+/* Styles for Next EDS 2.0 Components (vanilla CSS) */
+
+@layer eds-components {
+  /* Container - use neutral appearance by default */
+  .eds-input {
+    position: relative;
+    display: flex;
+    flex-direction: row;
+    align-items: center;
+    box-sizing: border-box;
```
torleifhalseth: This should probably be added to a reset file where we set box-sizing to border-box as default. This is common practise.

packages/eds-core-react/build/index.css
```diff
@@ -0,0 +1,122 @@
...
+    width: 100%;
```
torleifhalseth: Is it possible to remove this so that the input can easily specify width in context?

packages/eds-core-react/build/index.css
```diff
@@ -0,0 +1,122 @@
...
+    background: var(--eds-color-bg-canvas);
```
torleifhalseth: We actually have a unique background for input in our "concept" tokens: https://github.com/equinor/design-system/blob/main/packages/eds-tokens/build/css/variables.css#L12

packages/eds-core-react/build/index.css
```diff
@@ -0,0 +1,122 @@
...
+    border: 1px solid var(--eds-color-border-subtle);
+    border-radius: var(--eds-spacing-border-radius-none);
```
torleifhalseth: What is the value of specifying no radius? Is this to override browser defaults?

packages/eds-core-react/src/components/next/Input/input.css
```diff
@@ -0,0 +1,120 @@
...
+  .eds-field {
+    width: 100%;
+    border: none;
+    background: transparent;
+    outline: none;
+    color: var(--eds-color-text-strong);
+    font-family: var(--eds-typography-ui-body-font-family), sans-serif;
+    font-size: var(--eds-typography-ui-body-lg-font-size);
+    font-weight: var(--eds-typography-ui-body-lg-font-weight-normal);
+    line-height: var(--eds-typography-ui-body-lg-lineheight-squished);
```
torleifhalseth: We should use the generic font variables and resolve the values we want using data-attributes. This should also be aligned with the Figma design.

Tip: We have a selector using the font-family data attribute that applies font styles: https://github.com/equinor/design-system/blob/main/packages/eds-tokens/build/css/variables.css#L1101

But you could consider removing this selector and using the variables directly in your selector. You should discuss this with other devs ;)

packages/eds-core-react/src/components/next/Input/input.css
```diff
@@ -0,0 +1,120 @@
...
+    padding-block: var(--eds-selectable-space-vertical);
+    padding-inline: var(--eds-selectable-space-horizontal);
```
torleifhalseth: Perfect use of the new spacing variables 👍

packages/eds-core-react/src/components/next/Input/input.css
```diff
@@ -0,0 +1,120 @@
...
+    overflow: hidden;
+    white-space: nowrap;
+    text-overflow: ellipsis;
+    resize: none;
```
torleifhalseth, line 138: Should this be default?

packages/eds-core-react/src/components/next/Input/input.css
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  .eds-field::placeholder {
+    color: var(--eds-color-text-strong);
+  }
```
torleifhalseth: I would recommend using nesting to prevent repeating the selector and making it easier to read:


```css
  .eds-field {
    ::placeholder {
        color: var(--eds-color-text-strong);
    }
   }
```

packages/eds-core-react/src/components/next/Input/input.css
```diff
@@ -0,0 +1,120 @@
...
+
+  /* 
+   * Disabled field - override browser defaults for consistent styling
+   * Safari applies a default opacity and color to disabled inputs, so we need
+   * to explicitly set opacity: 1 and use -webkit-text-fill-color
+   * Note: We don't rely on inherited color context, so this works consistently
```
torleifhalseth: This could also be part of a reset css file.

Here is a reset I would look at: https://www.joshwcomeau.com/css/custom-css-reset/

packages/eds-core-react/src/components/next/Input/input.css
```diff
@@ -0,0 +1,120 @@
...
+    font-family: var(--eds-typography-ui-body-font-family), sans-serif;
+    font-size: var(--eds-typography-ui-body-md-font-size);
+    font-weight: var(--eds-typography-ui-body-md-font-weight-normal);
+    line-height: var(--eds-typography-ui-body-md-lineheight-default);
+  }
+
+  /* Text with neutral appearance should use neutral text color */
+  .eds-adornment [data-color-appearance='neutral'] {
+    color: var(--eds-color-text-subtle);
+  }
```
torleifhalseth: You should apply data-color-appearance in the markup to resolve the neutral color for adornment. This should be the default appearance so it should not be necessary.

packages/eds-core-react/src/components/next/Input/input.css
```diff
@@ -0,0 +1,120 @@
...
+
+  .eds-adornment--left {
+    left: 0;
+    padding-left: var(--eds-selectable-space-horizontal);
```
torleifhalseth: Because we used padding inline and block we should probably also use padding inline start and end.

packages/eds-core-react/src/components/next/Input/Input.tsx
```diff
@@ -0,0 +1,105 @@
...
+    const [leftAdornmentsRef, setLeftAdornmentsRef] = useState<HTMLDivElement>()
+    const [rightAdornmentsRef, setRightAdornmentsRef] =
+      useState<HTMLDivElement>()
+
+    // Calculate dynamic padding based on adornment widths
+    // Use manual width if provided, otherwise auto-measure from DOM
+    const leftWidth =
+      leftAdornmentsWidth ?? (leftAdornmentsRef?.clientWidth || 0)
+    const rightWidth =
+      rightAdornmentsWidth ?? (rightAdornmentsRef?.clientWidth || 0)
```
torleifhalseth: I would keep things simple. This adds some complexity that I am not sure we actually want/need.

packages/eds-core-react/src/components/next/Input/Input.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+    // Map invalid state to color appearance (internal use only)
+    const colorAppearance = invalid ? 'danger' : 'neutral'
+
+    // Fixed EDS 2.0 spacing values (internal - not exposed to users)
+    const selectableSpace = 'xs'
+    const spaceProportions = 'stretched'
```
torleifhalseth: I would move these outside the component but inside the same file:

// top of file, above the component
const SELECTABLE_SPACE = 'xs' as const
const SPACE_PROPORTIONS = 'stretched' as const

// inside component
data-selectable-space={SELECTABLE_SPACE}
data-space-proportions={SPACE_PROPORTIONS}

packages/eds-core-react/src/components/next/Input/Input.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+    const containerClasses = [
+      'eds-input',
+      className,
```
torleifhalseth: I would put the className prop last to provide consistent side-effects if we define overrides in the className selector. This is not good practise, but will probably be useful for some people.

packages/eds-core-react/src/components/next/Input/Input.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+    // Inline styles for dynamic adornment spacing
+    const fieldStyle: CSSProperties = {
+      paddingLeft: leftWidth
+        ? `calc(var(--eds-selectable-space-horizontal) + ${leftWidth}px)`
+        : undefined,
+      paddingRight: rightWidth
+        ? `calc(var(--eds-selectable-space-horizontal) + ${rightWidth}px)`
+        : undefined,
+    }
```
torleifhalseth: As mentioned, I would challenge the value of this added complexity.

torleifhalseth: I would not use inline style.

packages/eds-core-react/src/components/next/Input/Input.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+    return (
+      <div
+        className={containerClasses}
+        style={style}
```
torleifhalseth: I would not support using inline styles.

packages/eds-core-react/src/components/next/Input/Input.tsx
```diff
@@ -0,0 +1,105 @@
...
+        data-color-appearance={disabled ? 'neutral' : colorAppearance}
```
torleifhalseth: I would move this logic so that you control the appearance globally where you define colorAppearance.

packages/eds-core-react/src/components/next/Input/Input.tsx
```diff
@@ -0,0 +1,105 @@
...
+            {...leftAdornmentsProps}
+            ref={setLeftAdornmentsRef}
+            className="eds-adornment eds-adornment--left"
+          >
+            {leftAdornments}
+          </div>
+        )}
+        <Component
+          ref={ref as React.Ref<HTMLInputElement & HTMLTextAreaElement>}
+          type={Component === 'input' ? type : undefined}
```
torleifhalseth: Is this necessary? I would define the minimum and make it dynamic spreading props down to the input element.

packages/eds-core-react/src/components/next/Input/Input.types.ts
```diff
@@ -0,0 +1,34 @@
+import { ReactNode, CSSProperties, ComponentPropsWithoutRef } from 'react'
+
+type ElementComponent = 'input' | 'textarea'
+
+export type InputProps = {
```
torleifhalseth, line 3: I recommend having a look at the existing Input component where we extend InputHTMLAttributes. Then you don't have to specify these properties.

packages/eds-core-react/src/components/next/Input/Input.types.ts
```diff
@@ -0,0 +1,34 @@
...
+  /** Left adornments props */
+  leftAdornmentsProps?: ComponentPropsWithoutRef<'div'>
+  /** Right adornments props */
+  rightAdornmentsProps?: ComponentPropsWithoutRef<'div'>
+  /** Manually specify left adornments width. The width will be auto-measured if not defined */
+  leftAdornmentsWidth?: number
+  /** Manually specify right adornments width. The width will be auto-measured if not defined */
+  rightAdornmentsWidth?: number
+  /** Cast the input to another element */
+  as?: ElementComponent
```
torleifhalseth: Do we need this?

packages/eds-core-react/src/components/next/Introduction.mdx
```diff
@@ -0,0 +1,85 @@
...
+import { Input } from '@equinor/eds-core-react/next'
+
+function MyComponent() {
+  return <Input placeholder="Enter text" />
+}
+```
+
+### CSS Import
+
+**The CSS is automatically included** when you import components from `@equinor/eds-core-react/next`.
```
torleifhalseth: How? I think you would have to include the css manually 🤓

packages/eds-core-react/src/components/next/Introduction.mdx
```diff
@@ -0,0 +1,85 @@
...
+
+If you need to import CSS separately (e.g., for testing or custom builds):
+
+```tsx
+import '@equinor/eds-core-react/index.css' // Development
+import '@equinor/eds-core-react/index.min.css' // Production
+```
```
torleifhalseth: This part is necessary. I would recommend only providing one file index.css this should be minified.

packages/eds-core-react/src/components/next/Introduction.mdx
```diff
@@ -0,0 +1,85 @@
...
+
+### Density
+
+Control spacing density using the `data-density` attribute or `EdsProvider`.
+
+Available densities: `spacious` (default), `comfortable`.
+
+### Color Schemes
+
+Components automatically adapt to light and dark modes using `data-color-scheme` attribute.
```
torleifhalseth: I don't think we need this documentation as part of the new components. This is handled internally. I would keep the documentation as simple as possible without implementation details.

packages/eds-core-react/src/styles.css
```diff
@@ -0,0 +1,3 @@
+/* Styles for Next EDS 2.0 Components (vanilla CSS) */
+
+@import './components/next/Input/input.css';
```
torleifhalseth: Perfect. Here you can include other components in the future.

packages/eds-core-react/src/index.next.ts
```diff
@@ -0,0 +1,6 @@
+'use client'
```
torleifhalseth: Don't include 'use client'. We want our new components to be available in server side rendered components.

packages/eds-core-react/src/components/next/Input/Input.tsx
```diff
@@ -0,0 +1,105 @@
...
+          disabled={disabled}
+          readOnly={readOnly}
+          className="eds-field"
```
torleifhalseth: I find it more intuitive to reference this element as the input because that is what it is. Maybe you should consider switching the eds-input and eds-field?

packages/eds-core-react/src/components/next/Field/FieldDescription.tsx
```diff
@@ -0,0 +1,28 @@
...
+      family="ui"
+      size="sm"
+      lineHeight="default"
+      baseline="grid"
+      {...rest}
+    >
+      {children}
+    </TypographyNext>
+  )
+})
```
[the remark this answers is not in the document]
pomfrida: Agree

packages/eds-core-react/src/components/next/TextField/TextField.tsx
```diff
@@ -0,0 +1,72 @@
+import { forwardRef } from 'react'
+import { info_circle } from '@equinor/eds-icons'
+import type { TextFieldProps } from './TextField.types'
+import { Field, useFieldIds } from '../Field'
+import { Input } from '../Input'
+import { Tooltip } from '../../Tooltip'
+import { Icon } from '../Icon'
+import './text-field.css'
+
+export const TextField = forwardRef<HTMLInputElement, TextFieldProps>(
```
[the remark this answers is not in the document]
pomfrida: Sounds like a good suggestion

packages/eds-core-react/.storybook/preview.mjs
```diff
@@ -22,6 +22,8 @@ const preview = {
           'Icons',
           ['Introduction', 'Preview', 'Icon'],
           '*',
+          'Next (EDS 2.0)',
+          ['Introduction', '*'],
```
pomfrida: Probably a merge conflict here, EDS 2.0 already mentioned on top of the list (3rd)

packages/eds-core-react/src/components/next/Input/Input.stories.tsx
```diff
@@ -0,0 +1,513 @@
...
+    // Other
+    className: {
+      control: 'text',
+      description: 'Additional CSS class names for the container',
+      table: {
+        category: 'Other',
+      },
+    },
+  },
+  decorators: [
```
pomfrida, line 144: Really like how you've organised the Storybook setup, the argTypes with categories (Core, States, Adornments, Other) makes the controls panel super clear and easy to navigate. I'm inspired to do the same as well 😍

packages/eds-core-react/src/components/next/Input/Input.stories.tsx
```diff
@@ -0,0 +1,513 @@
...
+          background: '#f7f7f7',
+        } as React.CSSProperties
+      }
+    >
+      <Input {...args} placeholder="Custom background" />
+    </div>
+  )
+}
+
+export const ColorSchemes: StoryFn<InputProps> = () => {
```
pomfrida, line 454: Like this side by side preview of the dark & light themes!

packages/eds-core-react/src/components/next/Input/Input.tsx
```diff
@@ -0,0 +1,119 @@
...
+import { Icon } from '../Icon'
+import type { InputProps } from './Input.types'
+
+export const Input: OverridableComponent<InputProps, HTMLInputElement> =
+  forwardRef<
+    HTMLInputElement,
+    InputProps & InputHTMLAttributes<HTMLInputElement>
+  >(function Input(
+    {
+      invalid = false,
```
pomfrida, line 10: Figma uses "Error" for the validation state, but invalid is a good choice here since it aligns with the HTML standard (aria-invalid) and common React conventions. 👍

packages/eds-core-react/src/components/next/Input/Input.stories.tsx
```diff
@@ -0,0 +1,513 @@
...
+  placeholder: 'Placeholder text',
+  invalid: false,
+  disabled: false,
+  readOnly: false,
+}
+
+export const Types: StoryFn<InputProps> = () => (
+  <>
+    <Input
+      aria-label="Text input"
```
pomfrida, line 171: When someone uses Input standalone (without a label), they need to add aria-label manually and this might be easy to miss even though it's in the code examples. Maybe worth a short note in the docs?

packages/eds-core-react/src/components/next/Input/Input.tsx
pomfrida: aria-invalid is set before the spread, so a user could accidentally override it via props. Might be safer to move it after {...inputProps} to ensure it always stays in sync with the invalid prop? 🤔 Using the invalid state with aria-invalid=false would not make any sense in this case

packages/eds-core-react/src/components/next/Input/Input.tsx
```diff
@@ -0,0 +1,119 @@
...
+    },
+    ref,
+  ) {
+    const colorAppearance = invalid && !disabled ? 'danger' : 'neutral'
+
+    const showErrorIcon = invalid && !disabled
+    const hasStartAdornment = startText || startAdornment
+    const hasEndAdornment = endText || endAdornment
+
+    const containerClasses = ['eds-input-container', className]
```
pomfrida: className ends up on the container div here, not the input element itself. This might be unexpected for users. Could an inputProps pattern work (similar to MUI), so users can pass props directly to the input if needed?

packages/eds-core-react/src/components/next/Input/Input.types.ts
```diff
@@ -0,0 +1,26 @@
...
+  startAdornment?: ReactNode
+  /** Text at the end (e.g., "km", "%") - always neutral color */
+  endText?: string
+  /** Adornment at the end (icons, buttons, etc.) - inherits state color (red when invalid) */
+  endAdornment?: ReactNode
+  /** Render as input or textarea */
+  as?: 'input' | 'textarea'
+  /** Additional CSS class names */
+  className?: string
+}
```
pomfrida:
```suggestion
} & InputHTMLAttributes<HTMLInputElement>
```

InputHTMLAttributes<HTMLInputElement> is added in the component but not exported as part of InputProps. If someone imports this type to use elsewhere, they won't get the native input attributes like onChange, value, etc. and with the intersection in place, props like placeholder, disabled, type, readOnly, and className wouldn't need to be explicitly listed - they come for free from InputHTMLAttributes.

packages/eds-core-react/src/components/next/Input/input.css
```diff
@@ -0,0 +1,158 @@
...
+  /* Adornments container */
+  .eds-adornment {
+    display: flex;
+    align-items: center;
+    gap: var(--eds-typography-gap-horizontal);
+    flex-shrink: 0;
+  }
+
+  /* Adornment text (prefix/suffix) - always neutral via data-color-appearance */
+  .eds-adornment__text {
```
pomfrida, line 160: The prefix text ("NOK") doesn't appear vertically centered the same way as the icon on the right side.

packages/eds-core-react/src/components/next/TextField/TextField.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+    return (
+      <Field disabled={disabled}>
+        {label && (
+          <div className="eds-text-field__header">
+            <Field.Label htmlFor={inputId} indicator={indicator}>
+              {label}
+            </Field.Label>
+            {labelInfo && (
+              <Tooltip title={labelInfo} placement="top">
```
pomfrida: This is the v1 Tooltip so we have to remember to replace this when Tooltip 2.0 is ready

packages/eds-core-react/src/components/next/TextField/TextField.tsx
```diff
@@ -0,0 +1,72 @@
...
+          disabled={disabled}
+          invalid={invalid}
+          aria-describedby={getDescribedBy({
+            hasDescription: !!description,
+            hasHelperMessage: !!helperMessage,
+          })}
+          {...inputProps}
+        />
+        {helperMessage && (
+          <Field.HelperMessage id={helperMessageId}>
```
pomfrida:
```suggestion
          <Field.HelperMessage id={helperMessageId} role={invalid ? 'alert' : undefined>
```
When invalid is true, the helper message could have role="alert" so screen readers announce the error automatically when it appears

packages/eds-core-react/src/components/next/TextField/TextField.types.ts
```diff
@@ -0,0 +1,18 @@
...
+  /** Indicator text shown after label, e.g. "(Required)" or "(Optional)" */
+  indicator?: string
+  /** Descriptive text that provides additional context for the field */
+  description?: ReactNode
+  /** Helper or validation message shown below the input */
+  helperMessage?: ReactNode
+  /** Unique identifier for the field */
+  id?: string
+} & InputProps &
+  InputHTMLAttributes<HTMLInputElement>
```
pomfrida: If InputProps gets the intersection with InputHTMLAttributes (as suggested), this file could be simplified to just & InputProps since the native attributes would already be included.

pomfrida (commented):
Posting my comments so far, will continue tomorrow

packages/eds-core-react/src/components/next/Input/Input.stories.tsx
millus: Yes, I agree! Good point :) And preferably they use input with a "normal" label most of the time, but there are some use cases where aria-label is what they look for. I will make a note in the docs!

packages/eds-core-react/src/components/next/Input/Input.types.ts
millus: Oooh, good observation, thank you! Will fix 👍

packages/eds-core-react/src/components/next/Field/FieldDescription.tsx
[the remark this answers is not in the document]
millus: Deleted 👍  This was a bug after merging

packages/eds-core-react/.storybook/preview.mjs
millus: Yes, fixed now!

packages/eds-core-react/src/components/next/TextField/TextField.tsx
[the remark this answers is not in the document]
millus: Yes, good catch! Fixed now :)

packages/eds-core-react/src/components/next/Input/Input.stories.tsx
millus: Thank you ❤️

packages/eds-core-react/src/components/next/Input/Input.tsx
millus: Yes, good point! Fixed now :)

packages/eds-core-react/src/components/next/Input/input.css
millus: I think this one is good now!

packages/eds-core-react/src/components/next/TextField/TextField.tsx
millus: Very good point, thank you!

packages/eds-core-react/src/components/next/TextField/TextField.types.ts
millus: Fixed! Always nice with simplification 💯

packages/eds-core-react/src/components/next/TextField/TextField.tsx
millus: Yes, we should write it down somewhere smart 👍

packages/eds-core-react/src/index.next.ts
pomfrida: +1 to Torleif's comment. To support SSR in /next:
1. Remove 'use client'
2. Remove the CSS auto-import
3. Users import CSS manually: import '@equinor/eds-core-react/next/index.css'
The comment already describes this pattern, so it's just about removing the two lines that contradict

millus: I had forgot about this one, very good catch! Now fixed! :)

documentation/how-to/BETA_RELEASE_GUIDE.md
```diff
@@ -4,12 +4,13 @@ This guide explains how beta releases work for EDS 2.0 components under the `/ne
...
-| --------------------- | -------------------------------------------------------- |
-| Install beta          | `npm install @equinor/eds-core-react@beta`               |
-| Import beta component | `import { Button } from '@equinor/eds-core-react/next'`  |
-| Commit for beta       | `feat(next): add new component`                          |
-| View beta versions    | `npm view @equinor/eds-core-react versions \| grep beta` |
+| Action                | Command                                                    |
+| --------------------- | ---------------------------------------------------------- |
+| Install beta          | `npm install @equinor/eds-core-react@beta`                 |
+| Import beta component | `import { Button } from '@equinor/eds-core-react/next'`    |
+| Import CSS            | `import '@equinor/eds-core-react/next/index.css'`          |
```
pomfrida, line 11: Very good!

packages/eds-core-react/src/components/next/Input/Input.docs.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+## Usage
+
+```tsx
+import { Input } from '@equinor/eds-core-react/next'
+
+<Input placeholder="Enter text" aria-label="Description of input" />
+```
+
+> **⚠️ Accessibility Note:** When using Input without a visible label (e.g., in search fields or toolbars), you **must** provide an `aria-label` for accessibility. For form fields with visible labels, use the **TextField** component instead, which handles accessibility automatically.
```
pomfrida, line 38: Great!

packages/eds-core-react/src/components/next/Input/input.css
```diff
@@ -0,0 +1,159 @@
+@layer eds-components {
+  /* Container */
+  .eds-input-container {
+    position: relative;
```
vnys: Consider using `isolation: isolate` as an alternative if the purpose of using `position: relative` here is to create a stacking order. That would eliminate some of the side effects of using `position: relative`.

packages/eds-core-react/src/components/next/Input/input.css
```diff
@@ -0,0 +1,159 @@
...
+    border-radius: var(--eds-spacing-border-radius-none);
+    padding-inline: var(--eds-selectable-space-horizontal);
+
+    /* Hover - only for enabled inputs */
+    &:hover:not([data-disabled]):not([data-readonly]) {
+      border-color: var(--eds-color-border-strong);
+
+      & .eds-input {
+        color: var(--eds-color-text-strong);
+        -webkit-text-fill-color: var(--eds-color-text-strong);
```
vnys, line 29: What’s the purpose of `-webkit-text-fill-color` here?

packages/eds-core-react/src/components/next/Input/input.css
```diff
@@ -0,0 +1,159 @@
...
+        }
+      }
+    }
+
+    /* Focus state */
+    &:focus-within {
+      border-color: transparent;
+      outline: var(--eds-sizing-stroke-thick) solid
+        var(--eds-color-border-focus);
+      outline-offset: calc(-1 * var(--eds-sizing-stroke-thick));
```
vnys: 👍🏻

packages/eds-core-react/src/components/next/Input/Input.docs.mdx
vnys, line 38: 👍

packages/eds-core-react/src/components/next/Input/Input.docs.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+&nbsp;
```
vnys:
```suggestion
```

vnys: I’m just assuming this is an error.

packages/eds-core-react/src/components/next/Input/input.css
```diff
@@ -0,0 +1,159 @@
...
+    display: flex;
+    flex-direction: row;
+    align-items: center;
+    gap: var(--eds-typography-gap-horizontal);
+    box-sizing: border-box;
+    inline-size: 100%;
+    background: var(--eds-color-bg-canvas);
+    border: var(--eds-sizing-stroke-thin) solid var(--eds-color-border-subtle);
+    border-radius: var(--eds-spacing-border-radius-none);
+    padding-inline: var(--eds-selectable-space-horizontal);
```
vnys, line 16: padding-block is missing. The input field is now 32px, should be same as button, 36px.

packages/eds-core-react/src/components/next/Input/Input.docs.mdx
millus: Removed

packages/eds-core-react/src/components/next/Input/input.css
millus: For safari, added a comment 👍

packages/eds-core-react/src/components/next/Input/input.css
millus: Added padding-block: var(--eds-selectable-space-vertical) to the container. The height is now 38px total (36px content + 2px border), achieved by using data-selectable-space="sm" with data-space-proportions="squished" which gives 8px vertical padding. Combined with the 20px line-height from data-font-size="md", this gives:

8px top padding + 20px line-height + 8px bottom padding = 36px content height
Plus 2px border (1px each side) = 38px total
This matches the Figma design spec of 36px field height (without border).

packages/eds-core-react/src/components/next/Input/input.css
millus: Thank you for the suggestion, changed this now 👍

packages/eds-core-react/src/components/next/Input/input.css
```diff
@@ -0,0 +1,154 @@
...
+    padding-block: var(--eds-selectable-space-vertical);
+
```
eddiman, line 21:
```suggestion
    transition:
      background-color 150ms,
      border-color 150ms;
```

packages/eds-core-react/src/components/next/Input/input.css
```diff
@@ -0,0 +1,154 @@
...
+    flex-direction: row;
+    align-items: center;
+    gap: var(--eds-typography-gap-horizontal);
+    box-sizing: border-box;
+    inline-size: 100%;
+    background: var(--eds-color-bg-canvas);
+    border: var(--eds-sizing-stroke-thin) solid var(--eds-color-border-subtle);
+    border-radius: var(--eds-spacing-border-radius-none);
+    padding-inline: var(--eds-selectable-space-horizontal);
+    padding-block: var(--eds-selectable-space-vertical);
```
eddiman, line 17: I saw previous comment when it comes to height: Figma technically uses an equivalent of outline in its borders, it doesn't add to the height. The rendered height of the element has to match Figma exactly unfortunately. On proposal would be to offset the border in the vertical padding such as this. Other solution would be to change to using outline.

```
padding-block: calc(var(--eds-selectable-space-vertical) - 1px)
      calc(var(--eds-selectable-space-vertical) - 1px);
```

millus: Fixed, changed from border to outline! Thank you :)

packages/eds-core-react/src/components/next/Input/input.css
millus: Fixed! 👍

PR #4231, opened 2025-11-12
feat: new icons Horizontal unfold more + Horizontal unfold less 

### LINKED ISSUE #2388: New horizontal icon variants for unfold_more and unfold_less

In some situations where the user can show or hide interface parts, the icons `unfold_more` and `unfold_less` are often suitable. However, when we arrange the content horizontally, these icons might signify something slightly different than what's happening visually.

It's simple enough to rotate the icons in code, but in the Figma components, we would have to rotate the whole button as long as we don't use the icon as a standalone component.

**A clear and concise description of what you want to happen.**

It would be helpful to have horizontal variants of the mentioned icons (see example below).

image

pomfrida:
resolves #2388 
some other icon updates from figma as well (minor pixel adjustments)
also clarifying the figma broker readme to avoid future frustration

### DIFF
--- a/scripts/figma-broker/README.md
+++ b/scripts/figma-broker/README.md
@@ -8,7 +8,7 @@ Extracts design decisions into [design tokens] from Figma using the [Figma Web A
 ## Usage
 
 ### Setup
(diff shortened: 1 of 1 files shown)


pomfrida:
New Icons added:
- [x] Horizontal unfold more
- [x] Horizontal unfold less
- [ ] Drag Indicator
- [ ] Drag Handler
- [ ] Experiment
- [ ] Anything else @eddiman 

Some of these were not added while running the script so if I'm not doing it wrong, something is going on with Figma Broker.
Please verify @vnys

PR #4345, opened 2026-01-02
feat: adds chips in autocomplete

FredrikMWold:
This PR includes two main changes:

### 1. Refactor
- Use context to pass all ComboBox props.
- Move the logic into a custom hook.

### 2. Show selected items as chips
- Adds the ability to display selected items as chips inside the Autocomplete input (Most of these changes are contained in the `MultipleInput` component).

Right now, the `multiple` flag controls showing chips. How would you like to handle this new feature, the way it is now would be a breaking change. It could be separate flag specifically for showing chips?


FredrikMWold:
I thin all the things you mentioned should be fixed now. Including some improved keyboard interactivity

PR #4361, opened 2026-01-12
chore(figma-broker): sync icons from Figma

eddiman:
## Summary

Sync system icons from the Figma Assets file to update the icon library.

## Changes

- Updated SVG files in `assets/icons/system-icons/`
- Regenerated `packages/eds-icons/src/data.ts`
- Updated Storybook metadata in `packages/eds-core-react/stories/assets/icons/system-icons.json`


packages/eds-core-react/src/components/Search/__snapshots__/Search.test.tsx.snap
```diff
@@ -113,9 +113,10 @@ exports[`Search Matches snapshot 1`] = `
           <path
             class=""
             clip-rule="evenodd"
-            d="M11.224 10.691h.592l3.743 3.75-1.118 1.118-3.75-3.743v-.592l-.202-.21a4.85 4.85 0 0 1-3.173 1.177 4.875 4.875 0 1 1 4.875-4.875 4.85 4.85 0 0 1-1.177 3.173zM3.94 7.316a3.37 3.37 0 0 0 3.375 3.375 3.37 3.37 0 0 0 3.375-3.375 3.37 3.37 0 0 0-3.375-3.375 3.37 3.37 0 0 0-3.375 3.375"
+            d="M14.966 14.255h.79l4.99 5-1.49 1.49-5-4.99v-.79l-.27-.28a6.47 6.47 0 0 1-4.23 1.57 6.5 6.5 0 1 1 6.5-6.5c0 1.61-.59 3.09-1.57 4.23zm-9.71-4.5c0 2.49 2.01 4.5 4.5 4.5s4.5-2.01 4.5-4.5-2.01-4.5-4.5-4.5-4.5 2.01-4.5 4.5"
             data-testid="eds-icon-path"
             fill-rule="evenodd"
+            transform="scale(0.75)"
```
pomfrida, line 119: Interesting. Is this transform att expected?

eddiman: That is interesting, I didn't expect the snapshot update to change anything but the path data.

From what I found out, there is a fallback in place in the icon-component itself, when the svg icon doesn't have an 18px variant. In the new broker I chose to remove the small variants of the icons (in figma we dont use them and it was only a few icons that had it). Apparently Oddbjørn thought of this and wrote this transform scale so it scales down to 18px: [Icon.tsx:44](https://github.com/equinor/design-system/blob/main/packages/eds-core-react/src/components/Icon/Icon.tsx#L48)

So when an icon sets a size other than 24px, it calculates a scale and puts that inside the svg path.

pomfrida (approved):
NICE

PR #4388, opened 2026-01-21
feat: Radio 2.0  component

### LINKED ISSUE #4358: Update Radio Button in code
## Overview
Add the Radio component using vanilla CSS with EDS foundation tokens. This aligns with the EDS 2.0 migration strategy and follows the same patterns established by Checkbox 2.0.

### Implementation
- [x] Vanilla CSS with BEM naming convention
- [x] EDS foundation and variables tokens from `@equinor/eds-tokens`
- [x] Full accessibility support (aria-describedby, disabled states)
- [x] Support for dark mode via `data-color-scheme` attribute
- [x] Support for compact/comfortable density modes
- [x] Field component integration for label, helper text, and layout
- [x] Figma Code Connect mapping
- [x] Unit tests

pomfrida:
Closes #4358

## Summary
- Add Radio component to EDS 2.0 (`/next`)
- Vanilla CSS with BEM naming and EDS foundation tokens
- Full accessibility support, dark mode, and density modes
- Figma Code Connect mapping

### DIFF
--- a/packages/eds-core-react/src/components/next/Radio/Radio.figma.tsx
+++ b/packages/eds-core-react/src/components/next/Radio/Radio.figma.tsx
@@ -1,0 +1,16 @@
+import figma from '@figma/code-connect'
+import { Radio } from './Radio'
+
+figma.connect(
+  Radio,
+  'https://www.figma.com/design/dz0XQdc5j7AAtjXr1gTfVR/EDS-Core-Components?node-id=112-6322',
+  {
+    props: {
+      label: figma.string('Label'),
+      disabled: figma.boolean('Disabled'),
+    },
+    example: ({ label, disabled }) => (
+      <Radio label={label} disabled={disabled} name="radio-group" />
+    ),
+  },
+)
--- a/packages/eds-core-react/src/components/next/Radio/Radio.stories.tsx
+++ b/packages/eds-core-react/src/components/next/Radio/Radio.stories.tsx
@@ -1,0 +1,279 @@
+import { useState } from 'react'
+import type { ChangeEvent, ReactNode } from 'react'
+import { StoryFn, Meta } from '@storybook/react-vite'
+import { Stack } from './../../../../.storybook/components'
+import { Radio } from './Radio'
+import type { RadioProps } from './Radio.types'
+
+const meta: Meta<typeof Radio> = {
+  title: 'EDS 2.0 (beta)/Inputs/Selection Controls/Radio',
+  component: Radio,
+  args: {
+    label: 'Option',
+    disabled: false,
+  },
+  parameters: {
+    docs: {
+      description: {
+        component: `**⚠️ Beta Component** - This component is under active development and may have breaking changes.
+
+\`\`\`bash
+npm install @equinor/eds-core-react@beta
+\`\`\`
+
+\`\`\`tsx
+import { Radio } from '@equinor/eds-core-react/next'
+
+// Radio group
+<fieldset>
+  <legend>Select option</legend>
+  <Radio label="Option 1" name="group" value="1" />
+  <Radio label="Option 2" name="group" value="2" />
+</fieldset>
+
+// Without visible label (use aria-label)
+<Radio aria-label="Select row" name="table-row" />
+\`\`\`
+
+Radio buttons allow users to select one option from a set. Always use within a group with the same \`name\` attribute.
+`,
+      },
+      source: {
+        excludeDecorators: true,
+      },
+    },
+  },
+  argTypes: {
+    label: {
+      control: 'text',
+      description: 'Label text displayed next to the radio button',
+      table: {
+        category: 'Core',
+      },
+    },
+    disabled: {
+      control: 'boolean',
+      description: 'Disables radio button interaction',
+      table: {
+        category: 'States',
+        defaultValue: { summary: 'false' },
+      },
+    },
+    checked: {
+      control: 'boolean',
+      description: 'Controlled checked state',
+      table: {
+        category: 'States',
+      },
+    },
+    defaultChecked: {
+      control: 'boolean',
+      description: 'Initial checked state for uncontrolled usage',
+      table: {
+        category: 'States',
+      },
+    },
+    className: {
+      control: 'text',
+      description: 'Additional CSS class names for the input element',
+      table: {
+        category: 'Styling',
+      },
+    },
+    id: {
+      control: 'text',
+      description: 'HTML id attribute',
+      table: {
+        category: 'HTML Attributes',
+      },
+    },
+    name: {
+      control: 'text',
+      description: 'HTML name attribute for form submission',
+      table: {
+        category: 'HTML Attributes',
+      },
+    },
+    value: {
+      control: 'text',
+      description: 'HTML value attribute for form submission',
+      table: {
+        category: 'HTML Attributes',
+      },
+    },
+  },
+  decorators: [
+    (Story) => (
+      <Stack>
+        <Story />
+      </Stack>
+    ),
+  ],
+}
+
+export default meta
+
+const Wrapper = ({
+  children,
+  gap = 16,
+  ...rest
+}: {
+  children: ReactNode
+  gap?: number
+} & React.HTMLAttributes<HTMLDivElement>) => (
+  <div
+    style={{
+      display: 'flex',
+      flexDirection: 'column',
+      gap: `${gap}px`,
+      alignItems: 'flex-start',
+    }}
+    {...rest}
+  >
+    {children}
+  </div>
+)
+
+const AllVariants = ({ prefix = 'default' }: { prefix?: string }) => (
+  <>
+    <Radio label="Unchecked" name={`${prefix}-unchecked`} value="unchecked" />
+    <Radio
+      label="Checked"
+      name={`${prefix}-checked`}
+      value="checked"
+      defaultChecked
+    />
+    <Radio label="Disabled" name={`${prefix}-disabled`} disabled />
+    <Radio
+      label="Disabled checked"
+      name={`${prefix}-disabled-checked`}
+      disabled
+      defaultChecked
+    />
+  </>
+)
+
+export const Introduction: StoryFn<RadioProps> = (args) => {
+  return <Radio {...args} />
+}
+
+export const Spacious: StoryFn<RadioProps> = () => (
+  <Wrapper data-density="spacious">
+    <AllVariants prefix="spacious" />
+  </Wrapper>
+)
+
+export const Comfortable: StoryFn<RadioProps> = () => (
+  <Wrapper data-density="comfortable">
+    <AllVariants prefix="comfortable" />
+  </Wrapper>
+)
+
+export const ColorSchemes: StoryFn<RadioProps> = () => (
+  <div style={{ display: 'flex', gap: '32px' }}>
+    <div
+      data-color-scheme="light"
+      style={{ padding: '24px', background: 'var(--eds-color-bg-canvas)' }}
+    >
+      <Wrapper>
+        <h3 style={{ margin: 0 }}>Light Mode</h3>
+        <Radio label="Unchecked" name="light" value="unchecked" />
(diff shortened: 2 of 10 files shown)


packages/eds-core-react/src/components/next/Radio/Radio.test.tsx
```diff
@@ -0,0 +1,219 @@
...
+        <Radio
+          label="radio-test"
+          name="test"
+          className="custom-radio"
+          style={{ clipPath: 'unset' }}
+        />,
+      )
+      const radio = screen.getByLabelText('radio-test')
+      expect(radio).toBeInTheDocument()
+      expect(radio).toHaveClass('custom-radio')
```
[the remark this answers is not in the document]
pomfrida: The className going to the input element is intentional. We made the same change in Checkbox - letting className pass through via ...rest to the input is more intuitive and consistent with standard HTML behavior. If users need to style the wrapper, they can use CSS selectors like .my-class .eds-radio or compose with Field directly.

packages/eds-core-react/src/components/next/Radio/Radio.figma.tsx
```diff
@@ -0,0 +1,16 @@
+import figma from '@figma/code-connect'
+import { Radio } from './Radio'
+
+figma.connect(
+  Radio,
+  'https://www.figma.com/design/dz0XQdc5j7AAtjXr1gTfVR/EDS-Core-Components?node-id=112-6322',
```
eddiman, line 6:
```suggestion
  'https://www.figma.com/design/dz0XQdc5j7AAtjXr1gTfVR/EDS-Core-Components?node-id=18-3449',
```

Should refer to the actual set of the main component, not the section. We have to look at the actual implementation of code connect at a later point though.

packages/eds-core-react/src/components/next/Radio/Radio.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+const classNames = (...classes: (string | boolean | undefined)[]) =>
+  classes.filter(Boolean).join(' ')
+
+export const Radio = forwardRef<HTMLInputElement, RadioProps>(function Radio(
+  {
+    label,
+    disabled = false,
+    indicator,
+    helperMessage,
```
millus: Are we supporting indicator and helpermessage for Radio? or was this decided not to be on the selection controls yet?

packages/eds-core-react/src/components/next/Radio/Radio.tsx
```diff
@@ -0,0 +1,91 @@
...
+    id: providedId,
+    ...rest
+  },
+  ref,
+) {
+  const generatedId = useId()
```
millus, line 18: Should we use the usefieldids hook here? Im not sure what is correct, I did use it for Switch, but maybe overkill when its just for the label. It can be useId() is good here as well, we can talk and make the selection controls align on monday. This is not a blocker, both works fine.

packages/eds-core-react/src/components/next/Radio/Radio.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const inputId = providedId ?? generatedId
+  const helperMessageId = `${inputId}-helper`
```
millus: helpermessageId not needed if we are not supporting indicator and helpermessage for selection controls at this stage

packages/eds-core-react/src/components/next/Radio/Radio.tsx
```diff
@@ -0,0 +1,91 @@
...
+        data-space-proportions="squished"
+      >
+        {radioInput}
+        <Field.Label htmlFor={inputId} indicator={indicator}>
+          {label}
+        </Field.Label>
+        {helperMessage && (
+          <Field.HelperMessage id={helperMessageId}>
+            {helperMessage}
+          </Field.HelperMessage>
```
millus: Remove indicator and helpermessage

packages/eds-core-react/src/components/next/Radio/Radio.types.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export type RadioProps = {
+  /** Label for the radio button */
+  label?: ReactNode
+  /** If true, the radio button will be disabled */
+  disabled?: boolean
+  /** Indicator text shown after the label, e.g. "(Required)" or "(Optional)" */
+  indicator?: string
+  /** Helper message shown below the radio button, useful for additional context */
+  helperMessage?: ReactNode
```
millus: `disabled ` is a part of InputHtmlAttributes, so we dont need to have it explicit here. Indicator and helpermessage can be removed from here as well since (if) we are not supporting that yet.

packages/eds-core-react/src/components/next/Radio/radio.css
```diff
@@ -0,0 +1,151 @@
...
+  .eds-radio__input:focus {
+    outline: none;
+  }
+
+  .eds-radio:has(.eds-radio__input:focus-visible)::after {
+    content: '';
+    position: absolute;
+    inset: -2px;
+    border: 1px solid var(--eds-color-border-focus);
+    border-radius: 6px;
```
millus: The focus ring uses hardcoded values. In Switch we use outline for this.

Could this approach work here too? It would be more consistent and use proper tokens instead of hardcoded 1px, -2px, and 6px."

The outline approach is cleaner since it doesn't need ::after pseudo-element and automatically handles border-radius matching the element.

So to align and avoid hardcoded values, this could be used: 
```
outline: var(--eds-sizing-stroke-thin) solid var(--eds-color-border-focus);
outline-offset: var(--eds-sizing-stroke-thick);
```

packages/eds-core-react/src/components/next/Radio/radio.css
```diff
@@ -0,0 +1,151 @@
...
+    pointer-events: none;
+  }
+
+  .eds-radio--standalone:has(.eds-radio__input:focus-visible)::after {
+    border-radius: var(--eds-spacing-border-radius-pill);
+  }
```
millus: If we do the outline for focus, I dont think you need this part anymore as well.

packages/eds-core-react/src/components/next/Radio/radio.css
```diff
@@ -0,0 +1,151 @@
...
+
+  .eds-radio__icon {
+    position: relative;
+    pointer-events: none;
+    fill: var(--_radio-icon-color);
+    flex-shrink: 0;
+  }
+
+  .eds-radio[data-disabled='true'] .eds-radio__icon {
+    fill: var(--eds-color-border-neutral-medium);
```
millus: Nitpick: But should we avoid using static color variables, hence not have the `-neutral-` part in. I dont know how strict it is, but I think we should try to look for a way to keep to the dynamic way for colors. But sometimes it gets overly complex trying to do that.

packages/eds-core-react/src/components/next/Radio/radio.css
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  .eds-field.eds-radio[data-disabled='true'] .eds-field__label {
+    color: var(--eds-color-border-neutral-medium);
+    cursor: not-allowed;
+  }
+
+  .eds-field.eds-radio[data-color-appearance='accent'] {
+    --_eds-field-helper-color: var(--eds-color-text-neutral-subtle);
+  }
```
millus: Same here, I dont think we are suppose to use the static color variables with `-neutral-` in the name.

millus: But in the same time, when is something thats disabled going to be colorful, it should always be neutral? I think I read in slack that we are getting a token thats static specific for disabled so we can use that. So I could probably simplify Switch also then.

packages/eds-core-react/src/components/next/Radio/Radio.tsx
```diff
@@ -0,0 +1,91 @@
+/* eslint camelcase: "off" */
+import { forwardRef, useId } from 'react'
+import {
+  radio_button_selected,
+  radio_button_unselected,
+} from '@equinor/eds-icons'
+import { Field } from '../Field'
+import { Icon } from '../Icon'
+import type { RadioProps } from './Radio.types'
+import './radio.css'
```
millus: We dont need to have import for radio.css here, we can just have it in the global css. Lets align!

packages/eds-core-react/src/components/next/index.css
```diff
@@ -1,7 +1,7 @@
 /* EDS 2.0 Next Components - CSS */
 @import '@equinor/eds-tokens/css/variables';
+@import './Field/field.css';
+@import './Radio/radio.css';
 
 /* Define layer order - eds-components has lowest priority for easy user overrides */
 @layer eds-components;
```
millus: The `@layer eds-components;` should be on the top, before the imports.

packages/eds-core-react/src/components/next/Radio/Radio.stories.tsx
```diff
@@ -0,0 +1,266 @@
...
+
+export const Introduction: StoryFn<RadioProps> = (args) => {
+  return <Radio {...args} />
+}
+
+export const Spacious: StoryFn<RadioProps> = () => (
+  <Wrapper data-density="spacious">
+    <AllVariants />
+  </Wrapper>
+)
```
millus, line 164: In the spacious story, the one named checked and disabled checked is not checked when I run storybook locally.

packages/eds-core-react/src/components/next/Radio/Radio.stories.tsx
```diff
@@ -0,0 +1,266 @@
...
+export const GroupedRadio: StoryFn<RadioProps> = () => (
+  <fieldset>
+    <legend>Select your option</legend>
+    <Wrapper gap={8}>
+      <Radio label="Option 1" name="group" value="1" />
+      <Radio label="Option 2" name="group" value="2" />
+      <Radio label="Option 3" name="group" value="3" />
+    </Wrapper>
+  </fieldset>
+)
```
millus, line 223: I like the UX here, so we can see it in action how selecting one deselects another one 👍

packages/eds-core-react/src/components/next/Radio/radio.css
```diff
@@ -0,0 +1,151 @@
...
+    position: relative;
+    box-sizing: border-box;
+  }
+
+  .eds-radio--standalone {
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+    position: relative;
+    padding: 0.375rem; /* 6px for spacious: 24px icon + 12px = 36px */
```
millus: Is it possible to use calc and icon size token to avoid hardcoded values?

Several hardcoded values tied to icon sizing:

Line 15: padding: 0.375rem (spacious standalone)
Line 149: padding: 0.25rem (comfortable standalone)
Line 20: gap: 0.925rem (compensating for negative margin)

Could these be derived from icon tokens? e.g. `calc(var(--eds-typography-icon-size) * 0.25)` or similar? Would make them scale automatically with density changes. But then we also need to set a data-font-size attribute to make the `--eds-typography-icon-size` dynamic value work, so yeah, I just noticed the hardcoded values so just wanted to make a note on it. But it may be perfectly fine as it is.

packages/eds-core-react/src/components/next/Radio/Radio.tsx
pomfrida: As discussed on Slack: we align on selection controls (who only use label from field) to not use the useFieldHook there. It makes more sense to use that hook when we use multiple components from Field

packages/eds-core-react/src/components/next/index.css
pomfrida: Oh but don't you also get `[vite:css][postcss] @import must precede all other statements (besides @charset or empty @layer)` when running storybook then? I had to do it like this to avoid the warning in console. CSS spec allows @layer statements before @import, but looks like PostCSS doesn't support it yet

pomfrida: I'll just ignore the warnings

PR #4389, opened 2026-01-21
feat: Button 2.0 component

pomfrida:
## Summary

Adds a new Button component to EDS 2.0 (`/next`) with full Figma design spec compliance.

### Features

- **Variants:** `primary`, `secondary`, `ghost`
- **Sizes:** `small`, `default`, `large`
- **Tones:** `accent`, `neutral`, `danger`
- **Icon support:** 
  - Icons as children with automatic spacing
  - Icon-only mode via `icon` prop (square buttons)
  - Round icon buttons via `round` prop
- **Density modes:** Supports both spacious and comfortable density
- **Accessibility:** Full keyboard navigation and ARIA support

### Technical

- CSS-variable based styling with data attributes
- Figma Code Connect integration for both Button and Icon Button
- Comprehensive test coverage (63 tests)

## Notes: 
  Comfortable density mode caused a brain twist for me + Claude. We had to add CSS overrides because of an inconsistency in Figma's design specs:                                                  
                                                                                                                                                                                                   
  - Comfortable density uses font-size one step down (e.g., large uses MD font)                                                                                                                    
  - BUT line-height stays at the original size (large keeps LG line-height)                                                                                                                        
                                                                                                                                                                                                   
  Our typography tokens couple font-size and line-height together, so we needed manual overrides to decouple them for comfortable mode. Is this an ok approach in code or should tokens handle this?

### DIFF
--- a/packages/eds-core-react/src/components/next/Button/Button.figma.tsx
+++ b/packages/eds-core-react/src/components/next/Button/Button.figma.tsx
@@ -1,0 +1,125 @@
+import figma from '@figma/code-connect'
+import { Button } from './Button'
+
+/**
+ * Code Connect mapping for Button component.
+ *
+ * Maps Figma design variants to React component props.
+ * Run `npx figma connect publish` from packages/eds-core-react/ to publish.
+ *
+ * Figma component structure:
+ * - Button [EDS] (18:1193) → .Geometry Options → ⌘ Button
+ * - Icon Button [EDS] (18:1373) → .Geometry Options → ⌘ Icon Button
+ */
+
+// ============================================
+// Connection 1: Button [EDS] (18:1193)
+// Top-level wrapper with Tone + nested geometry + nested button
+// ============================================
+
+figma.connect(
+  Button,
+  'https://www.figma.com/design/dz0XQdc5j7AAtjXr1gTfVR/EDS-Core-Components?node-id=18-1193',
+  {
+    props: {
+      tone: figma.enum('Tone', {
+        Accent: 'accent',
+        Neutral: 'neutral',
+        Danger: 'danger',
+      }),
+      geometry: figma.nestedProps('.Geometry Options', {
+        size: figma.enum('Size', {
+          Large: 'large',
+          Default: 'default',
+          Small: 'small',
+        }),
+      }),
+      inner: figma.nestedProps('⌘ Button', {
+        variant: figma.enum('Variant', {
+          Primary: 'primary',
+          Outline: 'secondary',
+          Ghost: 'ghost',
+        }),
+        disabled: figma.enum('State', {
+          Disabled: true,
+        }),
+        leadingIcon: figma.boolean('Show Leading Icon', {
+          true: figma.instance('Leading Icon Item'),
+          false: undefined,
+        }),
+        trailingIcon: figma.boolean('Show Trailing Icon', {
+          true: figma.instance('Trailing Icon Item'),
+          false: undefined,
+        }),
+        label: figma.string('Label'),
+      }),
+    },
+    example: ({ tone, geometry, inner }) => (
+      <Button
+        variant={inner.variant}
+        size={geometry.size}
+        tone={tone}
+        disabled={inner.disabled}
+      >
+        {/* TODO: Replace with actual icon components, when Icon component is
+        available in Code Connect */}
+        {inner.leadingIcon}
+        {inner.label}
+        {inner.trailingIcon}
+      </Button>
+    ),
+  },
+)
+
+// ============================================
+// Connection 2: Icon Button [EDS] (18:1373)
+// Top-level wrapper with Tone + nested geometry (includes Round) + nested icon button
+// ============================================
+
+figma.connect(
+  Button,
+  'https://www.figma.com/design/dz0XQdc5j7AAtjXr1gTfVR/EDS-Core-Components?node-id=18-1373',
+  {
+    props: {
+      tone: figma.enum('Tone', {
+        Accent: 'accent',
+        Neutral: 'neutral',
+        Danger: 'danger',
+      }),
+      geometry: figma.nestedProps('.Geometry Options', {
+        size: figma.enum('Size', {
+          Large: 'large',
+          Default: 'default',
+          Small: 'small',
+        }),
+        round: figma.boolean('Round'),
+      }),
+      // @ts-expect-error - Code Connect's ConnectedComponent type doesn't match ReactElement
+      inner: figma.nestedProps('⌘ Icon Button', {
+        variant: figma.enum('Variant', {
+          Primary: 'primary',
+          Outline: 'secondary',
+          Ghost: 'ghost',
+        }),
+        disabled: figma.enum('State', {
+          Disabled: true,
+        }),
+        iconContent: figma.instance('icon'),
+      }),
+    },
+    example: ({ tone, geometry, inner }) => (
+      <Button
+        icon
+        variant={inner.variant}
+        size={geometry.size}
+        tone={tone}
+        round={geometry.round}
+        disabled={inner.disabled}
+        /* Remember to describe the action of the icon button for accessibility */
+        aria-label="[Describe action]"
+      >
+        {inner.iconContent}
+      </Button>
+    ),
+  },
+)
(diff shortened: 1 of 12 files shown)


packages/eds-core-react/src/components/next/Button/Button.stories.tsx
```diff
@@ -0,0 +1,451 @@
...
+}
+
+const meta: Meta<StoryArgs> = {
+  title: 'EDS 2.0 (beta)/Button',
+  component: Button,
+  parameters: {
+    docs: {
+      description: {
+        component: `
+Button component for triggering actions.
```
pomfrida:
```suggestion
Button component for triggering actions.

**⚠️ Beta Component** - This component is under active development and may have breaking changes.

\`\`\`bash
npm install @equinor/eds-core-react@beta
\`\`\`
```


Should be added to all beta stories to make it extra clear to users

packages/eds-core-react/src/components/next/Button/Button.stories.tsx
```diff
@@ -0,0 +1,451 @@
...
+      },
+    },
+  },
+}
+
+// ===== Color Appearances =====
+
+export const ColorAppearances: Story = {
+  render: () => (
+    <div style={{ display: 'flex', flexDirection: 'column', gap: '16px' }}>
```
pomfrida: Some inline styles are repeated. Consider extracting to a Wrapper component for consistency with other EDS stories

packages/eds-core-react/src/components/next/Button/Button.test.tsx
```diff
@@ -0,0 +1,514 @@
...
+    it('has no accessibility violations (square icon-only)', async () => {
+      const { container } = render(
+        <Button icon aria-label="Add">
+          <MockIcon />
+        </Button>,
+      )
+      expect(await axe(container)).toHaveNoViolations()
+    })
+  })
+})
```
pomfrida, line 515: 👍 Excellent test coverage!

packages/eds-core-react/src/components/next/Button/button.css
```diff
@@ -0,0 +1,121 @@
...
+
+  .eds-button[data-icon-only] {
+    aspect-ratio: 1;
+    padding: var(--eds-generic-space-vertical)
+      var(--eds-generic-space-horizontal);
+
+    /* TODO: Remove this when we revisit <Icon/> sizing*/
+    svg {
+      margin-block: 0 !important;
+      margin-inline: 0 !important;
```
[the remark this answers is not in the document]
pomfrida: ~~Hmm.. The !important workaround for Icon margins could potentially be solved by increasing selector specificity:~~
~~.eds-button[data-icon-only] > .icon {
  margin: 0;
 }~~
~~Alternatively, we should consider adding a standalone prop to Icon for cases where it's not inline with text. This could be addressed in a follow-up PR when revisiting Icon sizing if the suggestion above was not sufficient.~~

packages/eds-core-react/src/components/next/Button/Button.figma.tsx
```diff
@@ -0,0 +1,124 @@
...
+    },
+    example: ({ colorAppearance, geometry, inner }) => (
+      <Button
+        icon
+        variant={inner.variant}
+        size={geometry.size}
+        colorAppearance={colorAppearance}
+        round={geometry.round}
+        disabled={inner.disabled}
+        aria-label="Icon action"
```
pomfrida: Hmm aria-label="Icon action" is a placeholder that might be copied. Consider changing to a more obvious placeholder like "[Describe action]" or adding a code comment reminding developers to replace it with a meaningful label

packages/eds-core-react/src/components/next/Button/Button.stories.tsx
```diff
@@ -0,0 +1,451 @@
...
+  Add item
+</Button>
+
+// Icon after label
+<Button>
+  Next
+  <Icon data={chevron_right} aria-hidden />
+</Button>
+
+// Icon-only button (requires aria-label)
```
pomfrida: Good job focusing on accessibility here! 🙌

packages/eds-core-react/src/components/next/Button/button.css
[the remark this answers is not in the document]
pomfrida: Correction: Icon already resets margins when size prop is used (.icon[data-icon-size] { margin: 0 }). Recommended pattern: `<Icon data={add} size="md" />` in icon-only buttons. The workaround handles cases where users omit size.

packages/eds-core-react/src/components/next/Button/Button.tsx
```diff
@@ -0,0 +1,141 @@
...
+const sizeToIconSpace = {
+  small: '3xs',
+  default: '2xs',
+  large: 'xs',
+} as const
+
+const sizeToTypographySize = {
+  small: 'md',
+  default: 'lg',
+  large: 'xl',
```
vnys: This is not correct. The default size is `md`.

vnys: Also, in Figma the font-size is unchanged (MD) between small, default and large. In code it’s not, so it has to be updated in Figma.

vnys (changes requested):
- The font-size is too large, it should be md (`14px`) in the default button
- The large button should be `44px` tall, but it’s `44.72`.
- Changing the font-size property on the button only changes the icon, and not the label (which is consistently `16px`).
- We’re using `border-color-medium` for the text colour in the button, which we are not doing in text field. This has to be consistent. I assume this is what Alex is working on right now. 

For the disabled colour, I suggest for now that we add a token in the concept collection called `text-disabled` which references the medium neutral border colour.

packages/eds-core-react/src/components/next/Button/button.css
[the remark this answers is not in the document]
eddiman: The icon is supposed to follow font size, having users set the size each time they need an icon in a button sets up for wrong sizes or unnecessary time spent in finding the icon size in Figma.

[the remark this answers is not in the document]
pomfrida: Ok great, a potential future improvement could be a dedicated CSS selector in Icon for "non-inline" contexts, but the current approach works well!

packages/eds-core-react/src/components/next/Icon/icon.css
```diff
@@ -1,58 +1,60 @@
...
-.icon {
-  /*
-   * Set font-size first to establish the icon's size context.
-   * Then use 1em for width/height (relative to computed font-size).
-   * This avoids compound scaling (1.5em × 1.5em = 2.25em would be wrong).
-   */
-  font-size: var(--eds-typography-icon-size, 1.5em);
-  width: 1em;
-  height: 1em;
+@layer eds-components {
```
pomfrida, line 1: @layer needed to align with the rest of the next components (forgotten before)

packages/eds-core-react/src/components/next/Button/button.css
```diff
@@ -0,0 +1,347 @@
...
+}
+
+/* Icon-only: span should not affect layout */
+.eds-button[data-icon-only] > span {
+  display: contents;
+}
+
+/* ===== DENSITY OVERRIDES ===== */
+/*
+ * Comfortable density uses smaller font-sizes (one size down) and tighter spacing.
```
pomfrida, line 270: Comfortable density mode caused a brain twist for me + Claude. We had to add CSS overrides because of an inconsistency in Figma's design specs:

- Comfortable density uses font-size one step down (e.g., large uses MD font)
- BUT line-height stays at the original size (large keeps LG line-height)

Our typography tokens couple font-size and line-height together, so we needed manual overrides to decouple them for comfortable mode. Is this an ok approach in code or should tokens handle this?

eddiman: Optimally, the tokens should handle this, this slice has uncovered that there are a bit missing pieces and som misalignment in the code vs Figma tokens.
Personally, I think this is ok, _for now_ and I think we should merge this in. Then prioritize looking into fixing and aligning the tokens. It makes it easier to debug if we have all the components available in same branch.

PR #4458, opened 2026-02-02
docs: introduce Architecture Decision Records (ADRs)

### LINKED ISSUE #4457: Introduce Architecture Decision Records (ADRs)
## Background

As part of my coursework at **Arkitekturskolen**, I'm working on introducing Architecture Decision Records (ADRs) to our team. This issue tracks that work and invites discussion.

## What are ADRs?

ADRs are short documents that capture important architecture decisions along with their context and consequences. Each ADR describes a single decision and is stored in version control alongside the code.

**Benefits:**
- New team members understand *why* decisions were made
- Reduces repeated discussions about the same topics
- Creates a searchable history of architectural choices
- Decisions can be reviewed and evaluated over time

## What's been done

I've set up an initial ADR structure:

```
docs/adr/
  0000-template.md    # Template with explanatory comments
  0001-use-adr-for-architecture-decisions.md  # First ADR (meta!)
```

The template follows the [MADR](https://adr.github.io/madr/) format with helpful comments explaining each section.

## When to write an ADR

Suggested situations:
- Choosing a technology, framework, or library
- Making significant API design decisions
- Establishing patterns that affect multiple components
- Deciding on breaking changes
- Security or accessibility architecture decisions

**Rule of thumb:** If the decision is hard to reverse or affects multiple systems/teams, write an ADR.

## Feedback wanted

- Does the template make sense? Is anything unclear?
- Are there past decisions we should retroactively document?
- Should we adjust the process for proposing/accepting ADRs?

## Resources

- [MADR - Markdown Any Decision Records](https://adr.github.io/madr/)
- [Michael Nygard's original ADR article](https://cognitect.com/blog/2011/11/15/documenting-architecture-decisions)
- [ADR GitHub organization](https://adr.github.io/)

pomfrida:
## Summary

- Add ADR directory structure (`docs/adr/`)
- Add MADR template with explanatory comments
- Add ADR 0001: Decision to use ADRs
- Add ADR 0002: Use vanilla CSS with design tokens for EDS 2.0
- Add link to ADRs in README quick links

## Background

This is part of my coursework at **Arkitekturskolen** – introducing ADRs to the team.

See #4457 for discussion and context.

## What are ADRs?

Architecture Decision Records are short documents that capture important architecture decisions along with their context and consequences. Benefits include:

- New team members understand *why* decisions were made
- Reduces repeated discussions about the same topics
- Creates a searchable history of architectural choices

## Files added

| File | Purpose |
|------|---------|
| `docs/adr/0000-template.md` | MADR template with helpful comments |
| `docs/adr/0001-use-adr-for-architecture-decisions.md` | Meta-ADR documenting decision to use ADRs |
| `docs/adr/0002-use-vanilla-css-with-design-tokens-for-eds-2.md` | Documents EDS 2.0 styling approach |

## Template highlights

Based on [MADR 4.0](https://adr.github.io/madr/) with some adaptations:

- **Status**: `Proposed | Accepted | Rejected | Deprecated | Superseded`
- **Consequences** under Decision (not separate section) - clarifies difference from Pros/Cons
- **Confirmation** section - how we verify the decision is followed
- **Public by default** - only sensitive info goes in internal repo

## When to write an ADR

- Choosing a technology, framework, or library
- Making significant API design decisions
- Establishing patterns that affect multiple components
- Deciding on breaking changes

## Feedback welcome!

- Does the template make sense?
- Are there past decisions we should retroactively document?

Closes #4457

### DIFF
--- a/docs/adr/0002-use-vanilla-css-with-design-tokens-for-eds-2.md
+++ b/docs/adr/0002-use-vanilla-css-with-design-tokens-for-eds-2.md
@@ -29,24 +29,7 @@ For EDS 2.0, we need a styling approach that is future-proof and works seamlessl
 
 ## Options Considered
 
-### Option 1: Vanilla CSS with CSS custom properties
-
-Use plain CSS files with CSS custom properties (design tokens) for theming.
-
-**Pros:**
-
-- Zero runtime overhead
-- Native browser support, excellent DevTools experience
-- Works perfectly with SSR and RSC
-- CSS custom properties enable theming without JavaScript
-- Team already familiar with CSS
-
-**Cons:**
-
-- No automatic scoping (requires naming conventions like BEM)
-- Less type-safety than CSS-in-JS
-
-### Option 2: CSS Modules
+### CSS Modules
 
 Scoped CSS with automatic class name generation.
 
@@ -62,7 +45,7 @@ Scoped CSS with automatic class name generation.
 - Harder to override styles from consuming applications
 - Less flexible for theming compared to CSS custom properties
 
-### Option 3: Tailwind CSS
+### Tailwind CSS
 
 Utility-first CSS framework.
 
@@ -79,7 +62,7 @@ Utility-first CSS framework.
 - Utility classes don't align well with design token architecture
 - Harder for consumers to understand component styling
 
-### Option 4: Keep styled-components
+### Keep styled-components
 
 Continue with current CSS-in-JS approach.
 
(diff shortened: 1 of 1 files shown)


vnys (changes requested):
So the PR is actually about two separate things: 

1. Should we use ADRs, and specifically, should we use the Markdown variety. For this part of the PR I agree, I think it’s a great addition to our workflow and solves the issue we’ve had with internal documentation.
2. Should we use vanilla CSS, and specifically should we use a BEM-like syntax. For the first part, yes, I think we should. About BEM, though, I’m not totally convinced. One of the issues with BEM, is that we’re replicating the HTML structure in CSS, which means if we restructure the HTML we also need to change the CSS selectors. Another issue is that we end up with very long class-names.

Personally I like the SuitCSS naming convention. It’s similar to BEM, but less verbose:

1. Namespaces are lowercase
2. Components are PascalCase
3. Hyphens separate components and descendants
4. Double hyphens separate component and modifier
5. States are prefixed with verb.

So for example:

- `eds-Button`
- `eds-Button--primary`
- `eds-Button-label`
- `eds-Button.is-readonly`

I know some of these are obsolete because we use data-attributes, and perhaps we don’t end up with long class-names if we agree on how many levels we nest descendant-selectors, so maybe it’s not a problem – but I’d like to have a brief discussion at least.

Hjaf (commented):
LGTM 👍

PR #4472, opened 2026-02-06
fix: Code scanning alerts for workflows missing permissions

Hjaf:
Potential fix for [https://github.com/equinor/design-system/security/code-scanning/3](https://github.com/equinor/design-system/security/code-scanning/3)

In general, the fix is to add an explicit `permissions` block either at the top-level of the workflow (applying to all jobs that don't override it) or per job, and restrict `GITHUB_TOKEN` to the minimal scopes required. For this workflow, the `packages` job only needs to read the repository contents for checkout/cache and to run local commands. The `setup` job is a reusable workflow call; without seeing it, the safest minimal default is still read-only `contents` (and possibly `packages` if it pulls packages from GitHub Packages).

The single best fix without altering functionality is:
- Add a top-level `permissions` block after the `on:` section, setting `contents: read` (and optionally `packages: read` if your org uses GitHub Packages in setup). This constrains all jobs, including the `setup` reusable workflow call.
- If you know some job genuinely needs more (e.g., to create releases or comment on PRs), you would override `permissions` within that specific job. Since nothing here indicates write operations via the API, we keep all permissions read-only.

Concretely, in `.github/workflows/checks.yaml`, insert:

```yaml
permissions:
  contents: read
```

between the `on:` block and the `env:` block (after line 13 and before line 14), leaving the rest of the workflow unchanged.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -1,4 +1,6 @@
 name: Checks
+permissions:
+  contents: read
 on:
   workflow_dispatch:
   pull_request:
@@ -11,8 +13,6 @@ on:
   push:
     branches:
       - main
-permissions:
-  contents: read
 env:
   SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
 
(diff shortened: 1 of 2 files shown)


pomfrida (approved):
👍

Issue #4513, opened 2026-02-13
Review Button component CSS: density override calculations


eddiman:
## Summary

During a final review of the components, unexpected calculations were found in the Button component CSS related to density overrides. The current approach to density overrides needs discussion and potentially a rethink.

## Background

From Slack (2026-02-13):
- @vnys spotted unexpected calculations in the Button CSS during component review
- @pomfrida confirmed this is likely about the density overrides, and is unsure how to solve it differently
- Meeting scheduled for Monday 2026-02-17 at 12:00 to discuss

## Questions to resolve

- What are the unexpected calculations in the density overrides?
- Is there an alternative approach that avoids these issues?
- What are the trade-offs of the current implementation?

## Participants

- @vnys (raised the issue, component reviewer)
- @pomfrida (density override implementation context)
- @eddiman (design input)

PR #4514, opened 2026-02-13
refactor: simplify Button CSS by using min-heights

pomfrida:
## Summary
- Remove `<TypographyNext>` wrapper span from Button — children are now direct flex children
- Replace padding-derived sizing with explicit `height` values from Figma
- Eliminate all density overrides outside `@layer` (line-height overrides, asymmetric padding, icon-only padding calculations)
- Move comfortable density rules inside `@layer eds-components`

**Result:** 303 → 179 lines CSS (-41%), 0 `calc()`, 0 custom properties, 0 line-height overrides

## Context
See #4513 for detailed analysis of the density token mismatch and trade-offs.

## Test plan
- [x] All 63 Button tests pass (snapshot updated)
- [x] Storybook: all sizes, variants, icon-only, round — spacious and comfortable
- [x] Measured button dimensions in browser — all match Figma exactly

| | Spacious | Comfortable |
|---|---|---|
| sm | 24px | 20px |
| md | 36px | 24px |
| lg | 44px | 36px |
| icon-sm | 24×24 | 20×20 |
| icon-md | 36×36 | 24×24 |
| icon-lg | 44×44 | 36×36 |

### DIFF
--- a/packages/eds-core-react/src/components/next/Button/button.css
+++ b/packages/eds-core-react/src/components/next/Button/button.css
@@ -23,9 +23,7 @@
   /* ===== SIZE VARIANTS ===== */
   /*
    * Button height is controlled by min-height.
-   * padding-block provides vertical padding in fallback browsers;
-   * in modern browsers (@supports) it's removed so min-height alone
-   * controls height, and the span is centered by flexbox.
+   * The span keeps its natural line-height and is centered by flexbox.
    *
    * Spacious (default): sm=24px, md=36px, lg=44px
    * Comfortable:        sm=20px, md=24px, lg=36px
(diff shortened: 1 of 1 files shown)


packages/eds-core-react/src/components/next/Button/button.css
```diff
@@ -22,46 +22,62 @@
...
   /*
-   * Explicit height from Figma. Flexbox handles vertical centering.
-   * Horizontal padding and gap use design tokens.
+   * Figma uses padding + text-box-trim for content-driven button height.
+   * We follow the same model with progressive enhancement:
    *
-   * TODO: Replace hardcoded heights with --eds-selectable-height token
-   * when available from the token pipeline.
+   * 1. padding-block/inline from selectable-space tokens (all browsers)
+   * 2. min-height as fallback for browsers without cap unit support
```
vnys: No need to support browsers that don’t support the cap unit, it’s been supported in all major browsers for more than two years now.

pomfrida: Thanks! Updated now and ready for review ✅

PR #4553, opened 2026-02-20
fix(eds-core-react): prevent Datepicker day field from eagerly auto-advancing on "3"

### LINKED ISSUE #4477: Datepicker: Auto-formatting of day field prevents typing two-digit dates (e.g. 31)
## Description

When no date is selected in the Datepicker, typing `3` in the day field is immediately auto-formatted to `03`, preventing the user from entering two-digit dates like `30` or `31`.

This may be caused by one or both of the following:

1. **No date selected — eager auto-formatting:** The day field always auto-formats `3` to `03` because no month with more than 31 days exists, so it assumes the input is complete.
2. **Month-based validation too early:** If the Datepicker defaults to the current month (e.g. February) when no date is selected, the day field may limit input based on that month's max days (28/29), blocking `3x` entries entirely. This is also frustrating for users entering a full date (`dd.mm.yyyy`) in sequence — they may intend to change the month after entering the day.

## Steps to reproduce

1. Open a Datepicker with no pre-selected date
2. Click on the day field
3. Type the digit `3`
4. **Expected:** The field shows `3` and waits for the next digit
5. **Actual:** The field auto-formats to `03` and moves focus to the month field

## Expected behavior

The day field should wait for a second digit after `3` (valid values: `30`, `31`) before auto-formatting and moving focus to the month field.

## Clarification needed

Awaiting confirmation from the reporter on whether this happens regardless of month or specifically when February is selected/defaulted. See original report for context.

## Reported by

Reported via email from Datepicker end-users.

pomfrida:
## Summary
- Sets `placeholderValue` to January 1st (31 days) in `DateFieldSegments` when no date is selected
- Prevents React Aria from using the current month's day count as segment max (e.g. February = 28), which caused typing "3" to auto-advance focus since 30 > 28
- Adds tests covering the fix for both `en-US` and `no` locales

Fixes #4477

## Test plan
- [x] Verified in local Storybook: `aria-valuemax="31"` on day segment, focus stays after typing "3", typing "31" works
- [x] Added unit tests for focus behavior and `aria-valuemax`
- [ ] Verify in Storybook preview after deploy

### DIFF
--- a/packages/eds-core-react/src/components/Datepicker/DatePicker.spec.tsx
+++ b/packages/eds-core-react/src/components/Datepicker/DatePicker.spec.tsx
@@ -186,7 +186,6 @@ describe('DatePicker', () => {
       )
       const dayEl = screen.getByText('dd')
       await userEvent.click(dayEl)
-      expect(dayEl).toHaveFocus()
       await userEvent.keyboard('3')
       // Focus should stay on day field — "30" and "31" are still valid
       expect(dayEl).toHaveFocus()
@@ -200,7 +199,6 @@ describe('DatePicker', () => {
       )
       const dayEl = screen.getByText('dd')
       await userEvent.click(dayEl)
-      expect(dayEl).toHaveFocus()
       await userEvent.keyboard('3')
       expect(dayEl).toHaveFocus()
     })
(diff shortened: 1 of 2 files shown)


pomfrida:
@claude review

PR #4569, opened 2026-02-23
fix: use EDS Button for TextField info icon and fix helper message gap

### LINKED ISSUE #4568: Fix - Use EDS Button component for info icon trigger in TextField
## Description

The info icon tooltip trigger in the next TextField component uses a raw \`<button>\` element instead of the EDS \`Button\` component from \`/next\`.

**File:** \`packages/eds-core-react/src/components/next/TextField/TextField.tsx:39\`

\`\`\`tsx
// Current
<button
  type="button"
  className="eds-text-field__info"
  aria-label="More information"
>
  <Icon data={info_circle} size="xs" />
</button>
\`\`\`

Should be replaced with:

\`\`\`tsx
<Button variant="ghost" icon size="small" aria-label="More information">
  <Icon data={info_circle} size="xs" />
</Button>
\`\`\`

## Why
- Consistency with EDS component usage
- Button component handles styling, focus, and interaction states correctly
- Raw button requires manual CSS overrides (padding, border, background) that the EDS Button already handles

## Notes
- The custom \`.eds-text-field__info\` CSS in \`text-field.css\` can be simplified or removed once Button handles the styling

millus:
Fixes #4568

## Summary
- Replace raw `<button>` with EDS `Button` (ghost, icon, round, small) for the `labelInfo` tooltip trigger in `TextField`
- Simplify `text-field.css` — remove manual button overrides now handled by the `Button` component
- Fix `Field` helper message gap to 4px using density-aware tokens — applied to the preceding sibling via `:has()` to avoid `[data-font-family]` margin override
- Reduce `Input` focus ring gap from 2px to 1px to match Figma

## Test plan
- [ ] Verify info icon button renders correctly with hover/focus states in Storybook (light + dark mode)
- [ ] Verify tooltip appears on hover of info icon
- [ ] Verify TextField total height matches Figma (~84px with label + input + helper message)
- [ ] Verify helper message spacing is 4px from input
- [ ] Verify Input focus ring appears with 1px gap
- [ ] Verify disabled state styling is correct

### DIFF
--- a/packages/eds-core-react/src/components/next/Field/field.css
+++ b/packages/eds-core-react/src/components/next/Field/field.css
@@ -56,9 +56,9 @@
     color: var(--_eds-field-helper-color);
   }
 
-  /* Reduce gap between input and helper message to 3xs spacing.
-     Applied to input-container (not helper message) to avoid [data-font-family] margin: 0 override. */
-  .eds-field:has(.eds-field__helper-message) .eds-input-container {
+  /* Reduce gap before helper message to 3xs spacing.
+     Applied to the preceding sibling (not helper message) to avoid [data-font-family] margin: 0 override. */
+  .eds-field > :has(+ .eds-field__helper-message) {
     margin-block-end: calc(var(--eds-spacing-vertical-3xs) - var(--eds-generic-gap-vertical));
   }
 


packages/eds-core-react/src/components/next/Field/field.css
```diff
@@ -56,6 +56,12 @@
     color: var(--_eds-field-helper-color);
   }
 
+  /* Reduce gap between input and helper message to 3xs spacing.
+     Applied to input-container (not helper message) to avoid [data-font-family] margin: 0 override. */
+  .eds-field:has(.eds-field__helper-message) .eds-input-container {
+    margin-block-end: calc(var(--eds-spacing-vertical-3xs) - var(--eds-generic-gap-vertical));
+  }
+
```
pomfrida, line 64: Since this is TextField/Input specific, I think this makes more sense inside text-field.css

millus: I see what you mean, but I think the spacing needs to be on Field because thats where HelperMessage lives. But if HelperMessage is only relevant for Input, then I agree. But if its going to be used by other form components as well, then spacing should also be the same, and not input/textfield specific 👍

millus: I understand now, and agree! This is very specific to input-container, so I made the css more generic so it could be any component before helpermessage not just input, and kept it in field.css 👍

PR #4580, opened 2026-02-25
chore(config): auto-fix lint on edit via Claude Code PostToolUse hook

millus:
## Summary

Adds a `PostToolUse` Claude Code hook that automatically runs `eslint --fix` and `stylelint --fix` after every file edit, preventing lint and formatting failures from reaching CI.

- **`eslint --fix`** on `.ts`/`.tsx` files — covers both Prettier formatting rules and fixable ESLint rules (e.g. `react/jsx-no-target-blank`)
- **`stylelint --fix`** on `.css` files inside `next/` only — applies to EDS 2.0 vanilla CSS components

## Why project-level (`settings.json`) and not opt-in?

The hook is purely defensive — it fixes formatting silently and never blocks work. Having it on by default means all Claude Code users on this project benefit without needing to discover it. The performance cost is small (~1-3s per edited file).

**If you find it too slow or prefer to opt out**, add this to your personal `.claude/settings.local.json` (already gitignored):
```json
{
  "hooks": {
    "PostToolUse": []
  }
}
```

**Open question for the team:** Should this be opt-in via `settings.local.json` instead? Happy to flip it if that's preferred.

## Test plan

- [ ] Make a small formatting mistake in a `.tsx` file in a Claude Code session and verify it gets auto-corrected after the edit
- [ ] Or test manually: `echo '{"tool_name":"Edit","tool_input":{"file_path":"/absolute/path/to/file.tsx"}}' | CLAUDE_PROJECT_DIR=$(pwd) node .claude/hooks/format_hook.js`

### DIFF
--- a/.claude/hooks/format_hook.js
+++ b/.claude/hooks/format_hook.js
@@ -1,4 +1,4 @@
-const { execSync } = require('child_process')
+const { spawnSync } = require('child_process')
 const path = require('path')
 
 async function main() {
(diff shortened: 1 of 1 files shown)


pomfrida (dismissed):
I think this is such a great addition, that we really need! Great 👏 

Regarding no. 1 in Claude's review - The only scenario where this could actually be a problem is if someone adds a file with special characters in the filename and Claude edits it. Unlikely, but not impossible, so it's worth considering

PR #4601, opened 2026-03-02
feat: add Link component to next

pomfrida:
## Summary
- Add new Link component to EDS 2.0 (`next`)
- Supports `inline` (default) and `standalone` variants
- Standalone variant uses composition for icons (pass `<Icon>` as children)
- Includes Figma Code Connect mapping, tests (13 passing, incl. jest-axe), and Storybook stories

## Test plan
- [x] All 13 unit tests pass
- [x] Accessibility tests pass (jest-axe)
- [x] Lint passes
- [x] Visual review in Storybook (light + dark mode)

### DIFF
--- a/packages/eds-core-react/src/components/next/Link/link.css
+++ b/packages/eds-core-react/src/components/next/Link/link.css
@@ -3,7 +3,7 @@
     color: var(--eds-color-text-link);
     text-decoration: underline;
     text-decoration-thickness: var(--eds-sizing-stroke-thin);
-    text-underline-offset: var(--eds-sizing-stroke-thick);
+    text-underline-offset: var(--eds-spacing-vertical-4xs);
 
     transition: color 150ms ease-in-out;
   }
@@ -16,7 +16,7 @@
     border-radius: var(--eds-spacing-border-radius-rounded);
     color: var(--eds-color-text-strong);
     outline: var(--eds-sizing-stroke-thin) solid var(--eds-color-border-focus);
-    outline-offset: calc(var(--eds-sizing-stroke-thick) * 2);
+    outline-offset: var(--eds-spacing-vertical-3xs);
   }
 
   /* Standalone variant: flex layout for icon + text */
@@ -43,7 +43,7 @@
     content: '';
 
     position: absolute;
-    inset: calc(var(--eds-sizing-stroke-thick) * -2);
+    inset: calc(var(--eds-spacing-vertical-3xs) * -1);
 
     border: var(--eds-sizing-stroke-thin) solid var(--eds-color-border-focus);
     border-radius: var(--eds-spacing-border-radius-rounded);


packages/eds-core-react/.storybook/preview.css
```diff
@@ -9,6 +9,7 @@ legend {
 /* Storybook color scheme wrapper */
 .eds-storybook-wrapper {
   background-color: var(--eds-color-bg-neutral-surface);
+  color: var(--eds-color-text-strong);
```
pomfrida, line 12: Added this to support dark mode in stories (for text)

packages/eds-core-react/src/components/next/Link/link.css
```diff
@@ -0,0 +1,67 @@
+@layer eds-components {
+  .eds-link {
+    cursor: pointer;
+
+    color: var(--eds-color-text-link);
+    text-decoration: underline;
+    text-decoration-thickness: var(--eds-sizing-stroke-thin);
+    text-underline-offset: 3px;
```
millus: I think this is 2px in figma, and then maybe we could use a token similar to selection controls, when we use the thickness of the line to offset the line. So here it could be:

`text-underline-offset: var(--eds-sizing-stroke-thick);`

Image

packages/eds-core-react/src/components/next/Link/link.css
```diff
@@ -0,0 +1,67 @@
...
+    border-radius: 0;
+    outline: none;
+  }
+
+  .eds-link[data-variant='standalone']:focus-visible::after {
+    pointer-events: none;
+    content: '';
+
+    position: absolute;
+    inset: -4px;
```
millus: Could we use tokens instead of 4px here? so its similar/familiar to how we solve inset/offset in other next components. e.g.:

```
                             
  /* Matches outline-offset to align focus   
  ring with inline variant */
  inset: calc(var(--eds-sizing-stroke-thick)
  * -2);
```

packages/eds-core-react/src/components/next/Link/link.css
```diff
@@ -0,0 +1,67 @@
...
+
+  .eds-link:hover {
+    color: var(--eds-color-text-strong);
+  }
+
+  .eds-link:focus-visible {
+    border-radius: var(--eds-spacing-border-radius-rounded);
+    color: var(--eds-color-text-strong);
+    outline: var(--eds-sizing-stroke-thin) solid var(--eds-color-border-focus);
+    outline-offset: 4px;
```
millus: Could this be an alternative to the hardcoded 4px? So the value is more connected to what its trying to solve, since 2px is what we normally use for offset, in this case with link that already has the underline offset 2px, we need the focus ring to double that. If I understand these values correctly :)

```
/* Clears the text-underline-offset so     
  focus ring sits outside the underline */
  outline-offset:                            
  calc(var(--eds-sizing-stroke-thick) * 2);  
                
```

packages/eds-core-react/src/components/next/Link/link.css
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  /* Standalone variant: flex layout for icon + text */
+  .eds-link[data-variant='standalone'] {
+    position: relative;
+
+    display: inline-flex;
```
millus: This one is also added in the bottom part "outside layer". So maybe just necessary to have it one of the places 👍

packages/eds-core-react/src/components/next/Link/Link.types.ts
```diff
@@ -0,0 +1,11 @@
+import type { AnchorHTMLAttributes } from 'react'
+
+export type LinkVariant = 'inline' | 'standalone'
+
+export type LinkProps = {
+  /** Visual variant
+   * - `inline` (default): used within text, inherits surrounding font size
+   * - `standalone`: used on its own with flex layout, compose icons as children
+   */
```
millus, line 11: Should we consider this?

AnchorHTMLAttributes makes href optional.  
  
An `<a>` without `href` is not a link per HTML 
  spec — it won't be keyboard focusable and  
  won't have pointer cursor. The component
  should either make `href` required in the
  types, or at minimum document this. 

 ```
 export type LinkProps = {
    href: string  // required
    variant?: LinkVariant
  } & Omit<AnchorHTMLAttributes<HTMLAnchorEle
  ment>, 'href'>
```

File: packages/eds-core-react/src/components/next/Link/link.css [hunk 7]
```diff
@@ -0,0 +1,67 @@
...
+
+    color: var(--eds-color-text-link);
+    text-decoration: underline;
+    text-decoration-thickness: var(--eds-sizing-stroke-thin);
+    text-underline-offset: 3px;
+
+    transition: color 150ms ease-in-out;
+  }
+
+  .eds-link:hover {
```
millus, line 11: Should it be a .eds-link:visited styling as well? I cant see it in figma, but its very helpful to see what links youve already visited.

packages/eds-core-react/src/components/next/Link/Link.tsx
```diff
@@ -0,0 +1,24 @@
...
+  ref,
+) {
+  const classes = ['eds-link', className].filter(Boolean).join(' ')
+
+  return (
+    <a
+      ref={ref}
+      className={classes}
+      data-variant={variant}
+      data-font-family="ui"
```
millus: I like what you did for the data-font-size, maybe it could also be beneficial for data-font-family? Then we also can simplify the overrides in the css, for the display block issue, and the inline link will just inherit the font that is used in the context of where the link is used.

```
data-font-family={variant === 'standalone'
  ? 'ui' : undefined}
```

packages/eds-core-react/src/components/next/Link/link.css
```diff
@@ -0,0 +1,67 @@
...
+
+  .eds-link[data-variant='standalone'] .eds-icon {
+    color: var(--eds-color-text-strong);
+  }
+}
+
+/* Outside @layer to override typography.css [data-font-family] { display: block } */
+.eds-link {
+  display: inline;
+}
```
millus: Can remove this one part if we make data-font-family conditional, so it will not be effective for inline. If not, just ignore this comment and keep as is.

packages/eds-core-react/src/components/next/Link/Link.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+        gap: '16px',
+      }}
+    >
+      <Link href="#" variant="standalone">
+        <Icon data={arrow_forward} />
+        Link with start icon
+      </Link>
+      <Link href="#" variant="standalone">
+        Link with end icon
+        <Icon data={external_link} />
```
millus, line 69: Small thing — no guidance in the stories   
  for target="_blank" links. Since the 
  standalone + icon composition is a natural 
  fit for external links, it could be worth 
  adding a story showing the recommended     
  pattern (e.g. with rel="noreferrer" and an 
  external link icon).

packages/eds-core-react/src/components/next/Link/link.css
```diff
@@ -0,0 +1,67 @@
+@layer eds-components {
+  .eds-link {
+    cursor: pointer;
```
millus: Can be removed, I think `<a href>` gets cursor pointer automatically from the browser.

packages/eds-core-react/src/components/next/Link/Link.tsx
```diff
@@ -0,0 +1,24 @@
...
+import type { LinkProps } from './Link.types'
+
+export const Link = forwardRef<HTMLAnchorElement, LinkProps>(function Link(
+  { variant = 'inline', className, children, ...rest },
+  ref,
+) {
+  const classes = ['eds-link', className].filter(Boolean).join(' ')
+
+  return (
+    <a
```
millus: Total height for the component in figma says 15px, but are 19px in the browser. Could be a figma thing, just worth mentioning since it was something we had a focus on for the other components..

packages/eds-core-react/src/components/next/Link/link.css
```diff
@@ -0,0 +1,56 @@
+@layer eds-components {
+  .eds-link {
+    color: var(--eds-color-text-link);
+    text-decoration: underline;
+    text-decoration-thickness: var(--eds-sizing-stroke-thin);
+    text-underline-offset: var(--eds-sizing-stroke-thick);
```
vnys: Why is `--eds-sizing-stroke-thick` used for spacing? Shouldn’t this be `--eds-spacing-vertical-4xs`?

packages/eds-core-react/src/components/next/Link/link.css
```diff
@@ -0,0 +1,56 @@
...
+
+  .eds-link:hover {
+    color: var(--eds-color-text-strong);
+  }
+
+  .eds-link:focus-visible {
+    border-radius: var(--eds-spacing-border-radius-rounded);
+    color: var(--eds-color-text-strong);
+    outline: var(--eds-sizing-stroke-thin) solid var(--eds-color-border-focus);
+    outline-offset: calc(var(--eds-sizing-stroke-thick) * 2);
```
vnys: Shouldn’t this be `--eds-spacing-vertical-3xs`?

packages/eds-core-react/src/components/next/Link/link.css
```diff
@@ -0,0 +1,56 @@
...
+    border-radius: 0;
+    outline: none;
+  }
+
+  .eds-link[data-variant='standalone']:focus-visible::after {
+    pointer-events: none;
+    content: '';
+
+    position: absolute;
+    inset: calc(var(--eds-sizing-stroke-thick) * -2);
```
vnys: `--eds-spacing-vertical-3xs * -1`?

packages/eds-core-react/src/components/next/Link/Link.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+      Standalone link
+    </Link>
+  ),
+}
+
+export const WithIcon: Story = {
+  render: () => (
+    <Link href="#" variant="standalone">
+      Standalone link
+      <Icon data={external_link} />
```
millus: Nitpick: I think it would be more clear to have a normal arrow or some other icon here to separate the story from the external link story, so its more visually different use cases

pomfrida: Good point, but since we only have the external_link icon in Figma I think maybe this story is obsolete, so I'll just remove it (it's already covered in other stories)

PR #4612, opened 2026-03-03
fix: update Icon to follow next/ conventions

### LINKED ISSUE #4611: Update Icon to follow next/ conventions
The Icon component in `next/` has two inconsistencies compared to all other next components:

### 1. CSS class missing `eds-` prefix

**File**: `packages/eds-core-react/src/components/next/Icon/Icon.tsx:44`
```tsx
const classes = ['icon', className].filter(Boolean).join(' ')
```

All other next components use the `eds-` prefix (`eds-button`, `eds-link`, `eds-switch`, etc.). Icon should use `eds-icon`.

This also affects:
- `icon.css` selectors (`.icon` → `.eds-icon`)
- Any component targeting the Icon class (e.g. Link uses `.icon` in `link.css:55`)
- Tests referencing the class

### 2. CSS imported directly in component file

**File**: `packages/eds-core-react/src/components/next/Icon/Icon.tsx:3`
```tsx
import './icon.css'
```

Other next components rely on `index.css` for CSS imports. Icon (and TextField) import CSS directly in the component file. This should be removed since `index.css` already imports `./Icon/icon.css`.

pomfrida:
## Summary
- Add `eds-` prefix to Icon CSS class (`.icon` → `.eds-icon`), consistent with all other next/ components
- Remove direct CSS import in `Icon.tsx` — CSS is already imported via `next/index.css`
- Update Button CSS references from `.icon` to `.eds-icon`

Closes #4611

### DIFF
--- a/packages/eds-core-react/src/components/next/Checkbox/__snapshots__/Checkbox.test.tsx.snap
+++ b/packages/eds-core-react/src/components/next/Checkbox/__snapshots__/Checkbox.test.tsx.snap
@@ -20,7 +20,7 @@ exports[`Checkbox (next) matches snapshot 1`] = `
     >
       <svg
         aria-hidden="true"
-        class="icon eds-checkbox__icon eds-checkbox__icon--checked"
+        class="eds-icon eds-checkbox__icon eds-checkbox__icon--checked"
         data-icon-size="lg"
         data-testid="eds-icon"
         fill="currentColor"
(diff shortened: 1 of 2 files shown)


packages/eds-core-react/src/components/next/Button/button.css
```diff
@@ -1,18 +1,22 @@
 @layer eds-components {
```
pomfrida, line 1: Alphabetical order by stylelint

PR #4620, opened 2026-03-05
fix(eds-core-react): stabilise DatePicker popover size when navigating months

### LINKED ISSUE #4376: DatePicker month-transition buttons move around
#### General objective

When clicking the back/forward month transition buttons and the datepicker is rendered _above_ the input field, the back/forward month transition buttons will move around when moving through the months.

The _height_ of the popover will change when the amount of weeks in a month change. 

The _width_ of the popover will change when the _month name_ is extra long (September, December).

Image

Image

Image

It would be preferable to have the button _not_ move when quickly switching months using the back/forward month buttons.

Owner:
@AndreasPresthammer (Webstep) Farfetched team.

millus:
## Summary

Fixes #4376

The DatePicker popover was shifting/resizing when navigating between months because:
- The calendar height changed depending on how many weeks were in a month (4, 5, or 6)
- The calendar width changed based on the length of the month name (e.g. July vs September)
- The chevron icon in the title button shifted horizontally with the text

## Changes

- **`CalendarGrid`**: Always render 6 rows (the maximum any month can have). Extra rows show `visibility: hidden` dates from the adjacent month, keeping height consistent.
- **`CalendarHeader`**: Replaced the plain `Button` with a `TitleButton` styled component that has a `min-width: 13.1rem`, ensuring all month names (including the longest, September) fit without the calendar resizing. Longer month names in other locales can still grow beyond the minimum without breaking the layout.
- **`CalendarHeader`**: Override the button's inner `span` layout with `display: flex; justify-content: space-between` so the chevron is always pinned to the right edge regardless of text length.
- Reduced the gap between the title button and the Today button.

## Test plan

- [ ] Open DatePicker and navigate through months — verify height stays consistent (especially between 4-week months like February and 5/6-week months)
- [ ] Verify the calendar width stays consistent when navigating between short (May, July) and long (September, December) month names
- [ ] Verify the chevron icon always appears at the same rightmost position
- [ ] Test with `nb-NO` locale — month names should fit without issues
- [ ] Test in both comfortable and compact density
- [ ] Test DateRangePicker as well (uses the same components)

### DIFF
--- a/packages/eds-core-react/src/components/Datepicker/calendars/CalendarGrid.tsx
+++ b/packages/eds-core-react/src/components/Datepicker/calendars/CalendarGrid.tsx
@@ -1,7 +1,6 @@
 /* eslint-disable react/no-array-index-key */
 import { CalendarState, RangeCalendarState } from '@react-stately/calendar'
-import { AriaCalendarGridProps, useCalendarGrid, useLocale } from 'react-aria'
-import { getWeeksInMonth } from '@internationalized/date'
+import { AriaCalendarGridProps, useCalendarGrid } from 'react-aria'
 import { CalendarCell } from './CalendarCell'
 import { YearGrid } from './YearGrid'
 import { Dispatch, SetStateAction } from 'react'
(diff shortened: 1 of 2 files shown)


pomfrida (approved):
great 🚀

PR #4622, opened 2026-03-05
fix: DatePicker displays correct day for configured timezone

### LINKED ISSUE #4390: DatePicker displays incorrect day for given timezone
**Describe the bug**

DatePicker displays the day using the browser rather than the provided timezone.

For cases when there's enough gap between browser timezone and the one we use in our application to have different days, the component displays the system day, but only in display mode. When you click on a day to enable inline edit mode, the correct day is displayed.

1. Set `Europe/London` timezone for the DatePicker component
2. Set `America/New_York` as a browser's timezone
3. Select any day (e.g. `2025-01-22`) using DatePicker's calendar

**Actual behavior**

Day for the browser's timezone is displayed as an input value:
`2025-01-21`

**Expected behavior**

Selected day should be displayed as an input value:
`2025-01-22`

**Specifications**

- Version: 2.3.0
- Browser: Chrome 144.0.7559.97
- OS: OS X 10.15.7

**Additional context**

Reproduced it in storybook and highlighted important points on the screenshot.
Note, that in `aria-*` attributes the correct day is set.

Image

millus:
## Summary

- Fixes `DateSegment` using the browser's local timezone when formatting the displayed date instead of the configured `timezone` prop
- One-line fix: pass `timeZone: timezone` to `useDateFormatter` in `DateSegment.tsx`
- Affects both `DatePicker` and `DateRangePicker` (both use `DateSegment`)

**Root cause:** `useDateFormatter(formatOptions)` created a formatter without a timezone. When calling `formatter.formatToParts(state.value.toDate(timezone))`, the UTC timestamp was correctly computed for the configured timezone, but then displayed in the browser's local timezone — causing a wrong day to appear when there's enough offset between the two.

Closes #4390

## Test plan

- [ ] New regression test: `should display the date in the provided timezone, not the local timezone (bug #4390)` — passes a date that is Jan 22 midnight in `Pacific/Auckland` (UTC+13) but Jan 21 in UTC; verifies the component displays Jan 22
- [ ] All existing DatePicker tests still pass
- [ ] Manual: set `timezone="Europe/London"` with browser in `America/New_York`, select a date — displayed input should match selected day

### DIFF
--- a/packages/eds-core-react/src/components/Datepicker/DatePicker.tsx
+++ b/packages/eds-core-react/src/components/Datepicker/DatePicker.tsx
@@ -60,7 +60,7 @@ export const DatePicker = forwardRef<HTMLDivElement, DatePickerProps>(
     }: DatePickerProps,
     forwardedRef: RefObject<HTMLDivElement | null>,
   ) => {
-    timezone = timezone || defaultTimezone
+    timezone = timezone ?? defaultTimezone
     formatOptions = formatOptions ?? {
       year: 'numeric',
       month: '2-digit',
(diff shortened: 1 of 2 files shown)


pomfrida:
@claude review

PR #4637, opened 2026-03-09
feat: add Search component to next

### LINKED ISSUE #4434: Search - update code
Part of #4414

millus:
Closes #4434

## Summary
- Add new \`Search\` component to EDS 2.0 (\`next\`)
- Composes \`Field\` + \`Input\` with a search icon start adornment and a conditional clear button end adornment
- Handles both controlled and uncontrolled usage via \`onClear\` callback
- Hides native browser search cancel button (WebKit) in favour of custom clear
- \`<search>\` landmark is automatically named via \`aria-labelledby\` when a visible label is provided
- Includes Figma Code Connect, Storybook stories, and 14 passing tests (incl. jest-axe)

## Test plan
- [ ] All 14 unit tests pass
- [ ] Accessibility tests pass (jest-axe)
- [ ] Clear button appears when typing, hidden when disabled/readOnly
- [ ] Controlled usage: \`onClear\` fires and parent can reset \`value\`
- [ ] Uncontrolled usage: clear button clears input directly
- [ ] Visual review in Storybook (light + dark mode)
- [ ] Lint passes

### DIFF
--- a/packages/eds-core-react/src/components/next/Input/input.css
+++ b/packages/eds-core-react/src/components/next/Input/input.css
@@ -110,8 +110,7 @@
 
   /* Input/Textarea element */
   .eds-input {
-    /* Temporary alias until proper placeholder text token is available */
-    --_placeholder-text-color: var(--eds-color-border-strong);
+    --_placeholder-text-color: var(--eds-color-text-subtle);
 
     flex: 1;
     min-inline-size: 0;


packages/eds-core-react/src/components/next/Search/Search.figma.tsx
```diff
@@ -0,0 +1,21 @@
+import figma from '@figma/code-connect'
+import { Search } from '.'
+
+figma.connect(
```
Chibuzor-Nwemambu, line 4: Perhaps I'm not fully aware, but why do we need `figma.connect` in the code? I mean this entire file, why do we need it?

packages/eds-core-react/src/components/next/Search/Search.tsx
```diff
@@ -0,0 +1,141 @@
...
+          onChange={handleChange}
+          aria-describedby={getDescribedBy({
+            hasDescription: !!description,
+            hasHelperMessage: !!helperMessage,
+          })}
+          hideErrorIcon
+          startAdornment={
+            <Icon
+              data={searchIcon}
+              className="search-icon"
```
Chibuzor-Nwemambu, line 107: It seems this doesn't follow BEM. I think this is meant to be `eds-search__icon`? Following the BEM convention used in the rest of the project

packages/eds-core-react/src/components/next/Search/Search.tsx
```diff
@@ -0,0 +1,141 @@
...
+import './search.css'
+
+export const Search = forwardRef<HTMLInputElement, SearchProps>(function Search(
+  {
+    label,
+    description,
+    helperMessage,
+    id: providedId,
+    invalid,
+    disabled,
```
Chibuzor-Nwemambu, line 17: Should we default to false for invalid and disabled? similar to TextField?

packages/eds-core-react/src/components/next/Search/Search.figma.tsx
millus: We (should) have it in all the next components to connect with figma, to get the correct props visible in figma. Its not visible for Search yet in figma, until we merge.

A little more explanation from Claude:
In figma there is a "Code" tab  
  on the right panel — instead of generic CSS/properties, it shows 
  the actual React code snippet from the .figma.tsx file.          
                                                                   
  So for Button for example, clicking it in Dev Mode would show:   
  <Button variant="primary" size="default" tone="accent">
    Label                                                          
  </Button>                                                        
  ...with the real props filled in based on which variant you're
  inspecting.                                                      
                                                                   
  Without Code Connect, Figma just shows raw CSS and layer info —
  not useful for developers. With it, a developer inspecting a     
  design hands-off can copy-paste working React code.

packages/eds-core-react/src/components/next/Search/Search.tsx
millus: Good catch! But, per this discussion here: https://github.com/equinor/design-system/discussions/4639. We are trying out simpler class names over BEM way of writing it based on CUBE CSS https://cube.fyi/block.html.

packages/eds-core-react/src/components/next/Search/Search.tsx
millus: I think I did wrong in TextField, that its really not necessary to explicitly say false here, because the default is false. So its cleaner to just have it as is. And revisit and fix TextField (and others) later for consistency.

PR #4660, opened 2026-03-19
feat(eds-core-react): create button label typography css

### LINKED ISSUE #4659: Button label – typography proof of concept
Part of epic: [Typography system – CSS-first, no typography component](https://github.com/equinor/design-system/issues/4655#4655)
 
## Goal
 
Validate the CSS-first typography approach with hard-coded values in the Button component before investing in the token pipeline. This issue proves – or disproves – that `text-box` trimming, grid-snapping, `data-multiline`, and the two line-height scales work together correctly.
 
If something needs rethinking, it's cheap to do here.
 
> [!NOTE]
> Values are intentionally hard-coded in this issue. Extracting them to `typography.css` and Token Studio is tracked separately in Phase 2.
 
## Background
 
Currently Button wraps its label in `TypographyNext`:
 
```jsx
<TypographyNext
  as="span"
  family="ui"
  size={typographySize}
  lineHeight="squished"
  baseline="center"
>
  {children}
</TypographyNext>
```
 
The goal is to replace this with a plain `<span class="label">` whose typographic styles are owned entirely by the Button's own CSS, nested under `.eds-button`.
 
## Acceptance criteria
 
- [x] Button label is a `<span class="label">` – no `TypographyNext` wrapper
- [x] Single-line button: text is vertically centred within its grid-snapped bounding box
- [x] `data-multiline` button: text is baseline-aligned
- [x] Computed button height is a multiple of 4px at all sizes (sm, md, lg)
- [x] Two-line `data-multiline` button height is still a multiple of 4px
- [ ] `sm` size is always baseline-aligned (centred mode is not achievable at this size – see notes)
- [x] Styles work under both spacious and comfortable density (test with hard-coded values for each)
- [x] No regression on existing button variants (primary, secondary, ghost, danger)
 
### Class naming
 
Following Camilla's suggestion, sub-elements use simple scoped class names nested under the root component class – not BEM:
 
```css
.eds-button {
  & .label {
    /* typographic styles */
  }
}
```
 
```html
<button class="eds-button">
  <span class="label">Save changes</span>
</button>
```
 
### Space toggle pattern
 
The alignment mode is controlled via a space toggle – a bridging pattern until `if()` has broad Safari support. Document it in a CSS comment so future contributors don't remove it:
 
```css
.eds-button {
  & .label {
    /* Space toggle: --_align-centered / --_align-baseline
     * Controls text-box trimming mode (centred vs baseline).
     * Replace with CSS if() once broadly supported in Safari.
     * https://developer.mozilla.org/en-US/docs/Web/CSS/if */
    --_align-centered: initial; /* centred on by default */
    --_align-baseline: ;        /* baseline off by default */
 
    text-box: trim-both var(--_align-centered, ex)var(--_align-baseline, cap) alphabetic;
    line-height: /* hard-coded squished value */;
    font-size: /* hard-coded sm/md/lg value */;
  }
}
 
.eds-button[data-multiline] {
  & .label {
    --_align-centered: ;        /* centred off */
    --_align-baseline: initial; /* baseline on */
  }
}
```
 
### Grid-snapping
 
The cap-height must be snapped to the 4px grid for the height equation to hold:
 
```css
.eds-button {
  & .label {
    --_round-to-4: calc(round(1cap, 4px) - 1cap);
    padding-block: calc(var(--_round-to-4) + /* spacing token */);
  }
}
```
 
### The `sm` size exception
 
At `font-size-sm`, `round(1cap, 4px)` rounds *down*, which would require negative padding – not possible in CSS. `sm` is therefore always baseline-aligned regardless of `data-multiline`. This is not a bug – baseline alignment is visually correct for `sm` text in most contexts, and the `data-multiline` attribute on an `sm` button still works (it just has no additional effect on alignment).
 
### What to verify
 
Check computed height in browser DevTools at each size:
 
| Size | Padding | Expected cap-height | Expected total |
|---|---|---|---|
| sm | 8px | 8px | 24px |
| md | 12px | 12px | 36px |
| lg | 16px | 16px | 48px |
 
_(Actual values TBD – fill in once hard-coded values are settled.)_
 
## Out of scope
 
- Extracting values to `typography.css` – Phase 2
- Token Studio pipeline – Phase 3
- Other component migrations – Phase 2

vnys:
## Summary

This PR is a rewrite of the EDS 2.0 typography system intended to replace the typography component, covering the Button component and a new Foundation layer for element-level styles.

### Button typography

- Replaces the old Typography component with a plain `span.label` — no React component needed for text rendering
- Button height is derived from `padding-block + round(1cap, 4px)` — the cap-height of the label font, snapped to the 4px grid — giving predictable, grid-aligned heights without hardcoded values
- Padding is adjusted to compensate for the space above the cap-height and below the baseline that line-height introduces, keeping the text optically centred
- `data-multiline` prop: switches to baseline trimming (`text-box: trim-both ex alphabetic`) so wrapped text aligns to the baseline grid. This is an anti-pattern — button labels should always be short enough to fit on one line. The prop exists as a last resort only.
- `asChild` prop: renders any child element (e.g. `<a>`, router link) with Button styles applied via the `Slot` utility, merging props onto the child

### Modular type scale (`typography.css`)

Font sizes follow a geometric progression — five equal steps per octave, anchored at `lg` (= `--_base`). Changing `--_base` via `data-density` is the only thing needed to shift the entire scale:

| Density | Base |
|---|---|
| Relaxed | 1.15625rem (18.5px) |
| Spacious (default) | 1rem (16px) |
| Comfortable | 0.875rem (14px) |

All values are in rem and rounded to the 4px grid (0.25rem). Line-heights use an ease-out cubic curve that compresses leading as font size grows. Two scales are available: `default` (comfortable leading for prose) and `squished` (tight leading for UI controls like buttons and inputs).

### Foundation element styles (`elements.css`)

Drop-in baseline styles for semantic HTML elements, opt-in via the `.eds-elements` class. Add it to `<body>` for a full-page reset, or scope it to any container.

**Optical font-weight compensation**

Each heading level uses its own size-matched font-weight token. Emphasis elements (`strong`) and utility classes (`.eds-heading-bold`, `.eds-heading-light`) inherit the correct weight for their context automatically via pseudo-private variables — no targeting by heading level needed.

**Lobotomised owl flow spacing**

Adjacent block elements get spacing via the lobotomised owl selector, rather than per-element margins. Each element overrides `--_flow-space` to control the gap before itself.

**Baseline grid alignment**

Block text elements use `text-box: trim-both ex alphabetic` to strip whitespace above cap-height and below baseline. Where inset padding is also needed, the baseline offset and padding are combined via `calc()`.

### Typefaces

| Context | Typeface |
|---|---|
| UI labels and body copy | Inter (variable) |
| Headings | Equinor (brand typeface) |
| Code and monospace | Commit Mono |

### ADRs added

- `documentation/adr/0004-spacing-approach-for-eds-2.md`
- `documentation/adr/0005-typography-approach-for-eds-2.md`

## Test plan

- [x] No regression on existing button variants (primary, secondary, ghost, danger)
- [x] Single-line button: text is vertically centred within its grid-snapped bounding box
- [x] `data-multiline` button: text is baseline-aligned
- [x] Computed button height is a multiple of 4px at all sizes (sm, md)
- [x] Button styles work under relaxed, spacious, and comfortable density
- [x] `asChild` renders child element with Button styles merged correctly
- [x] Elements: `.eds-elements` scoping does not bleed into Storybook docs UI
- [x] Elements: headings, body text, tables, lists, and form elements render correctly
- [x] Elements: `strong` inside headings resolves the correct optical weight
- [x] Elements: flow spacing between adjacent block elements is consistent
- [x] Elements: baseline alignment holds across all heading and body sizes

Closes #4659


pomfrida:
Heads up: #4810 adds `asChild` to Button (next). When this PR lands and rewrites `button.css` to pure CSS typography, make sure asChild elements still get the correct font-family and font-size — since they bypass the `<TypographyNext>` label wrapper entirely.

We patched this in #4810 by setting `font-family` and `font-size` directly on `.eds-button` via tokens, which should be naturally absorbed by the `--_` variable approach here.

millus:
@claude please review

millus:
Thanks for the detailed review. A lot of the design decisions are explained in @vnys's notes: https://gist.github.com/vnys/a0f3f809a8f231c1bb590995ad9a9b4f

Quick summary of what's been addressed:

**Fixed in this PR:**
- Icon margin magic number → `--captured-size-2` with `inherits: true` + correct formula
- `eds-button__label` → `label` (CUBE CSS naming)
- `@function` cleanup in `typography.css`
- Input background token → `--eds-color-bg-input`
- `asChild` font-family/font-size set directly on `.eds-button` (Pomfrida's #4810)
- `multiline` prop now works with `asChild` (padding applies; baseline trimming is documented as not applying since there's no `.label` span)
- `--padding-top-baseline` and related variables renamed with `--eds-` prefix across all usages
- Debug grid `background-image` fix (undefined CSS variables invalidated the whole gradient)

**Intentional / deferred (see gist):**
- Density naming TODO — deferred, coordinated rename with Figma
- `relaxed-density.css` POC comment — stays until Token Studio is configured
- TODO stubs in `elements.css` — kept as reminders for a follow-up PR
- `white-space: pre-line` on root — intentional
- `typography.css` duplication — intentional, no `@function` in Safari yet

**Minor / style (not addressed, low priority):**
- `sizeToSelectableSpace` alias, IIFE in render, `text-align` inconsistencies — can be cleaned up in a follow-up

.claude/CLAUDE.md
```diff
@@ -128,6 +128,86 @@ One `eds-`-prefixed root class per component. Internal elements use simple class
...
+
+/* Enhancement — trims whitespace above/below the cap-height */
+@supports (text-box: trim-both ex alphabetic) {
+  padding-top: var(--padding-top-baseline);
+  padding-bottom: 0;
+  text-box: trim-both ex alphabetic;
+}
+```
+
+CSS `@function` (Chrome/Edge 128+) is a future enhancement — define it and comment it in once Safari ships support.
```
pomfrida, line 209: Very nice update to CLAUDE.md 👏

millus:
Follow-up issues created for the deferred TODOs:

- #4825 — Add highlight colour token for `mark` element (blocked on Brand Midnight Sun through colour generator)
- #4826 — Finalise density naming and values (blocked on Token Studio)

packages/eds-core-react/src/components/next/index.css
```diff
@@ -15,7 +15,11 @@
  * If you only load that, UI components will fall back to a generic sans-serif.
  */
 
+@import 'https://cdn.eds.equinor.com/font/eds-uprights-vf.css';
 @import '@equinor/eds-tokens/css/variables';
+@import './Foundation/registered-properties.css';
+@import './Foundation/typography.css';
+@import './Foundation/relaxed-density.css';
```
pomfrida, line 21: `relaxed-density.css` is opt-in by selector (`[data-density='relaxed'] ...`), so consumers who don't set `data-density='relaxed'` are unaffected — good. My remaining worry is the ones who *do* adopt it now: if they build layout around the current hardcoded values, they'll see visual regressions when the real tokens land in `@equinor/eds-tokens`. The 'relaxed' naming is also most likely going to change, if I'm not mistaken?

Since we've also agreed to stopgap with hardcoded values until tokens studio, could we pick one of:

1. Document it as experimental — short note in the PR description + a Storybook caveat that `relaxed` values are POC and will shift when tokens are defined. Simplest.
2. Drop the import from `next/index.css` until tokens exist 

Either is fine, just want to make sure we're signalling "don't build on these exact numbers or variables yet" somewhere visible.

packages/eds-core-react/src/components/next/index.css
```diff
@@ -15,7 +15,11 @@
...
 @import './Icon/icon.css';
 @import './Button/button.css';
 @import './Field/field.css';
```
pomfrida, line 24: The file is opt-in by selector (`[data-density='relaxed'] ...`), so consumers who don't set `data-density='relaxed'` are unaffected — good. My remaining worry is the ones who *do* adopt it now: if they build layout around the current hardcoded values, they'll see visual regressions when the real tokens land in `@equinor/eds-tokens`.

Since we've agreed internally to stopgap with hardcoded values, could we pick one of:

1. Document it as experimental — short note in the PR description + a Storybook caveat that `relaxed` values are POC and will shift when tokens are defined. Simplest.
2. Drop the `@import` from `next/index.css` until tokens exist — consumers who *really* need it can import the file directly and opt in explicitly.

Either is fine, just want to make sure we're signalling "don't build on these exact numbers yet" somewhere visible.

packages/eds-core-react/src/components/next/Button/button.css
```diff
@@ -1,62 +1,99 @@
...
-  .eds-button[data-selectable-space='md'] {
-    gap: var(--eds-typography-gap-horizontal);
-    min-height: 2.25rem; /* 36px */
-    padding-inline: var(--eds-selectable-space-horizontal);
-    font-size: var(--eds-typography-ui-body-md-font-size);
-  }
+    & .label {
+      font-family: var(--_font-family);
+      font-size: var(--_label-font-size);
+      line-height: var(--_label-line-height);
```
pomfrida, line 56: Nit: When `asChild` is used, the child isn't wrapped in `.label`, so this `line-height` never applies. The root `.eds-button` declares `font-family`, `font-size` and `color` so those cascade, but `line-height` falls back to browser default (`normal`) on the child's text. 

Two options:
- Hoist `line-height: var(--_label-line-height)` to the root `.eds-button`, or
- Expand the `multiline` JSDoc to call this out for *all* `asChild` usage, not just multiline.
What do you think is better?

packages/eds-core-react/src/components/next/index.css
```diff
@@ -15,7 +15,11 @@
  * If you only load that, UI components will fall back to a generic sans-serif.
  */
 
+@import 'https://cdn.eds.equinor.com/font/eds-uprights-vf.css';
```
pomfrida: This adds a blocking remote request to every consumer's first paint, with no opt-out. Two concerns:

1. `@import` at the top of a stylesheet blocks rendering until the remote sheet is fetched — no `font-display: swap` safety net.
2. Previously the comment above this block documented font-loading as opt-in; that contract changes here without a mention in the PR description.

Could we keep it opt-in here, or is this addition really necessary?

packages/eds-core-react/src/components/next/Button/button.css
```diff
@@ -1,62 +1,99 @@
...
     font-family: var(--eds-typography-ui-body-font-family);
+    font-size: var(--_label-font-size);
+    color: var(--_color);
+    text-align: start;
     text-decoration: none;
+    white-space: pre-line;
+
+    background-color: var(--_bg-color);
 
+    /* ===== LABEL ===== */
```
pomfrida: _nit:_ The actual label rules start a few lines down at `& .label { ... }`. Could we move the marker to just above the `& .label {` block?

documentation/adr/0005-typography-approach-for-eds-2.md
```diff
@@ -0,0 +1,108 @@
+# Typography approach for EDS 2.0
+
+- **Status:** Proposed
```
pomfrida: Btw remember to change from Proposed to Approved and update the date before merging

pomfrida (changes requested):
👏 Solid work on the typography system — the modular scale, text-box trimming, and cap-height-derived button heights are a big step forward, and the ADRs make the reasoning easy to follow.

Two things I'd like to resolve before merging:

1. **Unconditional CDN font import in `next/index.css`** — this is a behaviour change that affects every consumer (blocking first paint, no opt-out). Previously the font was explicitly opt-in. Could we keep it opt-in, or is there a reason it had to change?

2. **`relaxed-density.css` shipping as POC** — since we've agreed hardcoded values are a stopgap until tokens land, could we signal that somewhere visible (experimental note in docs, or drop the auto-import) so consumers don't build on values that will shift?

Left a few smaller suggestions inline — most are nits, feel free to skip them.

packages/eds-core-react/src/components/next/Button/button.css
millus: added label-line-height to root 👍

packages/eds-core-react/src/components/next/index.css
millus: Documented it in storybook 👍

packages/eds-core-react/src/components/next/index.css
millus: Documented it in storybook 👍

millus:
Addressed all review feedback:

- **CDN font import** — removed the `@import` from `next/index.css`, keeping font loading opt-in as before
- **`relaxed` density** — added a POC caveat to the density control description in both Typography and Elements stories: _"relaxed is a POC — values are hardcoded and will change when real tokens are defined"_
- **`line-height` on `asChild`** — hoisted `line-height: var(--_label-line-height)` to the root `.eds-button` so it cascades to `asChild` elements (RouterLink, `<a>`, etc.)
- **`/* ===== LABEL ===== */` comment** — moved to just above `& .label { }` where it belongs
- **ADR 0004 + 0005** — status updated from Proposed → Accepted, date → 2026-04-23

pomfrida (approved):
Very nice work 👏 
I'll be creating new tasks to track the deprecation of TypographyNext - and to replace it in the various components

PR #4751, opened 2026-03-31
feat: add Tooltip component to next/

### LINKED ISSUE #4439: [Tooltip] - update code
Part of #4419

millus:
## Summary

- Adds `Tooltip` component to `@equinor/eds-core-react/next`
- Uses native [Popover API](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API) (`popover="hint"`) for top-layer rendering — Escape + light-dismiss handled by the browser, does not close other popovers/menus. Safari falls back to `manual` (tooltip still works, no Escape-dismiss).
- Uses [CSS Anchor Positioning](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_anchor_positioning) for zero-JS positioning with automatic viewport-edge flipping via `flip-block`/`flip-inline` + explicit `@position-try` fallbacks
- Arrow is a single `::before` pseudo-element using a 16-point star `clip-path` polygon — all four arrow shapes pre-drawn, `margin: inherit` hides three and reveals the active one
- `box-shadow` on the tooltip rectangle only, always downward regardless of arrow direction
- WCAG 1.4.13 compliant: 100ms hide delay + cancel on mouse-enter-tooltip

## Browser support

CSS Anchor Positioning is [Baseline 2026](https://caniuse.com/css-anchor-positioning) (Chrome 125+, Firefox 135+, Safari 18.2+).

## Test plan

- [x] Hover/focus trigger shows tooltip
- [x] Escape dismisses tooltip (Chrome/Firefox)
- [x] All four placements render correctly with arrow
- [x] Arrow auto-flips when tooltip reaches viewport edge
- [x] `disabled` prop suppresses tooltip
- [x] Mouse can travel from trigger to tooltip without it closing (WCAG 1.4.13)
- [x] Passes axe accessibility tests

Closes #4439


packages/eds-core-react/src/components/next/Tooltip/tooltip.css
```diff
@@ -0,0 +1,127 @@
...
+
+  .eds-tooltip-anchor {
+    display: inline-block;
+    anchor-name: var(--tooltip-anchor-name);
+  }
+}
+
+/* @position-try must be outside @layer per spec. */
+
+@position-try --tooltip-bottom {
```
pomfrida: Nit: `@position-try` names are global per spec (not scoped by `@layer` or any selector). `--tooltip-bottom` etc. could collide with other libraries or future components on the same page. Consider prefixing to `--eds-tooltip-bottom`, `--eds-tooltip-top`, etc. — same convention as the `eds-` prefix on class names.

pomfrida (changes requested):
Great work on this! Tooltip would also need a `.figma.tsx` Code Connect file but that work is for after easter when we are back! 🐥

pomfrida (approved):
Nice work! 🎉

PR #4802, opened 2026-04-13
feat: add Chip component to EDS 2.0

pomfrida:
## Summary

New Chip component for EDS 2.0 (`@equinor/eds-core-react/next`).
Always interactive (`role="button"`); the consumer controls behaviour via props.

### Props
| Prop | Type | Default | Notes |
|------|------|---------|-------|
| `tone` | `'neutral' \| 'accent' \| 'success' \| 'info' \| 'warning' \| 'danger'` | `'neutral'` | |
| `variant` | `'default' \| 'outlined' \| 'high-contrast'` | `'default'` | |
| `selected` | `boolean` | `false` | Leading check on default chips; flips the arrow up on dropdown chips |
| `onDelete` | `(event) => void` | — | Makes the whole chip a delete button; shows trailing X |
| `dropdown` | `boolean` | `false` | Shows trailing arrow; pair with `selected` for the open state |

### Behaviour
- **Selectable (default):** `selected` toggles a leading check and `aria-pressed`
- **Deletable:** `onDelete` turns the whole chip into a remove button (click or Backspace/Delete); `onClick` is ignored while `onDelete` is set
- **Dropdown:** arrow flips ↓/↑ based on `selected`, mirrored by `aria-expanded` — consumer owns the menu
- **Custom:** mix arbitrary leading/trailing children with any of the above

### A11y
- `role="button"` + Enter/Space activation
- `aria-pressed` (toggleable) / `aria-expanded` (dropdown) / `aria-label="Remove"` via the close icon title
- No nested-interactive violations
- Respects `data-density="spacious|comfortable"` — icons scale with typography context

## Test plan
- [x] 57 unit tests (rendering, tones, variants, selected, deletable, dropdown, keyboard, axe)
- [x] Lint passes
- [x] Visual verification in Storybook (spacious/comfortable density, all four types)
- [x] Figma token comparison — all values match
- [ ] Review Figma Code Connect mapping

### DIFF
--- a/packages/eds-core-react/src/components/next/Chip/Chip.types.ts
+++ b/packages/eds-core-react/src/components/next/Chip/Chip.types.ts
@@ -52,9 +52,10 @@ export type ChipProps = {
    */
   selected?: boolean
   /**
-   * Shows a trailing close icon that removes the chip. The callback fires
-   * when the close icon is clicked — click propagation is stopped so the
-   * chip's `onClick` does not also fire. Omit to render a non-deletable chip.
+   * Shows a trailing close icon and turns the entire chip into a delete
+   * button — clicking anywhere on the chip fires this callback, and
+   * Backspace/Delete while focused does the same. When provided, `onClick`
+   * is ignored (a deletable chip has only one action).
    *
    * @example
    * ```tsx
--- a/packages/eds-core-react/src/components/next/Chip/Chip.types.ts
+++ b/packages/eds-core-react/src/components/next/Chip/Chip.types.ts
@@ -1,4 +1,4 @@
-import type { HTMLAttributes, ReactNode } from 'react'
+import type { HTMLAttributes, MouseEvent, ReactNode } from 'react'
 
 /**
  * Color tone for theming
@@ -37,8 +37,8 @@ export type ChipProps = {
    */
   variant?: ChipVariant
   /**
-   * Selected state. Shows a leading check icon when true.
-   * Use with `onClick` to toggle selection.
+   * Selected state. For default chips this shows a leading check icon.
+   * For dropdown chips this flips the trailing arrow up to indicate the menu is open.
    *
    * @default false
    *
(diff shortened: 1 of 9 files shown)


pomfrida:
@claude review

pomfrida:
@claude review

packages/eds-core-react/src/components/next/Chip/Chip.types.ts
```diff
@@ -0,0 +1,84 @@
...
+   * button — clicking anywhere on the chip fires this callback, and
+   * Backspace/Delete while focused does the same. When provided, `onClick`
+   * is ignored (a deletable chip has only one action).
+   *
+   * @example
+   * ```tsx
+   * <Chip onDelete={() => removeFilter(id)}>Active</Chip>
+   * ```
+   */
+  onDelete?: (
```
millus, line 70: The `onDelete` event type uses `HTMLElement` but this chip always renders as a `button`, so the events will always be `MouseEvent<HTMLButtonElement>` / `KeyboardEvent<HTMLButtonElement>`. Narrowing to `HTMLButtonElement` here would be more precise and let consumers type their handlers correctly without casting.
 
```ts
 onDelete?: (
  event: MouseEvent<HTMLButtonElement> | 
  KeyboardEvent<HTMLButtonElement>,
 ) => void
```

millus (dismissed):
Looks really good!! 😍 In figma it was a height difference of 2px between chip with icon and chip without, but in storybook/code I see that they are all aligned! So I think that is best 💯 Just added a comment on types, other than that its all really well put together 👍

packages/eds-core-react/src/components/next/Chip/Chip.types.ts
pomfrida: Hey that's a really good catch actually, I wonder how it passed both me and Figma MCP 😆 But I added a comment in Figma so the designers can align with code. And the input regarding the event types is gold, thank you so much 👏

PR #4805, opened 2026-04-13
feat: Add chipCount prop to limit the amount of chips shown

yusijs:
As discussed on [slack](https://equinor.slack.com/archives/CJT20H1B9/p1776079669764569)

This PR adds a new prop to the Autocomplete-component (chipCount), which allows limiting the amount of chips dynamically, while retaining the existing behaviour if left unspecified.


pomfrida:
@claude review

pomfrida:
Nice work @yusijs please check Claude review - I'll review it tomorrow myself as well

yusijs:
Fixed the comments from Claude @pomfrida. I didn't split the style on the chips into a variable, because I don't think they would necessarily need to be kept in sync (and it's only 2 lines).

pomfrida (approved):
Looks good! The core logic is solid and the focus fallback handles the edge cases. A couple of minor notes:

1. `isLastChip` checks against the full `selectedItems` array, so when `chipCount` is set it'll never be `true` for the last visible chip — meaning keyboard-deleting the last visible chip always falls back to focusing the input instead of the previous chip. Not a bug (focus doesn't get lost), but slightly different UX compared to the no-chipCount case. Something to be aware of if you want to polish it later.

2. The test is inside the "Scroll position and navigation memory" describe block — would read better in a chip-related block, but no big deal.

PR #4806, opened 2026-04-14
fix(eds-tokens-build): strip concept token self-references from static CSS

eddiman:
## Summary

Fixes a latent bug where concept tokens (bg-floating, bg-disabled, border-focus, text-link, etc.) produced self-referencing CSS declarations in the static build output, overwriting the correct light-dark() values from color-scheme.css.

### The bug

The static build includes concept tokens with `outputReferences: true`. Style Dictionary resolves the concept alias `{bg-floating}` to `var(--eds-color-bg-floating)` — the same variable being declared. When lightningcss bundles both color-scheme.css and static/variables.css into a single `:root`, the self-reference overwrites the correct `light-dark()` value.

Introduced in adb74eb9 but latent because the bundled `variables.css` was never rebuilt after that change. Discovered during elevation token work (#4728).

### The fix

Added a filter in the CSS combine step that strips self-referencing declarations (where `--foo: var(--foo)`) before writing static/variables.css. Concept tokens continue to appear in TS/JS/JSON exports — only the CSS output is affected.

See equinor/design-system-internal#258 for full write-up.

## Test plan

- [x] `pnpm run build:variables` completes without errors
- [x] `build/css/variables.css` has `light-dark()` values for concept tokens, not `var()` self-references
- [x] `build/css/color/static/variables.css` no longer contains concept token declarations
- [x] Verify TS/JS/JSON exports still include concept tokens


eddiman:
@claude

pomfrida (approved):
Nice fix 👏

Issue #4369, opened 2026-01-15
Impossible to create a generic wrapper around Autocomplete component


hchamre:
TypeScript errors prevent creating a generic wrapper around the `Autocomplete` component because of conditional types in its props.

### Minimal Reproducible Example

```typescript
type GenericProps<T> = {
    options: T[];
    labelFactory: (option: T) => string;
}

function GenericWrapper<T>(props: GenericProps<T>) {
    return (
        <Autocomplete<T>
            label="<label>"
            options={props.options}
            optionLabel={props.labelFactory}  // ❌ TS2322
        />
    );
}
```

```
error TS2322: Type '{ label: string; options: T[]; optionLabel: (option: T) => string; }' is not assignable to type 'T extends string | number ? { optionLabel?: ((option: T) => string) | undefined; } : T extends object ? { optionLabel: (option: T) => string; } : { optionLabel?: ((option: T) => string) | undefined; }'.
```

### Root Cause

It seems that when `T` is a generic type parameter without constraints, TypeScript cannot narrow down which branch of the conditional type applies to `Autocomplete`'s props. This prevents TypeScript from verifying that the always-passed `optionLabel` prop satisfies all possible branches of the conditional type.

### Real world example

This issue is preventing my team from upgrading to the latest version of `@equinor/eds-core-react`. Here is the actual component that is causing problems: https://github.com/equinor/ti-spine-auth/blob/main/src/Equinor.TI.Spine.Auth.PAP.UI/client/src/components/Common/Form/FormAutocomplete/FormAutocomplete.tsx

millus:
Hi, thanks for the report and for sharing the real-world example — helpful to see the full picture!

This is being addressed in #4808 (new `Autocomplete` in `next/`). The new component uses a simple unconditional type for `getOptionLabel`, so you can extend `AutocompleteProps<T>` directly in your wrapper without re-declaring all the props or needing `as any`.

One thing to note: the new component is single-select only for now, so teams using `multiple` won't be able to fully migrate yet. Multiselect (combobox with chips) is on the roadmap but out of scope for this PR.

PR #4808, opened 2026-04-14
feat: add Autocomplete component to next/

### LINKED ISSUE #3895: Autocomplete - Consider renaming the `onOptionsChange`
The `Autocomplete`-component has prop `onOptionsChange` as callback for change selected items changes. This is confusing, as it sound like the it signals changes to the list of `options` for the dropdown - which is already a prop named `options`.

It's not editing/changing the options list, but signal for selection. Consider naming it `onChange`, `onValueChange`, `onSelectChange` or something similar.

Ref.: https://github.com/equinor/design-system/blob/e9c2106e4c2b7f0fe8807ee4556dd07c4d133cd3/packages/eds-core-react/src/components/Autocomplete/Autocomplete.tsx#L285

### LINKED ISSUE #4209: Autocomplete - the control could be more friendly with screen readers
I'm testing the Autocomplete control using a screen reader (Windows narrator) and noticed that it could be improved using more aria attributes. 
I tested the search text box on gov.uk and narrator spoke this

Search, Combo Edit, collapsed, **When search suggestions are available use up and down arrows to review and enter to select. Touch device users, explore by touch or with swipe gestures., has popup, 
Type in 3 or more characters for search suggestions**
n
e
a
**No search suggestions found**
a
e
n
l
e
a
**Suggestions available
Search, lea, Combo Edit, expanded, has popup, 
5 search suggestions are available.
lea list, send rent and lease details, 1 of 5, selected,** 

Two things which stood out in the gov.uk version was that it told the user how to use the control, secondly, as suggestions got filtered/available the narrator also told how many suggestions were available.

I haven't tested how things are extensively triggered but i believe some of these attributes might be useful

aria-autocomplete="list"
aria-posinset="1"
aria-setsize="5"

another thing i noticed is that a block with the aria-live is used for "5 search suggestions are available" to make the screen reader announce this. I'm not sure if this is something the consumer of the autocomplete control should implement, if so, it should provide a callback so that the filtered results can be sent to that block.

### LINKED ISSUE #4383: Autocomplete - autoWidth sets wrong initial offset in popover
**Describe the bug**

I found that autoWidth sometimes messes up the popover positioning in Autocomplete. The initial scrollTop instead of 0 is right after the first element.

I tried to make a workaround, to do not use autoWidth and enforce the width but it is constantly overwritten to auto.

Image

0. The initial requirement is to have enough options to have a scrollable list in popover + autoWidth enabled.

```
  <Autocomplete
    label="Test"
    autoWidth
    options={["1", "2", "3", "4", "5", "6", "7", "8"]}
  />
```

1. Firstly focus on another field (i've found that it makes the reproduction 100% - right after the init it sometimes works sometimes not)
2. Then click on autocomplete input.
3. The list does not show the first element.

**Expected behavior**

I would like to use autoWidth and always see the first item in the list after opening the popover.

**Specifications**

- Version: 2.0.0
- Browser: Chrome 144.0.7559.60
- OS: OSX

### LINKED ISSUE #4433: [Autocomplete] Implement in code
## Summary

## Acceptance Criteria

### LINKED ISSUE #4671: Autocomplete - Allow users to press "enter" when entering a new option for the "inputs-autocomplete--add-new-option"

Its frustrating for users to add new option in the inputs autocomplete but pressing "enter" removes the written option rather than actually selecting

when there is only one option available have it auto highlighted so that pressing enter will select it

-None

**Additional context**

Image At this instance i would expect that pressing enter would add the new option 
[
Component in question](https://storybook.eds.equinor.com/?path=/story/inputs-autocomplete--add-new-option)

millus:
## Summary

New `Autocomplete` component in `packages/eds-core-react/src/components/next/`, implementing the EDS 2.0 single-select autocomplete design from Figma.

### Accessibility & ARIA
- Full WAI-ARIA combobox pattern: `role="combobox"`, `role="listbox"`, `role="option"`, `aria-expanded`, `aria-controls`, `aria-activedescendant`, `aria-autocomplete`, `aria-posinset`, `aria-setsize`
- Keyboard navigation: Arrow keys, Enter to select, Escape to close
- `aria-live` region announcing result count to screen readers
- Clear button with accessible label

### Core behaviour
- Filters options as the user types
- Controlled and uncontrolled value support
- States: disabled, readOnly, invalid
- Loading state with spinner (`loading` / `loadingText` props)
- `allowCustomValue` — lets users confirm a typed value not in the list

### API design
- **Generic `T`** — `options: T[]`, `value: T`, `onValueChange: (value: T) => void`. Fixes the TypeScript conditional type error from the old component that prevented teams from building generic wrappers (see #4369).
- **`getOptionLabel`** — controls what's shown in the dropdown, what appears in the input after selection, and what's searched as the user types. Required when `T` is not a string.
- **`getOptionValue`** — stable unique key for options when object references may change across renders (replaces `itemToKey`).
- **`optionDisabled`** — disable specific options from being selected.
- **`optionsFilter`** — custom filter function; return `true` for all options to disable local filtering for server-side search.
- **`renderOption`** — custom rendering per row (icons, subtitles, badges). `getOptionLabel` still controls filtering and input display.
- **`onInputChange`** — fired on every keystroke; use for async/server-side search.
- **`onValueChange`** vs **`onCustomValueConfirm`** — list selections and custom values handled by separate callbacks, keeping types unambiguous when `T = string`.

### Infrastructure
- CSS Anchor Positioning + Popover API for the dropdown (top-layer, light-dismiss, flip fallback)
- Dropdown visually connects to the input (shared border, squared corners)
- `Menu` / `MenuItem` components extracted for reuse

### Shared utilities
- Pure option helpers (`resolveOptionLabel`, `resolveOptionKey`, `isOptionDisabled`, `defaultOptionsFilter`) extracted to `next/utils/` so `NativeSelect`, `Combobox`, and future option-based components can reuse them without duplicating logic.

### Stories
- Introduction, Controlled, Invalid, Disabled, ReadOnly, ObjectOptions, CustomOptionRendering, DisabledOptions, AllowCustomValue, Loading, Async, WithReactHookForm, DensityModes
- Organised argTypes with categories, descriptions, and defaults

## Scope

This PR covers **single-select** only, matching the current Figma design. Multiselect (combobox with chips) is tracked separately.

## Intentional non-extraction: event handlers and keyboard navigation

The event handlers and keyboard navigation logic (`handleKeyDown`, `handleListOptionSelect`, popover helpers, etc.) are intentionally left inside the component rather than extracted into a shared hook. All of them close over component state and there is currently no second consumer. When `Combobox` (accessible custom dropdown) is added, this is the natural point to extract a `useComboboxBehavior` hook — doing it now would mean designing the hook API without a second use case to validate the shape.

## Test plan
- [x] Unit tests pass (53 tests)
- [x] Lint passes
- [x] Visual check in Storybook

## Closes

Closes #4433 — Implement Autocomplete in code
Closes #4209 — Screen reader improvements (aria-posinset/setsize, live region)
Closes #4671 — Allow Enter to confirm a new option in allowCustomValue
Closes #3895 — Rename confusing onOptionsChange callback (now onValueChange)
Closes #4383 — autoWidth causes wrong scroll offset on open

## Addresses

Addresses #4369 — The TypeScript generic wrapper error is fixed; teams using `multiple` won't be able to fully migrate until combobox with chips (multiselect) ships


packages/eds-core-react/src/components/next/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,452 @@
...
+    const customValueTyped =
+      allowCustomValue &&
+      isFiltering &&
+      inputValue.trim() !== '' &&
+      !filteredOptions.some(
+        (o) => o.toLowerCase() === inputValue.trim().toLowerCase(),
+      )
+
+    // Add item is always index 0 when allowCustomValue; regular options shift by 1
+    const addOptionIndex = allowCustomValue ? 0 : -1
```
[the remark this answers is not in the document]
millus: Its removed

millus:
@claude review

millus:
All review feedback has been addressed — thanks for the thorough review!

**Fixed:**
1. Clear button: replaced `aria-hidden` + `tabIndex={-1}` with the `inert` attribute — no layout shift, cleaner semantics
2. External `value` prop change now syncs the input text (added effect)
3. Options now handle both `onMouseDown` and `onClick` for AT that synthesises click without mousedown
4. Removed redundant `aria-live="polite"` — `role="status"` already implies it
5. WCAG 2.5.3: add-option visible text now shows `Add: {value}` to match the accessible name
6. Replaced deprecated `clip: rect(0,0,0,0)` with `clip-path: inset(50%)`
7. `HTMLElement.prototype.matches` mock now restored in `afterAll`
8. `aria-labelledby={inputId}` on listbox — works for both string and ReactNode labels
9. CSS Anchor Positioning: no fallback needed — now Baseline 2026 (Chrome 125+, Firefox 147+, Safari 26+)
10. Added comment in `menu.css` explaining the `--menu-anchor` custom property inheritance contract
11. Removed unused `AutocompleteOption` type export
12. `allOptions` and `filteredOptions` are now memoized with `useMemo`
13. Added comment to `Menu.tsx` clarifying it's a generic container (caller supplies `role`)
14. API aligned with industry conventions: `value` / `onValueChange` / `inputValue` pattern (following Base UI, designsystemet)

millus:
@claude review

millus:
@claude review

millus:
@claude review

millus:
@claude review dont look for nitpick now, just if there are blocking bugs or similar important things to fix before human review

millus:
@claude review dont look for nitpick now, just if there are blocking bugs or similar important things to fix before human review

packages/eds-core-react/src/components/next/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,630 @@
...
+
+  useEffect(() => {
+    if (!isOpen) return
+    const count = totalOptions
+    setAnnouncement(
+      count === 0
+        ? 'No results'
+        : `${count} result${count === 1 ? '' : 's'} available`,
+    )
+  }, [totalOptions, isOpen])
```
pomfrida: The live region announces "No results" while loading is true. When the parent passes an empty options array with loading={true} (as in the Loading story), totalOptions = == 0 and screen readers hear "No results" right when the spinner appears. The visible loadingText is in a role="presentation" MenuItem, so it's never announced.

Suggested fix:
```suggestion
 useEffect(() => {
  if (!isOpen) return
  if (loading) {
    setAnnouncement(loadingText)
    return
  }
  const count = totalOptions
  setAnnouncement(
    count === 0
      ? 'No results'
      : `${count} result${count === 1 ? '' : 's'} available`,
  )
}, [totalOptions, isOpen, loading, loadingText])
```

packages/eds-core-react/src/components/next/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -0,0 +1,719 @@
...
+        'Accessible label for the clear button. Override for localisation.',
+      table: { category: 'Selection', defaultValue: { summary: 'Clear' } },
+    },
+
+    // Callbacks
+    onValueChange: {
+      control: false,
+      description:
+        'Fired when the user selects an option from the list. Receives the full option value `T`.',
+      table: { category: 'Callbacks', type: { summary: '(value: T) => void' } },
```
pomfrida: The argTypes summary says (value: T) => void, but the actual type in
Autocomplete.types.ts:20 is (value: T | undefined) => void. The
| undefined is meaningful — consumers will hit it from onClear and
from the user clearing the input. Please align the story doc with  this
```suggestion
      table: { category: 'Callbacks', type: { summary: '(value: T | undefined) => void' } },
```

packages/eds-core-react/src/components/next/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,630 @@
...
+            ?.querySelector(`#${getOptionId(last)}`)
+            ?.scrollIntoView({ block: 'nearest' })
+          return
+        }
+        if (activeIndex === 0) {
+          // At first item — return focus to input
+          setActiveIndex(-1)
+          closeListbox()
+          return
+        }
```
pomfrida: Keyboard nav is otherwise solid, just one edge case where ArrowUp from the first option closes the listbox.

Specific sequence (verified in Storybook):
  1. Click the input — dropdown opens, nothing highlighted
  2. ArrowDown — highlight moves to first option (activeIndex = 0)
  3. ArrowUp — dropdown closes (aria-expanded=false, popover hidden)

_Note: ArrowUp from an empty highlight (step 1 → ArrowUp directly) wraps to the last item, which is correct. The issue is only the ArrowDown → ArrowUp edge case._

This is non-standard for the WAI-ARIA combobox + listbox pattern. The APG expects ArrowUp at the top to either wrap to the last item or stay put. Closing surprises users who pressed ArrowDown once too many and try to back up — they lose the dropdown instead of returning to the input.

Suggested behaviour: at activeIndex === 0, either wrap to the last option or clear activeIndex but keep the listbox open. What do you @millus or @eddiman think is best here? Personally I'm voting for wrapping to last option but I might be wrong about this. We should reserve closing for Escape and click-outside

pomfrida (changes requested):
Sooo good 👏 Visual match against Figma is tight: backgrounds, borders, radii, and the connected input+menu rounded rectangle all land exactly. Error and disabled states match without notes. Nice work on the details. 👌  

I just have a few requested changes if you could fix please 🙏

packages/eds-core-react/src/components/next/Autocomplete/Autocomplete.tsx
millus: I agree @pomfrida, escape and click-outside for closing, and the arrows for navigating the list, I went for the solution that one stop on the top item, not wrap, I tried both and it felt weird that i could wrap on top, but when I went the other way it stopped on the end of the list. So I thought for myself maybe its nice to know where the top and bottom of the list are, so the stop is intentional to know that one is on the top(or bottom) of the list. But not closing the list anymore with arrows 👍

pomfrida (dismissed):
Great work @millus 👏 💯

pomfrida (approved):

PR #4831, opened 2026-04-21
fix: register typography length properties and update docs

### LINKED ISSUE #4602: Add @property registration for --eds-typography-gap-horizontal
## Problem

`--eds-typography-gap-horizontal` is a unitless token (e.g. `6.5`), which means consumers must use `calc(var(--eds-typography-gap-horizontal) * 1px)` to convert it to a usable CSS length value.

Meanwhile, `--eds-typography-font-size` already has a `@property` registration with `syntax: '<length>'` in `typography.css`, so it can be used directly as `font-size: var(--eds-typography-font-size)` without the `* 1px` hack.

This inconsistency is confusing — some next components (Button, Input, TextField) already use `gap: var(--eds-typography-gap-horizontal)` without `* 1px`, which technically doesn't work since the browser treats the value as a string.

## Proposed solution

1. Add `@property` registration for `--eds-typography-gap-horizontal` (and potentially `--eds-typography-icon-size`) in `typography.css`
2. Update the token build pipeline to output values with units (e.g. `6.5px` instead of `6.5`)
3. Remove `* 1px` workarounds from component CSS

## References

- Current `@property` for font-size: `packages/eds-tokens/src/css/typography.css`
- Typography docs: `packages/eds-tokens/instructions/typography.md` (line 69 documents the `* 1px` pattern)
- Components using `gap: var()` without `* 1px`: Button, Input, TextField
- Component using `calc(var() * 1px)`: Link

pomfrida:
## Summary

- Register `@property` for `--eds-typography-gap-horizontal` and `--eds-typography-gap-vertical` with `syntax: '<length>'` — for consistency with the existing `--eds-typography-font-size` registration and correct interpolation during transitions.
- Update `instructions/typography.md` examples that still showed the outdated `calc(var(...) * 1px)` workaround.

## Context

Issue #4602 originally assumed the tokens were emitted as unitless values (e.g. `6.5`) and proposed updating the token pipeline to emit `px` units. Investigation showed that the CSS `var()` chain already resolves to valid length values — the primitives (`--eds-spacing-icon-md-gap-horizontal: 0.531rem`) have units via the `pxToRem` transform, and the typography tokens reference them transitively. No component currently uses `calc(* 1px)` for these tokens either.

The remaining concerns worth fixing:
1. **Typing**: Without `@property`, the browser treats these custom properties as untyped strings, which means animations/transitions on them fall back to discrete interpolation. Registering the gap tokens as `<length>` enables smooth interpolation and matches the pattern set by `--eds-typography-font-size`.
2. **Docs**: The `calc(* 1px)` pattern in `typography.md` is stale and misleading.

## Why `--eds-typography-icon-size` is NOT registered

An earlier revision of this PR also registered `--eds-typography-icon-size` with `initial-value: 24px`. That caused a visual regression on the Button density comparison story: icon-only Buttons in comfortable density rendered without padding.

Root cause: `icon.css` uses `font-size: var(--eds-typography-icon-size, 1.5em)` where the `1.5em` fallback provides proportional sizing when no `[data-font-size]` context is set. A registered `@property` with `inherits: true` always has a valid computed value (the initial-value), so the `var()` fallback never fires. Since `1.5em` is not a valid `initial-value` (must be computationally independent), there is no way to keep both the registration and the proportional fallback — the fallback wins.

## Test plan

- [ ] Run `pnpm run build:variables` and verify `build/css/variables.css` contains `@property` declarations for `--eds-typography-font-size`, `--eds-typography-gap-horizontal`, and `--eds-typography-gap-vertical` only
- [ ] Verify Storybook renders Link, Button, Input, TextField, TextArea with correct `gap` and Icon with correct size
- [ ] Verify icon-only Buttons in the Button > Density Comparison story render with padding in both spacious and comfortable density
- [ ] Verify no visual regressions at different `data-font-size` values

Closes #4602


pomfrida:
@claude review

millus (commented):
LGTM, buut I went trough the test plan, and one of the things was to look at Button, and in the story density comparison, something has happened. I dont know if its directly related, because its weird its just this one thats off. But the icon only button in comfortable density is wrong in this branch and not in the storybook that is live.

Image
VS.

Image

pomfrida:
@claude review

packages/eds-tokens/src/css/typography.css
```diff
@@ -1,21 +1,36 @@
...
- * preventing the mathematical operations needed for responsive icon sizing based on
- * the current text size.
+ * Registered custom properties for length-typed typography values.
+ *
+ * Typing these as <length> makes them safe to use directly in CSS length contexts
+ * (gap, font-size) and enables proper interpolation for transitions.
+ * --eds-typography-font-size in particular is used in tan(atan2(var(--eds-typography-font-size), 1px))
+ * for proportional icon sizing, which requires a registered length.
+ *
+ * Note: --eds-typography-icon-size is intentionally NOT registered. Icon.css uses
```
millus, line 9: Looks good now 👍

millus (approved):
LGTM 🚀

Issue #4836, opened 2026-04-23
refactor: migrate Chip label off TypographyNext


pomfrida:
Part of #4655. Follows the pattern established by Button in #4660.

`Chip` wraps its label in `<TypographyNext>`:

```tsx
// packages/eds-core-react/src/components/next/Chip/Chip.tsx:68
<TypographyNext as="span" className="label" family="ui" size="md" lineHeight="squished" baseline="center">
```

This is the closest analogue to Button — UI label, squished line-height, centred. The Button implementation is the template.

### Migration
- Replace with `<span className="label">{children}</span>`
- Move typography to `chip.css` — set `--_label-*` vars on `.eds-chip` and apply on `& .label`
- Remove `data-font-family`, `data-font-size`, `data-space-proportions` attributes from the button element (replaced by CSS vars)

### Acceptance
- [ ] `Chip.tsx` no longer imports `TypographyNext`
- [ ] Label uses `.label` span + pure CSS
- [ ] Visual parity with current Chip (all tones/variants/states)
- [ ] axe + existing tests still pass

Issue #4835, opened 2026-04-23
refactor: migrate Banner.Message off TypographyNext


pomfrida:
Part of #4655. Follows the pattern established by Button in #4660.

`Banner.Message` currently wraps text in `<TypographyNext>`:

```tsx
// packages/eds-core-react/src/components/next/Banner/Banner.tsx:30
<TypographyNext as="p" family="ui" size="md" baseline="center" lineHeight="default" tracking="normal" ...>
```

### Migration
- Render `<p className="eds-banner__message">` directly, no wrapper
- Apply `font-family`, `font-size`, `line-height` via `--eds-typography-ui-body-md-*` tokens in `banner.css`
- Since banner messages can wrap, apply `text-box: trim-both ex alphabetic` under `@supports` for baseline-grid alignment

### Acceptance
- [ ] `Banner.Message` no longer imports `TypographyNext`
- [ ] Typography handled by `banner.css` using tokens
- [ ] No visual regression (verify in Storybook across tones)
- [ ] axe + existing tests still pass

Issue #4655, opened 2026-03-16
Typography system – CSS-first, no typography component


vnys:
## Overview
 
The current typography system uses a `TypographyNext` wrapper component to handle baseline grid alignment, line-height scale selection, and font size. This approach was designed around Figma constraints and has leaked typographic concerns into component implementations — Button, for example, has to maintain a size mapping and coordinate with `TypographyNext`'s internal vocabulary (`family`, `lineHeight`, `baseline`) rather than owning its own typographic CSS.
 
With this epic we’re pivoting towards moving typography to CSS. Each component owns its typographic styles directly. The typography system's job becomes providing the token scale and base element defaults — not wrapping every text element in a component.
 
> [!IMPORTANT]
> This epic tracks the full migration. Work is sequenced in three phases to validate the approach cheaply before formalising it in the token pipeline.
 
## Background
 
See the [GitHub Discussion](https://github.com/equinor/design-system/discussions/4647) 
 
## The two-axis model
 
Every text element in EDS has two typographic axes:
 
**Axis 1 – Alignment**
- **Single-line (centred)** — default for UI labels (Button, Chip, Tab). Text floats vertically centred within its grid-snapped bounding box.
- **Multiline (baseline-aligned)** — default for prose and composition contexts (`p`, `h1`–`h6`, Input label stack). Text sits on the baseline of its grid-snapped bounding box.
 
Consumer API: `data-multiline` on components where the consumer knows content will wrap.
 
**Axis 2 – Line-height scale**
- **Normal** — generous line-height for prose. Used when the reader reads linearly.
- **Squished** — tight line-height for UI text. Used when the reader scans.
 
Line-height scale is derived from context — component and element defaults carry the decision. Consumers never declare it.
 
## The grid-conformant height system
 
Element height is always a multiple of 4px:
 
```
computed height = padding-block-start + round(cap-height, 4px) + padding-block-end
```
 
`round(1cap, 4px)` snaps the cap-height to the 4px grid. Every value in the chain — cap-height, line-height, padding — is a 4px multiple, so the sum is grid-conformant regardless of line count. [^1]
 
## Density themes
 
Two density themes: **spacious** (default) and **comfortable**. Both the typographic scale and line-height values change between themes. Components consume custom properties and respond to the active theme automatically through CSS inheritance.
 
```css
:root { /* spacious */ }
[data-density="comfortable"] { /* comfortable overrides */ }
```
 
### Phase 1 – Proof of concept
[...]

PR #4843, opened 2026-04-24
refactor: migrate Field off TypographyNext

### LINKED ISSUE #4837: refactor: migrate Field components off TypographyNext
Part of #4655. Follows the pattern established by Button in #4660.

Bundled migration for the three Field subcomponents that share `field.css` and visual rhythm — doing them together keeps token/var decisions consistent.

### Field.Label
```tsx
// packages/eds-core-react/src/components/next/Field/Field.Label.tsx:8
<TypographyNext as="label" family="ui" size="md" baseline="center" lineHeight="default" weight="normal" tracking="normal" ...>
```
Note: also contains `<span className="eds-field__indicator" data-font-size="sm">` — `data-font-size` is TypographyNext vocabulary and should be replaced by a dedicated `--_indicator-font-size` var in `field.css`.

### Field.Description
```tsx
// packages/eds-core-react/src/components/next/Field/Field.Description.tsx:10
<TypographyNext as="p" family="ui" size="sm" baseline="center" lineHeight="default" tracking="normal" ...>
```

### Field.HelperMessage
```tsx
// packages/eds-core-react/src/components/next/Field/Field.HelperMessage.tsx:26
<TypographyNext as="p" family="ui" size="sm" baseline="grid" lineHeight="default" tracking="normal" ...>
```
Note: uses `baseline="grid"` (not `center`), meaning it should use `text-box: trim-both` rather than the cap-height centring path. This is the only one of the three that is explicitly baseline-grid-aligned.

## Migration

- Render each as its native element with its `eds-field__*` class, no TypographyNext wrapper
- Apply typography tokens in `field.css`:
  - `.eds-field__label` → `--eds-typography-ui-body-md-*`
  - `.eds-field__description` → `--eds-typography-ui-body-sm-*`
  - `.eds-field__helper-message` → `--eds-typography-ui-body-sm-*` + `text-box: trim-both ex alphabetic` under `@supports`
- Descriptions can wrap, so apply `text-box: trim-both ex alphabetic` under `@supports` there too
- Replace `data-font-size="sm"` on the indicator with a CSS var

## Acceptance

- [ ] `Field.Label.tsx`, `Field.Description.tsx`, `Field.HelperMessage.tsx` no longer import `TypographyNext`
- [ ] Typography handled by `field.css`
- [ ] Indicator styling intact in Field.Label
- [ ] HelperMessage uses baseline-grid trimming
- [ ] No visual regression across Field stories
- [ ] axe + existing tests still pass

pomfrida:
Closes #4837. Part of #4655.

Follows the direction in [#4647](https://github.com/equinor/design-system/discussions/4647): each component owns its typography in CSS, via tokens.

## Summary

- `Field.Label`, `Field.Description`, `Field.HelperMessage` render as native `label` / `p` elements — no `TypographyNext` wrapper
- Typography applied in `field.css` via `--eds-typography-ui-body-*` tokens
- `font-family` set once on `.eds-field` root (inherits)
- `font-size` / `line-height` set per sub-element — `label` uses `md`, `indicator` / `description` / `helper-message` use `sm`
- `description` and `helper-message` share a selector-list block for the `sm` preset and `text-box: trim-both ex alphabetic` under `@supports` (baseline-grid alignment when wrapping)
- Indicator's `data-font-size="sm"` attribute removed — sizing comes from CSS now

## Notes on repetition

The CSS still repeats `font-size` / `line-height` once per size-tier (once for `md`, twice for `sm` — sharing via selector list). This matches Button's pattern and the "accept repetition" option in #4647. The broader mechanism question (PostCSS mixin vs utility classes vs `adoptedStyleSheets`) is better taken up after Chip (#4836) and Banner.Message (#4835) land — three datapoints inform that choice better than one.

## Test plan

- [x] `Field.test.tsx` — 29 / 29 pass
- [x] Visual check in Storybook across TextField, Checkbox, Radio, Switch, Autocomplete (all consumers of Field)
- [x] Check both density modes (`spacious`, `comfortable`)
- [x] Verify indicator still renders in `Field.Label` with correct size

### DIFF
--- a/packages/eds-core-react/src/components/next/Field/Field.Label.tsx
+++ b/packages/eds-core-react/src/components/next/Field/Field.Label.tsx
@@ -1,29 +1,17 @@
 import { forwardRef } from 'react'
-import { TypographyNext } from '../../Typography'
 import type { FieldLabelProps } from './Field.types'
 
 export const FieldLabel = forwardRef<HTMLLabelElement, FieldLabelProps>(
   function FieldLabel({ children, className, indicator, ...rest }, ref) {
     return (
-      <TypographyNext
+      <label
         ref={ref}
-        as="label"
-        family="ui"
-        size="md"
-        baseline="center"
-        lineHeight="default"
-        weight="normal"
-        tracking="normal"
         className={['eds-field__label', className].filter(Boolean).join(' ')}
         {...rest}
       >
         {children}
-        {indicator && (
-          <span className="eds-field__indicator" data-font-size="sm">
-            {indicator}
-          </span>
-        )}
-      </TypographyNext>
+        {indicator && <span className="eds-field__indicator">{indicator}</span>}
+      </label>
     )
   },
 )
--- a/packages/eds-core-react/src/components/next/Field/Field.Label.tsx
+++ b/packages/eds-core-react/src/components/next/Field/Field.Label.tsx
@@ -6,6 +6,7 @@ export const FieldLabel = forwardRef<HTMLLabelElement, FieldLabelProps>(
     return (
       <label
         ref={ref}
+        data-baseline="center"
         className={['eds-field__label', className].filter(Boolean).join(' ')}
         {...rest}
       >
--- a/packages/eds-core-react/src/components/next/Field/field.css
+++ b/packages/eds-core-react/src/components/next/Field/field.css
@@ -1,24 +1,28 @@
 @layer eds-components {
   .eds-field {
-    /* Layout - use CSS variable for configurable width */
-    display: flex;
-    flex-direction: column;
-    align-items: flex-start;
-    gap: var(--eds-generic-gap-vertical);
-    width: var(--eds-field-width, 100%);
     /* Private color variables */
     --_eds-field-label-color: var(--eds-color-text-strong);
     --_eds-field-description-color: var(--eds-color-text-subtle);
     --_eds-field-indicator-color: var(--eds-color-text-neutral-subtle);
     --_eds-field-helper-color: var(--eds-color-text-subtle);
+
+    /* Layout - use CSS variable for configurable width */
+    display: flex;
+    flex-direction: column;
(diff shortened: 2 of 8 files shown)


pomfrida:
@claude review

packages/eds-core-react/src/components/next/Field/field.css
```diff
@@ -50,12 +59,24 @@
     color: var(--_eds-field-description-color);
   }
 
-  /* HelperMessage styling */
   .eds-field__helper-message {
     margin: 0;
     color: var(--_eds-field-helper-color);
   }
 
+  .eds-field__description,
```
millus, line 75: Should we also take care of removing BEM when doing this? Or should that be a seperate task? These could be .description and .helper-message and .eds-field as the root class, then better nesting would be achieved as well. But this may be out of scope for this pr, I just wanted to point it out while we are going through the components.

packages/eds-core-react/src/components/next/Field/Field.Label.tsx
```diff
@@ -1,29 +1,18 @@
...
+      <label
         ref={ref}
-        as="label"
-        family="ui"
-        size="md"
-        baseline="center"
-        lineHeight="default"
-        weight="normal"
-        tracking="normal"
+        data-baseline="center"
```
millus, line 9: Should data-baseline still be here? The goal of this PR is to move Field off TypographyNext's data-attribute system. Keeping data-baseline="center" re-introduces that dependency. Consider inlining the equivalent text-box + padding styles directly in field.css, like you've done for .description and .helper-message.

millus (dismissed):
LGTM 🚀  Only a leftover data-attribute that probably can be removed as well?

packages/eds-core-react/src/components/next/Field/field.css
pomfrida: Sorry my bad! I should update CLAUDE.md with this as well 
Thanks ✨

PR #4809, opened 2026-04-14
chore: add pnpm.overrides for 13 Dependabot alerts

pomfrida:
## Summary

- Adds 8 `pnpm.overrides` entries to bump transitive dev-dependencies to patched versions, resolving 13 open Dependabot alerts
- All overrides target dev/build-only dependencies — no impact on published packages

| Override | Patched | Severity | Alerts |
|---|---|---|---|
| `flatted` | >=3.4.2 | high | #386 |
| `undici` | >=7.24.0 | high/med | #364–#369 |
| `yaml@^2` | >=2.8.3 | medium | #393 |
| `picomatch@^2` | >=2.3.2 | medium | #388 |
| `brace-expansion@^2` | >=2.0.3 | medium | #405 |
| `@isaacs/brace-expansion` | >=5.0.1 | high | #341 |
| `diff` | >=4.0.4 | low | #340 |
| `js-yaml` | >=4.1.1 | medium | #296 |

Ref #4781

## Test plan

- [x] `pnpm install` succeeds
- [x] All core-react tests pass (797/797)
- [x] CI passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4857, opened 2026-04-27
chore: scope js-yaml override and bump webpackbar to fix docs build

pomfrida:
## Summary

Fixes the Radix `docs` build failure that has been blocking dev deploys on most branches since the Dependabot overrides PR (#4809).

Two stacked bugs:

**1. `js-yaml` override too broad** — `pnpm.overrides` forced `js-yaml >=4.1.1` globally, which dragged `gray-matter@4.0.3` (uses `yaml.safeLoad`) onto v4 where `safeLoad` was removed. Docusaurus crashes on the first front-matter parse:

```
Error: Function yaml.safeLoad is removed in js-yaml 4. Use yaml.load instead.
    at gray-matter/lib/parse.js:12:17
```

Fix: scope the v3 patch to gray-matter only via `"gray-matter>js-yaml": "^3.14.1"`. The lockfile resolves to `js-yaml@3.14.2` (latest available v3.14.x).

**Security note:** CVE-2023-2251 (prototype pollution) — the alert that drove the original `>=4.1.1` override — was **not** backported to the v3 line, so `gray-matter` is now on a technically unpatched js-yaml. Practical risk here is low: `gray-matter` uses `yaml.safeLoad()` (restricted loader, does not execute the YAML tags the CVE targets), and the input is developer-controlled Markdown front-matter, not untrusted user data. Worth being aware of — Dependabot may re-flag the transitive path. If it does, document the exception with this rationale rather than re-broadening the override.

**2. `webpackbar@6.0.1` incompatible with `webpack@5.106.2`** — once #1 was fixed, the next failure surfaced:

```
ValidationError: ProgressPlugin options has an unknown property 'name', 'color', 'reporters', 'reporter'
    at webpackbar/.../ProgressPlugin.js:237
```

Webpack tightened the ProgressPlugin schema; webpackbar 6 still passes its old options. Bumping to `webpackbar@^7.0.0` resolves it (still declares `webpack: 3 || 4 || 5` as peer). Bonus: v7 drops several transitive deps (`chalk`, `figures`, `wrap-ansi`, `markdown-table`, `escape-string-regexp`) in favour of a single lightweight `ansis` lib.

## Verification

Ran the full Radix `docs` build chain locally:
```
pnpm run build:token-sync && pnpm run build:tokens-build && pnpm run build:icons \
  && pnpm run build:tokens && pnpm run build:utils && pnpm run build:core-react \
  && pnpm run build:docs
```
→ exits 0, `[SUCCESS] Generated static files in "build"`.

## Follow-up

Filed #4858 — CI does not currently run `build:docs` or the color-palette-generator app build, which is why both bugs slipped through with green PR checks. Long-term fix is to build the Radix Dockerfiles in CI so any failure that would break Radix breaks CI first.

## Test plan

- [x] Radix `docs` build for this branch passes in dev
- [x] Dependabot alerts for js-yaml and webpackbar remain closed (or if re-flagged on the gray-matter path, see security note above)
- [x] No regression in Storybook / core-react bundles

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,7 +111,9 @@
       "@isaacs/brace-expansion": ">=5.0.1",
       "diff": ">=4.0.4",
       "js-yaml": ">=4.1.1",
-      "lodash": ">=4.18.0"
+      "gray-matter>js-yaml": "^3.14.1",
+      "lodash": ">=4.18.0",
+      "webpackbar": "^7.0.0"
     }
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -14,7 +14,9 @@ overrides:
   '@isaacs/brace-expansion': '>=5.0.1'
   diff: '>=4.0.4'
   js-yaml: '>=4.1.1'
+  gray-matter>js-yaml: ^3.14.1
   lodash: '>=4.18.0'
+  webpackbar: ^7.0.0
 
 importers:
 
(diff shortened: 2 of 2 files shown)


pomfrida:
@claude review

PR #4869, opened 2026-04-28
feat(github): migrate storybook deployment from Azure to Radix

Hjaf:
Replace Azure Blob Storage + Front Door deployment with Radix platform deployment, matching the pattern used by docs and color-palette-generator.

- Add DockerfileStorybook (multi-stage: node build → nginx runtime)
- Add storybook-nginx.conf for SPA serving on port 3000
- Add storybook component to radixconfig.yaml with DNS aliases
- Remove publish_storybook.yaml workflow (Radix auto-deploys on push)
- Remove storybook deploy/CDN purge jobs from publish_core_react.yaml
- Remove storybook endpoint from _purge_cdn.yaml
- Delete obsolete packages/eds-core-react/Dockerfile.storybook
- Update documentation with new deployment model

---

**Note**

Radix only reads `radixconfig.yaml` from the Config Branch (`main`), even for `dev` environment builds ([Radix docs](https://radix.equinor.com/radix-config/)). New components on feature branches are ignored.

To enable the storybook build, we need to push [`radixconfig.yaml`](radixconfig.yaml) to `main` first. Components are only added to the config, no existing components are affected. 

This migration changes how dev storybook deployments work. Previously, storybook was deployed via a manually dispatched workflow, giving us control over when updates went live. With Radix, storybook rebuilds automatically on every push to `feat/*` (dev) and `main`(prod) branches. We need to merge the config and observe the impact before fully committing. If automatic rebuilds on every push prove too disruptive, we may need to explore alternative deployment patterns for dev (e.g. restricting the build trigger to specific branches or using manual promotion).


radixconfig.yaml
```diff
@@ -20,11 +20,18 @@ spec:
...
       component: docs
+    - alias: eds-storybook
+      environment: prod
+      component: storybook
   dnsExternalAlias:
     - alias: eds.equinor.com
       environment: prod
       component: docs
       useCertificateAutomation: true
+    - alias: storybook.eds.equinor.com
```
emirgens:
```suggestion
    remove
```

PR #4871, opened 2026-04-30
fix: correct Chip height to match Figma design spec

pomfrida:
## Summary

- Chip rendered at 32px (spacious) / 24px (comfortable) due to `padding-block` including the full line-height rather than just the cap-height
- Applies the same line-height compensation pattern already used by Button: subtract half the `(line-height − cap-height)` overflow from each side of the padding block
- Fixes spacious chips to 28px and comfortable chips to 20px, matching the Figma spec (node 3853-3622)

## What changed

`chip.css` — three pseudo-private custom properties added at the component root:

- `--_label-line-height` — squished line-height for the UI body-md text
- `--_label-height` — cap-height snapped to 4px grid (`round(1cap, 4px)`)
- `--_padding-block` — raw selectable-space-vertical token
- `--_padding-adjusted` — `padding-block − (line-height − cap-height) / 2`

`min-height` is set as a floor so the chip never shrinks below `padding × 2 + cap-height`.

Icon margins updated to use `--_padding-adjusted` so icons stay flush with the chip edge.

## Test plan

- [ ] Default chip: 28px spacious, 20px comfortable
- [ ] Deletable chip: same heights, delete icon correctly flush
- [ ] Dropdown chip: same heights, arrow icon correctly flush
- [ ] Label is vertically centred in both densities

### DIFF
--- a/packages/eds-core-react/src/components/next/Chip/chip.css
+++ b/packages/eds-core-react/src/components/next/Chip/chip.css
@@ -2,7 +2,7 @@
   .eds-chip {
     /* Typography metrics — same compensation pattern as Button */
     --_label-line-height: var(--eds-typography-ui-body-md-line-height-squished);
-    --_label-height: round(1cap, 4px);
+    --_label-height: round(1cap, 4px); /* Figma has a cap-rounded value but it is not exposed as a CSS token; 1cap gives the same result (Inter metrics + 4px grid) */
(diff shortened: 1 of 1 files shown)


pomfrida:
@claude review

Chibuzor-Nwemambu (approved):
Looks Great 🙌🏾

PR #4864, opened 2026-04-28
fix: emit explicit dark-scope rules instead of light-dark() in tokens

pomfrida:
## Summary

- Eliminates `light-dark()` from published `@equinor/eds-tokens` CSS by emitting explicit `[data-color-scheme="light"|"dark"]` scope rules + `prefers-color-scheme` media fallback during the build
- Restores subtree-scoped dark mode (broken in prod Storybook since the Vite 8 bump in #4759 — Vite 8/Rolldown runs lightningcss over imported CSS by default and polyfills `light-dark()` into a `:root`-baked `var()` pattern that breaks `[data-color-scheme]` scoping)
- Adds defense-in-depth lightningcss targets to the eds-core-react Storybook viteFinal so downstream consumers can follow the same pattern

## How it works

A new `build-dark-scope` step runs in `_build:css` between the bundle and minify steps. It scans `variables.css` for `--name: light-dark(L, D);` declarations, replaces each with `--name: L;` in place, and appends three blocks:

```css
[data-color-scheme="light"] { --eds-color-…: L; … }
[data-color-scheme="dark"]  { --eds-color-…: D; … }
@media (prefers-color-scheme: dark) {
  :root:not([data-color-scheme="light"]) { --eds-color-…: D; … }
}
```

With no `light-dark()` in the output, no downstream lightningcss/postcss/esbuild pass can polyfill it incorrectly. A build-time assertion guards against regression.

See the comment on #4863 for the full root-cause writeup.

## Test plan

- [x] eds-tokens-build unit tests pass (27/27)
- [x] eds-tokens build pipeline produces `variables.min.css` with: 0 `light-dark(`, 0 `--lightningcss-*` consumers, 2 `[data-color-scheme=dark]` rules, 2 `prefers-color-scheme` media queries
- [x] Visually verified in local Storybook production build: Switch dark-mode story renders correctly (track `rgb(82, 92, 101)` enabled / dark, was `rgb(225, 225, 225)` white; labels legible; disabled no longer brighter than enabled)
- [x] CI green
- [x] Reviewer spot-check: a few other components that use dual-mode tokens render correctly in dark scope (e.g. Button variants, Banner, Field components in the Storybook dark-mode wrapper)

Fixes #4863

PR #4873, opened 2026-04-30
fix: re-declare semantic color tokens on data-color-scheme scopes

pomfrida:
## Summary

Follow-up to #4864. The previous fix widened `:root` blocks containing `light-dark()` (primitive tokens) to also match `[data-color-scheme="light"]`, but didn't widen blocks containing only `var(--eds-color-*)` references (semantic tokens like `--eds-color-text-strong`). This caused semantic tokens to be inherited from `:root` with the wrong values when the OS color scheme differs from the explicit `[data-color-scheme]` scope.

**Symptom:** When the OS is in dark mode and a wrapper sets `[data-color-scheme="light"]` (e.g. Storybook's wrapper or a light-inside-dark subtree), Field labels/descriptions render as near-white text on white background — invisible.

**Root cause:** `--eds-color-text-strong: var(--eds-color-neutral-13)` is declared in a separate `:root, [data-color-appearance="neutral"]` block. The previous fix didn't widen that block. The semantic value is computed at `:root` (where the media query has set primitives to dark values) and inherited by the wrapper, which only re-declares primitives — not semantics.

**Fix:** Widen semantic `:root` blocks (`var(--eds-color-*)` references but no `light-dark()`) to match BOTH `[data-color-scheme="light"]` AND `[data-color-scheme="dark"]`. The `var()` re-resolves at each scope, picking up the local primitive values from the wrapper.

This is the workaround documented in [parcel-bundler/lightningcss#821](https://github.com/parcel-bundler/lightningcss/issues/821): "re-apply the `:root` theme variables on every element that changes `color-scheme`."

## Test plan

- [x] Unit tests for `build-dark-scope.ts` — added 3 new cases (compound selectors, semantic block widening, primitive blocks NOT widened with dark)
- [x] Build assertion — `assert-no-light-dark.mjs` now requires ≥3 occurrences each of light and dark scopes (color-scheme rule + primitive widening + semantic widening)
- [x] Storybook prod build verified with Playwright in 4 scenarios:
  - Mac DARK + SB Light + Field: label = `rgb(29, 29, 29)` ✓ (was the bug — near-white before)
  - Mac LIGHT + Dark scope + Switch: wrapper text-strong = `#f5fdff` ✓
  - Mac LIGHT + SB Light: label = `rgb(29, 29, 29)` ✓ (no regression)
  - Mac DARK + Dark scope + Switch (#4864 original): track = `rgb(82, 92, 101)` ✓ (no regression)


pomfrida:
@claude review

pomfrida:
**Side effect note (for transparency):**

This widening adds `[data-color-scheme="light"]` and `[data-color-scheme="dark"]` to the `:root, [data-color-appearance="neutral"]` block. There is one theoretical regression to be aware of:

If a consumer **nests** `[data-color-scheme="light"]` inside an element with `[data-color-appearance="accent"]` (or another non-neutral appearance), the nested scope element re-declares semantic tokens with neutral primitives — overriding the inherited accent appearance.

**Why it's theoretical**: searched the codebase — no EDS component, EdsProvider, Storybook decorator, or docs uses this nesting pattern. EDS components (Button, Checkbox, Radio, Input, Chip, Search) set `data-color-appearance` on themselves, not on an ancestor.

**Workaround for affected consumers (if any)**: set both attributes on the same element instead of nesting them:

```html
<!-- Works (cascade source-order: accent block wins) -->
<div data-color-scheme="light" data-color-appearance="accent">…</div>

<!-- Loses accent inside (regression) -->
<div data-color-appearance="accent">
  <div data-color-scheme="light">…</div>  <!-- resets to neutral -->
</div>
```

pomfrida:
@claude review

PR #4885, opened 2026-05-04
fix(eds-tokens): correct 5xl tracking-wide alias in UI Body mode

### LINKED ISSUE #4876: 5xl letter spacing for UI typography is incorrect

eddiman:
## Summary

- The Font family / UI Body mode aliased `5xl/tracking-wide` to `{typography.ui-body.5xl.font-weight-lighter}` instead of `{typography.ui-body.5xl.tracking-wide}`, producing a wide-tracking value of `300` (the lighter font weight) instead of the intended `~3.52`. Every other size (xs–4xl) already aliased correctly; the Header collection's 5xl was also already correct.
- Bug originated in the Figma variable mode, so patching only the synced JSON would regress on next sync. Fix was applied in Figma (Variables → 🅰️ Font family → UI Body → `5xl/tracking-wide` re-bound to `typography → ui-body → 5xl → tracking-wide`); this PR is the resulting synced JSON plus regenerated typography CSS and TS outputs.
- Net effect: `--eds-font-family-size-5xl-tracking-wide` now resolves to `var(--eds-typography-ui-body-5xl-tracking-wide)` and the TS `trackingWide` value flips from `300` to `3.5199999809265137`.

## Test plan

- [x] Verify `--eds-font-family-size-5xl-tracking-wide` resolves to the expected wide-tracking value in a browser/Storybook
- [x] Confirm no other typography variables drift in the diff
- [x] CI green

Closes #4876


Chibuzor-Nwemambu (approved):
Looks good 🚒

PR #4903, opened 2026-05-07
fix(eds-tokens): add composeTextStyle helper and remove broken typography axis TS files

### LINKED ISSUE #4901: TS typography tokens: unclear partial/full split and flat structure
**Describe the bug**

Surfaced while pulling the TS typography tokens into `eds-mobile`. Two issues with the typography tokens shipped from `@equinor/eds-tokens`:

### 1. Unclear semantic levels — "full" vs "partial" token files

The package exposes typography tokens in what appear to be two categories:

- **Full**: `font-family-ui.ts`, `font-family-header.ts` — these are the only files `eds-mobile` currently consumes for UI and Header typography.
- **Partial (?)**: every other file, e.g. `font-size-xs.ts`, `font-weight-normal.ts`.

The intent of the partial files is unclear. Initial assumption was that they compose into the full ones, but that doesn't hold up — `font-size-xs.ts`, for example, contains values like `iconSize`, `gapHorizontal`, and `gapVertical`, none of which appear in the full files.

We need to settle what these partial files are for:
- If they are not meant to be consumed by implementations, they should be removed from the package.
- If `eds-mobile` is misunderstanding their purpose, we should adopt them there instead.

### 2. Flat-ish token structure

Compared to spacing tokens, typography tokens are only semi-nested. A semantic value like `font size = sm, tracking = wide` is reached via `sm.trackingWide` rather than the more conventional `sm.tracking.wide` used elsewhere in the tokens.

The flat structure forces consumers to add a mapping layer to recover the expected shape. Example from `eds-mobile`:

```ts
export const PROP_TOKEN_FIELD_MAP = {
    weight: {
        bolder: "fontWeightBolder",
        lighter: "fontWeightLighter",
        normal: "fontWeightNormal",
    },
    lineHeight: {
        default: "lineHeightDefault",
        squished: "lineHeightSquished",
    },
    tracking: {
        normal: "trackingNormal",
        tight: "trackingTight",
        wide: "trackingWide",
    },
} as const satisfies PropTokenFieldMap;
```

**Expected behavior**

- Clear, documented intent for every typography token file shipped from `@equinor/eds-tokens` — either the partial files are first-class consumables (with usage guidance) or they are removed.
- Typography tokens follow the same nesting convention as spacing tokens (`sm.tracking.wide`) so consumers don't need a property-to-token mapping layer.

**Specifications**

- Package: `@equinor/eds-tokens`
- Reported by: `eds-mobile` consumers

**Additional context**

Original feedback from `eds-mobile`:

> Different semantic levels and/or unclear intention of tokens exposed (?):
> The tokens package exposes typography tokens in two categories: full and partial (?).
> The full ones are the `font-family-ui.ts` and `font-family-header.ts` files. The partial definitions are found in all other files like for example `font-size-xs.ts`, `font-weight-normal.ts`.
> The mobile implementation only uses the full files for UI and Header typography. It is unclear exactly what the intention is of the partial files, and how to apply them. I originally thought that the partial files together made up the full ones, but looking at `font-size-xs.ts` for example, I see values such as `iconSize`, `gapHorizontal` and `gapVertical`, neither of which are present in any of the full ones. We need to settle on what exactly is the intention of these token defs. If they are not to be considered for implementation, I suggest removing them from the token package. If we have misunderstood how they should be used, we should implement them in mobile instead.
>
> Flat-ish token structure:
> In relation to spacing token, typography tokens are only semi-nested. A semantic value such as `font size = sm, tracking = wide` is accessed with `sm.trackingWide` instead of `sm.tracking.wide` which is more commonly used elsewhere in the tokens. The flat structure means we need to provide an additional layer of mapping to achieve the desired structure.

eddiman:
Closes #4901.

## Summary

- Add `composeTextStyle` helper in `@equinor/eds-tokens` — a runtime composer that resolves the five typography axes (font-family × font-size × font-weight × line-height × tracking) for non-CSS consumers (React Native, SSR, design tooling). Sourced from the existing `build/ts/typography/font-family-{ui,header}.ts` matrix and `font-size-*.ts` size-axis extras, so Figma stays the single source of truth.
- Stop emitting TS for the `font-weight`, `line-height`, and `tracking` axes. Style Dictionary cannot represent runtime mode switching, so those files each contained one arbitrary cell from the family matrix and were silently wrong for any other combination. CSS output for these axes is unchanged.
- Documentation: rewrite the "TypeScript Tokens" section in `instructions/typography.md` around `composeTextStyle`, add a "Typography (non-CSS targets)" subsection to the README, and explain the rationale + build-order dependency in the package `CLAUDE.md`.

The exports map for `@equinor/eds-tokens/ts/typography/font-family-*` and `font-size-*` is unchanged — external consumers can still import the matrix directly when they need raw data.

A separate comment on issue #4901 documents the implementation risks (build-order dependency, `dist/types/src/` path drift, alternatives considered and rejected).

## Test plan

- [x] Unit tests for `composeTextStyle` (11 cases, vitest) — `pnpm --filter @equinor/eds-tokens run test`
- [x] Type-check the package — `pnpm --filter @equinor/eds-tokens run types`
- [x] Build the package end-to-end — `pnpm --filter @equinor/eds-tokens run build`, verify `dist/types/src/index.d.ts` exists and consumer imports resolve
- [x] Type-check a downstream consumer — `pnpm --filter @equinor/eds-core-react run types`
- [ ] Verify the published artifact via `pnpm pack` (deferred to release validation)
- [ ] Verify `eds-mobile` can drop its `PROP_TOKEN_FIELD_MAP` workaround in favour of `composeTextStyle` (out of scope for this PR — requires the new release to be available)


eddiman:
**Update — pushed a force-update consolidating the typography TS surface around `composeTextStyle`.**

Branch rewritten in place; commit hashes change. Same 3-commit shape:

1. `fix(eds-tokens): add composeTextStyle helper for non-CSS consumers`
2. `fix(eds-tokens): consolidate typography TS surface around composeTextStyle`
3. `docs(eds-tokens): document composeTextStyle and the typography TS surface`

### What changed since the original push

The `font-size-{xs..6xl}.ts` files were emitting a mix of:
- **Family-independent** values (`iconSize`, `gapHorizontal`, `gapVertical`) — genuinely useful, no duplicate elsewhere.
- **Family-dependent** values (`fontSize`, `lineHeight*`, `tracking*`, `fontWeight*`) — silently pinned to the UI Body family, exactly the same trap as the broken axis files we removed in commit 2.

This update removes the per-size TS files entirely and replaces them with a single `build/ts/typography/size-extras.ts` containing only the family-independent extras keyed by `fontSize`:

```ts
export const sizeExtras = {
  xs: { iconSize: 16, gapHorizontal: 6.5, gapVertical: 6.5 },
  sm: { iconSize: 18, gapHorizontal: 7.5, gapVertical: 7.5 },
  md: { iconSize: 20, gapHorizontal: 8.5, gapVertical: 8.5 },
  // ...
} as const
```

The helper now imports two family files plus this one extras file (down from twelve imports). The exports map drops `@equinor/eds-tokens/ts/typography/font-size-*` — the supported TS surface is now `composeTextStyle` plus `font-family-{ui,header}` and `size-extras` for raw access.

### How `size-extras.ts` is generated

The build pipeline runs the per-size font-size builds with TS emission enabled, then a post-step parses the just-emitted per-size TS files for `iconSize`/`gapHorizontal`/`gapVertical` and writes the consolidated `size-extras.ts`. The temporary per-size files are deleted before the build finishes.

This is pragmatic — Style Dictionary's resolved-value APIs don't ergonomically aggregate across multiple sources, and the per-size CSS files use `var(...)` references rather than literal pixels (`outputReferences: true`). Reading the temp TS gives us already-resolved numeric values for free. Documented in `packages/eds-tokens/CLAUDE.md`.

### Why this is better

- **One clear way to do typography in TS**: `composeTextStyle`. The two raw files are escape hatches, not invitations.
- **No more silent traps**: every TS export is either a complete matrix (`font-family-*`) or family-independent data (`size-extras`). No file claims to represent an axis it can't fully represent.
- **Cleaner stepping stone for future work**: if we later decide to merge size-extras into the family files (option 3a discussed in PR review), we already have a clean abstraction to move.

### Verified locally

- `pnpm --filter @equinor/eds-tokens run test` — 11/11 pass
- `pnpm --filter @equinor/eds-tokens run types` — clean
- `pnpm --filter @equinor/eds-tokens run build` — produces 3 typography modules in dist (`font-family-{ui,header}`, `size-extras`)
- `pnpm --filter @equinor/eds-core-react run types` — clean

### Risks comment on the issue

The risks comment on #4901 still applies — the new size-extras post-step is documented in the CLAUDE.md note about build-time aggregation, and the build-order dependency / `dist/types/src/` path drift / tests-need-build-files trade-offs are unchanged.

eddiman:
## Usage examples for `composeTextStyle`

### Web (CSS-shape, default)

```tsx
import { composeTextStyle } from '@equinor/eds-tokens'

const bodyMd = composeTextStyle({ fontFamily: 'ui', fontSize: 'md' })
// {
//   fontFamily: 'Inter',
//   fontSize: 14,
//   fontWeight: 400,
//   lineHeight: 20,
//   letterSpacing: 0,
// }

function Caption({ children }: { children: React.ReactNode }) {
  return <p style={composeTextStyle({ fontFamily: 'ui', fontSize: 'sm' })}>{children}</p>
}
```

### React Native

```tsx
import { Text, StyleSheet } from 'react-native'
import { composeTextStyle } from '@equinor/eds-tokens'

const styles = StyleSheet.create({
  heading: composeTextStyle({
    fontFamily: 'header',
    fontSize: '2xl',
    fontWeight: 'bolder',
    format: 'react-native',
  }),
  body: composeTextStyle({
    fontFamily: 'ui',
    fontSize: 'md',
    format: 'react-native',
  }),
})

export function Article() {
  return (
    <>
      <Text style={styles.heading}>Article title</Text>
      <Text style={styles.body}>Body copy in the UI face.</Text>
    </>
  )
}
```

The `react-native` format returns `fontWeight` as a string (`'400'`) so it slots straight into `<Text style>` without manual coercion.

### Drop-in replacement for the eds-mobile `PROP_TOKEN_FIELD_MAP`

The workaround that originally surfaced this bug:

```ts
// before — manually mapping prop names to picked cells
export const PROP_TOKEN_FIELD_MAP = {
  weight: { bolder: 'fontWeightBolder', lighter: 'fontWeightLighter', normal: 'fontWeightNormal' },
  lineHeight: { default: 'lineHeightDefault', squished: 'lineHeightSquished' },
  tracking: { normal: 'trackingNormal', tight: 'trackingTight', wide: 'trackingWide' },
} as const
```

becomes:

```ts
// after — let composeTextStyle do the picking
import { composeTextStyle } from '@equinor/eds-tokens'

const style = composeTextStyle({
  fontFamily: 'ui',
  fontSize: 'sm',
  fontWeight: 'bolder',
  lineHeight: 'squished',
  tracking: 'wide',
  format: 'react-native',
})
```

### With size extras (icon size + gap that scale with font-size)

For chip/button-like layouts where an icon sits next to text:

```tsx
import { composeTextStyle } from '@equinor/eds-tokens'

const chip = composeTextStyle({
  fontFamily: 'ui',
  fontSize: 'sm',
  includeSizeExtras: true,
})
// {
//   fontFamily: 'Inter',
//   fontSize: 12,
//   fontWeight: 400,
//   lineHeight: 16,
//   letterSpacing: 0,
//   iconSize: 18,        // ← matches the size axis
//   gapHorizontal: 7.5,  // ← matches the size axis
//   gapVertical: 7.5,
// }

function Chip({ icon, label }: { icon: React.ReactNode; label: string }) {
  return (
    <span
      style={{
        display: 'inline-flex',
        alignItems: 'center',
        gap: chip.gapHorizontal,
        fontFamily: chip.fontFamily,
        fontSize: chip.fontSize,
        lineHeight: `${chip.lineHeight}px`,
      }}
    >
      <span style={{ width: chip.iconSize, height: chip.iconSize }}>{icon}</span>
      {label}
    </span>
  )
}
```

### Defaults — only the two required axes

When you just want a sensible default, `fontFamily` and `fontSize` are the only required inputs. Weight defaults to `'normal'`, line-height to `'default'`, tracking to `'normal'`, format to `'css'`.

```ts
composeTextStyle({ fontFamily: 'header', fontSize: '4xl' })
// resolves to header / 4xl / normal / default / normal — no boilerplate
```

### Tracking `'loose'`

`'loose'` resolves to the same numeric value as `'wide'` because the Figma source aliases the two. Both are accepted so consumers don't have to guess; the helper documents this in TSDoc.

### Type narrowing on the return shape

`composeTextStyle` is overloaded so the return type matches your options without extra work:

```ts
// CssTextStyle
const a = composeTextStyle({ fontFamily: 'ui', fontSize: 'md' })

// CssTextStyleWithExtras
const b = composeTextStyle({ fontFamily: 'ui', fontSize: 'md', includeSizeExtras: true })

// ReactNativeTextStyle (string fontWeight)
const c = composeTextStyle({ fontFamily: 'ui', fontSize: 'md', format: 'react-native' })

// ReactNativeTextStyleWithExtras
const d = composeTextStyle({
  fontFamily: 'ui',
  fontSize: 'md',
  format: 'react-native',
  includeSizeExtras: true,
})
```

### Invalid input

Bad values throw a `RangeError` at runtime with a message naming the axis and the allowed set, so misuse fails loudly:

```ts
composeTextStyle({ fontFamily: 'display' as any, fontSize: 'md' })
// RangeError: composeTextStyle: invalid fontFamily "display".
//   Expected one of: ui, header.
```

pomfrida:
@claude review

HaakonSvane:
My two cents on this. Can't speak for much other than the exposed interface though.

There is very little to be gained from the consuming side of this by adopting this method in relation to what is already in place for the mobile implementation.

It has some issues like
- Typings being hard-coded and not derived from tokens themselves
- Lessened TS-support than if we access the tokens directly.

---

Just to make sure we understand the same problem to solve: the _variables_ in play in the typography tokens (excluding the extras for now) are `type`(ui vs heading) `weight`, `tracking` and `lineHeight` and `size`. This means we need a way to map those to actual styling values. 

Our current implementation looks like this in rough terms:

```ts

// ----------- Token defs
import { typography as header } from "@equinor/eds-tokens/ts/typography/font-family-header";
import { typography as ui } from "@equinor/eds-tokens/ts/typography/font-family-ui";

export const typographyToken = {
    header,
    ui,
} as const;

// ----------- Styling typography

// This is needed to "unflatten" the tokens
const TOKEN_FIELD_MAP = {
    weight: {
        bolder: "fontWeightBolder",
        lighter: "fontWeightLighter",
        normal: "fontWeightNormal",
    },
    lineHeight: {
        default: "lineHeightDefault",
        squished: "lineHeightSquished",
    },
    tracking: {
        normal: "trackingNormal",
        tight: "trackingTight",
        wide: "trackingWide",
    },
} as const

const styleToken = EDSStyleSheet.create(
  (token, { type, weight, tracking, lineHeight, size }) => {
    const fontFamilySize = token.typography[type].fontFamilySize;
    const sizeToken = fontFamilySize[size];

    return {
      text: {
        fontFamily: token.newTypography[type].typography.fontFamily,
        fontSize: sizeToken.fontSize,
        fontWeight: sizeToken[TOKEN_FIELD_MAP.weight[weight]],
        lineHeight: sizeToken[TOKEN_FIELD_MAP.lineHeight[lineHeight]],
        letterSpacing: sizeToken[TOKEN_FIELD_MAP.tracking[tracking]],
        color: token.colors.text.neutral.strong,
      },
    }
  }
)

---

Our only real need here is for the tokens to be nested to look something like what our map is doing at the moment:

```ts
export const typography = {
  documentation: 'header',
  typography: {
    fontFamily: 'Equinor',
  },
  fontFamilySize: {
    xs: {
      fontSize: 12,
      tracking: {
        tight: -1.149999976158142,
        normal: 0,
        wide: 1.149999976158142,
      },
      fontWeight: {
        lighter: 300,
        normal: 400,
        bolder: 500,
      },
      lineHeight: {
        default: 16,
        squished: 12,
      },
    },
    // ...
  },
} as const
```

With an implementation like this, we could infer types directly from the token, leading to a pretty decent type system:
```ts
type SizeToken = (typeof typography)['fontFamilySize']['xs']
type TypographyWeight = keyof SizeToken['fontWeight']                      // "ligther" | "normal" | "bolder"
type TypographyTracking = keyof SizeToken['tracking']                        // "tight" | "normal" | "wide"
type TypographyLineHeight = keyof SizeToken['lineHeight']                // "default" | "squished" 
type TypographySize = keyof (typeof typography)['fontFamilySize']  // "xs" | "sm" | ....
```

Chibuzor-Nwemambu:
Totally agree with you @HaakonSvane. For mobile specifically, `composeTextStyle` replaces our `TOKEN_FIELD_MAP` workaround but doesn't really eliminate the underlying issue which is that the types are still hard-coded rather than derived from the token shape.

The nested structure Haakon sketched would actually let us delete our mapping layer entirely. Is that something worth looking into, or is the intention to keep the flat shape long-term?

eddiman:
Closing this in favour of #4915.

Based on the feedback from @HaakonSvane and @Chibuzor-Nwemambu — the underlying point that the helper solved the wrong layer was correct. The new PR nests the tokens at the source so each size cell exposes `fontWeight` / `tracking` / `lineHeight` as nested objects (plus inlined `iconSize` / `gapHorizontal` / `gapVertical` extras), letting consumers derive variant types directly via `keyof` without a helper or mapping layer.

Net effect for the mobile codebase: `TOKEN_FIELD_MAP` can be deleted entirely, and no `composeTextStyle` is needed in between.

CSS output is unchanged. Closes #4901 from the new PR.

PR #4904, opened 2026-05-07
fix: forward child refs in Tooltip and TabList on React 18

### LINKED ISSUE #4898: [Tooltip] React 18 forwarded ref is silently dropped
**Describe the bug**
The tooltip reads the anchor child's ref via children.props.ref (React 19 convention) and unconditionally overrides the child's ref through cloneElement. Our project is on React 18, so the ref is not part of the props, causing any ref attached to a Tooltip-wrapped child to be silently dropped.

The following code should be a minimal case for a React 18 style forwarded ref. Here, `myButtonRef` will never get attached due to Tooltip incorrectly merging
```tsx
function MyButtonComp = (function (props: ButtonProps, ref: React.ForwardedRef<HTMLButtonElement>) {
   return <button ref={ref} />
}

const MyButton = React.forwardRef(MyButtonComp)

function SomeComponent(/* ... */) {
  const myButtonRef = React.useRef(null)
  
  // ...

  return (
     <Tooltip title="My tooltip" >
         <MyButton ref={myButtonRef} />
     </Tooltip>
  )
}
```

**Expected behavior**
I expect `myButtonRef` to be properly attached to the button element

**Specifications**

- Version: `react@18.3.1` , and `eds-core-react@2.4.0`
- Browser: Firefox, chrome
- OS: Windows

pomfrida:
## Summary

- Tooltip and TabList silently dropped forwarded child refs on React 18 because both components read only `child.props.ref` — but React 18 stores refs on `child.ref`, not in props. React 19 moved refs into props.
- Adds a `getElementRef` helper in `@equinor/eds-utils` that reads from both locations (`props.ref` first to avoid the React 19 `element.ref` deprecation warning), and uses it in both call sites.
- Peer dep declares `react: ^18 || ^19`, so both must work.

Closes #4898

## Test plan

- [x] `pnpm --filter @equinor/eds-utils test` — 29 passed (4 new for `getElementRef`: React 18 shape, React 19 shape, no ref, precedence)
- [x] `pnpm --filter @equinor/eds-core-react exec jest 'Tooltip|Tabs'` — 42 passed across 3 suites
- [x] Lint clean on all changed files
- [x] Visual smoke test in Storybook (Tooltip on Button, Tabs with-styled-component, Tabs with-panels keyboard nav) — pixel-equivalent to prod, no console warnings
- [x] Manual verification on a React 18 project (reviewer)

## Notes / follow-ups

- The repo already runs a `React 18 compatibility` CI job, but the bug slipped through because no existing test exercised ref-forwarding under React 18 specifically. The new `getElementRef` unit tests cover both React 18 and 19 element shapes via synthetic shapes (the test runtime is React 19). Adding an explicit ref-forwarding integration test that runs under both React versions would strengthen the safety net — worth a separate discussion/issue.
- Other `cloneElement` sites audited (`TabPanels`, `Accordion*`, `Banner/BannerIcon`, `Chip`, `ToggleButton`, `Menu/MenuList`, `next/Tooltip`) don't forward refs from children and are not affected. `next/Slot` already constructs its merged ref correctly without reading from the child element.

### DIFF
--- a/packages/eds-core-react/src/components/Tabs/TabList.tsx
+++ b/packages/eds-core-react/src/components/Tabs/TabList.tsx
@@ -11,7 +11,7 @@ import {
   isValidElement,
 } from 'react'
 import styled from 'styled-components'
-import { mergeRefs } from '@equinor/eds-utils'
+import { mergeRefs, getElementRef } from '@equinor/eds-utils'
 import { TabsContext } from './Tabs.context'
 import { Variants } from './Tabs.types'
 
@@ -96,7 +96,6 @@ const TabList = forwardRef<HTMLDivElement, TabListProps>(function TabsList(
       type ChildPropsWithRef = {
         value?: string | number
         disabled?: boolean
-        ref?: React.Ref<HTMLButtonElement>
         [key: string]: unknown
       }
 
@@ -106,7 +105,7 @@ const TabList = forwardRef<HTMLDivElement, TabListProps>(function TabsList(
         ? controlledActive === activeTab
         : $index === activeTab
 
-      const childRef = childProps?.ref || null
+      const childRef = getElementRef<HTMLButtonElement>(child) ?? null
       const tabRef =
         isActive && childRef
           ? mergeRefs<HTMLButtonElement>(childRef, selectedTabRef)
--- a/packages/eds-core-react/src/components/Tabs/TabList.tsx
+++ b/packages/eds-core-react/src/components/Tabs/TabList.tsx
@@ -105,7 +105,7 @@ const TabList = forwardRef<HTMLDivElement, TabListProps>(function TabsList(
         ? controlledActive === activeTab
         : $index === activeTab
 
-      const childRef = getElementRef<HTMLButtonElement>(child) ?? null
+      const childRef = getElementRef<HTMLButtonElement>(child)
       const tabRef =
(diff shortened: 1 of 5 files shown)


pomfrida:
@claude review

pomfrida:
@claude review

Chibuzor-Nwemambu (approved):
Looks good! 🙌🏾 

Tested locally and all tests pass:

  - pnpm --filter @equinor/eds-utils test — 29 passed (includes the 4 new getElementRef tests)
  - pnpm --filter @equinor/eds-core-react exec jest --testPathPatterns="Tooltip|Tabs" — 42 passed across 3 suites

Wasn't able to complete the manual React 18 verification since there's no React 18 project on hand to test against. The unit tests cover both React 18 and 19 element shapes via synthetic fixtures, but an actual runtime test would give stronger confidence here.

All in all, it looks good.

pomfrida:
> Wasn't able to complete the manual React 18 verification since there's no React 18 project on hand to test against. The unit tests cover both React 18 and 19 element shapes via synthetic fixtures, but an actual runtime test would give stronger confidence here.

Thank you! I forgot to cross that task out, I tested it myself. I just created a vite project, implementing our components and using them in a simple web app with react 18

PR #4912, opened 2026-05-11
feat: Accordion EDS 2.0

### LINKED ISSUE #4441: [Accordion] Implement in code
## Summary

## Acceptance Criteria

pomfrida:
## Summary

Adds the `Accordion` component to `@equinor/eds-core-react/next`, built on native `<details>`/`<summary>` for free accessibility, keyboard support, and CSS-only height animation.

## API

Compound component:

```tsx
<Accordion exclusive>
  <Accordion.Item defaultOpen>
    <Accordion.Header>Title</Accordion.Header>
    <Accordion.Panel>Content</Accordion.Panel>
  </Accordion.Item>
</Accordion>
```

- `exclusive` on the root: only one item open at a time via the native HTML `name` attribute on `<details>` — no JS state machine.
- `Accordion.Item` supports controlled (`open` + `onOpenChange`) and uncontrolled (`defaultOpen`) modes.
- `Accordion.Header` renders as `<summary>`; `Accordion.Panel` is the collapsible content.

## Notes

- All spacing and color values derived from EDS tokens — no magic numbers in CSS.
- Open/close animation uses `details::details-content` + `calc-size(auto, size)` (progressive enhancement; older engines fall back to instant toggle).
- Includes `prefers-reduced-motion` opt-out.

## Test plan

- [ ] Visual check in Storybook against Figma frame (Open/Closed × Default/Hover/Focus)
- [ ] Keyboard: Tab + Enter/Space toggles, focus ring visible
- [ ] Exclusive group: opening one closes the previously open item
- [ ] Controlled mode: parent `open` state stays in sync with user clicks
- [ ] Screen reader announces expand/collapse (native `<details>` behaviour)

### DIFF
--- a/packages/eds-core-react/src/components/next/Accordion/Accordion.stories.tsx
+++ b/packages/eds-core-react/src/components/next/Accordion/Accordion.stories.tsx
@@ -1,0 +1,110 @@
+import { useState } from 'react'
+import type { Meta, StoryFn, StoryObj } from '@storybook/react-vite'
+import { Accordion } from './Accordion'
+import type { AccordionProps } from './Accordion.types'
+
+const meta: Meta<typeof Accordion> = {
+  title: 'EDS 2.0 (beta)/Data Display/Accordion',
+  component: Accordion,
+  tags: ['beta'],
+  parameters: {
+    docs: {
+      description: {
+        component: `
+⚠️ **Beta Component** - This component is under active development.
+
+Built on the native \`<details>\`/\`<summary>\` elements for free accessibility,
+keyboard support and CSS-only open/close animation.
+
+\`\`\`tsx
+import { Accordion } from '@equinor/eds-core-react/next'
+
+<Accordion>
+  <Accordion.Item>
+    <Accordion.Header>Title</Accordion.Header>
+    <Accordion.Panel>Content</Accordion.Panel>
+  </Accordion.Item>
+</Accordion>
+\`\`\`
+        `,
+      },
+    },
+  },
+  argTypes: {
+    exclusive: {
+      control: 'boolean',
+      description:
+        'When true, only one item can be open at a time (native HTML group via the `name` attribute).',
+    },
+  },
+  args: {
+    exclusive: false,
+  },
+}
+
+export default meta
+
+export const Introduction: StoryFn<AccordionProps> = (args) => (
+  <Accordion {...args}>
+    <Accordion.Item defaultOpen>
+      <Accordion.Header>First section</Accordion.Header>
+      <Accordion.Panel>
+        Accordions help manage information density while keeping pages
+        scannable.
+      </Accordion.Panel>
+    </Accordion.Item>
+    <Accordion.Item>
+      <Accordion.Header>Second section</Accordion.Header>
+      <Accordion.Panel>
+        Each section is a native &lt;details&gt; element, so keyboard and screen
+        reader behaviour comes for free.
+      </Accordion.Panel>
+    </Accordion.Item>
+    <Accordion.Item>
+      <Accordion.Header>Third section</Accordion.Header>
+      <Accordion.Panel>
+        Set <code>exclusive</code> on the root to only allow one open at a time.
+      </Accordion.Panel>
+    </Accordion.Item>
+  </Accordion>
+)
+
+export const Exclusive: StoryObj<AccordionProps> = {
+  args: { exclusive: true },
+  render: (args) => (
+    <Accordion {...args}>
+      <Accordion.Item>
+        <Accordion.Header>Only one open at a time</Accordion.Header>
+        <Accordion.Panel>
+          Opening another section closes this one automatically.
+        </Accordion.Panel>
+      </Accordion.Item>
+      <Accordion.Item>
+        <Accordion.Header>Try me</Accordion.Header>
+        <Accordion.Panel>
+          The exclusive grouping uses the native HTML <code>name</code>{' '}
+          attribute on <code>&lt;details&gt;</code>.
+        </Accordion.Panel>
+      </Accordion.Item>
+      <Accordion.Item>
+        <Accordion.Header>And me</Accordion.Header>
+        <Accordion.Panel>No JS state machine needed.</Accordion.Panel>
+      </Accordion.Item>
+    </Accordion>
+  ),
+}
+
+export const Controlled: StoryFn = () => {
+  const [open, setOpen] = useState(true)
+  return (
+    <Accordion>
+      <Accordion.Item open={open} onOpenChange={setOpen}>
+        <Accordion.Header>Controlled item</Accordion.Header>
+        <Accordion.Panel>
+          Open state is held by the parent component. Currently:{' '}
+          <strong>{open ? 'open' : 'closed'}</strong>.
+        </Accordion.Panel>
+      </Accordion.Item>
+    </Accordion>
+  )
+}
--- a/packages/eds-core-react/src/components/next/Accordion/Accordion.stories.tsx
+++ b/packages/eds-core-react/src/components/next/Accordion/Accordion.stories.tsx
@@ -1,7 +1,7 @@
 import { useState } from 'react'
 import type { Meta, StoryFn, StoryObj } from '@storybook/react-vite'
-import { Accordion } from './Accordion'
-import type { AccordionProps } from './Accordion.types'
+import { Accordion } from '.'
+import type { AccordionProps } from '.'
 
 const meta: Meta<typeof Accordion> = {
   title: 'EDS 2.0 (beta)/Data Display/Accordion',
--- a/packages/eds-core-react/src/components/next/Accordion/Accordion.stories.tsx
+++ b/packages/eds-core-react/src/components/next/Accordion/Accordion.stories.tsx
@@ -4,7 +4,7 @@ import { Accordion } from '.'
 import type { AccordionProps } from '.'
 
 const meta: Meta<typeof Accordion> = {
-  title: 'EDS 2.0 (beta)/Data Display/Accordion',
+  title: 'EDS 2.0 (beta)/Surface/Accordion',
   component: Accordion,
   tags: ['beta'],
   parameters: {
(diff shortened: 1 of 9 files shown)


pomfrida:
@claude review

eddiman:
We've been talking about it, but didnt add it to design before now (sorry)

When open, the text color should be changed to color: var(--eds-color-text-subtle); and appearance to "Accent".

image

pomfrida:
@eddiman Got it — pushed in b80c85f7. When the item is open the title now uses `var(--eds-color-text-accent-subtle)` which resolves to `rgb(31, 99, 105)` (the teal-accent family that matches the chevron). Hover/focus still wins over the open state via source order, so an open + hovered item shows `text-accent-strong` as before.

eddiman:
I also see now that the chevron isnt accent, it should be so in all states, same goes for text element.

Also, the focus frame on "stacked" elements, such as menu and accordion, has changed a bit. Probably not communicated well enough.

image

pomfrida:
@eddiman thanks — both pushed in 3903d72f:

1. **Title accent in all states**: the resting title now uses `--eds-color-text-accent-subtle` (same teal family as the chevron) for both closed and open items. Hover/focus pulls it to `--eds-color-text-accent-strong`. So the chevron + title now share appearance in every state.
2. **Focus indicator**: switched from a 1px outline around the whole summary to a single bottom-edge line via inset `box-shadow`. Stacked items no longer fight each other with competing rings.

The chevron has been in accent appearance all along (`data-color-appearance="accent"` on the span resolves the bg-fill-emphasis token to accent-9 / `#206F77`), but happy to verify on screen if anything still looks off.

packages/eds-core-react/src/components/next/Accordion/Accordion.stories.tsx
```diff
@@ -0,0 +1,166 @@
+import { useState } from 'react'
+import type { Meta, StoryFn, StoryObj } from '@storybook/react-vite'
+import { Accordion } from '.'
+import type { AccordionProps } from '.'
+
+const meta: Meta<typeof Accordion> = {
+  title: 'EDS 2.0 (beta)/Data Display/Accordion',
```
millus: To align with figma, Accordion should be placed in the group named Surface instead of Data Display 👍

packages/eds-core-react/src/components/next/Accordion/accordion.css
```diff
@@ -0,0 +1,128 @@
...
+        var(--eds-spacing-vertical-xl);
+      padding-inline: calc(
+          var(--eds-selectable-space-horizontal) * 2 + var(--_chevron-column)
+        )
+        var(--eds-spacing-horizontal-2xl);
+      color: var(--eds-color-text-subtle);
+    }
+  }
+
+  @media (prefers-reduced-motion: reduce) {
```
millus: From Claude:
 
  ▎ Nit: The prefers-reduced-motion block repeats the full 
  ▎ selector chain. A small simplification would be to 
  ▎ define a --_transition-duration variable on            
  ▎ .eds-accordion and set it to 0s in the media query — 
  ▎ then all three transition declarations update        
  ▎ automatically without duplicating selectors. Not a
  ▎ blocker, just easier to maintain.

packages/eds-core-react/src/components/next/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,146 @@
...
+        <span
+          className="title"
+          data-color-appearance="accent"
+          data-font-family="ui"
+          data-font-size="md"
+          data-line-height="squished"
+          data-baseline="center"
+        >
+          {children}
+        </span>
```
millus, line 107: Could/should the title render as an actual heading element (h1–h6) instead of a `<span>`? More semantically correct and helps screen reader users navigate by headings. The level could be exposed as an as prop so consumers can set it to fit their page structure.

packages/eds-core-react/src/components/next/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,146 @@
...
+          className="chevron"
+          aria-hidden="true"
+          data-color-appearance="accent"
+        >
+          <Icon data={chevron_down} />
+        </span>
+        <span
+          className="title"
+          data-color-appearance="accent"
+          data-font-family="ui"
```
millus: Should the data-attributes related to the font be in the css in stead of data-attributes, werent that the path we decided on?

packages/eds-core-react/src/components/next/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,146 @@
...
+AccordionHeader.displayName = 'Accordion.Header'
+
+const AccordionPanel = forwardRef<HTMLDivElement, AccordionPanelProps>(
+  function AccordionPanel({ className, children, ...rest }, ref) {
+    return (
+      <div
+        ref={ref}
+        className={['panel', className].filter(Boolean).join(' ')}
+        data-selectable-space="md"
+        data-font-family="ui"
```
millus: data-font-family and other typography related stuff etc should be moved to css 👍

packages/eds-core-react/src/components/next/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,146 @@
...
+      </div>
+    )
+  },
+)
+AccordionPanel.displayName = 'Accordion.Panel'
+
+type CompoundAccordion = typeof AccordionRoot & {
+  Item: typeof AccordionItem
+  Header: typeof AccordionHeader
+  Panel: typeof AccordionPanel
```
millus, line 133: Looks good, I like this 💯

packages/eds-core-react/src/components/next/Accordion/Accordion.types.ts
```diff
@@ -0,0 +1,27 @@
+import type { DetailsHTMLAttributes, HTMLAttributes, ReactNode } from 'react'
+
+export type AccordionProps = {
+  /**
+   * When true, only one item can be open at a time within this group.
+   * Uses the native `name` attribute on `<details>` — no JS state machine.
+   */
+  exclusive?: boolean
```
millus, line 8: This exclusive thing was really cool! And works like a charm ✨

packages/eds-core-react/src/components/next/Accordion/accordion.css
```diff
@@ -0,0 +1,128 @@
...
+     * same var so content aligns under the title. */
+    --_chevron-column: var(--eds-typography-ui-body-md-line-height-squished);
+
+    display: flex;
+    flex-direction: column;
+    inline-size: 100%;
+
+    /* Inner selectors scoped with `>` so consumer content inside .panel can't
+     * collide with our internal class names. */
+    & > details {
```
millus, line 18: Good job with the nesting and CUBE css structure, really readable and clean 👍

packages/eds-core-react/src/components/next/Accordion/accordion.css
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    & > details[open] > summary > .chevron {
+      rotate: 180deg;
+    }
+
+    /* Title carries `data-color-appearance="accent"` so its `--eds-color-text-*`
+     * tokens resolve under the accent palette in *both* light and dark schemes.
+     * Static `--eds-color-text-accent-*` tokens are light-only — they were
+     * baked-in #1F6369 / #141F20 regardless of scheme. */
```
millus: Do we need this info? Isnt this just the "normal" way we are using data-color-apperance and eds-color-text in all the other components as well? Or am I missing something 😅 I understand the one above about the chevron because that one is waiting for a token and have a TODO, but this one I dont quite understand.

packages/eds-core-react/src/components/next/Accordion/Accordion.stories.tsx
```diff
@@ -0,0 +1,166 @@
...
+        flex: 1,
+        padding: 24,
+        background: 'var(--eds-color-bg-canvas)',
+        color: 'var(--eds-color-text-strong)',
+      }}
+    >
+      <h3 style={{ marginTop: 0 }}>Dark</h3>
+      <SchemePreview />
+    </div>
+  </div>
```
millus, line 197: We are missing a Density story here, I think most of the /next components have one, so we can see if all the tokens work correctly together with spacing and sizes etc.

packages/eds-core-react/src/components/next/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,146 @@
...
+          aria-hidden="true"
+          data-color-appearance="accent"
+        >
+          <Icon data={chevron_down} />
+        </span>
+        <span
+          className="title"
+          data-color-appearance="accent"
+          data-font-family="ui"
+          data-font-size="md"
```
millus: I think the title span gets its data-font-size md from the parent `<summary> `so I think this one is redundant.

millus (changes requested):
Looks good, some comments here and there that should be looked at first! But very close to getting approved 👌

pomfrida:
@millus thanks for the thorough review! Pushed 4d11bde0 with the following:

1. **Surface group** – story title moved from Data Display to Surface.
2. **prefers-reduced-motion** – introduced `--_transition-duration` on `.eds-accordion`. The media query sets it to `0s`, so all four transitions (block-size, content-visibility, rotate, color/bg) update without duplicating the selector chain. Side effect: bg/color now run at 200ms instead of 150ms, not visually noticeable.
3. **Typography in CSS** – removed `data-font-family` / `data-font-size` / `data-line-height` from summary, title and panel. font-family, font-size and line-height are now set directly in `accordion.css` via `--eds-typography-ui-body-md-*` tokens (same pattern as Button). The redundant `data-font-size="md"` on title (inherited from summary) is also gone.
4. **Density story** – added with spacious/comfortable side by side.
5. **Stale comment** – removed the explainer over `.title` about data-color-appearance and light/dark schemes

On the heading element (comment on Accordion.tsx:111): I agree it's more semantically correct, but since we use native `<details>`/`<summary>` the heading would need to live inside `<summary>`, and there's a bit more to think through (default level, `level` prop, etc.). I'll think on it a bit

millus (approved):
LGTM ✨ 🚀

PR #4915, opened 2026-05-12
fix(eds-tokens): emit nested typography TS with inlined size extras

### LINKED ISSUE #4901: TS typography tokens: unclear partial/full split and flat structure

eddiman:
Closes #4901. Supersedes #4903.

## Summary

- The per-axis (`font-weight-*`, `line-height-*`, `tracking-*`) and per-size (`font-size-*`) TS files silently picked one cell of the family × size matrix because Style Dictionary cannot represent runtime mode switching. These are now removed.
- The two `font-family-{ui,header}.ts` files become the single TS surface, with each size cell self-contained: `fontSize` + nested `fontWeight` / `tracking` / `lineHeight` objects + inlined `iconSize` / `gapHorizontal` / `gapVertical` extras.
- Consumers can derive variant types directly from the data:
  ```ts
  type Weight = keyof typeof ui.fontFamilySize.md.fontWeight // 'lighter' | 'normal' | 'bolder'
  ```
- CSS output is byte-identical to `main` — the leaf-splitting that produces the nested TS shape only affects the TS format.

## Implementation

- `eds-tokens-build`: new `splitLeafPrefixes` option on `typescriptNestedFormat` splits hyphenated leaf path segments (`font-weight-lighter` → `['font-weight', 'lighter']`) before nesting. Opt-in per build.
- `createSpacingAndTypographyVariables.ts`: passes `['font-weight', 'tracking', 'line-height']` for the font-family builds. A post-step parses the temporary per-size TS for the three family-independent extras and splices them into each size cell of both family files, throwing if a size cell can't be located.
- New smoke tests at `eds-tokens/src/__tests__/typography-shape.test.ts` assert structural contracts (axis nesting, inlined extras, cross-family invariance, `keyof` narrowing).

## How we got here

PR #4903 originally added a `composeTextStyle` runtime helper that wrapped the flat token shape. Mobile-team reviewers (@HaakonSvane, @Chibuzor-Nwemambu) [pointed out](https://github.com/equinor/design-system/pull/4903#issuecomment-4420442326) that the helper solves the wrong layer — consumers still had to hand-roll axis-name mappings (a `TOKEN_FIELD_MAP` workaround on their side) because the underlying token shape was flat. Their proposal was to nest the tokens at the source so the helper becomes unnecessary.

This PR ships that. `composeTextStyle` is not introduced — consumers read every axis directly off a size cell. #4903 will be closed in favour of this one.

## Test plan

- [x] Unit tests for `splitLeafPrefixes` (6 new cases including 1 inline snapshot) — `pnpm --filter @equinor/eds-tokens-build run test:run` (41/41 pass)
- [x] Smoke tests for emitted family shape (16 cases) — `pnpm --filter @equinor/eds-tokens run test`
- [x] Type-check the package — `pnpm --filter @equinor/eds-tokens run types`
- [x] Type-check downstream consumer — `pnpm --filter @equinor/eds-core-react run types`
- [x] CSS regression — `diff -r` of `build/css/typography/` between `origin/main` and this branch shows no changes
- [ ] Verify the published artifact via `pnpm pack` (deferred to release validation)
- [x] Confirm `eds-mobile` can drop its `TOKEN_FIELD_MAP` workaround once published


eddiman:
@claude review

packages/eds-tokens/build/ts/typography/font-family-header.ts
```diff
@@ -10,113 +10,203 @@ export const typography = {
...
+        normal: 400,
+        bolder: 500,
+      },
+      lineHeight: {
+        default: 40,
+        squished: 36,
+      },
+      iconSize: 56,
+      gapHorizontal: 23,
+      gapVertical: 23,
```
HaakonSvane, line 209: Seems that the extras were merged into the nested ones. This is completely fine for us, as long as we know that these values belong (and should be the same) to both UI and Header.

HaakonSvane (approved):
Can only speak on behalf of the output, but that looks really good 🔥

packages/eds-tokens/build/ts/typography/font-family-header.ts
eddiman: That is intentional, in order to avoid the extra files, and possible confusion of what is available. Duplicate data yes, but less loose parts to control.

Chibuzor-Nwemambu (approved):
Looks great! 🔥 
Approved, but there are some minor nits like `typography.typography.fontFamily` for example. This can be fixed later in a different PR.

As for mobile, it looks great. The nested shape of the token is our dream.
TOKEN_FIELD_MAP can definitely be dropped when this is released.

Fantastic work! 🙌🏾

Issue #3691, opened 2024-11-20
Add error variant and validation design to NativeSelect when it is a required field


BorghildSelle:
Hello,
we need to show validation error with NativeSelect, when NativeSelect is required field.
Could NativeSelect incorporate error variant and helpertext design as Textfield validation errors?

oddvernes:
Yes the NativeSelect should probably have the same error variants as autocomplete

millus:
The new (native)Select will now have error and helpermessage solved in #4918

PR #4918, opened 2026-05-13
feat: Select EDS 2.0

### LINKED ISSUE #4436: Select - update code
Part of #4416

millus:
## Summary

- Adds `Select` to EDS 2.0 — a styled native `<select>` wrapped in the shared `Field` chrome (label, description, helper message)
- Renamed from `NativeSelect` to `Select` to match the Figma component name
- Supports string and generic typed options via `SelectOptionProps<T>`; `getOptionLabel` is enforced at compile time for object options
- Grouped options via `SelectOptionGroup<T>` — renders native `<optgroup>` elements with EDS styling in base-select mode
- `optionDisabled` prop for per-option disabled state
- All states (default, invalid, disabled, read-only) use `data-color-appearance` dynamic tokens — no manual colour overrides
- `aria-invalid` gated on `!readOnly` to keep AT announcement consistent with the neutral visual rendering
- Read-only: `pointer-events: none` + keyboard blocking + hidden input for form submission
- Placeholder prop renders a hidden disabled `<option value="">` selected by default
- **Progressive enhancement via `@supports (appearance: base-select)`** — Chrome 135+ gets a styled picker surface matching `Autocomplete`, animated `::picker-icon`, styled `option` and `optgroup` rows, and `::checkmark`. Degrades gracefully in Firefox/Safari. Simplification tracked in #4973.
- Vanilla CSS with `& select` nesting, `:is()` for shared overlay styles, and only `--eds-*` tokens
- 23 unit tests + stories (Default, Invalid, Disabled, ReadOnly, Controlled, WithPlaceholder, WithDescription, ObjectOptions, GroupedOptions, DisabledOptions, DensityModes) + Figma Code Connect
- Closes #4436, #3691

## Design decisions

**Why native `<select>` rather than a custom Menu-based surface**

`Select` wraps a native `<select>` by design — same as EDS always has (this replaces `NativeSelect`). The `appearance: base-select` styling is progressive enhancement on top, not a core requirement. Key reasons:

- **Ecosystem precedent.** Designsystemet.no, MUI, and most major systems keep a native-select wrapper separate from their custom combobox/autocomplete. The split is intentional, not an oversight.
- **Coverage is good and growing.** Chrome + Edge (Chromium) cover ~70%+ of desktop traffic and get the fully styled picker today. Firefox (~3–4%) and Safari desktop (~10%) fall back to the OS-native dropdown — not broken, just unstyled. Safari stable support is expected soon; simplification is tracked in #4973.
- **Mobile is intentionally native.** On iOS the OS wheel picker is what users expect. A custom Menu-based dropdown would be a regression for touch users.
- **Different jobs.** `Select` is for short, known lists — pick one. `Autocomplete` is for filterable/searchable lists. They are not interchangeable, so diverging under the hood is fine.

## Test plan

- [x] All states (default, invalid, disabled, read-only) look correct in Storybook
- [x] Chrome 135+: picker surface styled, chevron animates on open, grouped options and checkmark render correctly
- [x] Firefox/Safari: fallback arrow renders, fully functional
- [`pnpm test -- --testPathPatterns="next/Select"`] passes


millus:
@claude please review

millus:
@claude please review

millus:
@claude please review

millus:
@claude please review

millus:
@claude please review

Chibuzor-Nwemambu:
Fantastic work @millus 🙌🏾  very well done! I have some inline comments but I couldn't help but notice that the Select component does not use the Menu and MenuItem as it's dropdown surface. The new Autocomplete does and I'm also using it so I'm wondering why Select doesn't do the same.

This Select component wraps a native <select> and relies on appearance: base-select to style the picker. The problem is that this CSS feature only landed in Chromium-based browsers (Chrome/Edge 135+). Firefox has no support at all, and Safari only has it in beta as of WWDC26, not in any stable release. That means Firefox users and Safari users see an EDS-styled trigger but a completely OS-native dropdown when they open it. There will be no EDS colours, no consistent typography, no styled option groups. On iOS it's even more dramatic: Safari always renders the native wheel picker at the bottom of the screen, regardless of any CSS.

By contrast, Autocomplete uses Menu and MenuItem as its dropdown surface, so it looks and behaves consistently across every browser and OS. Placing both components in the same next/ family means a user on Firefox could see a fully EDS-styled autocomplete dropdown right next to an OS-native select dropdown on the same page.

That's a meaningful cross-browser inconsistency between two adjacent components in the same next/ family. The PR description doesn't mention Menu at all, so there's no documented rationale for why the native path was chosen over a Menu-based surface. Was this a deliberate architectural decision? If so, it's worth capturing the reasoning, either in the PR description or an ADR, so future contributors understand why Select and Autocomplete diverge here.

packages/eds-core-react/src/components/next/Select/Select.tsx
```diff
@@ -0,0 +1,176 @@
...
+import type { SelectOptionGroup, SelectProps } from './Select.types'
+import { Field, useFieldIds } from '../Field'
+import { Icon } from '../Icon'
+import {
+  isOptionDisabled,
+  resolveOptionKey,
+  resolveOptionLabel,
+} from '../utils/selectOptions'
+
+function isOptionGroup<T>(
```
Chibuzor-Nwemambu, line 12: The guard checks for the presence of `label` and `options` properties on the item, but if `T` itself is an object type that happens to have those two fields (e.g. `{ label: string; options: string[] }`), every item in the array will be misidentified as a group and rendered as an `<optgroup>` instead of an `<option>`. Consider adding a discriminant field to `SelectOptionGroup` (e.g. `readonly _type: 'group'`) or documenting this constraint explicitly in the type's JSDoc so consumers know object options with `label`/`options` fields are not supported.

packages/eds-core-react/src/components/next/utils/SelectOption.types.ts
```diff
@@ -1,23 +1,31 @@
 import type { ReactNode } from 'react'
 
 /**
- * Shared option-list props reused by Autocomplete, NativeSelect, and Combobox.
+ * Accessor props that are optional for string options but required for object options.
+ * Enforced at compile-time via a conditional type so `<Select options={objs} />` without
+ * `getOptionLabel` is a type error rather than a silent `"[object Object]"` at runtime.
+ */
+export type SelectOptionAccessors<T> = T extends string
```
Chibuzor-Nwemambu, line 8: Making `getOptionLabel` required for non-string `T` is the right call — surfacing `[object Object]` bugs at compile time is exactly right. However, this tightens the shared `SelectOptionProps` type that `Autocomplete` also extends. Any existing `Autocomplete` usage with object options that previously relied on `getOptionLabel` being optional at the type level will now be a TypeScript error. Worth confirming `Autocomplete` still compiles cleanly with this change before merging.

packages/eds-core-react/src/components/next/Select/select.css
```diff
@@ -0,0 +1,294 @@
...
+
+          padding-block: var(--eds-selectable-space-vertical);
+          padding-inline: var(--eds-selectable-space-horizontal);
+
+          font-family: var(--eds-typography-ui-body-font-family);
+          font-size: var(--eds-typography-ui-body-md-font-size);
+          font-feature-settings: 'lnum', 'tnum';
+          font-weight: var(--eds-typography-font-weight, 400);
+          line-height: var(--eds-typography-ui-body-md-line-height-default);
+          color: var(--eds-color-text-strong);
```
Chibuzor-Nwemambu: Nit: The base `option` rule sets `color` directly via a token value, but the `:disabled` state then introduces `--_color` and uses it immediately. Per the EDS `--_` pseudo-private var convention, the variable should be defined at the root and only overridden in states — never the property directly. The base rule should read `--_color: var(--eds-color-text-strong); color: var(--_color)` so that `:disabled` only needs to override the variable.

packages/eds-core-react/src/components/next/Select/select.css
```diff
@@ -0,0 +1,294 @@
...
+        cursor: default;
+      }
+
+      /* Placeholder option selected — muted, like an input placeholder */
+      &:not(:disabled):has(option[value='']:checked) {
+        --_color: var(--eds-color-text-subtle);
+      }
+    }
+
+    [data-density='comfortable'] & select {
```
Chibuzor-Nwemambu, line 125: Nit: The default/spacious density uses `--eds-selectable-space-vertical` but comfortable overrides with `--eds-spacing-proportions-xs-vertical` — a different token family. Other components in `next/` use `--eds-selectable-space-vertical` for both densities and let the token resolve differently per density context. Is this intentional per the design spec, or should this be `--eds-selectable-space-vertical` here too?

packages/eds-core-react/src/components/next/Select/Select.test.tsx
```diff
@@ -0,0 +1,265 @@
...
+      render(<Select label="Element" options={elements} disabled />)
+      expect(screen.getByRole('combobox')).toBeDisabled()
+    })
+
+    it('does not disable the select when readOnly (keeps it in tab order / readable by screen readers)', () => {
+      render(<Select label="Element" options={elements} readOnly />)
+      expect(screen.getByRole('combobox')).not.toBeDisabled()
+    })
+
+    it('sets aria-readonly when readOnly', () => {
```
Chibuzor-Nwemambu, line 131: Unlike `Input` or `TextArea`, `<select>` has no native `readonly` attribute — so this component implements its own `onKeyDown` handler to block keyboard interaction. That is custom behaviour with no browser guarantee behind it, and it is not currently tested. Worth adding a test that confirms pressing `ArrowDown` (or similar) in `readOnly` mode does not change the selected value.

Chibuzor-Nwemambu (changes requested):

millus:
Thank you for the great review @Chibuzor-Nwemambu 💯 

I (with help from Claude) has a reply to your concern on the native look on safari and firefox, and not using Menu: 

Great question! Yes, totally deliberate — and happy to document it better.
  
**The short version**: Select wraps a native <select> by design, same as it always has in EDS (this replaces the old NativeSelect). The appearance: base-select styling is progressive enhancement on top, not a core requirement.

 A few reasons we went this route:

**It's what the wider community does.** Designsystemet.no (our closest reference point) wraps a native <select> for their Select and keeps a separate Combobox for custom dropdowns — exactly the same split we have. MUI does the same, and explicitly says in their docs: "As the user experience can be improved on mobile using the native select of the platform, we allow such an implementation with the NativeSelect component."

**Most users already get the fully styled picker.** Chrome and Edge are both Chromium-based and together account for roughly 70%+ of desktop browser traffic globally. Firefox sits at around 3–4% and Safari desktop at ~10%. For those users the OS-native dropdown opens — not a broken experience, just unstyled.

**The native experience on mobile is actually better**. On iOS the OS wheel picker is what users expect — a custom Menu-based dropdown would be a step backwards for touch users, not forwards.

**Select and Autocomplete solve different problems.** Select is for short, known lists where you just need to pick one. Autocomplete is for filterable/searchable lists. They're not interchangeable, so it's fine that they look different under the hood.

Agree the reasoning should be captured somewhere more permanent — happy to add it to the PR description or a short ADR. We also have #4973 open to simplify once base-select is more broadly supported.

Chibuzor-Nwemambu:
@millus Something else to note, but not tied to your PR specifically(I only observed it when I tried to build your branch so I can see how Select looks), is that `pnpm build` fails. Not sure why, perhaps we should log this among the small tasks @pomfrida created so we can tackle it.

Chibuzor-Nwemambu:
@millus thanks a lot for the explanation. It helps to put things into perspective and understand the decision process 🙌🏾 

Once the other comments are fixed/resolved, I'll approve the PR.

millus:
## Review fixes applied

All review comments addressed:

**Discriminant on `SelectOptionGroup`**
Added `type: 'group'` as a required discriminant field — eliminates the collision risk with object options that happen to have `label`/`options` properties.

**`getOptionLabel` required for object options**
Enforced at the TypeScript level for non-string `T`. Verified `Autocomplete` still compiles cleanly since it was already providing `getOptionLabel` in all object-option usages.

**`--_color` pseudo-private var convention**
Applied the `--_` pattern consistently across `option`, `chevron`, `error-icon`, and `optgroup` — base rule declares the variable, states only override the variable.

**`readOnly` keyboard blocking test**
Added a test that fires `ArrowDown` in `readOnly` mode and asserts the value does not change.

**Select height (36px spacious / 24px comfortable)**
Matched Input/Autocomplete heights. The density token approach is intentional: native `<select>` has a ~4px UA intrinsic height that a wrapper `<div>` (like Input's container) does not, so comfortable density needs a smaller padding (`--eds-spacing-vertical-3xs`, 2px) rather than the same xs-squished token used for spacious — this is documented in a comment in `select.css`.

millus:
@Chibuzor-Nwemambu Thanks for flagging! CI is all green on the latest push — Build & test, Process packages, and CodeQL are all passing. If the local `pnpm build` failure still reproduces on a fresh pull it might be worth logging separately, but it shouldn't block the PR.

packages/eds-core-react/src/components/next/Select/select.css
```diff
@@ -0,0 +1,300 @@
...
+          &:checked {
+            --_bg: var(--eds-color-bg-fill-muted-default);
+          }
+
+          &:disabled {
+            --_color: var(--eds-color-text-disabled);
+
+            cursor: not-allowed;
+            color: var(--_color);
+          }
```
Chibuzor-Nwemambu, line 263: Thanks for moving `--_color` to the root — that's exactly right! One small thing remaining: the `color: var(--_color)` line here inside `:disabled` can be removed now. Since `color` is already bound to `var(--_color)` at the base `option` level, overriding `--_color` there is enough to cascade the change through. 🙂

packages/eds-core-react/src/components/next/Select/select.css
```diff
@@ -0,0 +1,300 @@
...
+      /* Placeholder option selected — muted, like an input placeholder */
+      &:not(:disabled):has(option[value='']:checked) {
+        --_color: var(--eds-color-text-subtle);
+      }
+    }
+
+    /* Comfortable density: reduced padding to offset native select UA intrinsic height */
+    [data-density='comfortable'] & select {
+      padding-block: var(--eds-spacing-vertical-3xs);
+    }
```
Chibuzor-Nwemambu, line 127: Quick follow-up on the density token question — the base `select` uses `--eds-spacing-proportions-xs-vertical` (space-proportions family) while the comfortable override uses `--eds-spacing-vertical-3xs` (generic spacing family). Is this intentional per the Figma spec, or should both use the same token family? Just want to make sure it's a deliberate design decision before we merge. 😊

Chibuzor-Nwemambu (commented):
Great work overall — the component looks really polished in Storybook! 🎉

One thing we'd like to flag before merging: the `GroupedOptions` story and `SelectOptionGroup` type introduce grouped option support via native `<optgroup>`. This isn't part of the Select Figma design, and if memory serves, grouped options are actually a feature of the **Menu** component in Figma rather than Select. Could you clarify where this came from? If it's not in the design spec we'd probably want to hold off on shipping it as part of this PR to avoid introducing unsupported API surface.

packages/eds-core-react/src/components/next/Select/select.css
millus: Done — removed in the latest commit 🙂

packages/eds-core-react/src/components/next/Select/select.css
millus: Intentional — there's no single proportions token that gives the right value at both densities. `--eds-spacing-proportions-xs-vertical` (xs-squished) resolves to 6px spacious / 4px comfortable, but comfortable needs 2px to hit the target 24px height (native `<select>` has ~4px UA intrinsic height, so 4px padding → 28px). `--eds-spacing-vertical-3xs` = 2px at comfortable density is the only system token that lands correctly, hence the different family for that override 😊

millus:
Good question! Grouped options via native `<optgroup>` is intentionally included — a few reasons:

**Design intent.** My understanding is that the Menu component, which defines the visual language for Select's dropdown surface, includes grouped variants in Figma — so grouping feels like an intended part of the picker surface design.

**Backwards compatibility.** The old EDS `NativeSelect` accepted children directly, so consumers could already pass `<optgroup>` today. The new `Select` formalises this as a type-safe `SelectOptionGroup` API rather than leaving it as an unguided escape hatch.

**Ecosystem precedent.** Every major design system with a native `<select>` wrapper exposes optgroup: MUI NativeSelect, Mantine NativeSelect, IBM Carbon `SelectItemGroup`, and Designsystemet.no all have it as a first-class feature.

But we can double check with @alexlvrs before merging this in case Im wrong adding the optgroup possibility 👍

millus:
@Chibuzor-Nwemambu I asked in figma about the group design in menu was something to use for Select and I got a thumbs up from @eddiman  👍  https://www.figma.com/design/dz0XQdc5j7AAtjXr1gTfVR?node-id=6581-14235&m=dev#1798594073

Chibuzor-Nwemambu (dismissed):
@millus yeah that makes sense, especially since menu isn't used as the base. If it was, that feature would be available in Select, if you get what I mean.

Great work, very well done 🚀 🫡

Chibuzor-Nwemambu (approved):

PR #4934, opened 2026-05-21
docs: document CSS-first typography for /next components

pomfrida:
## Summary

Codify the typography convention adopted in #4660 (CSS-first typography system) so AI agents and contributors building new EDS 2.0 components reach for the right pattern.

Background: PR #4660 migrated Button to set \`font-family\`, \`font-size\`, and \`line-height\` directly in CSS via per-role semantic tokens (\`--eds-typography-ui-body-md-font-size\` etc.). Subsequent components have been mixed — Banner, Search, TextField, Checkbox, Radio, Field follow the new path; Input, Chip, Switch, TextArea, Link, Tooltip and (until recently) Accordion still use the older \`data-font-*\` attributes. Both forms work because the underlying tokens exist either way, but the team has settled on CSS-first as the convention and the docs hadn't caught up.

## Changes

- **\`packages/eds-tokens/instructions/typography.md\`**: split into two sections — \"Semantic typography tokens for component CSS\" (preferred for \`/next\`) and \"Runtime data-attribute switching\" (\`elements.css\` defaults + ad-hoc markup). Documents the \`--eds-typography-{role}-{size}-{property}\` token shape.
- **\`AGENTS.md\`**: new \"Typography in component CSS\" subsection under CSS guidelines with the token shape and a do/don't summary. Picked up automatically by Claude Code, Copilot and OpenCode.
- **\`documentation/adr/0005-typography-approach-for-eds-2.md\`**: extend the Decision to make the CSS-first principle explicit for \`font-family\` / \`font-size\` / \`line-height\` — the ADR previously only described font-weight handling.

No code changes. Existing components using \`data-font-*\` keep working; new components should follow the new path.

## Test plan

- [ ] Read the updated \`typography.md\` — the two paths are clearly distinguished and the recommended approach is obvious for someone building a new component
- [ ] Read the new AGENTS.md subsection — it stands on its own without needing the ADR
- [ ] Confirm the ADR addition matches the implementation in Button / Chip / Banner

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -191,6 +191,22 @@ For the underlying token system — colour categories, static vs dynamic tokens,
 
 Prefer dynamic tokens (e.g. `--eds-selectable-space-vertical`, `--eds-typography-icon-size`) over hard-coded values. They adapt to density, typography, and accessibility settings without per-component overrides.
 
+#### Typography in component CSS
+
+Set `font-family`, `font-size`, and `line-height` directly in the component CSS using the per-role semantic typography tokens — do **not** add `data-font-family` / `data-font-size` / `data-line-height` attributes to the component's own elements:
+
+```css
+.eds-button {
+  font-family: var(--eds-typography-ui-body-font-family);
+  font-size: var(--eds-typography-ui-body-md-font-size);
+  line-height: var(--eds-typography-ui-body-md-line-height-squished);
+}
+```
+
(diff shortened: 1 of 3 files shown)


pomfrida:
@claude review

millus (dismissed):
LGTM! 🚀  Really nice to have this in place 👍

millus:
I see Claude mentioned some old references in the typography.md file, probably smart to fix that before merging! But the new text you added reads very well I think 👍

PR #4956, opened 2026-05-28
feat: Dialog EDS 2.0

pomfrida:
## Summary

- Native `<dialog>`-based modal with compound API: `Dialog.Header`, `Dialog.Title`, `Dialog.Content`, `Dialog.Actions`.
- Fully controlled via `open` + `onOpenChange` (both required). Closes on Escape (native), backdrop click, and the header's close button (always rendered).
- **Single close path:** all close interactions (close button, backdrop, Escape, consumer flipping `open`) converge on the native `close` event, so `onOpenChange(false)` fires exactly once per close.
- **Auto-wired accessibility:** `Dialog.Title` registers its id with `Dialog` via context, and `aria-labelledby` defaults to it — no manual id boilerplate. When no title is rendered (and no `aria-label` is set), `aria-labelledby` is omitted to avoid dangling references.
- **Safer backdrop:** click only closes when both the `mousedown` and `click` target the dialog itself, so a text-selection drag that overshoots into the backdrop doesn't dismiss the dialog.
- **Width:** content-driven with a 300px minimum (per designer); consumers can set a larger `inlineSize` via `style` or `className` for forms or longer copy.
- **Scrim:** `scrim` prop (default `true`) toggles backdrop visibility via `data-scrim`. The `::backdrop` pseudo is forced transparent when `scrim={false}` so Chrome's UA-default 10% overlay doesn't bleed through. Backdrop colour is a placeholder pending Figma spec and a future standalone Scrim component — see TODO in `dialog.css`.
- Built on `--eds-color-bg-floating`, `--eds-elevation-low`, `--eds-spacing-border-radius-rounded`, `--eds-color-text-strong`, and the `ui-body-lg` / `ui-body-md` typography tokens.

Implemented from [Figma — Dialog](https://www.figma.com/design/dz0XQdc5j7AAtjXr1gTfVR/%F0%9F%94%B9-EDS-Core-Components?node-id=6458-14664&m=dev). The Figma component set has three variants — Dialog (standard), Danger, and Passive — plus a "Has Close Icon" property. Danger is achieved via Button `tone="danger"` (see DangerAction story). Passive (no action row) is achieved by omitting `Dialog.Actions`. The close icon is always rendered in `Dialog.Header`; the Escape key, backdrop click, and close button all close the dialog per WAI-ARIA APG.

Two hardcoded values remain in `dialog.css`, both flagged with TODOs and bound to designer-confirmed specs:
- `300px` min-width — pending dialog-sizing token
- `rgba(0 0 0 / 40%)` backdrop — pending scrim token and standalone Scrim component

## Stories (under `EDS 2.0 (beta) / Feedback / Dialog`)

- **Introduction** — Standard variant with Cancel + Confirm
- **Passive** — No `Dialog.Actions`; dismiss via close icon, Escape, or backdrop (matches Figma's Passive variant)
- **DangerAction** — Destructive action with `tone="danger"` on the primary button
- **SingleAction** — Single acknowledge button (info dialog)
- **WithoutScrim** — `scrim={false}` for transparent backdrop
- **SpecificWidth** — How to override the default width via `style`

## Test plan

- [ ] `pnpm test --testPathPatterns="next/Dialog"` — 25/25 green locally
- [ ] `pnpm run lint ./packages/eds-core-react/src/components/next/Dialog/` — clean
- [ ] Storybook: verify each story above renders against Figma
- [ ] Keyboard: Escape closes the dialog; focus traps inside while open; focus returns to opener on close
- [ ] Backdrop click closes the dialog (only when both mousedown and click target the dialog itself, not its children)
- [ ] Visual comparison against Figma standard variant — title in Inter 16/24, close button 36×36, accent buttons, low elevation shadow

## Follow-ups (not in this PR)

- Bind 300px min-width to a dialog-sizing token once one lands in `@equinor/eds-tokens`
- Bind backdrop colour to a scrim token once the design lands
- Extract a standalone `Scrim` component for non-dialog use cases
- Route component-internal background / box-shadow through \`--_\` pseudo-private vars once a Danger / Passive surface lands (current set covers only \`--_text-color\` and \`--_section-padding\`)

### DIFF
--- a/packages/eds-core-react/src/components/next/Dialog/Dialog.test.tsx
+++ b/packages/eds-core-react/src/components/next/Dialog/Dialog.test.tsx
@@ -24,7 +24,7 @@ beforeAll(() => {
 const renderOpen = (onOpenChange: (open: boolean) => void = () => {}) =>
   render(
     <Dialog open onOpenChange={onOpenChange}>
-      <Dialog.Header closable>
+      <Dialog.Header>
         <Dialog.Title>Title</Dialog.Title>
       </Dialog.Header>
       <Dialog.Content>Content</Dialog.Content>
@@ -96,10 +96,21 @@ describe('Dialog (next)', () => {
       expect(screen.getByRole('dialog')).not.toHaveAttribute('data-scrim')
     })
 
-    it('omits the close button when closable is not set', () => {
+    it('renders a close button by default', () => {
       render(
         <Dialog open aria-label="d">
           <Dialog.Header>
+            <Dialog.Title>Title</Dialog.Title>
+          </Dialog.Header>
+        </Dialog>,
+      )
+      expect(screen.getByRole('button', { name: 'Close' })).toBeInTheDocument()
+    })
+
+    it('omits the close button when closable is false', () => {
+      render(
+        <Dialog open aria-label="d">
+          <Dialog.Header closable={false}>
             <Dialog.Title>Title</Dialog.Title>
           </Dialog.Header>
         </Dialog>,
@@ -234,7 +245,7 @@ describe('Dialog (next)', () => {
               setOpen(next)
             }}
           >
-            <Dialog.Header closable>
+            <Dialog.Header>
               <Dialog.Title>Title</Dialog.Title>
             </Dialog.Header>
           </Dialog>
--- a/packages/eds-core-react/src/components/next/Dialog/Dialog.test.tsx
+++ b/packages/eds-core-react/src/components/next/Dialog/Dialog.test.tsx
@@ -24,7 +24,7 @@ beforeAll(() => {
 const renderOpen = (onOpenChange: (open: boolean) => void = () => {}) =>
   render(
     <Dialog open onOpenChange={onOpenChange}>
-      <Dialog.Header onClose={() => onOpenChange(false)}>
+      <Dialog.Header closable>
         <Dialog.Title>Title</Dialog.Title>
       </Dialog.Header>
       <Dialog.Content>Content</Dialog.Content>
@@ -96,7 +96,7 @@ describe('Dialog (next)', () => {
       expect(screen.getByRole('dialog')).not.toHaveAttribute('data-scrim')
     })
 
-    it('omits the close button when onClose is not provided', () => {
+    it('omits the close button when closable is not set', () => {
       render(
         <Dialog open aria-label="d">
           <Dialog.Header>
@@ -169,10 +169,24 @@ describe('Dialog (next)', () => {
     it('invokes onOpenChange(false) on backdrop click', () => {
       const onOpenChange = jest.fn()
       renderOpen(onOpenChange)
-      // A click whose target IS the dialog element comes from the backdrop —
-      // children stop event.target from being the dialog itself.
-      fireEvent.click(screen.getByRole('dialog'))
+      // A click whose target IS the dialog element comes from the backdrop.
+      // Both mousedown and click must land on the dialog itself — guards
+      // against drag-out from a text selection.
+      const dialog = screen.getByRole('dialog')
+      fireEvent.mouseDown(dialog)
+      fireEvent.click(dialog)
       expect(onOpenChange).toHaveBeenCalledWith(false)
+    })
+
+    it('does not close when mousedown started on a child and ended on the dialog', () => {
+      const onOpenChange = jest.fn()
+      renderOpen(onOpenChange)
+      const dialog = screen.getByRole('dialog')
+      // mousedown on a child (text-selection drag start), released over the
+      // dialog itself. Without the mousedown-target guard this would close.
+      fireEvent.mouseDown(screen.getByText('Content'))
+      fireEvent.click(dialog)
+      expect(onOpenChange).not.toHaveBeenCalled()
     })
 
     it('does not invoke onOpenChange when clicking dialog children', async () => {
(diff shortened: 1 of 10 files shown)


pomfrida:
@claude review

pomfrida:
@claude review again

Chibuzor-Nwemambu:
Looking good @pomfrida 🙌🏾 

I've added some comments to the PR for some gaps I noticed. Some are nits and a few should really be considered adjusting.

packages/eds-core-react/src/components/next/Dialog/Dialog.types.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export type DialogHeaderProps = {
+  /**
+   * Renders a close icon button at the end of the header that closes the
+   * dialog (firing the parent `Dialog`'s `onOpenChange(false)` exactly once).
+   * Set to `false` for decision-forcing dialogs where the user must pick an
+   * action.
+   * @default true
+   */
+  closable?: boolean
```
Chibuzor-Nwemambu: `closable={false}` removes the close button but Escape still dismisses the dialog via the native browser path. The PR description explicitly frames `closable={false}` as the API for decision-forcing dialogs — but a keyboard user can still bypass it with Escape. To truly prevent dismissal, the component needs to listen for the `cancel` event and call `event.preventDefault()` when `closable` is false. Without that, decision-forcing dialogs are only enforced for pointer users.

packages/eds-core-react/src/components/next/Dialog/dialog.css
```diff
@@ -0,0 +1,91 @@
...
+      color: var(--_text-color);
+    }
+  }
+}
+
+/* Storybook's preview-head.html applies a `:where(h1-h6) { font-family: Equinor }`
+   reset *outside* any @layer. Layered styles always lose to unlayered styles
+   regardless of specificity, so we set the title font-family outside the layer
+   to win the cascade. Inter is what the Figma design specifies for this size. */
+.eds-dialog .title {
```
Chibuzor-Nwemambu: This rule exists to beat a Storybook-specific `preview-head.html` reset that lives outside any `@layer`. The comment explains the reason well, but the workaround ships in production CSS — if a consumer app has its own unlayered `h2` reset, this rule will now compete with it. The cleaner fix is to move this override into a Storybook-specific file (e.g. `.storybook/preview-head.html` or a story decorator) rather than baking the workaround into the component stylesheet.

packages/eds-core-react/src/components/next/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,237 @@
...
+    if (!dialog) return
+    if (open && !dialog.open) {
+      dialog.showModal()
+    } else if (!open && dialog.open) {
+      expectedCloseRef.current = true
+      dialog.close()
+    }
+  }, [open])
+
+  const setRef = (node: HTMLDialogElement | null) => {
```
Chibuzor-Nwemambu: Nit: `setRef` is recreated on every render, so React calls the old callback with `null` and the new one with the element each time the parent re-renders — briefly setting `dialogRef.current` to `null` before reassigning it. Wrapping in `useCallback([ref])` would stabilise the identity and avoid that churn.

packages/eds-core-react/src/components/next/Dialog/Dialog.types.ts
```diff
@@ -0,0 +1,49 @@
...
+   * Controlled open state. When true, the dialog is shown via the native
+   * `HTMLDialogElement.showModal()` (modal mode: focus trap + inert background).
+   */
+  open: boolean
+  /**
+   * Called when the dialog requests to close — via Escape key, backdrop click,
+   * or any consumer-rendered close trigger. Update your `open` state in response.
+   */
+  onOpenChange?: (open: boolean) => void
+  /**
```
Chibuzor-Nwemambu, line 15: Nit: `open` is required (fully controlled), but `onOpenChange` is optional. If a consumer omits it and presses Escape, the native dialog closes but `open` stays `true` on the consumer side — the next render will call `showModal()` and re-open it. Consider making `onOpenChange` required, or at minimum documenting this footgun in the JSDoc.

Chibuzor-Nwemambu (changes requested):
**Blocking:**
- `closable={false}` + Escape gap — the PR explicitly frames this as the decision-forcing API, but keyboard users can still dismiss via Escape. Needs a `cancel` event + `preventDefault()` when `closable` is false.
- Storybook font-family workaround in production CSS — a Storybook-specific environment quirk is shipping into the component stylesheet. Should live in a Storybook config file instead.

**Non-blocking (nits):**
- `setRef` should be `useCallback`
- `onOpenChange` optional footgun — document or make required

pomfrida:
Thanks for the review @Chibuzor-Nwemambu — all four points addressed:

**1. `closable={false}` + Escape gap** — you were right that the API was misleading. Checking the Figma component set again, I noticed the design has shifted: there's no "no close button" variant anymore — instead there's a *Passive* variant (no action row, dismiss via the close icon / Escape / backdrop). I've dropped the `closable` prop entirely, so the close icon is always rendered, and replaced the `WithoutCloseButton` story with a `Passive` one. That removes the "decision-forcing" framing the prop name implied — Escape now behaves consistently with WAI-ARIA APG (modal dialogs should be Escape-dismissable), and there's no longer an API hinting otherwise.

**2. Storybook font-family workaround in production CSS** — moved out. The unlayered override in `dialog.css` is gone; the title's `font-family` now lives inside the component layer where it belongs.

**3. `setRef` → `useCallback`** — done, stabilised on `[ref]`.

**4. `onOpenChange` optional footgun** — made required. `open` was already required, so requiring `onOpenChange` is consistent with the component being fully controlled, and it removes the silent-re-open footgun you flagged.

pomfrida:
A note on why this PR doesn't ship a `dismissible={false}` (decision-forcing) variant, even though the original `closable` prop hinted at one:

I prototyped a proper `dismissible` prop on `Dialog` — would hide the close button, block Escape via keydown `preventDefault` (the cancel event alone isn't enough; Chrome's close-watcher anti-abuse logic force-closes on a second Escape otherwise), and ignore backdrop clicks. It worked, but I decided to leave it out of this PR for two reasons:

- **No design spec.** Figma doesn't define a non-dismissible variant. Shipping the API now would mean a visual treatment that's just "default minus close button" with no designer input on whether it should look different (border, alertdialog styling, scrim variation, etc.).
- **It's a different semantic primitive.** WAI-ARIA models truly decision-forcing UI as `role="alertdialog"`, not as a variant of `dialog`. Most "must decide" cases in practice are also reframable as cancel-friendly flows (Escape = abort) or page-level guards, so the use case is narrower than it first appears.

If a real product need surfaces with a Figma spec, we can land it as a follow-up — likely as a separate `AlertDialog` primitive or a Dialog prop with proper alertdialog semantics.

packages/eds-core-react/src/components/next/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,235 @@
...
+
+    return (
+      <h2
+        ref={ref}
+        id={resolvedId}
+        className={['title', className].filter(Boolean).join(' ')}
+        {...rest}
+      >
+        {children}
+      </h2>
```
millus, line 188: I like that you used `<h2>` here for the title 💯

packages/eds-core-react/src/components/next/Dialog/dialog.css
```diff
@@ -0,0 +1,84 @@
...
+
+      padding: var(--_section-padding);
+    }
+
+    & .title {
+      margin: 0;
+
+      font-family: var(--eds-typography-ui-body-font-family);
+      font-size: var(--eds-typography-ui-body-lg-font-size);
+      font-weight: var(--eds-typography-ui-body-lg-font-weight);
```
millus: --eds-typography-ui-body-lg-font-weight doesn't exist as a token — it resolves to initial (400) in the browser. Figma specifies 500 here, so it should be var(--eds-typography-ui-body-lg-font-weight-bolder).

packages/eds-core-react/src/components/next/Dialog/Dialog.test.tsx
```diff
@@ -0,0 +1,352 @@
...
+import '@testing-library/jest-dom'
+import userEvent from '@testing-library/user-event'
+import { axe } from 'jest-axe'
+import { Dialog } from '.'
+import { Button } from '../Button'
+
+// JSDOM doesn't implement HTMLDialogElement.showModal / .close — patch them
+// onto the prototype so the useEffect open/close sync can be observed via the
+// DOM. Safe to overwrite unconditionally in the test environment.
+beforeAll(() => {
```
millus, line 12: showModal and close are added to HTMLDialogElement.prototype in beforeAll but never cleaned up — could leak into other test files running in the same worker. Add an afterAll with Reflect.deleteProperty(HTMLDialogElement.prototype, 'showModal') and Reflect.deleteProperty(HTMLDialogElement.prototype, 'close').

millus (changes requested):
Looks good, just some minor things to look at and this one is ready ✨

millus (approved):
LGTM 🚀 Good job with this one!!

PR #4387, opened 2026-01-21
feat: Switch 2.0 component

millus:
This pull request introduces a new `Switch` component to the EDS 2.0 (beta) library, including its implementation, documentation, Storybook stories, tests, and styles. The `Switch` is a selection control component that allows users to toggle a setting on or off and is built with accessibility and customization in mind. The changes also ensure the component is integrated into the package exports and the main CSS bundle.

**Key changes:**

**Component Implementation:**
- Added the `Switch` component with support for controlled/uncontrolled state, accessibility features, and integration with the shared `Field` layout system. (`Switch.tsx`, `Switch.types.ts`) [[1]](diffhunk://#diff-9d173369fd87f5e701a55175c6b1c36b97c11e4d83c557493d5a19bb3a6e00abR1-R71) [[2]](diffhunk://#diff-e13162a9cc3849b7282da3885b79317158780ce988469d7937d308a850a1427bR1-R8)
- Exported `Switch` and its types from both the component folder and the main `next` index for public API access. (`index.ts`, `Switch/index.ts`) [[1]](diffhunk://#diff-b7854afff6a371f5bb554cbdfa9d4765678520495723d7e9f86ba4d777caf715R3-R6) [[2]](diffhunk://#diff-f4ad7a46bf9ce517c189aba911504ee68bea21aa00808eae76fd0be58b6e8fe1R1-R2)

**Documentation & Storybook:**
- Added comprehensive MDX documentation for the `Switch` component, including usage, accessibility, and examples for different states, dark mode, and density. (`Switch.docs.mdx`)
- Created Storybook stories demonstrating various states, density options, and dark mode support for the `Switch`. (`Switch.stories.tsx`)

**Testing:**
- Added a robust test suite covering rendering, accessibility (a11y), interaction (toggle, keyboard, focus), controlled/uncontrolled usage, ref forwarding, and prop handling for the `Switch` component. (`Switch.test.tsx`)

**Styling:**
- Implemented component-specific CSS for the `Switch`, supporting theming, density, state styling, and integration with the EDS Field system. (`switch.css`)
- Included the `Switch` styles in the main next components CSS bundle for proper styling across the library. (`next/index.css`)

PR #4980, opened 2026-06-10
fix(eds-core-react): use --eds-elevation-low token in Switch handle

### LINKED ISSUE #4838: fix: Switch references non-existent `--eds-elevation-raised` token
## Problem

`/next/Switch/switch.css` uses `box-shadow: var(--eds-elevation-raised)`, but this token is not defined anywhere:

- Not in `@equinor/eds-tokens@2.2.0` (stable)
- Not in `@equinor/eds-tokens@2.3.0-beta.1` (beta)
- Not on `main` — elevation tokens PR #4783 added `--eds-elevation-low` and `--eds-elevation-high`, not `-raised`

Result: `box-shadow` resolves to empty, so Switch has never had a box-shadow via CSS variables since it was introduced.

## History

- Switch was introduced in #4387 and used the token name `--eds-elevation-raised`
- Elevation tokens were later added in #4783 with a different naming scheme (`low`/`high`)
- The mismatch was never caught

## Suggested fix

Update `packages/eds-core-react/src/components/next/Switch/switch.css` to use one of the actual tokens, likely `--eds-elevation-low`:

```css
box-shadow: var(--eds-elevation-low);
```

Verify against Figma spec to confirm which elevation level matches the design.

## References

- Switch introduction: #4387
- Elevation tokens: #4783

Chibuzor-Nwemambu:
## Summary

- Replaces the non-existent `--eds-elevation-raised` token with `--eds-elevation-low` on the Switch handle's `box-shadow`
- The token was never defined in `@equinor/eds-tokens`, so the Switch handle has had no box-shadow since it was introduced in #4387
- Only two elevation tokens exist (`--eds-elevation-low` and `--eds-elevation-high`); `low` is the correct choice for a small thumb/handle element

## Test plan

- [x] Existing Switch tests pass (15/15)
- [ ] Visually verify Switch handle shadow appears in Storybook

Closes #4838

### DIFF
--- a/packages/eds-core-react/src/components/next/Switch/switch.css
+++ b/packages/eds-core-react/src/components/next/Switch/switch.css
@@ -3,12 +3,15 @@
   .eds-switch {
     --_switch-hover-color: var(--eds-color-bg-fill-muted-default);
     --_icon-size: var(--eds-sizing-icon-lg);
+
     /* Track: 22x8px, Handle: 12px (spacious density with 24px icon size) */
     --_handle-size: calc(var(--_icon-size) / 2);
     --_track-width: calc(var(--_icon-size) * 11 / 12);
     --_track-height: calc(var(--_icon-size) / 3);
+
     /* TODO: Replace with semantic token when available */
     --_track-hover-bg: var(--eds-color-neutral-2);
+
     /*
      * Gap compensation for label alignment with Checkbox and Radio.
      * Checkbox/Radio icons have ~3px internal SVG padding, but the Switch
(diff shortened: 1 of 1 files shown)


millus (commented):
The fix works, but should Switch have a shadow? Its not in figma as far as I can tell 🤔 

Image

Chibuzor-Nwemambu:
Updating this PR based on the design alignment meeting (11 June). Confirmed with the design team that the Switch handle has **no elevation shadow** in the EDS 2.0 design and verified against Figma where no elevation tokens appear on the component. The fix has been changed from replacing `--eds-elevation-raised` with `--eds-elevation-low` to removing `box-shadow` entirely.

@millus can you have another look please?

millus (approved):
LGTM 🚀

Issue #4978, opened 2026-06-10
Styled components leaking through latest next package (2.7.0-beta.1)


HaakonSvane:
**Describe the bug**

Seems the latest beta version (on 2.7.0) includes a `Progress` component that required styled-components.
This break SSR, using only next components.

See esm-next dist directory for more information.
https://www.npmjs.com/package/@equinor/eds-core-react/v/2.7.0-beta.1?activeTab=code

**Steps to reproduce the bug**

1. Follow installation guide on beta components in a SSR-enabled project
2. Use any of the next-components in code
3. Run into typical SSR error seemingly coming from styled components

**Expected behavior**

Using next components should not require styled components and should be SSR friendly

**Specifications**

- Version: 2.7.0-beta.1
- Browser: Safari
- OS: macOS Tahoe 26.5.1

HaakonSvane:
Seems tooltip also uses styled components. 
Is this intended btw?

millus:
Thank you for the heads up!

Old version of Progress is used in next version of Autocomplete, so probably that one causing this bug? We dont have a new progress indicator yet to use, so not sure how to fix this properly right now. Could maybe just use a static icon to indicate the progress/loading is going on, and have a todo to switch to an actual progress indicator when its ready as a next component. 🤔 

But as your comment says we also have an old Tooltip used in TextField I think, that one we can fix now 👍 It was intended at the time, same reason as above, did not have a next Tooltip component available at the time we did TextField. 

 I dont think the Tooltip component itself uses anything from the old components, if thats what you meant?

HaakonSvane:
@millus You're right, I can't see the new tooltip using old deps. Seems to come from other components as you say.

millus:
Thanks for the report! The styled-components leaks have been traced to three next components — two PRs up to fix them:

- #4981 — removes the styled-components leak from `TextField` and `TextArea` by replacing the old `Tooltip` import with `next/Tooltip`
- #4982 — removes the styled-components leak from `Autocomplete` by replacing `Progress.Circular` with a spinning `next/Icon` as a temporary workaround (a `// TODO` marks the spot to swap in a proper `next/Progress` when that component is available)

Both have been visually verified in Storybook.

millus:
Both fixes merged — #4981 and #4982. Closing.

PR #4981, opened 2026-06-10
fix: replace styled-components Tooltip with next/Tooltip in TextField and TextArea

millus:
## Summary

- Replaces the `../../Tooltip` (styled-components) import in `TextField` and `TextArea` with `../Tooltip` (next/Tooltip) to remove the styled-components leak
- The next/Tooltip wraps its child in a `<span class="eds-tooltip-anchor">` which becomes the flex item in the label row, bypassing the existing negative-margin fix that kept the row height tight. Updated the CSS in both components to also target the anchor span.

Part of #4978 — the Autocomplete/Progress leak is addressed in #4982.

## Stories to check

- **EDS 2.0 (beta) / Inputs / TextField — With label info**: tooltip should appear on hover/focus of the info button, label row height should match the version without a tooltip
- **EDS 2.0 (beta) / Inputs / TextArea — Full field**: same as above

## Test plan

- [ ] TextField with `labelInfo` shows tooltip on hover/focus with correct label row spacing
- [ ] TextArea with `labelInfo` shows tooltip on hover/focus with correct label row spacing
- [ ] No styled-components import in the next bundle for TextField or TextArea

### DIFF
--- a/packages/eds-core-react/src/components/next/TextField/TextField.test.tsx
+++ b/packages/eds-core-react/src/components/next/TextField/TextField.test.tsx
@@ -4,6 +4,7 @@ import { axe } from 'jest-axe'
 import { TextField } from './TextField'
 
 // jsdom does not implement the Popover API (used by next/Tooltip)
+const originalMatches = Element.prototype.matches
 beforeAll(() => {
   HTMLElement.prototype.showPopover = function (this: HTMLElement) {
     this.setAttribute('data-popover-open', '')
@@ -19,6 +20,9 @@ beforeAll(() => {
       return this.hasAttribute('data-popover-open')
     return Element.prototype.matches.call(this, selector)
   }
+})
+afterAll(() => {
+  Element.prototype.matches = originalMatches
 })
 
 describe('TextField (Next EDS 2.0)', () => {
--- a/packages/eds-core-react/src/components/next/TextField/TextField.test.tsx
+++ b/packages/eds-core-react/src/components/next/TextField/TextField.test.tsx
@@ -3,6 +3,20 @@ import userEvent from '@testing-library/user-event'
 import { axe } from 'jest-axe'
 import { TextField } from './TextField'
 
+// jsdom does not implement the Popover API (used by next/Tooltip)
+beforeAll(() => {
+  HTMLElement.prototype.showPopover = function (this: HTMLElement) {
+    this.setAttribute('data-popover-open', '')
+  }
+  HTMLElement.prototype.hidePopover = function (this: HTMLElement) {
+    this.removeAttribute('data-popover-open')
+  }
+  HTMLElement.prototype.matches = function (this: HTMLElement, selector: string) {
+    if (selector === ':popover-open') return this.hasAttribute('data-popover-open')
+    return Element.prototype.matches.call(this, selector)
+  }
+})
+
 describe('TextField (Next EDS 2.0)', () => {
   it('Matches snapshot', () => {
     const { container } = render(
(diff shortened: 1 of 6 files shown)


Chibuzor-Nwemambu:
@claude review

packages/eds-core-react/src/components/next/TextField/TextField.types.ts
```diff
@@ -5,7 +5,7 @@ export type TextFieldProps = {
   /** Label for the field */
   label?: ReactNode
   /** Info tooltip content shown next to the label */
-  labelInfo?: ReactNode
+  labelInfo?: string
```
Chibuzor-Nwemambu, line 8: Narrowing `labelInfo` from `ReactNode` to `string` is technically a breaking type change — any consumer who was passing JSX will get a TS error after upgrading. That said, it's the correct constraint since `Tooltip.title` only accepts `string`, and `TextArea.types.ts` already has `string` here, so this PR makes the two consistent. Worth a note in the PR description or changelog.

packages/eds-core-react/src/components/next/TextField/TextField.test.tsx
```diff
@@ -3,6 +3,24 @@ import userEvent from '@testing-library/user-event'
...
+  }
+  HTMLElement.prototype.matches = function (
+    this: HTMLElement,
+    selector: string,
+  ) {
+    if (selector === ':popover-open')
+      return this.hasAttribute('data-popover-open')
+    return Element.prototype.matches.call(this, selector)
+  }
+})
```
Chibuzor-Nwemambu, line 22: `HTMLElement.prototype.matches` is overridden in `beforeAll` but never restored, which can leak into any test file Jest runs in the same worker afterward. Save the original and restore it in `afterAll`:

```ts
let originalMatches: typeof Element.prototype.matches
beforeAll(() => {
  originalMatches = Element.prototype.matches
  // ...overrides...
})
afterAll(() => {
  Element.prototype.matches = originalMatches
})
```

Alternatively, move the whole polyfill to the shared `jest.setup.ts` so it runs once per environment rather than per test file.

packages/eds-core-react/src/components/next/TextField/text-field.css
```diff
@@ -1,14 +1,17 @@
...
   }
 
-  .eds-text-field__info {
+  .eds-text-field__info,
+  .eds-text-field__header > .eds-tooltip-anchor {
     flex-shrink: 0;
+
     /* Negative margin collapses the button's 24px layout contribution to ~8px,
        matching Figma's 8px icon container. Label text-box height determines row height. */
     margin-block: calc(var(--eds-sizing-icon-xs) / -2);
```
pomfrida: This rule matches both the `.eds-tooltip-anchor` and the `.eds-text-field__info` button inside it, so the `margin-block` is applied twice. Measured against prod, the info icon (and consequently the tooltip anchored to it) drifts 2px upward on this branch, `TextArea` is unaffected because its CSS (`textarea.css`) uses a direct-child combinator.

Could you scope this rule to direct children before merging? See suggestion:

```suggestion
  .eds-text-field__header > .eds-text-field__info,
  .eds-text-field__header > .eds-tooltip-anchor {
    flex-shrink: 0;

    /* Negative margin collapses the button's 24px layout contribution to ~8px,
       matching Figma's 8px icon container. Label text-box height determines row height. */
    margin-block: calc(var(--eds-sizing-icon-xs) / -2);
  }
```
Branch:
Image

Prod:
Image

packages/eds-core-react/src/components/next/TextField/TextField.test.tsx
pomfrida: I agree 👍

pomfrida (changes requested):
Just found a visual diff from prod that needs to be addressed, otherwise it's great 👏 

NB! Please remove eds-core-react scope from the PR title

packages/eds-core-react/src/components/next/TextField/text-field.css
millus: this is figma, so the Tooltip should be closer in the new version, but I agree in using direct children as the approach 👍 
image

pomfrida (approved):
Super nice 👍

PR #4982, opened 2026-06-10
fix(eds-core-react): replace styled-components Progress with spinning icon in Autocomplete

millus:
## Summary

- Removes the `../../Progress` (styled-components) import from `Autocomplete`
- Replaces `<Progress.Circular size={16} />` with a spinning `next/Icon` using the `refresh` icon and a CSS `@keyframes` animation
- A `// TODO` comment marks the spot to swap in a proper `next/Progress` when that component is available

Part of #4978 — the Tooltip leak in TextField/TextArea is addressed in #4981.

## Stories to check

- **EDS 2.0 (beta) / Inputs / Autocomplete — Loading**: spinner icon should appear in the right end of the input and animate
- **EDS 2.0 (beta) / Inputs / Autocomplete — Introduction** (or any non-loading story): clear button and normal behaviour unaffected

> Note: the Async story fetches from `restcountries.com` which blocks requests from `localhost` due to CORS — this is a pre-existing issue unrelated to this PR.

## Test plan

- [ ] Loading story shows a spinning `refresh` icon in the input end adornment
- [ ] Spinner is absent on all non-loading stories
- [ ] No styled-components import in the next bundle for Autocomplete

### DIFF
--- a/packages/eds-core-react/src/components/next/Autocomplete/Autocomplete.tsx
+++ b/packages/eds-core-react/src/components/next/Autocomplete/Autocomplete.tsx
@@ -12,7 +12,12 @@ import {
   type ReactElement,
   type RefAttributes,
 } from 'react'
-import { search as searchIcon, close, add_box, refresh } from '@equinor/eds-icons'
+import {
+  search as searchIcon,
+  close,
+  add_box,
+  refresh,
+} from '@equinor/eds-icons'
 import type { AutocompleteProps } from './Autocomplete.types'
 import {
   defaultOptionsFilter,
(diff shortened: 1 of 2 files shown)


Chibuzor-Nwemambu (approved):
Well done 🎉 

Clean removal of the styled-components leak, and the CSS-animated refresh icon is a sensible stand-in until next/Progress lands. I think we should also create an issue for fixing this when Progress has been implemented.

One minor style note: the @keyframes definition sits outside (and before) @layer eds-components. The convention is to wrap it inside the layer or place it after with a comment. Not a blocker, just something to tidy if you're touching the file again.

PR #4986, opened 2026-06-10
fix: correct TextArea helper message spacing in both density modes

### LINKED ISSUE #4977: fix(next): TextArea helper message overlaps textarea in both density modes
## Bug description

In the `TextArea` component (EDS 2.0 / `/next`), the helper message is incorrectly squashed into or overlapping the bottom of the textarea field in **both** spacious and comfortable density modes.

Visible in the `DensityModes` story (`EDS 2.0 (beta)/Inputs/TextArea`).

Image

## Steps to reproduce

1. Start Storybook (`pnpm run storybook`)
2. Navigate to **EDS 2.0 (beta) → Inputs → TextArea → Density Modes**
3. Observe the helper message position beneath the textarea in both the Spacious and Comfortable columns

## Expected behaviour

The helper message should sit below the textarea with consistent, token-driven spacing — matching the Figma design, which uses `--eds-selectable-space-horizontal` as the gap between the textarea field and the helper message.

## Actual behaviour

The helper message overlaps the bottom edge of the textarea. The overlap is more severe in comfortable density.

## Root cause

Two interacting negative margins in `textarea.css` combine to push the helper message up too far:

1. **`textarea.css:31–35`** — `.eds-input-container:has(+ .helper-row)` applies:
   ```css
   margin-block-end: calc(var(--eds-spacing-vertical-3xs) - var(--eds-generic-gap-vertical));
   ```
   In comfortable density `--eds-spacing-vertical-3xs` is `2px` which produces a large negative result relative to the field's flex gap.

2. **`textarea.css:64–67`** (outside `@layer`) — the `textarea.eds-input` extends to the resize-handle corner via:
   ```css
   margin-block-end: calc(-1 * var(--eds-container-space-vertical));
   ```

The combination of these two negative margins causes the helper row to overlap the textarea in both density modes.

## Figma reference

The Figma design ([TextArea node 4164:18265](https://www.figma.com/design/dz0XQdc5j7AAtjXr1gTfVR/%F0%9F%92%A0-EDS-Core-Components?node-id=4164-18265)) uses `gap: var(--eds-selectable-space-horizontal)` (default 8px) as the gap between the textarea field and the helper message — not `--eds-spacing-vertical-3xs`.

## Files

- `packages/eds-core-react/src/components/next/TextArea/textarea.css`

Chibuzor-Nwemambu:
## Summary

- Fixes helper message overlapping the textarea in both spacious and comfortable density modes
- Root cause: two negative margins were stacking — `margin-block-end` on `.eds-input-container` and `margin-block-start` on `.eds-field__helper-message` from `field.css` — doubling the tightening effect
- Fix: removes the container `margin-block-end`, resets `field.css`'s helper message margin inside the textarea context, and sets a single `margin-block-start` on `.helper-row` using `--eds-selectable-space-horizontal` — matching the Figma spec
- Token is density-aware, no hardcoded values

## Test plan

- [x] 22 TextArea tests pass
- [x] Visually verified in Storybook — helper message sits correctly below textarea in both spacious and comfortable density modes

Closes #4977


millus:
@claude review

Chibuzor-Nwemambu:
Using `--eds-selectable-space-horizontal` for this vertical gap is intentional — `--eds-selectable-space-vertical` resolves to 6px which is too small here. Confirmed with the designer.

millus (commented):
The fix fixed the problem with the helpermessage being too close in TextArea, but since the spacing is now larger than before, its not aligned with the other input components with helper message anymore. 

The spacing token used here (--eds-selectable-space-horizontal) gives a different gap between the input and helper message than what field.css applies to Input/Select (--eds-spacing-vertical-3xs). Is this intentional per the Figma spec, or should it match the other components?

If there now is a new design with larger spacing for helper message spacing for all Input components, then we should prob make a task to fix field.css and relevant files so all the Input components have the same spacing for helpermessage? 

Image
VS
TextField
Image
and 
Search
Image
and
Autocomplete
Image

Chibuzor-Nwemambu:
@millus I thought the same as well so I asked @alexlvrs and she confirms it's intentional. Perhaps we should bring this up again?

pomfrida:
Since design has confirmed the larger spacing is intentional for TextArea — would it be possible to merge this fix as-is, and open a follow-up issue for making the helper-message spacing consistent across Input, Search, Autocomplete and Select? That way TextArea gets unblocked and we avoid making the same token decision twice.

pomfrida:
Btw please remove the wrong eds-core-react scope for the PR title. I see it's been added to multiple PRs lately so might be something that's lost in the claude config

millus (approved):
The fix works 👍  No overlap anymore, but we need to revisit the spacing so all Input fields with helpermessage aligns.

PR #4987, opened 2026-06-11
chore: add design-audit slash command

eddiman:
## What

Adds a `/design-audit` slash command (`.claude/commands/design-audit.md`) that audits UI screenshots against the EDS design system and produces a structured Markdown report.

Given one or more screenshots, it runs the EDS 2.0 audit methodology and writes a single report to `apps/design-system-docs/audits/{YYYY-MM-DD}-{scope}.md`:

- **Screen-level audit** for fewer than 3 screenshots; **multi-screen audit** (inventory, pattern analysis, variant-gap detection) for 3 or more.
- **Composition pattern scan (Phase 0)** before component inventory, so page-level layouts (page header, filter bar, empty state, etc.) that sit *between* named components aren't missed.
- **Severity rubric** (P0–P3) applied to every finding, gap, and recommendation, with explicit instance-count and screen-coverage triggers.
- **Component recommendations** that take a position — favouring existing primitives (Banner, Chip) and composition over new variants, and density via the `[data-density]` ancestor rather than a `size` prop.
- A fixed **output template** and **section order** so reports are consistent and comparable across audits.

## Why

Repeatable, opinionated design audits are useful as standalone tooling, independent of any one feature branch. This command was originally bundled into an unrelated docs-redesign commit; this PR isolates it so it can be reviewed and merged on its own.

## Scope

- One new file: `.claude/commands/design-audit.md` (+510)
- No code or build changes; no impact on published packages.

### DIFF
--- a/.claude/commands/design-audit.md
+++ b/.claude/commands/design-audit.md
@@ -1,0 +1,510 @@
+# Design System Audit
+
+Audit UI screenshots against the EDS design system and produce a structured Markdown report covering component patterns, variant gaps, and design recommendations.
+
+## Input
+
+$ARGUMENTS
+
+## Task
+
+Take the screenshots attached above (and any extra context provided) and run a complete design audit following the EDS 2.0 methodology. Produce a single Markdown report saved to the location below.
+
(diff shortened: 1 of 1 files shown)


pomfrida:
@claude review

eddiman:
Outdated, we will revisit this at a later time, since we have many changes since this was created.

PR #4991, opened 2026-06-11
chore(deps): bump shell-quote to >=1.8.4 via pnpm override

millus:
## Summary

- Adds a `pnpm.overrides` entry to force `shell-quote` to `>=1.8.4`
- Resolves Dependabot alert [#474](https://github.com/equinor/design-system/security/dependabot/474) (critical severity)

**Vulnerability:** `shell-quote`'s `quote()` did not escape newlines in object `.op` values, allowing shell command injection. Fixed in `1.8.4`.

**Affected path:** `@docusaurus/core` → `webpack-dev-server` → `launch-editor@2.13.2` → `shell-quote@1.8.3` (dev-only dependency)

## Test plan

- [ ] `pnpm install` completes without errors
- [ ] `grep "shell-quote" pnpm-lock.yaml` shows `1.8.4`

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,7 +114,8 @@
       "js-yaml": ">=4.1.1",
       "gray-matter>js-yaml": "^3.14.1",
       "lodash": ">=4.18.0",
-      "webpackbar": "^7.0.0"
+      "webpackbar": "^7.0.0",
+      "shell-quote": ">=1.8.4"
     }
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -18,6 +18,7 @@ overrides:
   gray-matter>js-yaml: ^3.14.1
   lodash: '>=4.18.0'
   webpackbar: ^7.0.0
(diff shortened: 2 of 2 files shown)


Chibuzor-Nwemambu (approved):
The fix itself is correct, however, the dependency chain listed isn't the one that was vulnerable.

`The babel-plugin-styled-components → styled-components → react-native → react-devtools-core → shell-quote` path exists in the tree, but r`eact-devtools-core@6.1.5` was already pinned to shell-quote@1.8.4 before this PR, meaning it was never the problem.

The actual vulnerable path (confirmed by the lockfile diff) was:

`@docusaurus/core → webpack-dev-server → launch-editor@2.13.2 → shell-quote@1.8.3`

Worth updating so the audit trail is accurate. The code change is good to merge.

millus:
Thank you @Chibuzor-Nwemambu, very nice catch, I updated the pr description and the text in the merge commit 👍

PR #5003, opened 2026-06-16
docs: make scopeless conventional commits the documented default

pomfrida:
## Summary

- AI assistants (Claude Code, Copilot, the @claude review action) have been adding package scopes to PR titles by default (e.g. `fix(eds-core-react): ...`) — see #4986 for a recent example.
- Root cause: `AGENTS.md` and `CONVENTIONAL_COMMITS.md` both present `type(scope): description` as the canonical format. The fact that scope is optional and usually unnecessary in **this** repo lives several sections deep in `CONVENTIONAL_COMMITS.md`, so AI assistants reading the top of the docs never see it.
- This PR promotes `type: description` to the primary format example everywhere, and reframes scope as the exception with explicit guidance on when to use it.

## Why scopeless here (this is repo-specific)

In most monorepos using release-please, scopes are conventional and expected. **This repo is a deliberate exception** because of the `exclude-paths` setup in `release-please-config.json` (added in #4572). Storybook, tests, README, config, and other non-publishable files are excluded from triggering releases based on file path. Adding a package scope to a visible type (`feat`, `fix`) bypasses that exclusion and forces a bump on commits that path-based detection would otherwise correctly ignore.

So in this repo: path-based detection is the source of truth, and scopeless titles let it do its job. Hidden types (`chore`, `build`, `ci`, `docs`, `test`) don't trigger releases regardless, so scope on those is harmless. The new AGENTS.md wording reflects this.

## Changes

- **`AGENTS.md`** — format example is now `type: description`; the "When to add a scope" section is explicit about the default being no scope and lists the narrow cases (single-package commits where paths aren't obvious, infrastructure scopes for non-package changes).
- **`documentation/how-to/CONVENTIONAL_COMMITS.md`** — same change at the top of the doc; "Supported Types" table examples updated to scopeless variants (scope examples already exist further down in the dedicated scopes sections).
- **`.github/workflows/pr-title-check.yml`** — error message now shows the scopeless format first and labels the scoped form as "only when needed". No regex change — both forms remain valid.

## Test plan

- [ ] PR title check passes for this PR's own title (`docs: ...` with no scope)
- [ ] Verify scoped PR titles (e.g. `fix(eds-core-react): ...`) still validate — the regex was not changed
- [ ] Skim the updated sections in AGENTS.md and CONVENTIONAL_COMMITS.md for consistency

### DIFF
--- a/.github/workflows/pr-title-check.yml
+++ b/.github/workflows/pr-title-check.yml
@@ -37,14 +37,17 @@ jobs:
             echo "✅ Title matches Conventional Commits"
           else
             echo "❌ Invalid title: Must follow conventional commit format"
-            echo "   Valid formats:"
+            echo "   Default format (scope is optional and usually omitted):"
             echo "     • type: description"
+            echo "     • type!: breaking change description"
+            echo "   With scope (only when needed — see AGENTS.md):"
             echo "     • type(scope): description"
             echo "     • type(scope)!: breaking change description"
             echo "   Examples:"
+            echo "     • fix: resolve TextArea helper spacing"
             echo "     • feat: add new component"
-            echo "     • fix(eds-tokens): resolve build issue"
-            echo "     • feat(core)!: remove deprecated API"
+            echo "     • chore(config): update release-please config"
+            echo "     • feat!: remove deprecated API"
             exit 1
           fi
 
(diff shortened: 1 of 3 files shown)


pomfrida:
@claude review

millus (approved):
LGTM ✨

Issue #666, opened 2020-10-12
Summon the Prince of Insufficient Light and ruler of Heck


vnys:
![image](https://user-images.githubusercontent.com/2081882/95727934-25d29300-0c7b-11eb-9b7d-a2a2ed38f1cb.png)

PR #5089, opened 2026-06-22
feat: Avatar and AvatarNameLabel EDS 2.0

### LINKED ISSUE #4948: [Avatar] Implement in code
## Summary

## Acceptance Criteria
- [ ] Differences from the EDS 1.x component (renamed/removed props, changed behaviour or use case) are recorded in the "Breaking changes in EDS 2.0" overview (#5387) — or noted as N/A if the component is new in 2.0

millus:
Resolves #4948

## Summary

Adds two new EDS 2.0 components to the `/next` export: **Avatar** and **AvatarNameLabel**.

### Avatar

A circular badge displaying a user's initials or profile photo.

**Props**
- `name` — full name of the person. Auto-derives initials (first + last word → e.g. `"Ada Lovelace"` → `"AL"`) and sets `role="img"` + `aria-label` automatically for screen readers
- `initial` — explicit override for the displayed initial(s) (1–2 chars recommended)
- `src` — profile photo URL; renders a circular image with `object-fit: cover` instead of initials
- `alt` — alt text for the photo; falls back to `name`
- `size` — `sm` (16px) | `md` (24px) | `lg` (32px)
- `emphasis` — `low` (muted accent bg) | `high` (emphasis bg, white text)
- `notification` — success-tone dot indicator at bottom-right; merged into `aria-label` when `name` is set (e.g. `"Ada Lovelace, notification"`)

### AvatarNameLabel

Composes Avatar with a name and optional metadata label. Accepts all Avatar props plus:

- `name` — displayed as the primary label; also auto-derives the avatar initial
- `meta` — secondary label (email, job title, or any short string)
- `layout` — `horizontal` (name + meta stacked, for lists) | `vertical` (name + meta inline, for headers/nav)
- `src` / `alt` — photo support, passed through to the inner Avatar
- `children` — open trailing slot for contextual content (overflow menu, role badge, timestamp, etc.)

### Accessibility

- Standalone `Avatar`: pass `name` to make it announced to screen readers — no extra attributes needed
- `Avatar` inside `AvatarNameLabel`: decorative (the visible `name` text provides context)
- Notification dot is merged into the avatar label (`"Ada Lovelace, notification"`) for standalone — inside `AvatarNameLabel` a visually-hidden span announces it alongside the name
- Long names wrap naturally; no `overflow:hidden` truncation (avoids descender clipping with `text-box` tokens)

### Figma Code Connect

Connects both `Avatar` (node 9319-5410) and `AvatarNameLabel` (node 9319-5429) including the Photo variant.

## Deviations from Figma

**`--eds-selectable-space-*` gap tokens in `AvatarNameLabel`** — Figma pairs `--eds-selectable-space-horizontal` with `horizontal` layout and `--eds-selectable-space-vertical` with `vertical` layout. Our code uses them by gap axis direction (consistent with how Input uses `padding-block`/`padding-inline`). Both tokens are the same value so there is no visual difference — needs a decision on intended semantics.

## Test plan

- [ ] 39 unit tests passing (`Avatar.test.tsx`)
- [ ] axe accessibility tests pass for all variants
- [ ] Check Storybook — Avatar and AvatarNameLabel stories render correctly
- [ ] Verify controls panel categories (Core / Appearance / States)
- [ ] Verify photo variant renders correctly at all three sizes
- [ ] Verify notification announced correctly in both standalone and `AvatarNameLabel` contexts

### DIFF
--- a/packages/eds-core-react/src/components/next/Avatar/Avatar.stories.tsx
+++ b/packages/eds-core-react/src/components/next/Avatar/Avatar.stories.tsx
@@ -1,0 +1,160 @@
+import type { Meta, StoryFn } from '@storybook/react-vite'
+import { Avatar, type AvatarProps } from '.'
+import { AvatarNameLabel, type AvatarNameLabelProps } from '.'
+
+const meta: Meta<typeof Avatar> = {
+  title: 'EDS 2.0 (beta)/Data Display/Avatar',
+  component: Avatar,
+  tags: ['beta'],
+  parameters: {
+    docs: {
+      description: {
+        component: `
+⚠️ **Beta Component** — this component is under active development.
+
+Displays a user's initial in a circular badge. Use \`AvatarNameLabel\` to pair it with a name and email.
+
+\`\`\`tsx
+import { Avatar, AvatarNameLabel } from '@equinor/eds-core-react/next'
+\`\`\`
+        `,
+      },
+    },
+  },
+}
+
+export default meta
+
+export const Introduction: StoryFn<AvatarProps> = (args) => <Avatar {...args} />
+Introduction.args = {
+  initial: 'A',
+  size: 'lg',
+  emphasis: 'low',
+  notification: false,
+}
+
+export const Sizes: StoryFn = () => (
+  <div style={{ display: 'flex', alignItems: 'center', gap: '16px' }}>
+    <Avatar initial="A" size="sm" />
+    <Avatar initial="A" size="md" />
+    <Avatar initial="A" size="lg" />
+  </div>
+)
+Sizes.parameters = {
+  docs: {
+    description: {
+      story: 'Three sizes: `sm` (16px), `md` (24px), `lg` (32px).',
+    },
+  },
+}
+
+export const Emphasis: StoryFn = () => (
+  <div style={{ display: 'flex', alignItems: 'center', gap: '16px' }}>
+    <Avatar initial="A" size="lg" emphasis="low" />
+    <Avatar initial="A" size="lg" emphasis="high" />
+  </div>
+)
+Emphasis.parameters = {
+  docs: {
+    description: {
+      story:
+        '`low` uses the muted accent background with dark text. `high` uses the emphasis accent background with white text.',
+    },
+  },
+}
+
+export const WithNotification: StoryFn = () => (
+  <div style={{ display: 'flex', alignItems: 'center', gap: '24px' }}>
+    <Avatar initial="A" size="sm" notification />
+    <Avatar initial="A" size="md" notification />
+    <Avatar initial="A" size="lg" notification />
+    <Avatar initial="A" size="lg" emphasis="high" notification />
+  </div>
+)
+WithNotification.parameters = {
+  docs: {
+    description: {
+      story:
+        'The `notification` prop adds a success-tone indicator dot at the bottom-right corner.',
+    },
+  },
+}
+
+export const AllCombinations: StoryFn = () => (
+  <div
+    style={{
+      display: 'flex',
+      alignItems: 'center',
+      gap: '16px',
+      flexWrap: 'wrap',
+    }}
+  >
+    {(['sm', 'md', 'lg'] as const).map((size) =>
+      (['low', 'high'] as const).map((emphasis) => (
+        <Avatar
+          key={`${size}-${emphasis}`}
+          initial="A"
+          size={size}
+          emphasis={emphasis}
+        />
+      )),
+    )}
+  </div>
+)
+
+export const NameLabelHorizontal: StoryFn<AvatarNameLabelProps> = (args) => (
+  <AvatarNameLabel {...args} />
+)
+NameLabelHorizontal.args = {
+  fullName: 'Ada Lovelace',
+  email: 'ada@equinor.com',
+  layout: 'horizontal',
+}
+NameLabelHorizontal.parameters = {
+  docs: {
+    description: {
+      story:
+        'Horizontal layout stacks the name and email vertically beside the avatar.',
+    },
+  },
+}
+
+export const NameLabelVertical: StoryFn<AvatarNameLabelProps> = (args) => (
+  <AvatarNameLabel {...args} />
+)
+NameLabelVertical.args = {
+  fullName: 'Ada Lovelace',
+  email: 'ada@equinor.com',
+  layout: 'vertical',
+}
+NameLabelVertical.parameters = {
+  docs: {
+    description: {
+      story:
+        'Vertical layout places the name and email on the same line — compact for use in dense lists.',
+    },
+  },
+}
+
+export const NameLabelWithNotification: StoryFn = () => (
+  <AvatarNameLabel
+    fullName="Ada Lovelace"
+    email="ada@equinor.com"
+    notification
+  />
+)
+
+export const NameLabelWithSlot: StoryFn = () => (
+  <AvatarNameLabel fullName="Ada Lovelace" email="ada@equinor.com">
+    <span aria-label="Settings" role="img">
+      ⚙
+    </span>
+  </AvatarNameLabel>
+)
+NameLabelWithSlot.parameters = {
+  docs: {
+    description: {
+      story: 'The `children` prop renders into the trailing slot.',
+    },
+  },
+}
--- a/packages/eds-core-react/src/components/next/Avatar/Avatar.stories.tsx
+++ b/packages/eds-core-react/src/components/next/Avatar/Avatar.stories.tsx
@@ -1,6 +1,10 @@
 import type { Meta, StoryFn } from '@storybook/react-vite'
-import { Avatar, type AvatarProps } from '.'
-import { AvatarNameLabel, type AvatarNameLabelProps } from '.'
+import {
+  Avatar,
+  AvatarNameLabel,
+  type AvatarProps,
+  type AvatarNameLabelProps,
+} from '.'
 
 const meta: Meta<typeof Avatar> = {
   title: 'EDS 2.0 (beta)/Data Display/Avatar',
--- a/packages/eds-core-react/src/components/next/Avatar/Avatar.stories.tsx
+++ b/packages/eds-core-react/src/components/next/Avatar/Avatar.stories.tsx
@@ -1,3 +1,4 @@
+import React from 'react'
 import type { Meta, StoryFn } from '@storybook/react-vite'
 import { more_vertical } from '@equinor/eds-icons'
 import { Button } from '../Button'
@@ -362,16 +363,19 @@ const ListWithDividers = ({
 }: {
   children: React.ReactNode[]
   maxWidth?: string
-}) => (
-  <div style={{ maxWidth }}>
-    {children.map((child, i) => (
-      <div key={i}>
-        <div style={{ padding: '8px 0' }}>{child}</div>
-        {i < children.length - 1 && <Divider />}
-      </div>
-    ))}
-  </div>
-)
+}) => {
+  const items = React.Children.toArray(children)
+  return (
+    <div style={{ maxWidth }}>
+      {items.map((child, i) => (
+        <div key={(child as React.ReactElement).key}>
+          <div style={{ padding: '8px 0' }}>{child}</div>
+          {i < items.length - 1 && <Divider />}
+        </div>
+      ))}
+    </div>
+  )
+}
 
 export const NameLabelWithSlot: StoryFn = () => (
   <div style={{ display: 'flex', flexDirection: 'column', gap: '24px' }}>
--- a/packages/eds-core-react/src/components/next/Avatar/Avatar.stories.tsx
+++ b/packages/eds-core-react/src/components/next/Avatar/Avatar.stories.tsx
@@ -82,6 +82,40 @@ Introduction.parameters = {
   },
 }
 
+export const Photo: StoryFn = () => (
+  <div style={{ display: 'flex', alignItems: 'center', gap: '16px' }}>
+    <Avatar
+      src="https://i.pravatar.cc/150?img=47"
+      name="Ada Lovelace"
+      size="sm"
+    />
+    <Avatar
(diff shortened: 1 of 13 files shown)


millus:
@claude please review

millus:
@claude please review

millus:
@claude please review

millus:
@claude please re-review, is this good to go now? no blockers, ready for human review?

packages/eds-core-react/src/components/next/Avatar/AvatarNameLabel.tsx
```diff
@@ -0,0 +1,60 @@
+import { forwardRef } from 'react'
+import type { AvatarNameLabelProps } from './AvatarNameLabel.types'
+import { Avatar } from './Avatar'
+
+function deriveInitials(name: string): string {
+  const words = name.trim().split(/\s+/)
+  if (!words[0]) return ''
+  if (words.length === 1) return words[0][0].toUpperCase()
+  return (words[0][0] + words[words.length - 1][0]).toUpperCase()
+}
```
pomfrida, line 10: `deriveInitials` is defined identically in both `Avatar.tsx` and `AvatarNameLabel.tsx`. Move it to a single source so the two can't drift. Either `Avatar/utils.ts` (component-local) or `components/next/utils/` (shared) — the rest of `/next` keeps small helpers component-local, so a `utils.ts` next to `Avatar.tsx` and an import from both files would match existing conventions.

packages/eds-core-react/src/components/next/Avatar/Avatar.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+      render(<AvatarNameLabel name="Ada" size="sm" />)
+      expect(screen.getByText('Ada')).toBeInTheDocument()
+    })
+
+    it('renders with emphasis prop without error', () => {
+      render(<AvatarNameLabel name="Ada" emphasis="high" />)
+      expect(screen.getByText('Ada')).toBeInTheDocument()
+    })
+  })
+
```
pomfrida, line 212: These tests only assert that the visible name still renders — they don't verify that `size` or `emphasis` actually reach the inner `Avatar`. The comment says "Testing Library's no-node-access rule prevents querying the inner div", but `screen.getByText('AL').closest('.eds-avatar')` (or a `data-testid` on the inner Avatar) would let us assert `data-size` / `data-emphasis` directly. As written, the inner Avatar could be rendered with hard-coded defaults and these tests would still pass.

Suggestion: 
```tsx
it('passes size to the inner Avatar', () => {
  render(<AvatarNameLabel name="Ada Lovelace" size="sm" />)
  const avatar = screen.getByText('AL').closest('.eds-avatar')
  expect(avatar).toHaveAttribute('data-size', 'sm')
})
```

packages/eds-core-react/src/components/next/Avatar/avatar.css
```diff
@@ -0,0 +1,177 @@
...
+    height: var(--_size);
+    border-radius: var(--eds-spacing-border-radius-pill);
+
+    background-color: var(--_bg);
+
+    &[data-size='sm'] {
+      --_size: var(--eds-sizing-icon-xs); /* 16px */
+      --_font-size: var(--eds-typography-ui-body-xs-font-size);
+      --_line-height: var(--eds-typography-ui-body-xs-line-height-default);
+      --_dot-size: 8px;
```
pomfrida, line 32: Hmm I see all other `/next` components scope their size attribute (`data-icon-size`, `data-selectable-space`). `data-size` is the first plain variant in `/next` and can collide with ancestor styles or future tokens that key off `[data-size]`. Suggest renaming to `data-avatar-size` for consistency. (`data-emphasis` is fine — Badge already uses it.)

packages/eds-core-react/src/components/next/Avatar/avatar.css
```diff
@@ -0,0 +1,177 @@
...
+    &[data-layout='vertical'] {
+      & .content {
+        flex: initial;
+        flex-shrink: 0;
+      }
+
+      & .names {
+        flex: initial;
+        flex-direction: row;
+        flex-shrink: 0;
```
pomfrida, line 130: In `data-layout="vertical"` the `.content` and `.names` blocks both have `flex-shrink: 0`, with `white-space: nowrap` on `.names` and no truncation fallback. Verified in Storybook via the `NameLabelEdgeCases` story (which uses deliberately long names like `"Bartholomew Featherstonehaugh"`):

- At a **375px viewport**, the `.names` row is 44px wider than the body — the email gets clipped on the right edge (macOS hides the scrollbar by default, so visually it looks like the text just disappears; Windows / Linux / iOS would show a horizontal scrollbar on `<body>`)
- At a **240px viewport**, the row is 347px wide inside a 176px container — ~50% of the email content is hidden

Acknowledged that this only triggers with the combination of (a) a narrow container and (b) long names, and that the layout is documented as "intended for wider contexts". But once both conditions hit, the email is invisible with no indicator that more text exists, which is worse than truncating with an ellipsis (where the `…` at least signals "there's more"). Suggest adding a defensive truncation so the component degrades gracefully — note that **both** `.content` and `.names` need to be allowed to shrink, otherwise the truncation doesn't engage:

```css
&[data-layout='vertical'] {
  & .content {
    flex: 1 1 auto;     /* was: flex: initial */
    flex-shrink: 1;     /* was: 0 — must let parent shrink */
    min-width: 0;
  }

  & .names {
    flex: 1 1 auto;     /* was: flex: initial */
    flex-shrink: 1;     /* was: 0 */
    min-width: 0;       /* required for text-overflow inside flex */
    /* flex-direction: row, white-space: nowrap kept */
  }

  & .meta {
    overflow: hidden;
    text-overflow: ellipsis;
    min-width: 0;
  }
}
```

Verified locally — with the patch applied at 375px viewport, body overflow drops from 44px to 0px and the email shows as `"b.featherstoneha…"`. Alternative: drop `white-space: nowrap` to allow wrapping (preserves information at the cost of multi-line).

packages/eds-core-react/src/components/next/Avatar/avatar.css
```diff
@@ -0,0 +1,177 @@
+@layer eds-components {
+  .eds-avatar {
+    /* Size + font tokens — overridden per data-size */
+    --_size: var(--eds-sizing-icon-2xl); /* 32px — Large */
+    --_font-size: var(--eds-typography-ui-body-md-font-size);
+    --_line-height: var(--eds-typography-ui-body-md-line-height-default);
+
+    /* Dot sizes have no token equivalent in the design — hardcoded per Figma spec */
+    --_dot-size: 12px;
```
pomfrida, line 9: Acknowledged in the comments as "no token equivalent". Could you eave hard-coded but maybe reference the Figma spec node id in the comment so future readers can verify, in case it would be added later

packages/eds-core-react/src/components/next/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,62 @@
...
+    className,
+    ...rest
+  },
+  ref,
+) {
+  const classes = ['eds-avatar', className].filter(Boolean).join(' ')
+  const resolvedInitial = initial ?? (name ? deriveInitials(name) : undefined)
+  // When src is provided the <img> element carries the accessible name — skip role/aria-label on the div
+  const label =
+    !src && name ? (notification ? `${name}, notification` : name) : undefined
```
pomfrida, line 29: When `name` is not provided, the `notification` dot is `aria-hidden` and silent. That's a UX trap — a consumer might use `<Avatar src=... notification />` for "you have an unread message" and screen-reader users would never know. 

Two options:

1. Require `name` when `notification` is true (type-level enforcement).
2. Always render a visually-hidden "Notification" span when `notification` is true, regardless of `name`.

Option 2 matches what `AvatarNameLabel` already does (line 51 of `AvatarNameLabel.tsx`).

packages/eds-core-react/src/components/next/Avatar/AvatarNameLabel.tsx
```diff
@@ -0,0 +1,60 @@
...
+      .join(' ')
+
+    return (
+      <div ref={ref} className={classes} data-layout={layout} {...rest}>
+        <div className="content">
+          <Avatar
+            size={size}
+            emphasis={emphasis}
+            src={src}
+            alt={alt ?? ''}
```
pomfrida, line 42: `alt={alt ?? ''}` is passed to the inner Avatar, but `Avatar` itself does `alt={alt ?? name ?? ''}` (line 36 of `Avatar.tsx`) — so the `alt` prop on `AvatarNameLabel` does work via override, but the doc comment in `AvatarNameLabel.types.ts:27` says it defaults to `""` so the visible name isn't announced twice. Good, but worth a one-line code comment here explaining *why* we force `''` rather than letting `Avatar`'s own fallback to `name` run, since the visible name covers the image.

packages/eds-core-react/src/components/next/Avatar/Avatar.stories.tsx
```diff
@@ -0,0 +1,518 @@
...
+  </div>
+)
+NameLabelWithSlot.parameters = {
+  docs: {
+    description: {
+      story:
+        'The `children` prop renders into a trailing slot to the right of the name. It is open — use it for anything contextual to that person, like an action button, a role badge, or a timestamp. The examples here are just starting points.',
+    },
+  },
+}
```
pomfrida, line 518: Not a blocker for this PR — `Avatar.stories.tsx` is fully consistent with how the other `/next` stories (`Button`, `Badge`, `Icon`, `Input`) handle demo styling. Flagging it as a broader observation rather than a change request: across `/next` stories we routinely inline raw `#hex` colours and `'12px'` / `'16px'` pixel values for demo scaffolding (section headings, gaps, sample backgrounds). Examples outside this PR:

- `Icon.stories.tsx` — `color: '#666'`, `background: '#f5f5f5'`
- `Input.stories.tsx` — `background: '#f7f7f7'`, `'--eds-color-neutral-1': '#fff'`
- `Badge.stories.tsx` — `fontSize: '14px'`, `fontSize: '16px'`

Storybook is one of the most-visible surfaces of the design system, so demo styling that bypasses the token system is a bit of a "do as I say, not as I do" signal. Worth a separate conversation/PR about whether we want to tighten this up across the board (an `.sb-section-heading` utility or a shared story-helper component would cover most of the repeated patterns). So I think I will add a task for conforming it to our tokens for us to tackle after we have landed the new token structure.  No action needed in this PR.

packages/eds-core-react/src/components/next/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,62 @@
...
+      ref={ref}
+      className={classes}
+      data-color-appearance="accent"
+      data-size={size}
+      data-emphasis={emphasis}
+      {...a11yProps}
+      {...rest}
+    >
+      {src ? (
+        <img className="photo" src={src} alt={alt ?? name ?? ''} />
```
pomfrida, line 43: The CSS sizes the image with `position: absolute; width: 100%; height: 100%`, so layout is stable, but adding explicit `width` / `height` attributes (matching the rendered pixel size) helps browsers reserve space before CSS loads and is a small CLS improvement. Optional.

packages/eds-core-react/src/components/next/Avatar/Avatar.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+      render(<AvatarNameLabel name="Ada" />)
+      expect(screen.getByText('A')).toBeInTheDocument()
+    })
+
+    it('uses provided initial over derived one', () => {
+      render(<AvatarNameLabel name="Ada Lovelace" initial="X" />)
+      expect(screen.getByText('X')).toBeInTheDocument()
+    })
+
+    it('renders slot right when children provided', () => {
```
pomfrida, line 172: Test asserts that the slot content renders, but not that it ends up in `.slot-right`. A `.toHaveClass` check on the closest container would catch a regression where someone moved children into `.names`.

packages/eds-core-react/src/components/next/Avatar/AvatarNameLabel.tsx
```diff
@@ -0,0 +1,60 @@
...
+            initial={src ? undefined : derivedInitial}
+            notification={notification}
+          />
+          <div className="names">
+            <span className="full-name">{name}</span>
+            {meta && <span className="meta">{meta}</span>}
+            {notification && (
+              <span className="notification-label">Notification</span>
```
pomfrida, line 50: Good a11y instinct: when the avatar is decorative inside `AvatarNameLabel`, the notification can't ride on the avatar's aria-label, so you've added a visually-hidden span instead. Two different solutions for two different contexts, and both are right.

pomfrida (changes requested):
Really solid PR overall. The component split between `Avatar` and `AvatarNameLabel` is clean, the accessibility model is well-thought-out across both contexts (the `notification` merged into `aria-label` for standalone, visually-hidden span for the composed component — two different solutions for two different contexts, and both right), the pseudo-private `--_` variable pattern is applied correctly for size scaling, and the Figma Code Connect coverage for both components (including the Photo variant) is more than the bar. Tests are organised into Rendering / Variants / Notification / Accessibility blocks with a `jest-axe` assertion per meaningful variant — exactly what AGENTS.md asks for. Thanks also for pre-flagging the `--eds-selectable-space-*` deviation in the description.

A few inline comments split across three buckets:

- 🔴 **Blockers (please address before merge):** `deriveInitials` duplicated across both files (1), and the `size` / `emphasis` pass-through tests in `AvatarNameLabel` don't actually assert anything (2)
- 🟡 **Worth a look:** vertical-layout overflow at narrow viewports with long names (3, verified in Storybook with a concrete CSS fix proposed), generic `data-size` attribute name (4), and a few smaller a11y / token observations 
- 🟢 **Nits / follow-up:** stylistic cleanups and one broader observation about hardcoded values in `/next` stories that isn't specific to this PR

PR #5091, opened 2026-06-22
fix: propagate density-aware spacing aliases to comfortable mode

### LINKED ISSUE #5090: Density-aware spacing aliases not propagated to comfortable mode
## Problem

Spacing alias tokens like `--eds-generic-gap-vertical`, `--eds-selectable-space-vertical`, and `--eds-container-space-vertical` are declared only on `:root` (and `[data-vertical-gap="*"]` etc.). They reference density-aware primitives like `var(--eds-spacing-vertical-xs)`, but because CSS custom properties resolve `var()` at the declaring element, the aliases get the **spacious** value computed at `:root` and inherit that static value to descendants.

When a descendant gets `data-density="comfortable"`, the primitive `--eds-spacing-vertical-xs` correctly updates to `.375rem`, but the alias `--eds-generic-gap-vertical` stays at the `.5rem` value inherited from `:root`.

## Reproduction

Storybook → TextField → "Density Modes" story → inspect the comfortable section:

- `--eds-spacing-vertical-xs`: `.375rem` ✓
- `--eds-generic-gap-vertical`: `.5rem` ❌ (should be `.375rem`)

## Impact

Field-based components (TextField, Select, Search, Autocomplete, TextArea etc.) use `gap: var(--eds-generic-gap-vertical)` for vertical layout between Label / Description / Input / Helper Message. Figma shows these gaps should shrink in comfortable density (8px → 6px), but they stay at 8px due to this bug.

## Where the fix belongs

`packages/eds-tokens/src/generate-variables/createSpacingAndTypographyVariables.ts`

The `densityComfortableFilter` and `buildDensityDictionary` for comfortable should emit alias tokens (`generic-gap-*`, `selectable-space-*`, `container-space-*`) in the `[data-density="comfortable"]` selector block, not only the primitives.

Expected output in `build/css/spacing/comfortable.css`:

```css
[data-density="comfortable"] {
  /* …primitives as today… */
  --eds-spacing-vertical-xs: .375rem;

  /* …plus aliases that are missing today: */
  --eds-generic-gap-vertical: .375rem;
  --eds-selectable-space-vertical: .375rem;
  /* …and others using density-aware primitives */
}
```

## Discovered while

Following up PR #5080 (alignment of helper-message spacing across Field-based inputs). The PR is correct as-is; once this token bug is fixed, all Field-based inputs will automatically get density-aware gaps without further component changes.

pomfrida:
## Summary

Fixes #5090. Spacing aliases (`--eds-generic-gap-vertical`, `--eds-selectable-space-vertical`, `--eds-container-space-vertical`, `--eds-page-space-vertical`, `--eds-spacing-proportions-{size}-*`, `--eds-{selectable,container,page}-gap-*`) were declared at `:root` and referenced density-aware primitives. CSS `var()` resolves at the declaring element, so the spacious value was inherited as a string and never re-resolved when a descendant got `data-density="comfortable"`.

Density CSS now re-emits each bare alias with the resolved comfortable / spacious value, so density propagates through the alias chain.

Also fixes a latent filter bug: `name.includes('-container')` combined with `!name.startsWith('container-space')` wrongly excluded any token whose name contains `-container`, because `token.name` carries the `eds-` prefix at filter time and the `startsWith` allow-list never matched. The check now uses `endsWith('-container')` to match the intended icon-container pattern precisely.

## Verified

Storybook → TextField → Density Modes, computed values:

| Token | Spacious | Comfortable |
|---|---|---|
| `--eds-generic-gap-vertical` | `.5rem` | `.375rem` ✓ |
| `--eds-selectable-space-vertical` | `.5rem` | `.375rem` ✓ |
| `--eds-container-space-vertical` | `1rem` | `.75rem` ✓ |
| `--eds-spacing-proportions-md-vertical` | `1rem` | `.75rem` ✓ |

## Test plan

- [ ] Inspect `build/css/spacing/comfortable.css` and `spacious.css` — both should contain the bare aliases with resolved values
- [ ] Verify `variables.min.css` has no broken cascade
- [ ] In Storybook, check Field-based components (TextField, Select, Search, Autocomplete, TextArea) in both density modes
- [ ] Confirm `--eds-typography-icon-size` etc. still work in components that set `data-font-size` on themselves
- [ ] Check Banner, Button, Chip, Accordion render correctly in comfortable

## Notes

- See [#5090 investigation comment](https://github.com/equinor/design-system/issues/5090#issuecomment-4767850027) for the architectural drøfting (why this bug existed) and side-effect analysis.
- Typography aliases turned out to be fine under documented usage — they're declared inside `[data-font-size="X"]` (not `:root`) so density propagates naturally. See follow-up comment on the issue.
- A comment in `index-variables.css` documents the import-order requirement: density CSS must come before per-axis alias files for the cascade to resolve correctly.


pomfrida:
@claude review

millus (approved):
Reviewed the code — root cause diagnosis and implementation look correct. The `endsWith('-container')` fix is right, the bare aliases in the aggregated JSON are correct, and the import order comment in `index-variables.css` is clear.

Verified in Storybook (worktree, `fix/density-aware-aliases` branch) via Playwright:

- TextField, Select, TextArea, Autocomplete, Button, Chip, Accordion — all confirm `--eds-generic-gap-vertical: .375rem` in comfortable vs `.5rem` in spacious
- `--eds-container-space-vertical: .75rem` comfortable / `1rem` spacious on all field-based components
- All 46 unit tests pass

All test plan items checked off. ✓

PR #5125, opened 2026-06-25
chore: make release-please exclude-paths repo-root-relative

pomfrida:
## Problem

release-please keeps adding a spurious `eds-core-react` release (most recently `2.7.0` with **Dialog EDS 2.0**) to every release PR, even though Dialog was only released to beta (`eds-core-react-next@v2.8.0-beta.1`). This has required manual cleanup of each release PR (e.g. #5106, #5116).

## Root cause

release-please resolves `exclude-paths` **relative to the repository root** and matches them by prefix — a file is only excluded when it starts with `${excludePath}/` (see [`commit-exclude.ts`](https://github.com/googleapis/release-please/blob/main/src/util/commit-exclude.ts) and the [manifest docs](https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md): *"path segment should be a folder relative to repository root"*).

Our config listed them **package-relative**, e.g. `"src/components/next"`. The actual file path is `packages/eds-core-react/src/components/next/Dialog/Dialog.tsx`, which does **not** start with `src/components/next/`. So the exclusions never matched and excluded **nothing** — every next-only `feat`/`fix` leaked into the parent `eds-core-react` release.

## Fix

Prefix every `exclude-path` with its package path so it resolves correctly from the repo root. This affects all packages (not just core — `eds-tokens`, `eds-data-grid-react`, etc. had the same broken paths).

After this merges, release-please will correctly exclude the `next` subfolder, storybook, tests, and config files. The pending spurious core release in #5116 should disappear on the next release-please run.

No behaviour change for legitimate source commits — `exclude-paths` can only ever *exclude*.

### DIFF
--- a/.github/release-please-config.json
+++ b/.github/release-please-config.json
@@ -7,24 +7,24 @@
       "package-name": "@equinor/eds-core-react",
       "component": "eds-core-react",
       "exclude-paths": [
-        "src/components/next",
-        ".storybook",
-        "stories",
-        "tests",
-        ".gitignore",
-        ".cachebust",
-        "rollup.config.js",
-        "jest.config.cjs",
-        "jest.setup.ts",
-        "tsconfig.json",
-        "tsconfig.build.json",
-        "tsconfig.test.json",
-        "playwright.config.ts",
-        "Dockerfile.storybook",
-        "nginx.conf",
-        "README.md",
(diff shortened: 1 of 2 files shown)


pomfrida:
@claude review

Issue #5108, opened 2026-06-24
Establish a new Token build & publishing pipeline


emirgens:
## Description
Rebuild the token build & publishing pipeline from the ground up with proper **CI/CD**, so the team genuinely owns it. Current state: a pipeline exists but has no CI/CD (built locally then shipped), it's heavily generated, and docs are scattered — leading to silent build drift and "zombie tokens."

## Scope
- Rebuild the pipeline with real **CI/CD** (no more local-build-then-ship).
- Tokens always stay updated automatically once migrated to Tokens Studio.
- GitHub sync is pull-based (CI pulls via the Tokens Studio CLI).
- Eliminate zombie tokens and build drift; the team owns and understands the pipeline.
- Tooling specifics depend on the Tokens Studio exploration (#4800).

## Definition of Done
- [x] New pipeline with CI/CD in place
- [x] Automated token updates from Tokens Studio
- [x] No more local-only builds / silent drift
- [x] Zombie tokens eliminated
- [x] Pipeline documented and owned by the team

Timeline: ships in the eds-tokens beta first (current beta naming); stable major not this year.

pomfrida:
## Status + next step: the Tokens Studio CI trigger is live

A release in Tokens Studio now fires a `repository_dispatch` (`tokens-release`) that runs `.github/workflows/tokens_studio_release.yaml` ([first successful run](https://github.com/equinor/design-system/actions/runs/28928718702)). This is the entry point for the pull-based GitHub sync in the scope above. The workflow currently only logs the payload, which is minimal (`{"triggered_by": "tokens-studio"}` — no release version included), so the workflow must pull the token state itself.

Plan for making it pull for real:

### 1. CI authentication — OIDC, no service token needed
- Per the [CI/CD triggers docs](https://documentation-v2.tokens.studio/integrations/ci-cd-triggers.html), OIDC is the authentication method used by CI Integrations — no API keys to store. The trust config is part of the CI Integration already set up in Studio: verify the **Repository** (`equinor/design-system`) and **Subject Pattern** (e.g. `repo:equinor/design-system:*`) fields there.
- On the GitHub side the workflow needs `id-token: write` in `permissions` — without it the `ACTIONS_ID_TOKEN_REQUEST_TOKEN` env var the CLI auto-detects doesn't exist.
- A service account token (`STUDIO_SERVICE_TOKEN`) remains available as a fallback, but shouldn't be needed.

### 2. Commit `.studio.json`
[...]

PR #5153, opened 2026-07-07
ci: add repository_dispatch listener for tokens studio releases

pomfrida:
## What

Adds a minimal GitHub Actions workflow that listens for the `repository_dispatch` event type `tokens-release`, sent by the outbound CI trigger configured in Tokens Studio (Integrations → CI Triggers).

## Why

First step of the Tokens Studio → GitHub pipeline (#5108). The workflow intentionally only logs the incoming `client_payload`:

- `repository_dispatch` only triggers workflows on the default branch, so nothing can be verified until a listener is merged to `main`
- the payload structure Studio sends is undocumented — logging it tells us what the real pipeline (pull → build → PR) has to work with

The payload is passed via env vars (not inline `${{ }}` interpolation in the script) to avoid script injection, and the job runs with `contents: read` only.

## Testing

After merge, the wiring can be tested without Studio:

```
gh api repos/equinor/design-system/dispatches -f event_type=tokens-release
```

and end-to-end by creating a test release in Tokens Studio.


pomfrida:
@claude review

Issue #5154, opened 2026-07-08
Automated git backup of Tokens Studio tokens via studio CLI


pomfrida:
## Why

The Tokens Studio platform currently has **no undo, rollback, or restore functionality** — only a read-only version history of releases ([version history docs](https://documentation-v2.tokens.studio/releases/version-history.html)). Designers have already made mistakes that were permanently applied, and since plugin changes push to Studio in real time, errors propagate immediately.

Until the platform ships undo, we need our own safety net outside the platform.

## What

Set up a scheduled job that pulls all token sources from Tokens Studio into a git repository, so we get diffs, history, and recovery independent of the platform:

- `studio tokens pull` on a schedule (e.g. GitHub Actions cron), committing changes with a timestamped message
- Authentication via service account token or OIDC (`--ci`), see [service account tokens](https://documentation-v2.tokens.studio/integrations/service-account-tokens.html) — prefer OIDC, otherwise store the token in the CI provider's secret store
- Decide target: dedicated backup repo vs. a directory in this monorepo (note: `.studio.json` config model and safety rubric are documented in `documentation/agent-instructions/TOKENS_STUDIO.md`, added in #5151)
- Recovery path documented: how to restore token state in Studio from a given backup commit

## Notes

- The CLI is installed as a devDependency of `packages/eds-tokens` (PR #5151)
- Related: token pipeline rebuild #5108
- Webhooks or `studio tokens watch` could make this event-driven later; a simple cron is enough as a first safety net

PR #5155, opened 2026-07-08
ci: pull tokens from Tokens Studio on release

pomfrida:
## What

Extends `.github/workflows/tokens_studio_release.yaml` from logging the `repository_dispatch` payload to actually pulling the token state and opening a PR with any changes:

- `setup` job via the reusable `_setup.yml` (same sparse-checkout + cache pattern as `sync-figma-to-tokens.yml`), checking out `packages/eds-tokens`
- `pull-tokens` job runs `pnpm --filter @equinor/eds-tokens exec studio tokens pull --ci` and creates a PR via `peter-evans/create-pull-request` (skipped when there is no diff)
- CI auth is **OIDC** (`id-token: write` + `--ci`) — the CLI auto-detects the GitHub Actions OIDC token and Studio validates it against the CI integration's repository/subject pattern. No service token stored.

Also:

- Commits `packages/eds-tokens/.studio.json` (currently pointing at the **eds-test-3** project, branch `main`, raw format — will be repointed when the production project is ready)
- Fixes a factual error in `documentation/agent-instructions/TOKENS_STUDIO.md`: the config file uses a `configurations` key with a `ref` object, not the `sources` shape the doc showed; the repo-state snapshot line now describes the CI setup

## Notes

- The workflow only runs from the default branch (`repository_dispatch` limitation), so the first real test is the next Tokens Studio release after this merges. This is also the first live test of the OIDC direction (workflow → Studio API); the trigger run so far only proved Studio → GitHub.
- PRs created by the workflow use the built-in `GITHUB_TOKEN` and therefore don't get CI runs — same limitation as the existing Figma sync; can be upgraded to a GitHub App token later.
- Part of #5108; the release-triggered pull commits also cover the backup safety net described in #5154.

### DIFF
--- a/.github/workflows/tokens_studio_release.yaml
+++ b/.github/workflows/tokens_studio_release.yaml
@@ -3,16 +3,34 @@ name: Tokens Studio release
 # CI Triggers). Studio sends a repository_dispatch event with this event
 # type when a release is created. Note: repository_dispatch only triggers
 # workflows on the default branch.
+#
+# The payload does not include the release version, so the workflow pulls
+# the current state of the sources configured in
+# packages/eds-tokens/.studio.json and opens a PR with any changes.
 on:
   repository_dispatch:
     types: [tokens-release]
 jobs:
-  handle-release:
-    name: Handle Tokens Studio release
-    runs-on: ubuntu-latest
-    timeout-minutes: 5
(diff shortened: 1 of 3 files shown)


pomfrida:
@claude review

PR #5160, opened 2026-07-10
ci: run Tokens Studio CSS export in release workflow

pomfrida:
Adds a step to the `Tokens Studio release` workflow that runs the saved **EDS** export configuration (CSS Variables) via `studio exports run EDS --ci` and writes the generated CSS to `packages/eds-tokens/src/tokens/css/`.

This is the first step towards replacing the local Style Dictionary build with platform-generated CSS: the release PR now contains both the raw token sets (existing `studio tokens pull`) and the generated CSS output. The generated directory is not yet wired into the package `exports` map, so nothing changes for consumers.

Authentication reuses the existing inbound OIDC CI integration (read-only), so no new secrets are needed.


pomfrida:
@claude review

pomfrida:
@claude review

Issue #5139, opened 2026-07-01
ADR: Token variable architecture — Figma structure (typography & spacing)


eddiman:
## Description
Write the ADR defining the **variable structure** for the redefined token system's typography, spacing, corner-radius, and density — the primitive, mode-bearing *mapping*, and mode-free *semantic* variable layers plus the text-style layer. **Tokens Studio is the canonical source of truth; the Figma variables and text styles are synced from it.**

Parent: #4963. Colour and the code/output layer are separate ADRs.

## Scope
- **Three layers, one-way references** (primitive → mode-bearing mapping → mode-free semantic → text-style); no sibling / cross-collection / semantic→semantic / mapping→mapping refs.
- The **mode-free semantic layer** is the only consumed/published layer; all mode behaviour lives in the mapping layer beneath it.
- **Density** is the mode-bearing mapping collection (`compact` / `comfortable` / `relaxed`), with independent `typography` / `spacing` / `corner-radius` groups, via primitive step-shifting.
- **Mode-less `Font`** (family + weight); weight composed at the text-style layer.
- Body typography = ui sizes × 1.5 (not a token).
- Naming: spacing uses rem-hundredths; the type scale uses a 100-step index; semantic is lowercase-hyphenated.

## Definition of Done
- [ ] ADR written (Markdown in the `design-system` repo, `documentation/adr/`)
- [ ] Layer diagram included
- [ ] One-way reference rule documented
- [ ] Reviewed by design + engineering
- [ ] Status moved Proposed → Accepted

**Status:** in review — `design-system` repo `documentation/adr/0006-token-variable-architecture-spacing-typography.md` (Proposed); PR #5161.

Issue #5107, opened 2026-06-24
Re-define Typography, Spacing & Corner Radius tokens


emirgens:
## Description
The core model change from the 23 Jun decision. Decouple typography, spacing, and corner radius into three independent variable collections built directly on the primitives, and redefine typography as plain text styles.

## Scope
- **Three decoupled collections** — typography, spacing, corner radius — no cross-collection links; references flow one way to the primitives.
- **Typography → plain text styles** exposed directly in Figma's type list (sizes visible), split into two sets: **UI text** and **body text** (body with higher, WCAG-compliant line-height). Drop the typography component.
- **Spacing applied directly** (xs/sm/md/lg…), with px values visible. Spacing and typography vary only with **density** (and font: Equinor/headings vs Inter/body).
- **Remove text-trim** — it makes line-height misleading and doesn't exist in the mobile stack; removing it keeps web + mobile on one system.
- **New Figma file** for the new variables (mapping already started); leave the old file alone.
- Settle the 0.5px rounding question.
- Result imported to Tokens Studio.

## Definition of Done
- [x] Three decoupled collections defined (typography / spacing / corner radius)
- [x] Typography shipped as text styles (UI + body sets), component dropped
- [x] Text-trim removed
- [x] px values visible on spacing + type
- [x] New Figma file created with variables applied to a representative set
- [x] Imported to Tokens Studio

eddiman:
All DoD items complete — three decoupled collections (typography / spacing / corner radius), typography shipped as UI + body text styles with the component dropped, text-trim removed, px values exposed, new Figma file with variables applied, imported to Tokens Studio. Closing.

Issue #4743, opened 2026-03-30
📏 Rename density modes (compact / comfortable / relaxed)


emirgens:
## Description
The 23 Jun decision renamed the density modes to `compact` / `comfortable` (the current naming was flagged as confusing). Since then a third mode has been retained as **`relaxed`** (originally for the mobile POC), so the canonical set is **`compact` / `comfortable` / `relaxed`** — see ADR 0007 (token variable architecture). This supersedes the earlier plan to drop the third mode / "spacious".

## Scope
- Rename existing density modes to **compact** / **comfortable** / **relaxed**.
- Retire the "spacious" label (the third mode is now **relaxed**).
- Ensure density (plus font) is the only thing that drives spacing/type variation.
- Validate across representative components.

## Definition of Done
- [x] Density modes named compact / comfortable / relaxed
- [x] "Spacious" label retired
- [x] Tokens updated — validation across representative components moved to equinor/design-system#5119 (Phase 2), where the components actually move onto these tokens
- [x] Updated token set published
- [x] Rename rationale documented — [ADR 0007](https://github.com/equinor/design-system/blob/main/documentation/adr/0007-token-variable-architecture-spacing-typography.md)

eddiman:
## Status: design captured in ADR 0007

_Edit 2026-08-19: this comment originally cited **ADR 0006** — that's the flat class names ADR. The token variable architecture ADR is **0007**, and it is now **Accepted** (status header fixed in #5348). Corrected below; see the closing comment for where the remaining work went._

The definition for this is captured in **ADR 0007 (token variable architecture)**, which is now **Accepted**.

### Defined in ADR 0007

- **Naming:** `Compact` · `Comfortable` · `Relaxed`. The 23-Jun two-mode "compact/comfortable" plan is superseded — `Relaxed` (originally the mobile-POC density) is canonical; the new set simply doesn't carry a "spacious" label.
- **Mechanism:** density works by **step-shifting** which primitive each token references — `Compact` = one step down, `Comfortable` = base, `Relaxed` = one step up — clamped at the scale ends, applied uniformly. It's the only density-bearing collection; density is set once per surface.
- **Scope:** `density/spacing/*`, `density/corner-radius/*`, `density/typography/{ui,header}/…`.
- **Figma:** the `density` collection (compact/comfortable/relaxed) is already built in the new *EDS Redefined Foundation* file (built 1 Jul, alongside #5107).

### Rollout model

[...]

PR #5161, opened 2026-07-10
docs: 📝 add token variable architecture ADR (typography & spacing)

eddiman:
## What

Adds ADR `0007-token-variable-architecture-spacing-typography`, documenting the variable structure for the redefined EDS token system — **typography, spacing, corner-radius, and density**.

## Why

Foundational sub-task of the token redefinition epic (#4740): it establishes the shared mental model the rest of the redefinition (typography/spacing #5107, density #4743) builds on. Tracked by #5139.

## How

Three layers with one-way references — **primitive → mode-bearing mapping → mode-free semantic → text-style**, plus naming conventions. **Tokens Studio is the canonical source of truth; the Figma variables and text styles are synced from it.** Colour (#4742 / #4744) and the code/output layer are explicitly out of scope and get their own ADRs.

Refs #5139.


documentation/adr/0007-token-variable-architecture-spacing-typography.md
```diff
@@ -0,0 +1,109 @@
...
+
+- **Colour.** The colour collection's variable structure and state-naming live in a separate track (#4742 / #4744) and will get **their own ADR**. This document does not decide the colour collection; the typography/spacing structure here does not depend on the colour outcome.
+- **The code / output layer** (Tokens Studio export, build pipeline, CSS + TS shape) — still firming, to be captured in its own ADR once the first export cycle has run.
+
+This is a foundational sub-task of the colour & token epic (#4740): typography/spacing/corner-radius (#5107) and density (#4743) build on the shape defined here. The type scale and primitive value layer are the foundation this structure sits on.
+
+## Decision — Figma variable structure
```
pomfrida, line 97: Did you follow the `0000-template.md`? This ADR skips Decision Drivers and Options Considered, and uses top-level `Consequences` and `References` instead. An Options Considered section would mostly be a restructuring of existing content, and it's where future readers will look for "what else was on the table". Suggest aligning with the template so it's consistent across all ADRs

documentation/adr/0007-token-variable-architecture-spacing-typography.md
```diff
@@ -0,0 +1,109 @@
...
+  │ Figma text-style layer   (header/* · ui/* · body/* + bolder)      │
+  │   bind SEMANTIC font-size · line-height · family · weight          │
+  └───────────────┬───────────────────────────────────────────────────┘
+                  │ one-way ▼
+  ┌───────────────┴───────────────────────────────────────────────────┐
+  │ SEMANTIC  (single "default" mode — MODE-FREE, PUBLISHED)           │
+  │   spacing/{none,4xs…3xl}                                           │
+  │   corner-radius/{none,rounded,rounded-outer,pill}                  │
+  │   typography/{ui,header}/{size}/{font-size,line-height}            │
+  │   font-family/{ui,heading} · font-weight/{lighter,normal,bolder}   │
```
pomfrida: `│   font-family/{ui,heading} · font-weight/{lighter,normal,bolder}   │`

Here we have `typography/{ui,header}/…` ut the family tokens are `font-family/{ui,heading}`. If the actual variables use both spellings, the ADR is faithfully recording a trap worth fixing at the source; if it's a typo, please unify before this becomes the canonical reference.

pomfrida: The semantic diagram (line 31) and point 7 (line 67) say `font-weight/{lighter,normal,bolder}`, but the Font mapping here and the primitives (line 46) say `weight/{…}`. Since the Naming section says mapping layers are namespaced by collection (`font/…`), spelling out the exact full paths once (e.g. `font/weight/bolder` → semantic `font-weight/bolder`) would remove the ambiguity.

documentation/adr/0007-token-variable-architecture-spacing-typography.md
```diff
@@ -0,0 +1,109 @@
...
+   - **Header weight varies by step** (`xs`/`sm` → `font-weight/normal`, `md`→`4xl` → `font-weight/bolder`). This variation lives in the **header text-style bindings**, not in tokens — there is no per-step header weight var.
+   - **"Bolder" is composition, not a parallel group.** The `ui/bolder/*` text styles are the ui size ramp × `font-weight/bolder`; no separate `ui-bolder` variable group exists, and adding one (or per-step weight vars) is the anti-pattern this rule forbids.
+
+### Naming
+
+- **Primitives:** `primitives/<domain>/<step|key>`. **Spacing** (and corner-radius, which aliases spacing) uses the rem-relative hundredths convention — `name = round(px / 16 × 100)`, ties resolving **down** (`12.5→12`, `37.5→37`); e.g. spacing `100 = 16px`, `25 = 4px`. **The type scale uses a separate 100-step index** — `type-scale/{inter,equinor}/{100…1100}`, where the number is the scale step (not a px-derived name) and values are rounded to whole px (e.g. `type-scale/inter/500 = 16px`, the base).
+- **Mapping layers are namespaced by where they live** — `density/…`, `font/…`, `colour/…` — so a reference makes its source (and whether it's density-dependent) obvious.
```
pomfrida: This bullet says semantic names "map cleanly to CSS custom properties", but the existing CSS tokens are `--eds-color-*` (US spelling), while the diagram (line 32) presents `colour/*` as a token path. Presumably this is just British-English — but a one-line clarification would prevent someone implementing the `colour` spelling in the variable set.

pomfrida (changes requested):
Nice ADR just a few small comments below 👍

documentation/adr/0007-token-variable-architecture-spacing-typography.md
eddiman: Good call — restructured to match `0000-template.md`: added **Decision Drivers** and **Options Considered** (the latter now captures what was on the table — the superseded two-layer draft, the two-mode density plan, and the per-step-weight anti-pattern), nested **Consequences** + **Confirmation** under **Decision**, and renamed **References → Related**. Pushed in 72ed87d.

documentation/adr/0007-token-variable-architecture-spacing-typography.md
eddiman: It was a genuine split rather than a typo: the type-size groups use `header` (`typography/header/*`) but the family token was `heading` in Tokens Studio. Rather than just document the trap, I unified it at the source — renamed `family.heading` → `family.header` and `font-family.heading` → `font-family.header` (via PATCH so UUIDs/scopes are preserved; on a branch pending review + merge). The ADR now uses `header` throughout. Pushed in 72ed87d.

eddiman: Spelled out the full chains in Naming: `primitives/weight-scale/bolder` (500) → mapping `font/weight/bolder` → semantic `font-weight/bolder`, and the family equivalent. Also corrected the diagram — the Font mapping row now reads `family/{ui,header} · weight/{…}` (it previously implied `font-family/…` at the mapping layer) and the primitive row `weight-scale`/`lineheight-scale`. So `weight/*` (mapping) vs `font-weight/*` (semantic) is now explicit. Pushed in 72ed87d.

documentation/adr/0007-token-variable-architecture-spacing-typography.md
eddiman: Checked the source — there are no `colour/*` tokens; Tokens Studio already uses US `color` (e.g. `text.color-on-light`). The British spelling only ever lived in the diagram. Unified the ADR to `color` everywhere and added a Naming line noting `color` is used across Figma variables, CSS `--eds-color-*`, and TS keys. Pushed in 72ed87d.

pomfrida (approved):
LGTM 👍

PR #5166, opened 2026-07-13
chore: generate TypeScript tokens in the Tokens Studio release workflow

### LINKED ISSUE #5164: Generate TypeScript tokens in the Tokens Studio pipeline
The Tokens Studio platform has no TypeScript/JavaScript export format (supported formats: css, scss, dtcg, figma, swift, android, compose), so the new token pipeline covers CSS via the saved `EDS` export configuration but leaves the TS token outputs unsolved.

Today `@equinor/eds-tokens` exposes `./ts/color/*`, `./ts/spacing/*` and `./ts/typography/*`, generated by the legacy Style Dictionary build. The new pipeline needs to produce equivalent TS output by generating it locally from the raw/DTCG JSON that `studio tokens pull` already commits.

Scope:

- Map which of the current TS exports are actually used by consumers, to decide how much surface the new pipeline must reproduce (the typography TS output is intentionally minimal — see `packages/eds-tokens/CLAUDE.md` for the rationale)
- Write a small codegen script (JSON in, `.ts` files out) that runs on the pulled token files — no Style Dictionary dependency
- Wire the script into the `Tokens Studio release` workflow after the pull/export steps

Related: #5108

pomfrida:
Closes #5164

The Tokens Studio platform has no TypeScript export format, so this adds a small dependency-free codegen script that combines the two exports the release workflow already produces:

- **DTCG export** (new `EDS-DTCG` configuration, added as a workflow step) provides the token tree structure and `$type` metadata — its values are unresolved `{alias}` references and colour formulas
- **CSS export** provides the evaluated concrete values; the script dereferences the `var(--eds-*)` chains per dimension context (colour scheme × density)

`scripts/generate-ts-tokens.mjs` then emits `as const` TypeScript modules to `packages/eds-tokens/src/tokens/ts/`, mirroring the export layout. Modules that resolve differently per colour scheme (currently `semantic/`) are automatically split into per-scheme files.

Values are converted to React Native friendly forms for `equinor/design-system-mobile` (the only known TS-token consumer, see #5164):

- `oklch()` → hex via CSS Color 4 §13.2 gamut mapping (chroma bisection, not naive channel clipping) — verified identical to lightningcss for all 199 unique colours in the current export
- `px` dimensions → unitless numbers

Also fixes the workflow referencing the CSS export by name: the configuration was renamed `EDS` → `EDS-CSS` in Tokens Studio, which would have broken the next release run. Both exports are now referenced by ID, which is rename-proof.

Verified locally against real platform exports: 10 generated modules, `tsc --noEmit --strict` passes. Like the CSS output, the generated TS is not yet wired into the package `exports` map.

Note for the mobile migration: the new density dimension is comfortable/compact/relaxed — `spacious` (which mobile imports today) no longer exists.


pomfrida:
@claude review

pomfrida:
@claude review

PR #5175, opened 2026-07-14
ci: add scheduled git backup of Tokens Studio tokens

pomfrida:
Closes #5154

Tokens Studio has no undo/rollback — only a read-only version history of releases — and plugin changes push to the platform in real time, so designer mistakes propagate immediately. The release workflow only snapshots at release moments; everything in between is uncovered.

## What

- **New workflow `tokens_studio_backup.yaml`**: hourly cron (`23 * * * *`, plus `workflow_dispatch`) that runs `studio tokens pull --ci` for all sources in `packages/eds-tokens/.studio.json` and commits changes to the orphan branch `tokens-studio-backup` (one directory per source alias). Runs with no changes make no commit. Timestamped commit messages give an hourly-granularity recovery history.
- **Auth**: GitHub OIDC against the existing inbound CI Integration — scheduled runs always execute on `main`, so the token subject matches (manual dispatches must also pick `main`) the integration's subject pattern. No new secrets.
- **Rename-proof**: aliases and output dirs are read from `.studio.json` with `jq`, so the planned `eds-test-3` → `eds` rename (#5171) requires no workflow change.
- **Docs**: new "Backup & recovery" section in `documentation/agent-instructions/TOKENS_STUDIO.md`, including the manual recovery path (the CLI has no push command — restore = re-import JSON in Tokens Studio, coordinated with the designers).

## Notes

- The `tokens-studio-backup` branch is seeded manually (orphan, README only) so the workflow's checkout can always assume it exists. Never merge that branch.
- Failure alerts via the standard Slack step — an hourly safety net that fails silently is no safety net.

## Verification (after merge)

1. Trigger via `workflow_dispatch` → green run, snapshot commit on `tokens-studio-backup`
2. Trigger again immediately → green run, "No token changes since last backup", no commit
3. Next hourly run after a real token change in Studio → readable diff commit

PR #5176, opened 2026-07-14
ci: simplify Tokens Studio backup workflow

pomfrida:
Follow-up to #5175. The backup job reused the release workflow's setup (Node + pnpm + full `pnpm install --force` of the monorepo) purely for consistency — two minutes every hour just to obtain the one `studio` binary.

## What

- Drop `Install Node.js`, `Setup pnpm`, `Cache pnpm-store` and the full workspace install.
- Install the CLI standalone instead: version resolved from `packages/eds-tokens/package.json` (single source of truth), installed with `npm install --prefix` **outside** the pnpm workspace (npm fails on `workspace:` protocol deps inside it), cached keyed on the version so warm runs skip npm entirely.
- Pull/sync/commit/push logic and the Slack failure alert are unchanged.
- Docs: note in `TOKENS_STUDIO.md` § Backup & recovery that this is the one sanctioned exception to "never install the CLI with npm".

Warm runs go from ~2 min to ~20–30 s.

## Trade-off

The CLI version spec comes from `package.json` (`^0.1.7`), and since the install is cached on the spec string, the CLI is effectively pinned to the patch resolved on the first cold run — it only moves when the spec changes. That stability is intentional for an unattended backup.

## Verification (after merge)

Manual `workflow_dispatch` from `main`: first run installs and caches the CLI, pulls, and reports "No token changes since last backup" (or commits a snapshot if designers changed something); a second run should be fully warm.

### DIFF
--- a/.github/workflows/tokens_studio_backup.yaml
+++ b/.github/workflows/tokens_studio_backup.yaml
@@ -7,6 +7,11 @@ name: Tokens Studio backup
 # commits changes to the orphan branch `tokens-studio-backup`, giving us
 # diffs, history and a recovery point independent of the platform.
 # Recovery instructions: documentation/agent-instructions/TOKENS_STUDIO.md
+#
+# The CLI is installed standalone with npm outside the pnpm workspace —
+# a full workspace install just to obtain one binary is not worth two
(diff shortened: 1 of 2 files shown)


pomfrida:
@claude review

Issue #5123, opened 2026-06-25
Align mobile on the unified typography


eddiman:
## Description
Removing text-trim lets web and mobile share **one** typography set, avoiding a separate mobile typography set + component library. Confirm the impact and align mobile.

## Scope
- Confirm mobile can adopt the new (text-trim-free) typography tokens.
- Avoid a separate mobile typography / component library.
- Coordinate with EDS Mobile.

## Definition of Done
- [ ] Mobile impact of text-trim removal confirmed
- [ ] Mobile aligned on the unified typography set
- [ ] No separate mobile typography library required

eddiman:
We need to set a date with Marco to talk about the typescript exporter. They have updated it since we started with Tokens Studio in june, but what remains is the conversion of numbers in the string keys, so we dont have to do to Object.key() lookup.

PR #5178, opened 2026-07-15
docs: add token code / output architecture ADR

### LINKED ISSUE #5140: ADR: Token variable architecture — code / output layer
## Description
Write the ADR defining the **code / output architecture** that consumes the Figma variable structure (see the Figma-structure ADR under #4963): Tokens Studio export, build pipeline, and the CSS + TypeScript output shape.

Parent: #4963. Depends on the Figma-structure ADR and the Tokens Studio exploration (#4800, Marco Krenn sessions). To be written once the first CSS/TS export cycle has run — the direction is still firming, so it is deliberately not folded into the Figma ADR.

## Scope (to decide)
- Output targets: CSS custom properties + nested TypeScript module from one source.
- Path → nesting / naming transform (slash to nested keys, hyphen to camelCase, digit-leading tiers to JS-safe keys).
- One resolved tree per density mode.
- How the generator composes `Font` into typography; how `body` (ui × 1.5) is emitted.
- CSS specificity strategy (`@layer` vs `:where()`).
- Source-of-truth: Tokens Studio vs Figma; CLI / local validation; CI sync.
- Design vs. code ownership boundary.

## Definition of Done
- [ ] ADR written (Markdown in `docs/architecture-decisions/`)
- [ ] Reviewed by design + engineering
- [ ] Agreed as the shared model for the token build/output

pomfrida:
## Description

Adds ADR 0009 — the **code / output architecture** for the redefined token system (Tokens Studio export, build pipeline, CSS + TypeScript output shape). This is the sibling of ADR 0007 (Figma variable structure, #5161) and the umbrella over ADR 0008 (TS-token generation).

Key decisions recorded:

- **Source of truth & sync:** Tokens Studio is canonical; pull-based consumption via the `studio` CLI, release-triggered CI, export configs referenced by ID, hourly backup branch.
- **Output targets:** raw sets + CSS export (the single evaluation source) + DTCG export, with TypeScript generated by combining DTCG structure and CSS values (mechanics in ADR 0008).
- **Naming transform:** one deterministic path → `--eds-*` / nested camelCase mapping, incl. digit-leading tiers (`2xl` → `twoXl`); the `heading`/`header` discrepancy is flagged as a Studio-source fix.
- **One resolved tree per density mode**; per-colour-scheme splitting is data-driven.
- **Font composition & body typography:** mechanism decided (family/weight composed per role; `body/*` = ui × 1.5 as derived output in CSS + TS), implementation pending — the 1.5 ratio stays code-owned per the ADR 0007 ownership boundary.
- **CSS specificity:** published bundle wrapped in `@layer eds-tokens` with a shared `@layer eds-tokens, eds-elements, eds-components;` order statement. The legacy-unlayered-beats-layered collision risk is recorded as a consequence with a required collision check.

All claims about the generated output are verified against the `tokens-studio-release` branch.

Status is **Proposed** — moves to Accepted after design + engineering review per the issue's DoD.

Closes #5140

## Follow-ups (separate issues, not in this PR)

- Implement font/body emission in `generate-ts-tokens.mjs` + CSS output
- Implement the `@layer` bundle step and wire `src/tokens/*` into the package `exports` map


pomfrida:
@claude review

eddiman (changes requested):
Read this against ADR 0007/0008, the newer ADR PRs (#5195, #5196, #5199, #5200), and the
actual generated output on `main`. The architecture is right and it answers all seven
scope bullets from equinor/design-system#5140 — my comments are about staleness,
overlap with the ADRs that landed after this one, and one gap I think matters.

**What this lands, and I want it on the record:** decision point 5. Emitting
`font-family` + `font-weight` into the per-role typography trees, with per-step header
weight coming from the composition rather than from tokens, closes the partial-typography
trap — the one where a dev (or an agent) reaches for `typography/ui/md/line-height` to
style a paragraph and silently gets UI leading, because the correct answer ("ui font-size
+ literal 1.5") exists nowhere in the published surface. That was my biggest single worry
about the redefinition as scoped. Good that it's decided here even with implementation
pending.

I also checked the naming transform against the real export — `background/surface/accent/default/hover`
→ `--eds-background-surface-accent-default-hover` is exactly what `css/semantic/default.css`
emits, and the split-only-when-values-differ rule matches reality (`ts/semantic/{light,dark}.ts`
split, `ts/font/default.ts` doesn't). The claims hold up.

---

## Needs resolving before merge

**1. The `0009` slot is taken.** `0009-temporary-next-subpaths-for-eds-tokens-beta.md`
merged on 21 Jul via #5194, and 0010–0013 are claimed by #5199, #5195, #5196 and #5200.
This wants to be 0014.

**2. Decision point 1 is now ADR 0011's territory — and 0011 is Accepted.** #5195 decides
the `repository_dispatch: tokens-release` trigger, OIDC auth, exports-referenced-by-ID,
and the hourly `tokens-studio-backup` orphan branch, in more depth than this. Suggest
pointing at it the way this already points at 0008, rather than restating it — otherwise
we get two sources of truth on the pipeline and they'll drift.

**3. Decision point 6 (`@layer eds-tokens`) conflicts with ADR 0010 and with what
shipped.** `packages/eds-tokens/scripts/generate-css-bundle.mjs` is on `main` today, the
bundle contains no `@layer`, and #5199 says explicitly that "the bundle adds no behaviour
of its own on top of the generated files." Two open ADRs decide the bundle step
differently. I don't have a strong view on which way it goes — layering the published
bundle is a genuinely good idea and the `!important`-inversion consequence is honest — but
the two need to agree, and if `@layer` wins then 0010's "pure concatenation" framing has
to give.

**4. The required collision check is aimed at the wrong collision.** Confirmation asks for
a *legacy-vs-new* `--eds-*` name comparison. The collision that actually bit us is *inside*
the new output: Tokens Studio's CSS export flattens `.` to `-`, so semantic `border.focus`
lands on the same custom property as scheme `border-focus`. Where the semantic token
aliases the other one you get `--eds-border-focus: var(--eds-border-focus)` — self-
referential, resolves to no value. That was the Button/Chip focus ring. `border-disabled`
and `text-disabled` have the same shape, and all three are still in
`css/semantic/default.css` on `main` (`generate-css-bundle.mjs` documents them in its own
header comment; #5221 closed without clearing them).

Two asks: add the intra-export check to Confirmation alongside the legacy one, and record
the trap — **`tsctl resolved` will not show this**, because Studio resolves by token name
where the two names are distinct, so the platform looks healthy while the built CSS is
broken. Verification has to happen on built CSS output. There's a `check-collisions.mjs`
that exits non-zero and can gate the pipeline.

**5. "Values cannot drift between CSS and TypeScript" is overstated for colour.** CSS ships
wide-gamut `oklch()`; TS ships sRGB hex via gamut mapping. For the out-of-sRGB steps —
high-chroma reds and oranges at high lightness, worst in dark mode — web and React Native
will render visibly different colours. The real and still-strong claim is *one evaluation
source, structurally identical output*; "identical values" isn't quite true and mobile will
eventually notice. Suggest softening the driver and the matching consequence.

---

## The gap I think matters: the publish boundary isn't stated

ADR 0007 point 3 is explicit that primitives, density and font are hidden — nobody consumes
a mode-bearing variable directly, and the mode-free semantic layer is the only published
surface. Decision point 2 here commits all four generated directories as output without
saying which of them are *published* versus internal, and the shipped bundle currently
carries `--eds-density-spacing-md`, `--eds-primitives-spacing-100`, `--eds-family-header`
and `--eds-weight-bolder`, with `ts/density/*` and `ts/primitives/*` as importable modules.

That turns the intentional duplication into two valid-looking paths for the same value.
Humans will mostly pick the semantic one; agents won't reliably — `--eds-density-spacing-md`
reads as more specific and it's right there in the package. We removed dynamic colour partly
because agents kept fighting a paradigm they couldn't hold, and I'd rather not hand them a
fresh ambiguity in the replacement.

ADR 0010 says the granular files stay internal, which points the right way, but the bundle
concatenates all of them regardless. This ADR is the natural place to state the boundary:
which directories are the consumer contract, which are reference/recovery, and whether the
mapping-layer names appear in the published bundle at all. Cheap to decide now, breaking to
change after beta consumers land on them.

---

## Smaller things

- **TS mode-as-file vs mode-as-key.** Point 4 emits one file per density, and that's what
  shipped. The RN plan recommended density/scheme as top-level object keys
  (`spacing[density].md`) so a provider resolves both axes once and swaps at runtime with a
  single lookup — RN has no cascade, so mode is runtime state there, not a build-time
  choice. Per-file works, and tree-shaking is a fair argument for it if an app pins one
  density. Was that a deliberate call? If so it's worth a line in the options, since it's
  the one place the output shape diverges from what mobile asked for.
- **No iOS / Jetpack Compose story.** equinor/design-system#5123 wants the same resolved
  per-mode tokens feeding Swift and Compose. Not asking for it in this ADR, but naming the
  intended format list now is what keeps the token shape from having to change later.
- **Reference-chain lint isn't in Confirmation.** ADR 0007 leaves the one-way
  primitive → mapping → semantic rule to code review, noting a pipeline lint would be
  better. Once expanded colour lands, review won't hold it — the pipeline ADR is the
  natural home for that check.
- `src/tokens/<alias>/` in the diagram reads as a placeholder; the actual directory is
  `src/tokens/raw/`.
- Composed spreadable `textStyle` objects still aren't in the output. Point 5 gets most of
  the way there via the per-role trees, so this is much smaller than it was — flagging it
  as a follow-up rather than a gap in this ADR.

---

Substance is good and I'm happy with the direction. The renumber, the 0011 overlap, and the
`@layer`-vs-0010 conflict are mechanical once decided; the publish boundary is the one I'd
like settled in the ADR itself rather than deferred. Happy to pair on that bit if it's
easier to talk through than to comment on.

pomfrida:
Thanks for the thorough read @eddiman — all points should now be in.

552e91e3 (from before my break) covers the bulk: renumbered to 0014, decision point 1 defers the pipeline to ADR-0011 instead of restating it, the intra-export collision check is added to Confirmation alongside the legacy one (with the "platform resolution won't show this — verify on built CSS" trap recorded in Consequences), the drift claim is softened to one evaluation source / structurally identical output with the oklch-vs-hex colour exception spelled out, and the publish boundary is now decision point 7 (contract = semantic + per-role typography + per-density trees; everything else is resolution plumbing, never API). The smaller points landed too: mode-as-file vs mode-as-key is recorded as a deliberate call with the tree-shaking rationale, Swift/Compose is named in the format list, the reference-chain lint is in Confirmation, the diagram says raw/, and spreadable textStyle objects are flagged as follow-up.

On @layer vs 0010: the two already agree — ADR-0010 got the matching line on #5199 back on 5 Aug (7a5b70c8a: the layer wrap is the bundle step's only added behaviour, keeping the bundle a pure function of the source files), and 0010 was marked Accepted today after beta validation of the bundle. Point 6 here reflects that.

260b3cd refreshes what went stale over the summer: ADR-0011 is Accepted (#5195 merged), the heading→header rename has landed in the export, and the border-focus/border-disabled/text-disabled duplicates were cleared from main by the #5280 restructure — the consequence now records that as a content fix, not a structural one, so the check stays required.

eddiman:
Re-reviewed against `260b3cdd`. All six points from my 5 Aug review are addressed — thanks for the thorough pass, and for the summer refresh on top.

Rewriting this one in plain language, since the ADR is going to be read by people who weren't in the token-pipeline weeds with us.

**The thing I most wanted checked, I checked myself.** Back in August three token names were broken: `border-focus`, `border-disabled` and `text-disabled` each existed twice, and in one case a token pointed at itself, which means it resolves to nothing — that's what killed the focus ring on Button and Chip. You say #5280 cleared them. It did. I looked at the generated CSS on `main` and those names are gone, and none of the 1,231 declarations in the bundled `variables.css` point at themselves any more. The 207 repeated names that *are* in there are just the three density files each declaring the same set, which is how it's supposed to work. Your framing is right too: that was a fix to the token *content*, not to the machinery that allowed it, so the check stays required.

**Point 7 is the important addition.** It says which parts of the output are a promise to consumers and which are internal wiring we can rename whenever we like. That matters because the package ships far more names than we intend people to use, and there's no way to tell them apart by looking. Worth stating now — once beta consumers start depending on an internal name, changing it becomes a breaking change.

Five things left. None of them change the architecture.

---

### 1. The pipeline diagram is missing a step

The diagram shows Tokens Studio exporting CSS into `src/tokens/css/`, and then that going into the bundle. So you'd reasonably conclude the committed CSS is exactly what Studio produced.

It isn't. A script called `widen-semantic-scope.mjs` runs in between and rewrites those files. It changes which elements the semantic tokens apply to — from just the page root to the page root *plus* anything marked with a colour scheme — which is what makes dark mode work on a subtree instead of only on the whole page (#5226). The ADR doesn't mention this anywhere.

To be fair to what point 6 actually says: the claim "the layer wrap is the bundle step's only added behaviour" is **correct**. The rewriting happens before the bundler, and the bundler itself really does just glue files together. But this ADR is the document that describes what the pipeline produces, and there's a step between Studio and the committed files that it doesn't describe.

While you're there — one detail worth a sentence. The bundler deliberately puts the colour-scheme files **last**. That's stricter than what point 6 says ("defaults before modes"), and it's there for a reason: if two files declare the same name with equal weight, the last one wins, and the colour-scheme values need to be the winners. Right now nothing depends on it, because the duplicate names are gone. But a future `@layer` change that reorders files could bring the focus-ring bug back, and the only place this rule is currently written down is a comment inside a script.

### 2. Point 7 covers five of the seven output folders

The build emits seven: `color-scheme`, `colors`, `density`, `elevation`, `font`, `primitives`, `semantic`. Point 7 says semantic, font and density are the promise, and primitives and colour-scheme are internal. That leaves `colors/` and `elevation/` unassigned — and today the beta package publishes all seven.

Both should be internal:

- `ts/colors/*` is the raw colour ladder (`dark.blue.1` through `12`) — the ingredients, not the finished thing.
- `ts/elevation/*` is the individual pieces of a shadow: how far down it sits, how blurred it is, what colour. The finished shadow already exists in the semantic set as `elevation.high` and `elevation.low`. Its own description in Tokens Studio says exactly that: *"Semantic face over the `shadow.*` parts in `elevation/default`, mirroring how `spacing.*` faces `density.spacing.*`."* And the generated `ts/semantic/{light,dark}.ts` gives you both a ready-made CSS `boxShadow` string and a React-Native-shaped `layers[]` array. So nobody downstream needs the loose pieces.

Worth noting why this differs from `density`, which point 7 *does* publish: density is published because an app switches between compact and comfortable at runtime and React Native needs all three. Elevation has one version. There's nothing to switch between.

### 3. The ADR says it's based on a branch that no longer exists

Lines 13, 103 and 156 say everything is grounded in the `tokens-studio-release` branch. That branch is created fresh for each release PR and deleted when it merges — it isn't in the repo right now, so a reader can't go and check. The output lives on `main` at `packages/eds-tokens/src/tokens/`. Since `260b3cdd` was specifically about refreshing the claims against main, the grounding sentence should point there too.

### 4. The collision check in Confirmation doesn't exist yet

Confirmation says a duplicate-name check gates the release workflow. There's no such step in `tokens_studio_release.yaml`, and no collision check anywhere on `main` — the only matches are the bundler comment that describes the problem and some lockfile noise. The `check-collisions.mjs` I pointed you at in August is in local tooling, not in this repo.

That's fine for a Confirmation section, which is partly about what we intend to enforce. But it should either carry a tracking issue like the other pending items, or say plainly that it isn't built yet — otherwise someone reads it as a guarantee that's already in place.

### 5. Two small things about ADR-0010

The link `./0010-single-bundled-css-entry-for-eds-tokens-3.md` will be dead until #5199 merges, since the file isn't on main yet. And this ADR calls 0010 "proposed" in both point 6 and Related, but it's now marked Accepted on its branch. Whichever of the two PRs merges second can clean this up.

---

Separate from this PR, but worth an issue so the next person doesn't work out the wrong story from scratch: the header comments in `widen-semantic-scope.mjs` and `generate-css-bundle.mjs` still describe the three duplicate names as a live problem and point at #5221 for tracking. #5221 is closed and #5280 cleared the names.

Leaving my changes-requested in place only because items 1 and 2 touch the decision text itself. Both are small, and I'm happy with where this landed.

pomfrida:
Thanks — all five in 5ad6c87d0. Widen step and colour-scheme-last order are now in the diagram and point 6, colors/ and elevation/ are internal in point 7, grounding points at main, and the collision check is marked not built yet with #5407 tracking it. #5408 covers the stale script comments you flagged.

documentation/adr/0014-token-code-output-architecture.md
```diff
@@ -0,0 +1,168 @@
+# Token code / output architecture for the redefined token system
+
+- **Status:** Proposed
```
eddiman:
```suggestion
- **Status:** Accepted
```

when we merge it in

eddiman (dismissed):

Issue #5141, opened 2026-07-01
fix: pin EDS 2.0 beta releases to a fixed 3.0.0-beta.N series


pomfrida:
## Problem

The `/next` (EDS 2.0) beta channel of `@equinor/eds-core-react` is versioned along the stable 2.x line instead of as a prerelease series of the upcoming EDS 2.0 major.

Current state: `@beta` = `2.8.0-beta.1`, `@latest` = `2.6.1`. Two issues:

1. **Wrong target.** EDS 2.0 is documented to graduate as `3.0.0` (see `documentation/how-to/BETA_RELEASE_GUIDE.md` → batch graduation to a single major `v3.0.0`), but the betas are labelled `2.x.0-beta.N` — i.e. previews of stable 2.x minors that actually come from EDS 1.0 work, not EDS 2.0. The version misrepresents what it is a preview of.
2. **The counter doesn't accumulate.** Each `feat(next)` bumps the base minor (`release-type: simple` + `prerelease-type: beta`, no way to pin the base), so the published history is `2.6.0-beta.0 → 2.7.0-beta.1 → 2.8.0-beta.1` — a new base almost every time, with the beta counter resetting/jumping instead of climbing.

## Desired behaviour

Follow the standard major-beta-cycle pattern (React/Node style): a **fixed base `3.0.0`** with an **incrementing counter** — `3.0.0-beta.1`, `3.0.0-beta.2`, `3.0.0-beta.3`, … regardless of whether commits are `feat(next)` or `fix(next)` — until EDS 2.0 graduates to stable `3.0.0`.

The pinned `3.0.0-beta.N` must be consistent **everywhere**: npm (`@beta` dist-tag), `version.txt`, the release-please manifest, the `/next` CHANGELOG headings, and the git tags.

## Why this needs custom automation

release-please has no versioning strategy that keeps the base pinned while incrementing the prerelease counter across `feat` commits (`versioning: prerelease` still rolls the minor on `feat`; `always-bump-patch` strips the prerelease). So the beta number must be owned outside release-please's commit-type-driven bump: the publish workflow computes `3.0.0-beta.N` and writes it back to all release artifacts.

## Scope

- [ ] Reset the `/next` beta target to `3.0.0-beta.0` (`release-please-config.json`, `release-please-manifest.json`, `version.txt`).
[...]

Issue #5120, opened 2026-06-25
📦 eds-tokens packaging & release (beta flags → 3.0.0)


eddiman:
## Description
Package and release strategy for the redefined tokens. Ship under **beta flags** during the transition, then a full major release — deliberately **not rushed** (last time tokens were pushed out of beta too quickly; this time the beta is verified against real components first).

## Scope
- During transition: `eds-core-react` continues under a **beta flag**; `eds-tokens` gets the new token structure under a **beta flag**.
- Iterate and verify the beta against real components, gather feedback, adjust before going stable.
- **Full release:** EDS core (react) + tokens released together as one major (likely `3.0.0`) — a single clear breaking change for both.
- Keep the existing EDS 2.0 tokens in place but **mark them "do not use."**
- Pure component-consumers: no change.

## Definition of Done
- [x] Beta-flagged eds-tokens published and verified against components
- [x] Old tokens marked "do not use"
- [x] Release notes / changelog prepared

Timeline: stable release **not this year** — beta first, iterate on feedback.

pomfrida:
Implementation notes from the (now closed) duplicate #5172, so they live here:

- The beta flag can mirror the existing `eds-core-react-next` setup in `.github/release-please-config.json` (`prerelease-type: beta`) — publish `3.0.0-beta.x` under the npm dist-tag `beta` while 2.x stays `latest`, so testers keep their import specifiers unchanged (`npm i @equinor/eds-tokens@beta`)
- On the beta line: add the pipeline-generated directories (`src/tokens/{css,dtcg,ts}` from the Tokens Studio release workflow) to the package `files` field and repoint the `exports` map specifiers (`./css/variables` etc.) — today none of the generated output is published
- Decide the fate of the `./ts/*` subpath exports with the mobile team — `equinor/design-system-mobile` imports six TS modules, and the new density dimension has no `spacious` (see #5164 for the usage mapping)
- npm trusted publisher allows only one workflow per entry — a new/changed publish workflow for the beta line needs its trusted-publisher entry updated
- Blocked by the token architecture rework (#5139) landing, so the beta does not ship names that immediately change again

pomfrida:
All Definition of Done items are complete. The final item was resolved in #5191, published as `eds-tokens@2.3.2` on `latest`: the package README now carries a transition notice (CSS variables supported today but renamed in the next major; the legacy `tokens` object deprecated), and the `tokens` export has `@deprecated` JSDoc.

[...]

PR #5180, opened 2026-07-16
ci: set up automatic beta releases for eds-tokens

pomfrida:
## What

Puts the **whole `@equinor/eds-tokens` package into beta** during the Tokens Studio rewrite and sets up the automatic release chain: every release is a **pinned `3.0.0-beta.N`** on the npm `beta` dist-tag. There is no stable release channel during the transition — no stable tokens releases are needed until the rework graduates as `3.0.0`, and consumers on `latest` simply stay on the last stable `2.3.1`.

**The chain:** Tokens Studio release → `tokens_studio_release.yaml` opens a `feat:` PR → merge → release-please adds an `eds-tokens` bump to the release PR → merge → `trigger_publish.yml` dispatches `publish_tokens.yaml` with `npm-tag=beta` → publishes `3.0.0-beta.N` with provenance. Any other `fix`/`feat` commit touching the package cuts a beta the same way.

Part of #5120.

## Changes

- **`release-please-config.json`**: the `packages/eds-tokens` entry becomes the beta line itself — `release-type: simple` with `versioning: prerelease` + `prerelease: true` + `prerelease-type: beta` (no separate `eds-tokens-next` component; tags and the existing CHANGELOG continue seamlessly). `src/tokens/README.md` is excluded from release detection.
- **`release-please-manifest.json`** + seeded `packages/eds-tokens/version.txt`: base pinned at `3.0.0-beta.0` (first published release will be `3.0.0-beta.1`)
- **`tokens_studio_release.yaml`**: PR commit/title switches from `chore:` to `feat:` so pipeline PRs trigger releases
- **`publish_tokens.yaml`**: `beta` tag option; sets the version from `version.txt` and injects the generated `src/tokens/{css,dtcg,ts}` into `files`/`exports` (temporary additive `./next/*` subpaths) at publish time, with a fail-fast guard if the pipeline output doesn't exist yet
- **`trigger_publish.yml`**: `trigger-tokens` dispatches `npm-tag=beta` during the transition (flip back to `latest` at graduation)
- **Docs**: `BETA_RELEASE_GUIDE.md`, `release-please-config.md`, `src/tokens/README.md`, `eds-tokens/CLAUDE.md`

## Why `release-type: simple` (version.txt) instead of `node` (package.json)

Four packages (`eds-core-react`, `eds-utils`, `eds-lab-react`, `eds-data-grid-react`) depend on `@equinor/eds-tokens` via `workspace:^`, and `pnpm publish` rewrites that to the version in eds-tokens' `package.json`. Keeping `package.json` at `2.3.1` means stable releases of those packages keep depending on stable tokens; the beta version lives in `version.txt` and is applied by the publish workflow at publish time.

## Why `versioning: prerelease` + `prerelease: true`

`prerelease-type: beta` alone does **not** increment the beta counter — the default strategy bumps the base and carries the suffix through unchanged (the exact drift from #5141: `2.6.0-beta.0 → 2.7.0-beta.1 → 2.8.0-beta.1`). Verified by running release-please's actual versioning code:

- without `prerelease: true`: a `fix` on `3.0.0-beta.0` releases **stable `3.0.0`**
- with this config: `fix`, `feat` and `feat!` all produce `3.0.0-beta.1 → beta.2 → …` — the base never moves

At graduation, flip `prerelease: false` and switch the dist-tag back to `latest`, and the same config emits stable `3.0.0`. (Side finding: the same config would fix the core-react drift in #5141.)

## Trade-offs

- An urgent stable 2.x patch during the transition is a manual publish from the `eds-tokens@v2.3.1` tag — accepted, since no stable tokens work is planned
- The first beta's changelog may sweep up feat/fix commits made since `2.3.1`

## Verification

- JSON + workflow YAML validated; release-please schema accepts the new fields and the factory passes them through to the versioning strategy (runtime-tested, see above)
- npm side needs no changes: the `beta` dist-tag already exists (`2.3.0-beta.3`, semver-below `3.0.0-beta.N`) and the trusted-publisher entry stays valid since the same workflow publishes
- Live verification after merge: trigger the pipeline (`gh api repos/equinor/design-system/dispatches -f event_type=tokens-release`), merge its `feat:` PR, verify the release PR shows `eds-tokens: 3.0.0-beta.1`, merge, and check `npm view @equinor/eds-tokens dist-tags`

## Out of scope (follow-ups for #5120)

- Repointing the stable specifiers (`./css/variables` etc.) to the Tokens Studio output so consumers keep imports unchanged — needs the first pipeline output in-repo plus a bundling step
- Marking the old tokens "do not use"
- Applying the same pinning fix to `eds-core-react-next` (#5141)


pomfrida:
@claude review

pomfrida:
Thanks for the review! All three points are addressed in bb03ce0df, which also reworks the PR to a simpler model after a design discussion: eds-tokens needs no stable releases during the Tokens Studio rewrite, so the stable/beta split is gone — the whole package is now the beta line (single `eds-tokens` component, `release-type: simple`, pinned `3.0.0-beta.N`, `latest` stays on 2.3.1 until graduation).

1. **Missing `exclude-paths`**: `packages/eds-tokens/src/tokens/README.md` is now excluded on the (single) eds-tokens entry, so docs-only edits there don't bump the beta counter. `version.txt` moved to the package root and the seeded `src/tokens/CHANGELOG.md` is gone — the existing package CHANGELOG continues instead.
2. **Raw TypeScript in `./next/ts/*`**: noted in `src/tokens/README.md` that the subpath ships raw `.ts` modules and requires a TS-aware consumer (same as the existing stable `./ts/*` exports, which also ship `.ts` files from `build/`).
3. **Concurrent publishes**: resolved structurally by the rework — with a single release channel there is no scenario where one release-PR merge dispatches two simultaneous `publish_tokens.yaml` runs.

pomfrida:
@claude review

pomfrida:
@claude review

PR #5191, opened 2026-07-20
docs: mark old token systems as deprecated and transitional

pomfrida:
## Summary

Addresses the remaining Definition of Done item in #5120: *Old tokens marked "do not use"*.

Marks the two existing token systems in `@equinor/eds-tokens` ahead of the Tokens Studio-based replacement:

- **README** (ships to npm): a transition notice at the top of Usage. The CSS variables lose the "(Recommended)" label and are flagged as transitional — still the supported option for production today (`eds-core-react` 2.x requires the CSS bundle), but the variable names will change in the next major. The legacy `tokens` JavaScript object is marked deprecated — do not use in new code. The beta line is flagged as testing/feedback only, not production.
- **`src/base/index.ts`**: `@deprecated` JSDoc on the `tokens` export, so consumers get the deprecation hint in their editors via `dist/types`.
- **`package.json`**: `2.3.1` → `2.3.2` so the marking can reach the stable line — the current publish setup only releases on the beta dist-tag, and npmjs.com shows the README of `latest`.
- **CHANGELOG**: manual `2.3.2` entry (release-please does not generate changelog entries for manual stable publishes).

### Why the wording is differentiated

An unconditional "do not use" would leave teams building today in limbo: the replacement only exists on the beta line and is not production-ready. So the legacy JS object gets a hard deprecation, while the CSS variables are marked "supported today, names will change in the next major — keep direct references few and centralised".

### Release-please safety

- `eds-tokens` uses `release-type: simple`, which tracks the version in `version.txt` — the manual `package.json` bump does not interfere with the beta line (currently `3.0.0-beta.1`).
- `docs:` is a hidden type, so this PR triggers no release bump.

## After merge

1. Run the **Publish tokens** workflow manually with npm-tag `latest` to publish `2.3.2` (docs-only, legacy build unchanged).
2. Verify the npm page shows the updated README, then tick the DoD checkbox in #5120.

`npm deprecate` on 2.x is deliberately **not** used now — it would print install warnings for every consumer, including transitively via `eds-core-react` 2.x which still requires these tokens. That is the right tool at 3.0.0 graduation instead.

### DIFF
--- a/apps/design-system-docs/docs/about/getting-started/develop/getting_started_development.md
+++ b/apps/design-system-docs/docs/about/getting-started/develop/getting_started_development.md
@@ -54,16 +54,20 @@ For consistent styling, install and use design tokens:
 npm install @equinor/eds-tokens
 ```
 
-```javascript
-import { tokens } from '@equinor/eds-tokens'
+Import the CSS variables and use them in your styles:
 
-const { typography, colors } = tokens
+```css
+@import '@equinor/eds-tokens/css/variables';
 
-const styles = {
-  fontSize: typography.heading.h1.fontSize,
-  color: colors.interactive.primary__resting.rgba,
+.my-component {
+  background-color: var(--eds-color-bg-neutral-surface);
+  color: var(--eds-color-text-neutral-strong);
 }
 ```
+
+:::note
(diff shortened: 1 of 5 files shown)


pomfrida:
@claude review

PR #5194, opened 2026-07-20
docs: add ADR for temporary next subpaths during eds-tokens beta

pomfrida:
## Summary

Documents the decision (already in effect since PR #5180) to publish the Tokens Studio pipeline output under temporary additive `./next/{css,dtcg,ts}/*` subpaths during the eds-tokens beta, injected into `files`/`exports` at publish time — instead of replacing the legacy build on the final specifiers immediately, or asking testers to install two package versions side by side.

## Contents

- **`documentation/adr/0009-temporary-next-subpaths-for-eds-tokens-beta.md`** — full context, the three options considered, and the exit plan: when the last `/next` component has migrated to the new variable names, a beta release drops the legacy `build/` output and moves the Tokens Studio output onto the final specifiers — inside the beta line, before graduation, so at least one beta exposes the exact 3.0.0 export surface.
- **`.github/workflows/publish_tokens.yaml`** — the comment above the publish-time injection now describes the full lifecycle and links the ADR.
- **`packages/eds-tokens/src/tokens/README.md`** — same lifecycle note and ADR link for anyone reading the generated-output docs.

## Notes

- Docs/comments only — no behaviour change; the injection itself landed in #5180.
- One open decision is called out in the ADR's consequences: the final specifier layout (granular `./css/*` only, or also a bundled convenience entry) must be settled before the flip.

### DIFF
--- a/.github/workflows/publish_tokens.yaml
+++ b/.github/workflows/publish_tokens.yaml
@@ -106,7 +106,11 @@ jobs:
       # The Tokens Studio pipeline output (src/tokens/{css,dtcg,ts}) is only
       # published on the beta line, additively under ./next/* subpaths. The
       # stable specifiers (./css/variables etc.) keep serving the legacy
-      # build until the new output replaces it at graduation.
+      # build while /next components migrate to the new variable names;
+      # once that migration completes, a beta release drops the legacy build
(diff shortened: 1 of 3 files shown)


pomfrida:
@claude review

PR #5199, opened 2026-07-20
docs: add ADR for the single bundled css entry in eds-tokens 3.0.0

pomfrida:
## Summary

ADR-0010 (status **Accepted**) settles the open consequence from ADR-0009 (#5194): the 3.0.0 CSS export surface mirrors 2.x — **one bundled `./css/variables` entry**, while the granular generated files stay unpublished internal pipeline artifacts.

Key verified facts behind the decision:

- Publishing granular files would freeze the internal file layout into public API while the token architecture is still being reworked
- Payload difference between the full bundle and a minimal subset is under 1 kB min+gzip (4.6 vs 3.7 kB)
- eds-core-react inlines the token CSS into its own dist, so only direct token consumers see the specifiers
- CSS variables fail silently, so multi-file granular imports are an error-prone setup surface

Colour-scheme behaviour is governed by ADR-0004 (attribute-driven, set once at root): the bundle is a plain concatenation (deliberately not minified: byte-stable output, reviewable release-PR diffs, no extra toolchain dependency — costs ~0.2 kB gzipped) with no added behaviour and no `prefers-color-scheme` fallback. 2.x's automatic system-dark was inherited from `light-dark()` (#3447) via the #4864 bug-fix rewrite, not a design decision, and does not carry over — flagged as a migration-guide item.

## Validation (completed on the beta line)

The ADR was proposed 2026-07-20 and is marked **Accepted** as of 2026-09-04:

- The bundle step (`generate-css-bundle.mjs`, #5201) is live; the bundle has shipped as `next/css/variables.css` in every beta since `3.0.0-beta.2`
- The /next Button and Chip token migrations (#5222, #5225) import the bundle
- The open question (mode-file set changing shape) resolved itself: the step globs `src/tokens/css/`, and the added `elevation` set flowed through without touching it

The original acceptance criterion (a beta exposing the exact 3.0.0 surface) gated this ADR on ADR-0009's flip — a mechanical export-map change that validates nothing decided here — and was dropped; see the Validation section in the ADR.

Related: ADR-0009 (#5194), companion ADR-0013 (#5200), epic #4740, #5108

## Tracking issues

- #5201 — implements the bundle step this ADR decides (done)
- #5202 — beta validation and marking the ADR Accepted before the flip (superseded: validation is documented in the ADR itself; the flip mapping is ADR-0009's exit plan)
- #5203 — migration-guide consequences (import line, colour-scheme changes)


pomfrida:
@claude review

millus:
The `variables.css` bundle exists and has shipped in `@equinor/eds-tokens@3.0.0-beta.2` through `beta.6`, and `generate-css-bundle.mjs` is live and references this ADR. The technical implementation is in place.

However, the ADR's own validation criteria state it should be marked Accepted after the bundle has been exercised by beta consumers — not just published. Currently the only usage is internal, through the draft Button and Chip migration PRs which haven't merged yet. No external consumers have picked up the bundle in practice.

Holding off on merging until the migration PRs land and the bundle has seen some real consumer usage, at which point the status can be updated from Proposed to Accepted.

pomfrida:
@millus Fair point on the criteria — but rather than holding the PR, I've updated the ADR itself (d87df49). The old acceptance criterion tied this ADR's status to ADR-0009's flip, which is a mechanical export-map change that doesn't validate anything decided *here*. What this ADR actually decided — one bundled entry, granular files internal — has been exercised already: the bundle has shipped in every beta since 3.0.0-beta.2 and is used by the Button/Chip migrations (#5222/#5225), and the open question about mode-file shape resolved itself when the elevation set was added and flowed through the globbed bundle step untouched. So the ADR is now marked Accepted with the validation notes inline, and I'd merge it as such.

millus (approved):
LGTM ✨

Issue #4863, opened 2026-04-28
fix: Switch has poor contrast in dark mode; disabled state appears more prominent than enabled


pomfrida:
**Describe the bug**

Reported by a user evaluating the EDS 2.0 beta Switch for an app with dark mode:

> Heisann, jeg tenkte å bruke Switch knappen fra EDS beta versjonen.
> Vi har dark mode i applikasjonen vår og ser knappen skal ha dark mode,
> men ser ut som om det kanskje har skjedd noe feil?

In the dark-mode story (https://storybook.eds.equinor.com/?path=/docs/eds-2-0-beta-inputs-selection-controls-switch--docs):

- Enabled (unchecked & checked) switches and their labels render with very low contrast against the dark background — the track blends into the background and the label text is barely legible.
- Disabled switches render with a **white** track, making them visually *more* prominent than the enabled states. This is the inverse of the expected hierarchy.

**Steps to reproduce the bug**

1. Open the Switch docs page in Storybook
2. Scroll to the "Dark mode" example
3. Compare the enabled rows ("Dark mode switch", "Checked in dark mode") with the disabled rows ("Disabled in dark mode", "Disabled checked")

**Expected behavior**

- Enabled switches and labels meet WCAG 2.1 AA contrast against the dark surface
- Disabled switches appear visually *less* prominent than enabled ones

**Additional context**

Likely related to how the Switch picks up tokens in dark mode. Suspect tokens used in `packages/eds-core-react/src/components/next/Switch/switch.css`:

- Track default: `--eds-color-bg-fill-muted-default`
- Handle default: `--eds-color-bg-fill-emphasis-default`
- Handle disabled: `--eds-color-bg-disabled` (resolves bright in dark mode → causes the inverted hierarchy)
- Label disabled: `--eds-color-text-disabled`

Worth checking whether these tokens have correct dark-mode values, or whether the Switch needs to use different semantic tokens.

Related: #4838 (separate Switch styling bug — missing `--eds-elevation-raised` token).

Screenshot: _to attach via GitHub UI_

pomfrida:
## Root cause

The Switch dark-mode bug is a downstream symptom of how Vite 8 processes the published `@equinor/eds-tokens` CSS.

**Timeline:**
[...]

PR #5200, opened 2026-07-20
docs: add ADR for attribute-scoped colour-scheme switching over light-dark()

pomfrida:
## Summary

ADR-0013 (status **Accepted**) records a decision that is already implemented in both token pipelines but was only documented in a commit message, a CLAUDE.md pitfall note, and a README section: **token CSS expresses colour schemes exclusively as attribute-scoped custom-property blocks, and `light-dark()` must never appear in published CSS.**

Why it deserves a permanent record:

- The #4863 production regression is the evidence: downstream bundlers (Vite 8 / Rolldown / lightningcss without targets) polyfill `light-dark()` into a `:root`-baked `var()` pattern that silently breaks subtree-scoped dark mode. #4864 rewrote the 2.x output to explicit scopes and added the `assert-no-light-dark` guard.
- `light-dark()` accepts only colour values, so density/spacing/typography modes need attribute scoping regardless — one switching mechanism for every axis (per ADR-0004).
- The 3.0 Tokens Studio output is attribute-scoped by construction (zero `light-dark(` occurrences). Without an ADR, the per-scheme duplication looks like something a future maintainer or code agent should "modernize" back into `light-dark()` — reintroducing the bug class.

The ADR also captures the consequences: no automatic system-dark (apps set `data-color-scheme` themselves), the FOUC/inline-script setup-docs requirement, reusing the assertion guard on the future 3.0 bundle, and an explicit revisit criterion.

Related: ADR-0004, ADR-0010 (#5199), #4863 / #4864

## Tracking issues

- #5201 — reuses the `assert-no-light-dark` guard on the 3.0 bundle
- #5203 — documents the colour-scheme consequences (no automatic system-dark, FOUC/inline-script setup note) in the migration guide
- #5204 — trims the 2.x README's light-dark section to reference this ADR

### DIFF
--- a/documentation/adr/0013-attribute-scoped-custom-properties-not-light-dark.md
+++ b/documentation/adr/0013-attribute-scoped-custom-properties-not-light-dark.md
@@ -1,0 +1,113 @@
+# Use attribute-scoped custom properties, not `light-dark()`, for colour-scheme switching in token CSS
+
+- **Status:** Accepted
+- **Date:** 2026-07-20
+- **Decision makers:** Frida Erdal, EDS Core Team
+
+## Context
+
+EDS publishes design tokens as CSS custom properties, and colour tokens
+carry different values per colour scheme. There are two mainstream ways to
+express that in published CSS:
+
+1. **`light-dark(lightValue, darkValue)`** — one declaration per token;
+   the browser picks the value from the element's `color-scheme` property.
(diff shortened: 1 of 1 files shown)


pomfrida:
@claude review

millus (approved):
LGTM ✨

PR #5206, opened 2026-07-21
ci: set cache headers and remove deploy downtime for storybook publish

### LINKED ISSUE #5205: Storybook prod serves stale mixed-version bundles after deploy (48h cache on index.html and runtime files)
## Bug

After each Storybook deploy, users who visited https://storybook.eds.equinor.com before the deploy can get a broken Manager UI:

```
TypeError: api.getNavAvailability is not a function
```

Reported 21 July 2026 by a user on Chrome and Edge (Safari worked, dev environment worked). A hard refresh fixed it, confirming the root cause below.

## Root cause

All blobs are served with `cache-control: public, max-age=172800` (48 hours) — including `index.html` and the un-hashed Storybook manager files (`sb-manager/runtime.js`, `sb-manager/globals-runtime.js`). After a redeploy, a browser can hold a cached bundle from the previous deploy while fetching a fresh one for another file, producing a mixed-version Manager UI that crashes.

`publish_storybook.yaml` uploads with `az storage blob upload-batch` without setting cache headers, so the 48h value comes from the storage account / Front Door defaults (the `x-fd-int-roxy-purgeid` response header shows Front Door in front).

The "Delete all existing blobs" step makes the window worse: users with a cached old `index.html` also get 404s on the deleted hashed chunks until their cache expires.

## Suggested fix

Set differentiated `Cache-Control` headers at upload time in `publish_storybook.yaml` (e.g. via `--content-cache-control` on separate `upload-batch` calls, or `az storage blob update` after upload):

- `no-cache` (or short `max-age` + `must-revalidate`) on `index.html`, `iframe.html`, `sb-manager/**` and `sb-addons/**` — ETag/Last-Modified already exist, so revalidation is cheap 304s
- Long `max-age` + `immutable` on content-hashed assets (`assets/*-<hash>.js`)

Verify that Front Door does not override the origin cache headers; adjust its caching rules if it does.

## Workaround

Hard refresh (Ctrl+Shift+R) or clearing site data for storybook.eds.equinor.com; the problem also self-heals when the 48h cache expires.

pomfrida:
Closes #5205

## Problem

Storybook prod serves every file with Front Door's default `cache-control: public, max-age=172800` (48h) because the upload sets no cache headers — including `index.html` and the un-hashed `sb-manager`/`sb-addons` bundles. After each deploy, browsers with a warm cache mix bundles from two different deploys and the Manager UI crashes with `api.getNavAvailability is not a function` (reported by a user on Chrome and Edge; hard refresh fixed it).

The workflow also deleted **all** blobs before uploading, taking the site down for the duration of the upload on every deploy.

## Changes to `publish_storybook.yaml`

- **Content-hashed `assets/*`**: uploaded with `Cache-Control: public, max-age=31536000, immutable` — filenames change on every content change, so caching forever is safe
- **Everything else** (`index.html`, `iframe.html`, `sb-manager/**`, `sb-addons/**`, …): uploaded with `Cache-Control: no-cache` so browsers revalidate on every load — cheap 304s via the ETag/Last-Modified headers already served
- **Zero-downtime deploy**: the delete-all-before-upload step is replaced with upload-with-overwrite followed by deletion of stale blobs only (those with `lastModified` older than deploy start). Hashed assets are uploaded before `index.html`, so the live page never references chunks that don't exist yet
- The stale-blob comparison is done in bash rather than a JMESPath query — JMESPath only defines ordering operators for numbers, so a date-string query could silently match nothing
- The existing `purge-cdn` job is unchanged; it handles the Front Door edge cache, this PR handles browser caches

## Verification

- YAML parses; `bash -n` passes on all run steps
- After merge, the push-to-main dev deploy allows checking headers with `curl -sI` (`no-cache` on `index.html`, `immutable` on `assets/*`)
- On the next production publish, verify Front Door honors the origin headers on https://storybook.eds.equinor.com — if it still serves `max-age=172800`, the Front Door caching rule overrides origin and needs an infra follow-up

### DIFF
--- a/.github/workflows/publish_storybook.yaml
+++ b/.github/workflows/publish_storybook.yaml
@@ -191,13 +191,23 @@ jobs:
 
           # Every deploy renames all content-hashed assets, so the whole
           # previous assets/* set is stale each run — parallelise the deletes.
+          # Cleanup is best-effort: the new content is already live at this
+          # point, and blobs that survive a transient delete failure are
+          # re-detected as stale on the next deploy. Don't fail the job.
+          set +e
           xargs -r -P 8 -I{} az storage blob delete \
             --container-name '$web' \
             --name '{}' \
             --connection-string "$AZURE_STORAGE_CONNECTION_STRING" \
             --output none < /tmp/stale-blobs.txt
+          rc=$?
+          set -e
 
-          echo "✅ Stale blob cleanup completed. Deleted $COUNT stale blobs."
+          if [ "$rc" -eq 0 ]; then
+            echo "✅ Stale blob cleanup completed. Deleted $COUNT stale blobs."
+          else
+            echo "⚠️ Some stale deletes failed (rc=$rc); orphans will be retried on the next deploy"
+          fi
 
       - name: logout 🔓
         run: az logout
--- a/.github/workflows/publish_storybook.yaml
+++ b/.github/workflows/publish_storybook.yaml
@@ -123,6 +123,8 @@ jobs:
           # after the upload (zero-downtime deploy: upload with --overwrite
           # first, delete stale blobs afterwards). Deleting by name set is
           # deterministic — timestamps would race the Azure server clock.
+          # Must run before the mv of assets/ below, or assets/* drops out of
+          # the set and the cleanup deletes all live assets.
           (cd packages/eds-core-react/storybook-build && find . -type f | sed 's|^\./||') \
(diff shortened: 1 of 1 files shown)


pomfrida:
@claude review

pomfrida:
@claude review

pomfrida:
@claude review

pomfrida:
@claude review

PR #5209, opened 2026-07-21
feat: add bundled variables.css to the Tokens Studio release pipeline

### LINKED ISSUE #5201: Add a bundled variables.css to the Tokens Studio release workflow
Part of #4740 (tooling track, alongside #5108).

Implement the bundle step decided in ADR-0010 (#5199): `tokens_studio_release.yaml` concatenates the generated CSS files in `packages/eds-tokens/src/tokens/css/` and minifies the result into a committed `src/tokens/css/variables.css`.

**Requirements**

- Plain concat + minify — the bundle adds no behaviour on top of the generated files (colour-scheme behaviour is governed by ADR-0004; no `prefers-color-scheme` fallback, per ADR-0013 / PR #5200)
- Glob the `css/` directory rather than hard-coding the file list, so future export-set changes are followed automatically
- Reuse the `assert-no-light-dark` guard from 2.x on the bundled output as a regression check
- The committed file becomes importable on the beta line as `@equinor/eds-tokens/next/css/variables.css` through the existing `./next/css/*` wildcard — no change needed in `publish_tokens.yaml`

**Done when** a beta release ships the bundle and it can be imported with one line.

pomfrida:
Closes #5201. Implements step 1 of the ADR-0010 (#5199) validation plan.

## What

- **New script `packages/eds-tokens/scripts/generate-css-bundle.mjs`** — globs `src/tokens/css/` recursively (excluding only the output file itself), concatenates the files in sorted path order, and writes the committed `src/tokens/css/variables.css` with a do-not-edit header. Plain concatenation — the bundle adds no behaviour on top of the generated files (colour-scheme behaviour is governed by ADR-0004; no `prefers-color-scheme` fallback, per ADR-0013).
- **Deliberately not minified** (deviation from the original ADR-0010/#5201 wording "concat + minify" — the ADR text in #5199 is updated accordingly): the committed file stays a pure function of the source files — no minifier-version churn in release-PR diffs, no third-party tool in the bundle step, and diffs stay reviewable line by line (the Tokens Studio release PRs get no CI, so the human diff review is the only check). Cost is ~0.2 kB gzipped (4.8 vs 4.6 kB); consumers that bundle minify with their own tooling anyway.
- **`scripts/assert-no-light-dark.mjs` parameterised** — optional file path and scheme-scope threshold arguments; defaults are unchanged so the 2.x `_build:css` chain behaves exactly as before. The threshold is validated (a non-numeric value would otherwise make the comparisons silently pass). The bundle script runs the guard on the output as a regression check (threshold 1 — the new export has exactly one block per scheme).
- **New workflow step in `tokens_studio_release.yaml`** — runs `generate:css-bundle` after the TS token generation, so every Tokens Studio release PR regenerates the bundle alongside the nine source files.
- **`src/tokens/css/variables.css` committed now** — generated from the current token state so the bundle ships in the next beta without waiting for the next Tokens Studio release. It becomes importable as `@equinor/eds-tokens/next/css/variables.css` through the `./next/css/*` wildcard `publish_tokens.yaml` already injects — no publish-workflow change needed.

## Notes

- The `comfortable` density is scoped at `:root` in the bundle (not `[data-density=comfortable]`) because it is the base density of the saved EDS-CSS export — this mirrors the source files as-is.
- Concat order is alphabetical by path: `var()` references are late-bound so order does not affect resolution; sorting only keeps the committed artifact deterministic.

## Verification

- Two consecutive runs produce byte-identical output (determinism + self-exclusion)
- Guard passes on the new bundle and, with defaults, still passes on the 2.x `build/css/variables.min.css`; fails correctly on inputs containing `light-dark(`, missing a scheme scope, or given a non-numeric threshold
- Bundle contains one `[data-color-scheme=\"light\"]` and one `\"dark\"` scope, `compact`/`relaxed` density scopes, and no `light-dark(`
- 44.5 kB raw / 4.8 kB gzipped


pomfrida:
@claude review

pomfrida:
@claude review

PR #5218, opened 2026-07-22
chore: replace HeaderMdx with Typography in icons preview story

pomfrida:
## Why

The `icons-preview--preview` story crashes in production Storybook with `TypeError: Cannot read properties of undefined (reading 'secondary')` — reproduced with a clean browser profile, so it is baked into the deployed build.

Storybook 10.5.0 (#5185) changed MDX heading anchors for keyboard accessibility (storybookjs/storybook#34368): `HeaderMdx` now renders a Storybook-UI button that reads `theme.color.secondary` from the manager theme. The canvas iframe has no Storybook `ThemeProvider`, so the story throws on render.

## What

Replace `HeaderMdx` with an EDS `Typography` h2 carrying the same `id`, so anchor deep-links keep working. Only the hover copy-link button is lost. Docs-blocks components should not be rendered in canvas stories — a comment in the story now says why.

Closes #5213

PR #5219, opened 2026-07-22
ci: use shared cache-header deploy action for all storybook publishes

### LINKED ISSUE #5212: Storybook cache-header fix (#5206) does not cover the release publish workflows
## Problem

The Manager UI crash from #5205 (`api.getNavAvailability is not a function`) is still occurring on https://storybook.eds.equinor.com after #5206 was merged.

#5206 only changed `publish_storybook.yaml`, but push-to-main runs of that workflow deploy to the **development** environment. Production Storybook is actually deployed by the `publish-storybook` job in `publish_core_react.yaml`, dispatched by `trigger_publish.yml` with `environment=production` on every release merge (verified: prod blob `last-modified` matches the release run for #5135, 20 July 09:34 UTC).

That job — and the equivalent jobs in `publish_lab.yaml` and `publish_data_grid.yaml` — uses `tibor19/static-website-deploy@v4`, which sets no `Cache-Control` headers, so Azure Front Door falls back to its 48h default (`public, max-age=172800`, confirmed with `curl -sI` on 22 July). Warm browser caches mix bundles from two deploys after each release and the Manager UI crashes.

## Fix

Extract the deploy logic from #5206 (immutable cache headers on hashed `assets/*`, `no-cache` on the rest, overwrite-then-delete-stale for zero downtime) into a shared composite action and use it in all four workflows:

- `publish_storybook.yaml` (replace inline script with the action — single source of truth)
- `publish_core_react.yaml`
- `publish_lab.yaml`
- `publish_data_grid.yaml`

## Verification after merge

- `curl -sI` on prod after the next release: `no-cache` on `index.html`, `max-age=31536000, immutable` on `assets/*`
- Confirm Front Door honors origin headers (if prod still serves `max-age=172800`, the Front Door caching rule overrides origin and needs an infra change)

Refs #5205, #5206

pomfrida:
## Why

The Manager UI crash from #5205 kept happening on https://storybook.eds.equinor.com after #5206 was merged. #5206 only changed `publish_storybook.yaml`, but push-to-main runs of that workflow deploy to the **development** environment — production Storybook is deployed by the `publish-storybook` job in `publish_core_react.yaml`, dispatched with `environment=production` on every release merge (verified: prod blob timestamps match the release run for #5135). That job — and the same jobs in `publish_lab.yaml` / `publish_data_grid.yaml` — used `tibor19/static-website-deploy@v4`, which sets no `Cache-Control` headers, so Azure Front Door falls back to its 48h default (`public, max-age=172800`, confirmed with curl) and warm browser caches mix bundles from two deploys after each release.

## What

- New composite action `.github/actions/deploy-storybook` with the deploy logic from #5206: immutable cache headers on hashed `assets/*`, `no-cache` on everything else, zero-downtime overwrite-then-delete-stale, and a guard refusing cleanup on an empty deploy set. One addition: the `assets/` upload is skipped when the build has no such directory.
- All four publish workflows use the action — `publish_storybook.yaml` (replacing its inline script, so there is a single source of truth), `publish_core_react.yaml`, `publish_lab.yaml`, `publish_data_grid.yaml`.
- The publish jobs check out `.github` explicitly (sparse), because the setup job's sparse checkout only includes `packages/apps/scripts` and the workspace cache glob skips dot-directories.
- The `az login`/`logout` steps in `publish_storybook.yaml` are dropped: `--connection-string` is complete authentication for the data plane on its own.

## Verification after merge

- Dispatch `publish_core_react.yaml` with `environment=production` (or wait for the next release), then `curl -sI` prod: expect `no-cache` on `index.html` and `max-age=31536000, immutable` on `assets/*`.
- If prod still serves `max-age=172800`, the Front Door caching rule overrides origin headers and needs an infra change outside this repo.

Closes #5212

### DIFF
--- a/.github/actions/deploy-storybook/action.yml
+++ b/.github/actions/deploy-storybook/action.yml
@@ -1,0 +1,130 @@
+name: 'Deploy Storybook to Azure Blob Storage'
+description: >-
+  Zero-downtime upload of a static Storybook build to the $web container with
+  correct Cache-Control headers: content-hashed assets/* are cached forever,
+  everything else must be revalidated on every load. Without these headers
+  Azure Front Door applies its 48h default TTL and warm browser caches mix
+  bundles from two deploys, crashing the Manager UI (#5205, #5212).
+inputs:
+  folder:
+    description: 'Path to the built Storybook (static website root)'
+    required: true
+  connection-string:
+    description: 'Connection string for the target Azure Storage account'
+    required: true
+runs:
+  using: 'composite'
+  steps:
+    - name: Deploy to Azure Blob Storage 🚀
+      shell: bash
+      env:
+        AZURE_STORAGE_CONNECTION_STRING: ${{ inputs.connection-string }}
+        FOLDER: ${{ inputs.folder }}
+      run: |
+        set -euo pipefail
+        echo "Deploying to Azure Blob Storage..."
+
+        if [ ! -d "$FOLDER" ]; then
+          echo "❌ Error: folder '$FOLDER' not found"
+          exit 1
+        fi
+
+        FILE_COUNT=$(find "$FOLDER" -type f | wc -l)
+        echo "Uploading $FILE_COUNT files..."
+
+        # Capture the full set of blob names this deploy produces. Blobs not
+        # in this set are orphans from a previous deploy and are cleaned up
+        # after the upload (zero-downtime deploy: upload with --overwrite
+        # first, delete stale blobs afterwards). Deleting by name set is
+        # deterministic — timestamps would race the Azure server clock.
+        # Must run before the mv of assets/ below, or assets/* drops out of
+        # the set and the cleanup deletes all live assets.
+        (cd "$FOLDER" && find . -type f | sed 's|^\./||') \
+          | LC_ALL=C sort > "$RUNNER_TEMP/deployed-blobs.txt"
+
+        # 1) Content-hashed assets: cache forever. Uploaded first so the new
+        #    index.html never references chunks that don't exist yet.
+        if [ -d "$FOLDER/assets" ]; then
+          az storage blob upload-batch \
+            --destination '$web' \
+            --destination-path assets \
+            --source "$FOLDER/assets" \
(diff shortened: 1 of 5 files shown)


pomfrida:
@claude review

pomfrida:
@claude review

pomfrida:
Review follow-ups addressed in 67511fb:

- `action.yml` now documents the az CLI runner dependency, that static-website hosting / container access policy are pre-existing account configuration the action does **not** re-apply (unlike `tibor19/static-website-deploy`), and that the stale cleanup requires one storage account per environment.
- All four workflows now state explicitly that the sparse `.github` checkout must stay the **first** step (checkout cleans the workspace and would wipe the restored build).

On the 🔴 environment-scoping question — confirmed with production data that `AZ_STORYBOOK_CONNECTION_STRING` resolves to separate storage accounts per environment:

- 22 July 09:34 UTC: manual `workflow_dispatch` with `environment=production` → prod (`storybook.eds.equinor.com`) `index.html` last-modified 09:38.
- 22 July 10:37 UTC: push to main (#5218 merge) → `environment: development` → dev (`s478stedsstorybookdev.z16.web.core.windows.net`) `index.html` last-modified 10:41, while prod stayed at 09:38.

A dev deploy does not touch the prod container, so the sole-writer assumption behind the stale cleanup holds per environment.

Bonus: the 09:34 production dispatch also proves Azure Front Door honors origin `Cache-Control` — prod now serves `no-cache` on `index.html`/`sb-manager/**` and `public, max-age=31536000, immutable` on `assets/*`. This PR is still needed so the next release (which deploys via `publish_core_react.yaml`) doesn't regress prod to the 48h default again.

pomfrida:
Addressed the remaining 🟢 in f482df7: dropped the redundant `--connection-string` flag from every `az storage` command — the CLI reads `AZURE_STORAGE_CONNECTION_STRING` from the step env, and passing it as an argument would expose the secret in process args on the runner. A comment in the action now documents why the flag must not be reintroduced.

PR #5222, opened 2026-07-22
refactor: migrate Button (next) colours to beta tokens

pomfrida:
Resolves #5208 (phase 1 of #5119). First component migrated onto the Tokens Studio beta colour tokens — the test case for the migration pattern.

**Updated 2026-08-06** after tokens release #5280: the branch is merged with `main` and rewritten from the interim `--_tone-N` step indirection to the semantic `interactive` token group that release added. The Storybook wiring has landed on `main` (#5224/#5239), so the diff is now `button.css` only.

**Updated 2026-08-07**: design confirmed in #5283 that the missing neutral text/icon interactive tokens are deliberately out of scope — neutral interactive text uses `--eds-text-primary` by design, so the `--eds-text-primary` mapping in this PR is the intended final state, not a stand-in. That was the last token gap Button was waiting on.

**Token availability for consumers**: the interactive token group is published — `@equinor/eds-tokens@3.0.0-beta.4` (cut right after #5280 merged) ships it in the Tokens Studio bundle under the publish-time `./next/css/*` export (ADR-0009: consumers import `@equinor/eds-tokens/next/css/variables.css`). Note that the legacy `./css/variables` export does *not* contain the group, so apps on the /next beta need the `next/css` import. Prod Storybook predates the #5224/#5239 wiring; the next Storybook deploy from `main` picks the tokens up automatically.

## Changes

`button.css`:

- All colours come from the semantic interactive group:
  - Primary fill: `--eds-background-interactive-{tone}-emphasis-{default,hover,pressed}`; label: `--eds-text-on-emphasis`
  - Secondary/ghost hover/pressed backgrounds: `--eds-background-interactive-{tone}-muted-{hover,pressed}`; text: `--eds-text-interactive-{tone}-default`
  - Secondary border: `--eds-border-interactive-{tone}-emphasis-default`
  - Focus: `--eds-border-interactive-focus`
  - Disabled: `--eds-{background,text,border}-interactive-disabled` — the legacy-token fallbacks and the TODO(#4741) are gone. If #4741 ships alpha-channel values they land as value changes on these same tokens, with no component change needed.
- A small `--_` indirection remains at the component root purely so `data-color-appearance` can switch the accent/neutral/danger token set; states read those variables.

## Design notes

- The `text-interactive` group deliberately has no neutral variant (confirmed in #5283) — the neutral tone maps secondary/ghost text to `--eds-text-primary` (neutral-12) by design.
[...]

Issue #5221, opened 2026-07-22
Author import-ready Tokens Studio JSON for the missing beta colour tokens


pomfrida:
Part of #4740, follow-up from the Button migration test case (#5208).

The Button migration surfaced gaps and bugs in the Tokens Studio beta export that must be fixed at the source. The `studio` CLI is pull-only, so the changes are authored as token JSON in the same structure as the pulled raw sets (`packages/eds-tokens/src/tokens/raw/`) and imported via the Studio UI by design.

## Scope

Author proposal JSON covering:

- Alpha-channel disabled tokens (#4741) — as **explicit values**, not modify formulas (only the CSS export evaluates colour formulas)
- A semantic action/fill group for buttons (the current `background-surface-accent-*` values are unsuitable as solid fills; Button consumes steps 9/10/11 directly in the meantime)
- Accent text/border semantic tokens (secondary/ghost Button currently uses steps 12/8)
- Neutral/danger equivalents of the accent-only semantic surface group (tone support)
- Fix the `border-focus` self-reference in the semantic set (concept vs semantic name collision; `border-disabled` has the same collision with two conflicting values)
- Step-based state naming per #4744, reusing existing colour values per #4742

## Notes

- The platform has no undo — changes go onto a Studio work branch with design review before merge and release
- Alpha percentages need design sign-off to become canonical

## Definition of Done

- [ ] Import-ready JSON authored and shared with design
- [ ] Names and values reviewed by design
- [ ] Imported to Tokens Studio on a work branch and shipped in a beta release
- [ ] Verified after `studio tokens pull` that the `src/tokens/css/` output contains and resolves the new tokens

pomfrida:
Additional export/palette finding from the Chip migration (equinor/design-system#5225), dark scheme:

**The dark step scale looks mis-generated around the subtle ladder (steps 2/4/6):**

- Step 6 is *identical* to step 3 in every dark scale (e.g. `--eds-dark-north-sea-3` and `--eds-dark-north-sea-6` are both `oklch(0.47 …)`), and step 4 (L 0.52) is *lighter* than step 6. The semantic subtle ladder `--eds-background-surface-*-subtle-{default,hover,pressed}` = steps 2/4/6 therefore resolves to L 0.226 → 0.52 → 0.47 in dark: a huge jump on hover and a pressed state lighter than hover. (The legacy light palette has a similar duplicate, 6 == 4, so this may be inherited generator behaviour rather than new.)
[...]

Issue #5208, opened 2026-07-21
[Button] Migrate to new beta colour tokens


pomfrida:
Part of #5119 (phase 1 — colour). First component migrated onto the redefined tokens from the eds-tokens beta, serving as the test case for the migration pattern.

## Scope

- Switch Button (`/next`) onto the expanded static colours — expanded scale + step-based state names (#4742, #4744), replacing dynamic colour tokens with static ones where the migration prescribes it
- Apply the new alpha-channel disabled states (#4741)
- Consume tokens via the bundled `next/css/variables.css` entry (#5201) — this doubles as validation input for ADR-0010 (#5202)

## Definition of Done

- [ ] Button uses the new beta colour tokens (no references to the old colour tokens)
- [ ] Disabled states use the alpha-channel tokens
- [ ] Visual regression checked against current Button
- [ ] Learnings captured to guide the remaining component migrations

Issue #5119, opened 2026-06-25
🧩 Component migration (colour, then spacing & typography)


eddiman:
## Description
Migrate EDS components onto the redefined tokens. Per the 23 Jun decision, **all current (and future) EDS 2.0 components migrate** — no prioritised subset. Pure component-consumers see no change. Run in two phases, **colour first**.

## Phase 1 — switch to the expanded static colours (first)
- Move components onto the **expanded static colours** — expanded scale + reworked step-based state names (see #4742 and #4744), shipped alongside the current set.
- Includes the dynamic → static swap and applying the new **alpha-channel disabled states** (#4741).

## Phase 2 — switch to new spacing & typography tokens
- Apply the decoupled spacing / typography tokens.
- **Remove text-trim and reset component paddings** to fit all sizes (the work the text-trim change creates).

## Definition of Done
- [ ] Phase 1: all components on the expanded static colours
- [ ] Phase 2: all components on new spacing & typography tokens; text-trim removed; paddings reset
- [ ] Density modes (`compact` / `comfortable` / `relaxed`) validated across representative components — moved here from #4743. Components consume `--eds-density-*`; the legacy `data-density="spacious"` scheme does not survive the rewrite.
- [ ] Visual regression tested
- [ ] No consumer-facing breakage for pure component-consumers

pomfrida:
Suggest adding a Phase 1 DoD point: disabled styling is keyed off the native :disabled state (:has() on wrappers), not a prop-set data-disabled attribute — so controls inside a disabled fieldset render correctly. This was fixed for Checkbox/Radio/Switch in #5406; #5412 tracks the remaining Input-family components and points to their migration tasks here, so it's cheapest to do while each component's CSS is being migrated anyway.

eddiman:
**Phase 2 scope additions from the dev/design sync 17 September 2026.** The typography side of phase 2 was one line before this; it now has three concrete pieces, because #4655 closes and its remaining work moves here.

- **#5477, re-scoped.** The base element stylesheet, written on the redefined typography tokens with no text-box trim, giving `h1`-`h6` and `p` their default size and flow spacing. The decision was to write this fresh rather than ship Victor's existing `elements.css` as a stopgap, so #5481 is dropped.
- **#5501.** A CSS class per text style, applied on top of those defaults. This is the code counterpart of the Figma text styles, and it is what a consumer uses when the element default is not the size the design needs. It had no issue before.
[...]

PR #5224, opened 2026-07-22
chore: load the eds-tokens beta bundle in storybook for the /next migration

pomfrida:
Part of #5119. Infrastructure for the component-by-component colour migration (first consumers: Button #5222, Chip) — extracted from #5222 so component PRs can merge independently.

## Changes

- **`.storybook/preview.css`** — imports the Tokens Studio beta bundle alongside the legacy one, so both variable sets resolve during the mixed migration period (ADR-0009). Relative path because the `./next/*` exports are injected at publish time only. Imported **before** the legacy bundle so legacy wins the 8 `--eds-typography-header-*-font-size` name collisions.
- **`.storybook/preview.mjs`** — mirrors the colour scheme onto `<html>` for EDS 2.0 stories. The beta semantic tokens are declared at `:root` and reference scale aliases that only exist under `[data-color-scheme]`; custom properties substitute where they are declared, so the attribute must sit on the root element for the semantic layer to resolve (captured as a learning on #5208, export-side fix tracked in #5221).

No component or published-package changes — Storybook config only.

### DIFF
--- a/packages/eds-core-react/.storybook/preview.css
+++ b/packages/eds-core-react/.storybook/preview.css
@@ -1,3 +1,9 @@
+/* Tokens Studio beta bundle (ADR-0009: published consumers import
+   @equinor/eds-tokens/next/css/variables.css; the ./next/* exports are
+   injected at publish time only, hence the relative path locally).
+   Imported BEFORE the legacy bundle so legacy wins the name collisions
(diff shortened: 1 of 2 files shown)


pomfrida:
@claude review

PR #5225, opened 2026-07-22
refactor: migrate Chip (next) colours to beta tokens

### LINKED ISSUE #5223: [Chip] Migrate to new beta colour tokens
Part of #5119 (phase 1 — colour). Second component onto the redefined tokens from the eds-tokens beta, following the Button test case (#5208, PR #5222).

Chip has no disabled state, so unlike Button it can be migrated completely without waiting for the alpha-channel disabled tokens (#4741). It also exercises parts of the beta semantic layer Button never touched: the `selected` state maps onto the semantic selected groups (`background-surface-accent-selected-*`, `border-selected`), which validates whether those groups work in practice.

## Scope

- Switch Chip (`/next`) onto the expanded static colours — step-based state names, using the `--_tone-N` indirection pattern established in Button for `tone` support
- Use semantic tokens where they exist and fit (selected groups, `border-focus`, canvas); flag new gaps as learnings
- Visual regression checked against current Chip (all variants × tones × selected × light/dark)

## Definition of Done

- [ ] Chip uses the new beta colour tokens (no references to the old colour tokens)
- [ ] Selected states validated against the semantic selected groups
- [ ] Visual regression checked against current Chip
- [ ] Learnings captured on #5208/#5221 if new gaps surface

pomfrida:
Closes #5223. Part of #5119 (phase 1 — colour). Second component onto the Tokens Studio beta tokens, following the Button test case (#5222). Chip has no disabled state, so it migrates **completely** — no legacy colour references remain.

## Changes

`chip.css` only.

### Token mapping

All six tones consume the semantic `*-interactive-*` background groups via `--_` slots; `data-color-appearance` switches the tone at the chip root, with neutral as the default fallback.

- **Fills:** `background-interactive-{tone}-{muted,emphasis}-{default,hover,pressed}` for all three variants × all six tones.
- **Label text:** `--eds-text-on-muted-{tone}` (shade 12). Neutral maps to `--eds-text-primary` — the interactive group has no neutral text variant by design (#5283).
- **Icons:** `--eds-icon-on-default-{tone}` (shade 10). The text/icon shade difference (12 vs 10) is carried by the on-muted vs on-default surface distinction in the token architecture, not by text-vs-icon.
- **High-contrast text/icon:** `--_text-emphasis` / `--_icon-emphasis` slots, pinned to `--eds-{text,icon}-on-emphasis-neutral` for **all** tones — **verified against the Chip Figma component** (node 3853-3622): the design uses one shared flat-white text style (`#FFFFFF` = `neutral.15`) across all six tones, and every high-contrast fill matches `background-interactive-{tone}-emphasis-default` byte-for-byte. The per-tone `on-emphasis-{tone}` tokens (tinted near-whites) are not used by the Chip design.
- **Outlined border:** `--eds-border-interactive-{tone}-emphasis-{default,hover,pressed}` — the border steps along the state ladder (9 → 10 → 11) together with the fill, per the Figma bindings and ADR-0016's state-progression rule.
- **Focus ring:** `--eds-border-interactive-focus`, matching Button.

### Focus ring: transparent gap (was painted)

The focus ring previously used a two-layer `box-shadow` with the gap painted in `--eds-background-canvas` — a hardcoded guess at the backdrop that mismatches on any other surface. It now uses `outline` + `outline-offset` like every other `/next` component (Button, Radio, Checkbox, Switch, Accordion, Link), so the gap is genuinely transparent. **Ring geometry verified against Figma** (Button [EDS] `State=Focus`): the design specifies a 1px ring at 1px offset bound to `border/interactive/focus` — Chip's `stroke-thin` ring matches the design spec exactly (Button's 2px `stroke-thick` ring is the deviation from the design, not the reference). To free `outline`, the outlined variant's border moved to an inset `box-shadow` (visually identical). A `forced-colors` fallback keeps the outlined border visible in Windows High Contrast Mode, where `box-shadow` is stripped; the focus ring itself now also survives forced colors, which the old box-shadow ring did not.

## Visible changes (design-authored, not local choices)

- **Muted resting fills are shades 1/2/3** (previously 3/4/5 in beta.6). This is the deliberate repoint done in Tokens Studio 17–18 Aug as part of the on-surface restructure: resting chips read as quiet/flat against `background.surface`, gaining fill on hover/pressed. Confirmed intentional with the token team. **Note:** this deliberately supersedes the resolution of #5287 — `background-interactive-neutral-muted-default` again equals `background-canvas`. Recorded here so it isn't rediscovered as a regression a third time.
- **Neutral chips darken slightly:** `--eds-text-primary` / `--eds-icon-primary` moved from `neutral.12` to `neutral.13` in the token release.
- Everything else renders identically to the previous verified state — the text/icon token renames are value-identical (verified token-by-token), and outlined + high-contrast are pixel-equal against the deployed Storybook in both colour schemes.

## APCA (computed fresh, W3 0.1.9, from current `main` token values)

Label text against the chip's own fill (Lc, light / dark):

| Tone | muted default | muted hover | muted pressed | emphasis default | emphasis hover | emphasis pressed |
|---|---|---|---|---|---|---|
| neutral | 94.7 / 104.3 | 86.9 / 104.0 | 81.2 / 86.1 | 82.8 / 72.2 | 88.6 / 83.0 | 93.7 / 92.4 |
| accent | 89.9 / 98.7 | 82.2 / 98.4 | 76.6 / 80.1 | 81.5 / 73.5 | 87.5 / 84.0 | 92.9 / 93.2 |
| success | 89.8 / 99.5 | 82.1 / 99.2 | 76.6 / 80.4 | 80.5 / 74.7 | 86.7 / 85.0 | 92.4 / 94.0 |
| info | 89.6 / 97.2 | 82.2 / 96.9 | 76.5 / 78.7 | 81.6 / 72.5 | 87.6 / 82.8 | 93.1 / 91.9 |
| warning | 89.0 / 93.1 | 82.0 / 92.8 | 76.3 / 75.5 | 83.4 / 69.4 | 89.0 / 79.2 | 94.0 / 88.1 |
| danger | 87.0 / 87.0 | 79.2 / 86.7 | 73.3 / 69.9 | 83.0 / 59.4 | 88.8 / 70.8 | 94.0 / 81.5 |

Outlined variant against `background.surface` (the backdrop components are designed to sit on):

| Tone | text (light / dark) | icon (light / dark) | border (light / dark) |
|---|---|---|---|
| neutral | 98.8 / 105.5 | 98.8 / 105.5 | 77.5 / 71.1 |
| accent | 93.6 / 100.0 | 82.4 / 83.8 | 76.1 / 72.4 |
| success | 93.2 / 100.8 | 81.6 / 84.9 | 75.2 / 73.7 |
| info | 93.7 / 98.5 | 82.6 / 82.5 | 76.3 / 71.4 |
| warning | 94.4 / 94.3 | 84.0 / 78.6 | 78.1 / 68.1 |
| danger | 94.5 / 88.3 | 83.8 / 69.6 | 77.7 / 57.5 |

Muted resting fills measure Lc 0.0 against `background.surface` in both schemes (below the APCA clamp) — the quiet resting state described above. Dark-mode emphasis (high-contrast variant) is the lowest group in the set, with danger at Lc 59.4; these values are token-level and would change if design opts into the per-tone `on-emphasis` mapping.

## Verification

- 45/45 Jest tests pass.
- Browser check (local Storybook vs deployed prod Storybook, `ToneVariantMatrix`, both colour schemes): outlined and high-contrast render identically; default-variant fills show the deliberate muted repoint; focus ring verified on default and outlined variants (blue ring, transparent gap, coexists with the inset border).

## Learnings

- `selected` has no colour styling in Chip (check icon only), so the semantic selected groups (`background.interactive.*.selected.*`) were **not** exercised here — a later component or a Chip design update needs to validate those.
- The tone icon/text asymmetry the earlier inline comment attributed to text-vs-icon is expressed by the `on-muted` (12) vs `on-default` (10) surface split in the new token architecture.


pomfrida:
@claude review

pomfrida:
Re. review point 1 (text colour on default/outlined moving from `--eds-color-text-subtle` to `--_tone-12`): this is not a subtle → primary change. The legacy `--eds-color-text-subtle` was a *dynamic* token that resolved to step 12 of the active appearance scale — the published bundle defines it as `var(--eds-color-{tone}-12)` per `data-color-appearance` scope (e.g. `--eds-color-text-subtle: var(--eds-color-neutral-12)` under the neutral appearance). `--_tone-12` preserves that mapping exactly; only the semantic name is gone, since the beta export has no equivalent dynamic token yet.

The absolute step-12 values differ marginally because the whole beta palette is regenerated, but that applies to every step and is covered by the visual parity check against prod (see PR description).

Points 2 (`--eds-text-on-accent` naming on danger/warning) and 3 (tone-block repetition) are known: the naming gap is tracked in #5221, and the repetition matches the Button pattern by design.

pomfrida:
**Correction to the parity claim above:** the visual check against prod turned out to only cover light mode for Chip (dark was compared local-only). I've now completed the dark-mode comparison against prod with computed colours, and it is *not* full parity:

**Matches in dark**
- Outlined: text (step 12, L 0.91 vs legacy ≈0.90) and border (step 8, L 0.76) match.
- Default/outlined text: step 12 matches legacy.
- High-contrast fill: step 9 (L 0.82) matches legacy exactly.

**Deviations in dark**
1. **Default-variant resting fill is dramatically darker.** Legacy uses step 3, which in dark is L 0.47 — a clearly visible fill. The beta semantic subtle ladder (`--eds-background-surface-*-subtle-default` = step 2) is L 0.226 in dark, nearly identical to the canvas (step 1 = L 0.19), so default chips almost disappear into the background. The "step 2 vs 3" deviation noted in the PR description is trivial in light but large in dark.
2. **The dark scale looks like it has a generation bug**: step 6 (L 0.47) is identical to step 3, and step 4 is L 0.52 — so the dark hover/pressed ladder becomes 0.226 → 0.52 → 0.47 (huge jump on hover, pressed *lighter* than hover). This sits in the semantic subtle ladder itself (2/4/6), not in this PR's mapping.
3. **High-contrast text**: legacy is near-black in dark (rgb 3,3,3); beta `--eds-text-on-accent` resolves to north-sea-14 (L 0.33) — noticeably lighter on the pastel fills. Same class of deviation already flagged for light.

Items 1–3 are token-export issues rather than mapping mistakes — the PR deliberately follows the semantic subtle ladder. Flagged in #5221 for the token/design side. Whether this PR should stay on the semantic ladder (and let the fix land in the tokens) or pin steps per scheme is a design call.

pomfrida:
Converted to draft after massive tokens change today - needs updating

pomfrida:
Re-verified against local Storybook after the token realignment (4a813cc7a, merged main in first so the branch has the renamed tokens).

Quick background for readers without token context: every colour (neutral, accent, danger, …) comes as a row of 15 numbered shades, from lightest to darkest. The chip just picks shade numbers — shade 2 for the resting fill, 3 for hover, 4 for pressed — the same numbers the designers' own semantic tokens point to.

- **Light mode looks right.** The chips are one shade lighter than prod across the board (that's what the new design prescribes), but side by side the difference is barely visible.
- **The high-contrast variant is correct in both light and dark.**
- **Dark mode looks wrong, but the chip code isn't the problem — the dark shade values are.** In the dark palettes the shades don't get gradually darker the way they should: shade 2 is very dark, shade 3 is suddenly much lighter, and shade 6 is identical to shade 3. The result: a resting chip is nearly invisible against the background, and hovering makes it jump in brightness. That's the palette bug tracked in #5282 — once the dark shades are fixed, the chip picks them up automatically, no code change needed.
- **One question for the #5282 rework:** in dark mode, should resting chips be subtle (close to the background colour) or clearly grey like they are in production today? The palette fix alone doesn't decide that.

Holding this in draft until #5282 is resolved — the beta components are actively being tested, and dark mode shouldn't go out looking broken.

pomfrida:
Update after today's push: Chip is now on the semantic `*-interactive-*` groups, following the pattern from Button (#5222) — details in the refreshed PR description.

Good news on #5283: the success/info interactive tokens have landed in Tokens Studio and are in #5284. Their values are identical to the raw steps Chip mirrors in the meantime, so once #5284 merges the swap is a pure rename.

One new thing that needs a design call: `background-interactive-neutral-muted-default` resolves to neutral-1 — the same value as `--eds-background-canvas`. That makes the resting neutral default-variant chip an invisible pill on canvas in light mode (only the text shows). Button never hits this because ghost is transparent at rest, but Chip is a filled pill. Is neutral-1 intended for filled muted surfaces, or should it sit one step up? Happy to split this into its own issue if that's easier to track.

millus:
Regarding #5283, we are not using interactive colors on the Chip text, me and Edvard dont see the use for it right now, but can change later if we were wrong. We are using static --eds-text-success (etc.) for the text colors, since the text dont change on hover/pressed, just the background color. Its updated in figma with the new success and info colors from #5284 which we merged in yesterday. Just fyi, I did write the same but a bit longer explanation on #5283 as well, so we have track of the decisions both places.

millus:
Rebased onto beta.6 (tokens-studio-release #5331) — both original blockers are now resolved:

**#5282 (non-monotonic dark scale):** The dark neutral palette is now monotonic. Steps 3–6 progress correctly in lightness (0.47 → 0.52 → 0.58 → 0.595), so hover and pressed no longer jump unexpectedly.

**#5287 (neutral chip invisible on canvas):** `background-interactive-neutral-muted-default` now resolves to neutral-3 (L 0.47 in dark), clearly distinct from canvas at neutral-1 (L 0.19). The resting chip is visible in both light and dark mode.

Also updated text and icon tokens to use the separate `--eds-text-{tone}` and `--eds-icon-{tone}` static tokens (rather than the interactive variants), following the decision in #5283 that chip text colour does not change on hover/pressed.

millus:
@claude review

eddiman:
Heads-up from verifying #5283 — this PR's text colours will break on the next token pull.

`chip.css` sets `--_text: var(--eds-text-success)` and `var(--eds-text-info)`. Those exist in the repo's current token snapshot, but they no longer exist in Tokens Studio `main` or the published Figma library. `text.{success,info}` was replaced by `text.on-default|on-muted|on-emphasis.{tone}` as part of the same round that added the interactive success/info tokens this PR consumes.

`packages/eds-tokens/studio.lock` was last pulled `2026-08-12T14:11Z`, so `main` hasn't caught up yet — which is why it still builds. Once `studio tokens pull` runs, `--eds-text-success` and `--eds-text-info` disappear and these declarations fall back to whatever the cascade gives them.

Pick the replacement by what the text sits on:

- `--eds-text-on-muted-success` / `--eds-text-on-muted-info` for the muted (default) chip
- `--eds-text-on-emphasis-success` / `--eds-text-on-emphasis-info` for the emphasis variant

Worth checking the other tones in the file for the same pattern — `--eds-text-accent`, `--eds-text-danger` and `--eds-text-warning` are gone the same way.

The rest of the migration verifies clean: the interactive background and border tokens this PR uses are live in both Studio and Figma, and there are no raw shade numbers left in `chip.css`.

eddiman:
Correction to my comment above — the `--eds-text-{tone}` part is right, but I under-called the scope and the replacement mapping was too simple. Full check against live Tokens Studio `main`:

**Disappears on the next pull (all used in this diff):**

| Currently | Replacement |
|---|---|
| `--eds-text-{accent,success,info,warning,danger}` | `--eds-text-on-{default,muted,emphasis}-{tone}` |
| `--eds-icon-{accent,success,info,warning,danger}` | `--eds-icon-on-{default,muted,emphasis}-{tone}` |
| `--eds-text-on-emphasis` | `--eds-text-on-emphasis-{tone}` |
| `--eds-icon-on-emphasis` | `--eds-icon-on-emphasis-{tone}` |

I missed the icon tokens and the two flat `on-emphasis` ones entirely. The flat ones matter most — `text.on-emphasis` / `icon.on-emphasis` went from single tokens (`neutral.15`) to six-tone groups, so the high-contrast variant breaks for **every** tone, not just success and info.

**Survives unchanged:** `--eds-text-primary`, `--eds-icon-primary` (both now `neutral.13`, were `neutral.12`), `--eds-background-canvas`, `--eds-border-interactive-focus`, and every `--eds-background-interactive-{tone}-*` / `--eds-border-interactive-{tone}-emphasis-default` in the diff.

**On the mapping — I said two surfaces, there are three.** `on-default` is shade 10, `on-muted` is 12, `on-emphasis` is 15, and Chip renders on all three:

- `default` variant → muted fill (shades 1/2/3) → wants `on-muted`
- `outlined` variant → transparent, sits on the canvas → wants `on-default`
- `high-contrast` variant → emphasis fill (shade 9) → wants `on-emphasis`

The tone blocks set `--_text` / `--_icon` once and `high-contrast` overrides them, so `default` and `outlined` currently share one value but now want different ones. Either accept one value across both, or move `--_text` / `--_icon` down into the variant blocks. Your call which — flagging it rather than prescribing.

One knock-on: the inline comment at the top of the file says text and icon "differ for accent: shade 12 vs shade 10". Under the new tokens that distinction is carried by `on-muted` (12) vs `on-default` (10) instead of by text-vs-icon — `text.on-muted.accent` and `icon.on-muted.accent` are both `accent.12`. Worth rewording when you touch it.

None of this is urgent for review — `main` is still on the 12 Aug snapshot, so the current diff builds fine. It only bites when the token pull lands.

eddiman:
Concrete fixes. Twelve custom properties in `chip.css` stop existing when `studio tokens pull` runs. Here are the exact edits.

**1. Base block** (the neutral defaults, ~line 40) — add two lines, change nothing existing:

```css
    --_text: var(--eds-text-primary);
    --_icon: var(--eds-icon-primary);
+   --_text-emphasis: var(--eds-text-on-emphasis-neutral);
+   --_icon-emphasis: var(--eds-icon-on-emphasis-neutral);
```

**2. Each of the five tone blocks** — replace the two `--_text` / `--_icon` lines with four. Shown for `accent`; same shape for `success`, `info`, `warning`, `danger`:

```css
    &[data-color-appearance='accent'] {
      /* … background/border lines unchanged … */
-     --_text: var(--eds-text-accent);
-     --_icon: var(--eds-icon-accent);
+     --_text: var(--eds-text-on-muted-accent);
+     --_icon: var(--eds-icon-on-default-accent);
+     --_text-emphasis: var(--eds-text-on-emphasis-accent);
+     --_icon-emphasis: var(--eds-icon-on-emphasis-accent);
    }
```

**3. The `high-contrast` variant block** — point at the tone-scoped vars instead of the flat tokens:

```css
    &[data-variant='high-contrast'] {
-     --_text: var(--eds-text-on-emphasis);
-     --_icon: var(--eds-icon-on-emphasis);
+     --_text: var(--_text-emphasis);
+     --_icon: var(--_icon-emphasis);

      background-color: var(--_bg-emphasis);
      outline: none;
    }
```

This works because both selectors are attribute selectors on the same element and the variant block sits after the tone blocks in source order, so its `--_text` wins while `--_text-emphasis` still resolves from whichever tone applied.

**Nothing else changes.** Every `--eds-background-interactive-*` and `--eds-border-interactive-*` line in the diff is fine, as are `--eds-text-primary`, `--eds-icon-primary`, `--eds-background-canvas` and `--eds-border-interactive-focus`.

**On visual impact:**

- `default` and `outlined` variants render identically to today. `--eds-text-{tone}` is shade 12 and `--eds-text-on-muted-{tone}` is also shade 12; `--eds-icon-{tone}` is shade 10 and `--eds-icon-on-default-{tone}` is also shade 10. Straight rename, same values — and it preserves the text/icon asymmetry your inline comment describes.
- `high-contrast` **does** change for non-neutral tones. Today all five tones share the flat `neutral.15`; per-tone tokens give `{tone}.15`. That looks like the intent behind splitting them, but if you want byte-identical output for now, point all five `--_text-emphasis` / `--_icon-emphasis` at the `-neutral` variants instead.

Unrelated to this PR but landing in the same pull: `--eds-text-primary` and `--eds-icon-primary` shift `neutral.12` → `neutral.13`, so neutral chips darken slightly either way.

One caveat — I derived the custom-property names from the token paths (`text.on-muted.accent` → `--eds-text-on-muted-accent`), which matches how every other token in the file is emitted. Worth a quick grep against the generated `variables.css` once the pull lands.

pomfrida:
Applied your mapping in d77555f — text to on-muted, icons to on-default, plus --_text-emphasis/--_icon-emphasis slots for the high-contrast variant. We pinned those slots to the -neutral tokens for now to keep today's flat high-contrast look; if the per-tone split was meant to reach components, say the word and it's one line per tone. The inline comment about the text/icon shade asymmetry is reworded as you suggested, and the focus ring now uses outline with a transparent gap like the rest of /next (cae618e), so the painted-canvas gap is gone entirely. Fresh APCA numbers for the new values are in the updated PR description.

pomfrida:
Follow-up on the high-contrast question: checked the Chip component in Figma (node 3853-3622) and the design settles it. All six tones share one flat-white text style (#FFFFFF = neutral.15), and every high-contrast fill matches background-interactive-{tone}-emphasis-default exactly. So the -neutral pinning is the design, not a stopgap — no per-tone switch needed. Comment in chip.css updated to say so (6b80519).

pomfrida:
@claude review

pomfrida:
Addressed the review in 545b86a:

- **Transition trimmed to background-color only** — good catch on the outline-color fade; the focus ring appeared as a 150ms fade from the label colour instead of instantly blue.
- **Source-order comments added** above the tone blocks and the high-contrast block (the (0,2,0) tie noted in both reviews).
- **Colour comment rewritten** — the 12-vs-10 split is target-driven per the rule documented in #5330 (text aims for Lc 90 → step 12, icons for Lc 60 → step 10), which is more precise than both my earlier surface framing and the value-preserving framing.
- **--_border-interactive renamed to --_border** for slot consistency.
- **PR description now records explicitly that #5287's resolution is deliberately superseded** by the 1/2/3 repoint (confirmed with the token team), so it isn't rediscovered again.

On the **focus ring width**: checked the design rather than matching Button — the Button [EDS] Focus state in Figma specifies a 1px ring at 1px offset (stroke bound to border/interactive/focus, align outside). Chip's stroke-thin ring matches that spec; Button's stroke-thick is the deviation. Leaving Chip at 1px.

The **ring colour contrast** and the **dark-mode hover delta** are real and token-level — measured fresh in APCA and being raised with the token team separately.

pomfrida:
Ran a full binding-level reconciliation against the Chip [EDS] Figma component set (all 54 Composition=Default variants, comparing boundVariables via the REST API rather than rendered hex, with the known flat-name rename lag treated as equivalent).

**Result: everything matches** — default variant (all tones × states), high-contrast, outlined resting, text and fills — **except one real gap in the code, now fixed in de69d33:** Figma steps the outlined border along the state ladder (border.interactive.{tone}.emphasis 9 → 10 → 11) on hover/pressed, and chip.css kept it static at default. The border now follows the fill, and box-shadow is back in the transition list for it.

Two things for design to correct in the Figma bindings (no code impact):

1. **Outlined Hover and Active bindings are crossed** for all six tones: State=Hover binds muted/pressed + border/pressed, State=Active binds muted/hover + border/hover — the inverse of the default variant, and it contradicts ADR-0016's "ladders resolve in one direction, no inversions" rule. chip.css follows the ADR direction (hover → hover, pressed → pressed); the Figma bindings look swapped.
2. **Tone=Success, Variant=High Contrast, State=Hover** has its label text bound to icon/on-emphasis instead of text/on-emphasis (value-identical white, so purely a hygiene fix).

pomfrida:
Tokens are still under progress, keeping these as draft until after user testing and tokens have stabilised. We've been going a lot of back and forth here now so dev is pausing the migration atp

Issue #5226, opened 2026-07-23
Emit the semantic token layer under the [data-color-scheme] scopes


pomfrida:
Part of #5119 / #5108. Structural export problem found during the Button migration (#5208), separate from the token-content work in #5221.

## Problem

The beta export declares the semantic layer (`--eds-text-on-accent`, `--eds-background-surface-*`, …) in a `:root` block, while the scale aliases it references (`--eds-accent-N`, `--eds-neutral-N`, …) only exist under `[data-color-scheme="light"|"dark"]`. CSS custom properties substitute where they are **declared**, so on `:root` the references are guaranteed-invalid unless `data-color-scheme` sits on `<html>` itself.

Consequences (empirically confirmed in the Button migration):

- Subtree colour-scheme switching (`<div data-color-scheme="dark">`) does not work for the semantic layer — only the aliases declared in the scope rule re-resolve. This breaks the intent of ADR-0004/ADR-0013.
- Storybook needs a workaround (#5224): the colour scheme is mirrored onto `document.documentElement` for EDS 2.0 stories.
- The layer split amplifies the name-collision bugs (`border-focus` self-reference, `border-disabled` double value — content fixes tracked in #5221).

## Proposed fix

Emit the semantic layer under the same `[data-color-scheme]` scopes as the aliases (per scheme), which preserves subtree switching. The alternative — giving the aliases `:root` defaults — sacrifices subtree switching for semantic tokens and should only be chosen deliberately.

This is an export/pipeline change (export config / bundle step in `tokens_studio_release.yaml`), not a token-content change.

## Definition of Done

- [ ] Decision recorded: semantic layer emitted under the scheme scopes (or the `<html>`-attribute requirement explicitly documented as the contract)
- [ ] Export/pipeline updated accordingly
- [ ] Subtree colour-scheme switching verified with a migrated component (Button/Chip story wrapper, without the html mirror)
- [ ] Storybook workaround from #5224 removed if no longer needed

pomfrida:
**Decision (DoD item 1):** the semantic layer is emitted under the colour-scheme scopes, not documented as an `<html>`-attribute contract.

[...]

PR #5227, opened 2026-07-23
chore: author import-ready Tokens Studio JSON for the missing beta colour tokens

pomfrida:
Part of #5221, follow-up from the Button migration test case (#5208).

## What

Import-ready Tokens Studio proposal sets in `packages/eds-tokens/token-proposals/5221-beta-colour-gaps/`, generated deterministically from the current `studio tokens pull` by `build-proposal.mjs` (re-run it after a new pull to rebase the proposal). The three JSON files are complete **replacement** sets for their counterparts in `src/tokens/raw/`, so the diff against the raw sets is exactly the proposal. `README.md` in the folder is the design-review document: change list with rationale, computed-value table, open questions, and import instructions.

## Content

- **Alpha step twins** `{accent,neutral,danger}-alpha.{3,7}` in the scheme sets as explicit hex8 values (#4741 — the platform only evaluates colour formulas in the CSS export, so `modify` formulas are not an option). Each twin composites to its opaque step exactly over white (light) / black (dark); deviation over the actual canvas is ≤ ±12/255 and reported per twin. Colour math matches `scripts/generate-ts-tokens.mjs`.
- **Semantic fill group** `background.fill.{tone}.{default,hover,pressed,disabled}` → steps 9/10/11 + alpha-3 (the values the migrated Button consumes directly today).
- **Neutral/danger equivalents** of the accent-only `background.surface` group, mirrored step for step.
- **Tone text/border**: `text.{tone}` → step 12, `border.{tone}` → step 8. `text.accent` repointed from `{info.8}` (link blue, still available as `text-link`) to `{accent.12}`.
- **Name-collision fixes**: the semantic layer becomes the single owner of `border-focus` (→ `{info.7}`, removing the self-reference), `border-disabled` and `text-disabled` (→ `{neutral-alpha.7}`); the scheme sets drop their duplicates. `text-disabled` had the same double-value collision as `border-disabled` and is fixed the same way.

The generated artifacts are added to `.prettierignore` so they keep the studio-pull JSON style and diff cleanly against `src/tokens/raw/`.

## Verification

`build-proposal.mjs` resolves all 376 references across the proposal sets, checks that no flattened `--eds-*` name is declared by both the semantic and a scheme set, and asserts each alpha twin composites back to its opaque step exactly.

## Out of scope

- The `:root`-scoping export problem (#5226) — this PR is token content only.
- The Studio UI import itself: after design review, the sets are imported as replacements on a Studio work branch (removals must apply), released to beta, and verified after `studio tokens pull`.


pomfrida:
@claude review

pomfrida:
Closing as this was solved in Tokens Studio instead.

PR #5228, opened 2026-07-23
api change proposal for dialog component

FredrikMWold:
This PR proposes a change to the `Dialog` API.

The `Dialog` component becomes a pure context provider, while dialog-specific props are moved to a new `Dialog.Popup` component. It also introduces `Dialog.Trigger` and `Dialog.Close`, both of which support `asChild` and integrate with the internal dialog context to control the open state.

With this API, the dialog can manage its own state internally, removing the need for consumers to control it externally.

```tsx
<Dialog>
  <Dialog.Trigger>Open dialog</Dialog.Trigger>
  <Dialog.Popup>
    <Dialog.Header>
      <Dialog.Title>Dialog title</Dialog.Title>
    </Dialog.Header>
    <Dialog.Content>
      This is a short description of the action the user is about to take.
    </Dialog.Content>
    <Dialog.Actions>
      <Dialog.Close>Close</Dialog.Close>
      <Button onClick={() => {}}>Confirm</Button>
    </Dialog.Actions>
  </Dialog.Popup>
</Dialog>
```

This pattern is even more beneficial for components such as `Popover` and `Menu`. In addition to managing their own state internally, they can also manage the ref used to anchor the popup. `Popover.Trigger` can own this logic and, through `asChild`, attach the required ref to any element without exposing that implementation detail to consumers.

### DIFF
--- a/packages/eds-core-react/src/components/next/Dialog/Dialog.stories.tsx
+++ b/packages/eds-core-react/src/components/next/Dialog/Dialog.stories.tsx
@@ -33,23 +33,22 @@ See the stories below for usage patterns.
 export default meta
 
 export const Introduction: StoryFn = () => {
-  const [open, setOpen] = useState(false)
   return (
     <>
-      <Button onClick={() => setOpen(true)}>Open dialog</Button>
-      <Dialog open={open} onOpenChange={setOpen}>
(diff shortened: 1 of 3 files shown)


pomfrida:
@claude review

PR #5239, opened 2026-07-27
fix: emit the beta semantic token layer under the color-scheme scopes

### LINKED ISSUE #5226: Emit the semantic token layer under the [data-color-scheme] scopes

pomfrida:
Closes #5226

## Problem

The Tokens Studio CSS export declared the semantic layer (`--eds-background-*`, `--eds-text-*`, …) in a `:root` block, while the scale aliases it references (`--eds-accent-N`, `--eds-neutral-N`, …) only exist under `[data-color-scheme="light"|"dark"]`. CSS custom properties substitute where they are **declared**, so the semantic tokens resolved once at the root and never re-resolved inside `[data-color-scheme]` subtrees — subtree colour-scheme switching (a dark panel inside a light app, per ADR-0004) was broken for the semantic layer, and Storybook needed an `<html>`-mirror workaround (#5224).

## Fix

The semantic layer — the only layer that references the scheme-scoped scale aliases — is widened to `:root, [data-color-scheme]` by a new post-export step, `scripts/widen-semantic-scope.mjs`, chained before `generate-css-bundle.mjs` in the `generate:css-bundle` package script (so the release workflow needs no change). It is thereby also declared on any element carrying the attribute, and `var()` references substitute against the nearest scope. The script is idempotent and fails loudly if the export layout changes.

The saved **EDS-CSS** export configuration keeps the default `rootSelector: ":root"`. An earlier revision of this PR set the export's global `rootSelector` instead, but that widened *every* non-dimensional layer and regressed the density cascade (the density base declares the same `--eds-density-*` names as the compact/relaxed variants, so re-declaring it on `[data-color-scheme]` elements clobbered a `[data-density]` ancestor's values — caught in review). The export format has no per-layer selector, hence the repo-side step.

- Token content is byte-identical — the diff vs `main` is the semantic layer's selector line (source + bundle) only
- `generate:ts-tokens` output is unchanged (the generator parses declarations, not selectors)
- The `<html>`-mirror `useEffect` in the Storybook decorator is removed; the wrapper div's own `data-color-scheme` is now sufficient

**Collision handling (names tracked in #5221):** `border-focus`, `text-disabled` and `border-disabled` are declared in both the color-scheme layer and the semantic layer (including a `--eds-border-focus: var(--eds-border-focus)` self-reference). Both blocks apply to `[data-color-scheme]` elements at equal specificity, so the bundler concatenates the color-scheme files **last** — the scheme-specific values win and every resolved value matches the pre-widening state. This matters concretely: the incoming Chip (#5225) and Button (#5222) migrations use `var(--eds-border-focus)` for their focus rings, and letting the semantic self-reference win would have dropped focus outlines inside scoped subtrees. The duplicate names themselves remain a token-content bug to fix upstream (#5221).

## Storybook wrapper canvas

The wrapper's background token — the legacy `--eds-color-bg-neutral-surface` — is only declared under `:root, [data-color-scheme="light"]` in the 2.x bundle (no dark value), so the story canvas stayed light in dark mode. Rather than patching the legacy pipeline (being replaced, not maintained), the wrapper now uses the beta semantic tokens `--eds-background-container-canvas-default` / `--eds-text-primary`.

## Verification

Checked in a running Storybook from this branch (computed values via the browser, no `<html>` attribute present):

| Context | Value |
| --- | --- |
| `--eds-background-container-canvas-default` in light wrapper | `oklch(98% 0 0)` |
| … inside nested `<div data-color-scheme="dark">` | `oklch(19% .0014 242.96)` |
| `--eds-density-spacing-md` at root (comfortable) | `16px` |
| … on a `[data-density="compact"]` container | `12px` |
| … on a `data-color-scheme` child inside that container | `12px` (ancestor density preserved — review finding 1) |

Wrapper canvas follows the toolbar toggle: dark gives `background: oklch(0.19 …)` / `color: oklch(0.91 …)`, light `oklch(0.98 …)` / `oklch(0.37 …)`.


pomfrida:
@claude review

pomfrida:
Review findings addressed in cb4119b8a:

**1. Density regression — fixed.** The saved EDS-CSS export configuration is reverted to the default `rootSelector: ":root"` (updated on the platform), and the widening is now done repo-side for the semantic layer only: `scripts/widen-semantic-scope.mjs`, chained before the bundler in the `generate:css-bundle` package script, so the release workflow is untouched. The export format has no per-layer selector (checked `exports schema --format css`), hence the post-export step. Verified in Storybook: `--eds-density-spacing-md` on a `data-color-scheme` child inside a `[data-density="compact"]` container stays `12px` (was clobbered to `16px` with the global widening), while subtree colour-scheme switching for the semantic layer still works.

**2. Token collisions — documented, deferred to #5221.** The three duplicate names (`border-focus` self-reference, `text-disabled`, `border-disabled`) are upstream token-content bugs already tracked in #5221 (called out in #5226's problem statement as well). The bundle-script header now records that the widened semantic block wins those at equal specificity and why ordering can't fix it. The suggested no-duplicate assertion would fail today because the duplicates exist; worth adding once #5221 lands — noted there.

**3. Minor points — fixed.** The ADR-0013 reference is dropped from the decorator comment (the ADR is still unmerged in #5200), the wrapper comment now says the legacy bundle scopes its *scheme-dependent* tokens rather than all of them, and the bundle header's "source order does not affect resolution" claim is corrected. Consumer-override reset on scoped elements is now limited to the semantic layer, which is the layer where per-subtree re-declaration is the intended behaviour.

**Not verified / legacy collisions:** with the widening narrowed to the semantic layer, the overlap surface against legacy is the semantic names only — the committed-source scan finding no overlap should carry over, and the "legacy wins collisions" import order in preview.css is unaffected for every layer that stays `:root`-scoped.

pomfrida:
@claude review

pomfrida:
Round-two findings addressed in bbdc0e44c:

**1. Prettier fragility — fixed, with one correction to the premise.** The selector is now matched with whitespace-tolerant regexes and emitted in Prettier's canonical one-selector-per-line form, with the pattern shared between both scripts via `scripts/semantic-scope.mjs` (also resolves the duplicated-constant minor). However: the generated directory was never Prettier-conformant to begin with — the export emits double-quoted attribute selectors (`[data-color-scheme="dark"]`) and no trailing newline, so `prettier --check src/` fails on the same 33 files on `main` too; this PR's files were not the first. The right fix for that class of problem is the same one `build/*` already has: `src/tokens/*` is now in `.prettierignore` (root and package-local), since the directory is machine output and any hand-formatting would be reverted by the next export run. The package `prettier:check` passes for the first time. The tolerant regexes stay as defence in depth.

**2. Semantic glob gap — fixed.** The widen script now widens every `semantic/*.css` it finds and the bundler asserts every one of them, mirroring its own directory glob. A rename still fails loudly; an addition can no longer ship narrow.

**3. Bundle-level regression test + eds-tokens tests not running in CI — agreed, but deferred to its own issue.** Wiring `test:tokens` into the root `test` script changes what the Checks workflow runs for every PR (and interacts with the build-before-test ordering that `typography-shape.test.ts` already depends on), which is beyond this PR's scope. Will open a dedicated issue covering the missing `test:tokens` wiring plus the `variables.css` scope-assertion test modelled on `density-aliases.test.ts`.

Verified after the changes: the chained `generate:css-bundle` is idempotent and reproduces the committed bundle, the bundler exits 1 on an unwidened semantic file before writing, `generate:ts-tokens` output is unchanged, and `assert-no-light-dark` still passes.

pomfrida:
@claude review

pomfrida:
Round-three items addressed in 5e41f852e:

**1. Recursive mismatch — fixed.** The widen script now globs `semantic/` recursively, matching the bundler's assertion glob. Verified with a nested `semantic/extra/nested.css`: previously the chain deadlocked with a no-op remedy message; now the file is widened and the chained `generate:css-bundle` succeeds.

**2. Non-canonical whitespace — normalised.** A selector matching the tolerant regex but not the canonical form is rewritten to `WIDE`, so the committed bytes are a function of content only (verified with the `:root,   [data-color-scheme]\t{` variant).

**3. Workflow description — updated.** The release workflow's bundle step is renamed/documented and the auto-PR body now notes that the `semantic/*.css` selector diff comes from `widen-semantic-scope.mjs` (#5226), so tokens-release PR reviewers aren't left wondering why the export "changed" a selector.

**4. Deferred follow-ups — now traceable:** #5246 (run the eds-tokens vitest suite in CI) and #5247 (regression test asserting the `variables.css` scope structure, modelled on `density-aliases.test.ts`). The two `.prettierignore` copies stay: the package-local file also covers ad-hoc `prettier --write src/` runs from the package directory, which a `--ignore-path` flag on the package script alone would not.

pomfrida:
One more hardening after a pre-merge worst-case pass (48353e6d4): the round-two review's finding 2 turned out to matter concretely — the incoming Chip (#5225) and Button (#5222) migrations both use `var(--eds-border-focus)` for their focus rings, and with alphabetical concatenation the widened semantic block's `--eds-border-focus` **self-reference** won on `[data-color-scheme]` elements, computing to the guaranteed-invalid value and dropping focus outlines inside scoped subtrees.

The bundler now concatenates `color-scheme/*` after every other layer, so the scheme-specific values win the three #5221 collisions and every resolved value matches the pre-widening state. Verified in the browser: `--eds-border-focus` resolves to the light/dark blue per subtree (per-subtree switching for it now works too), and `text-disabled`/`border-disabled` keep their scheme values. Cross-layer name overlaps were re-checked declared-name against declared-name: semantic ∩ {colors, font, primitives, density} = 0, so the reorder affects exactly the three tracked names.

PR #5320, opened 2026-08-11
chore: subtree-import design-system-mobile into the monorepo

Chibuzor-Nwemambu:
Part of #5138 (step 1: Subtree import).

Targets `5138-mobile-repo-migration`, the integration branch collecting every step of this migration before it goes to `main` as a single reviewed PR.

## Summary

Structural-only import of `design-system-mobile` into this monorepo, split so each destination keeps its own scoped commit history instead of nesting the whole foreign repo under one path:

- `packages/eds-mobile-components/` ← `design-system-mobile`'s `packages/components`
- `apps/mobile-storybook/` ← `design-system-mobile`'s `apps/storybook`

## How this was done

A plain `git subtree add --prefix=packages/eds-mobile-components <mobile-repo> main` merges the foreign repo's *entire* tree under one prefix — that would have nested the whole mobile monorepo (its own `packages/`, `apps/`, `pnpm-workspace.yaml`, `turbo.json`, root `package.json`, etc.) inside a single path. Instead:

1. Cloned `design-system-mobile` standalone.
2. In that clone: `git subtree split --prefix=packages/components -b components-split` and `git subtree split --prefix=apps/storybook -b storybook-split`, producing synthetic branches scoped to just each subdirectory's own history.
3. Fetched both split branches into this repo and ran `git subtree add --prefix=packages/eds-mobile-components components-split` / `--prefix=apps/mobile-storybook storybook-split`.

Each import commit carries only its own directory's real history (92 commits apiece, with some overlap where a single mobile commit touched both directories, e.g. release commits).

## Not included (deferred to later steps in #5138)

- Dropping Turbo / hand-sequencing build scripts
- Folding in `eslint-config-eds-mobile` / `tsconfig-eds-mobile` as workspace packages — `packages/eds-mobile-components/tsconfig.json` currently extends a path (`../tsconfig/base.json`) that doesn't exist yet in this repo; expected until that step lands
- npm trusted publisher configuration
- release-please + publish workflow wiring
- Mobile docs via workspace link in `eds-core-react`
- Repo metadata (CODEOWNERS, templates, labels, branch protection)
- Issue transfer / archiving the old repo

## CI note

`checks.yaml` only runs on PRs targeting `main`, so this PR against `5138-mobile-repo-migration` won't trigger it. Known-broken state (tsconfig extends, lint on Expo Router's default-export routes) is expected until the drop-Turbo/fold-configs steps land on the integration branch — everything needs to be green before the integration branch's own PR into `main`.

## Merge strategy

This PR will squash-merge into `5138-mobile-repo-migration` per the repo-wide squash-only setting — fine, the original 97 commits stay permanently visible on this PR's Commits tab and via `refs/pull/<this-PR>/head`. The one-time exception being arranged with the team is for the final `5138-mobile-repo-migration` → `main` merge, using a merge commit there instead so each migration step lands as its own distinct commit in `main`'s history; that setting gets reverted right after.


Chibuzor-Nwemambu:
@claude review

PR #5329, opened 2026-08-12
chore: drop Turbo, hand-sequence mobile scripts, align pnpm/Node/React

Chibuzor-Nwemambu:
Part of #5138 (step 2: Drop Turbo, hand-sequence scripts).

Targets `5138-mobile-repo-migration`, same as the subtree-import PR — collecting every migration step there before it goes to `main` as one reviewed PR.

## Summary

Hand-sequences root scripts for the two mobile packages imported in step 1, fixes a few things that surfaced while getting `test:mobile` actually green, and addresses the full round of findings from the `@claude` review plus @millus's question.

## Changes

- **Root scripts:** `build:mobile`, `watch:mobile`, `check-types:mobile`, `dev:mobile`, `start:mobile` — the middle two build `packages/eds-mobile-components` first via `pnpm --filter`, mirroring the existing `tokens → eds-core-react` hand-sequencing pattern. Named `<verb>:mobile` to match the convention already used for libraries (`build:core-react`, `test:core-react`), not the `<pkg>:<verb>` app convention (`demo:dev`) — `test:mobile` was already shaped this way. `dev:mobile` resolves to `expo run:ios` (macOS/Xcode-only); `start:mobile` is the cross-platform Metro-only alternative, documented in both `CLAUDE.md` files. No `turbo.json`/dependency existed to actually drop — the subtree-split import (see the previous PR) never pulled in mobile's own Turbo config in the first place.
- **`test:mobile`** added and wired into the root `test` aggregate, so CI's `pnpm run test` actually covers mobile (it previously wouldn't have, even once everything else worked).
- **pnpm/Node alignment:** `packageManager` bumped `10.15.0` → `10.28.0`, `engines.node: ">=22"` added — matches `design-system-mobile`'s own requirements, which the original plan claimed were "already matching" but weren't. Also matches an existing de facto standard: `packages/eds-tokens-build`/`eds-tokens-sync` already declared `node: ">=22"` themselves.
- **React version alignment:** `react`/`react-dom` bumped from an exact `19.2.0` pin to `19.2.8` on `apps/mobile-storybook`, matching what the rest of this monorepo already uses. `react-test-renderer` added as an explicit devDependency on `packages/eds-mobile-components` (previously undeclared — silently worked in the source repo only because its `.npmrc` uses `node-linker=hoisted`). A `pnpm.overrides` entry (`"react-test-renderer@19": "19.2.8"`, scoped to the v19 line per review feedback) collapses a duplicate copy `jest-expo` otherwise bundles internally at `19.2.0`.
- **CLAUDE.md fixes** in both mobile packages: dev commands and paths that were correct in `design-system-mobile`'s own root tooling, but don't exist in this monorepo; corrected a claim that `lint:all` meaningfully covers this package (it runs without crashing, but applies root's style rules instead of this package's own config — thousands of Prettier-noise findings, not real signal).
- **Fixed a real regression this PR introduced**: `eslint-plugin-import` was peering to `eslint-import-resolver-typescript@3.10.1` instead of the root's own `4.4.5` — confirmed via `git show` against the pre-PR lockfile that this didn't exist before this PR (most likely the pnpm minor bump changing dedupe tie-breaking). Fixed with a `pnpm.overrides` entry; smoke-tested both real consumers of the old v3 (`eslint-config-expo`, `eslint-config-next`) to confirm nothing broke.
- **Stale `pnpm@10.15.0` references** fixed repo-wide: `AGENTS.md`, `README.md`, and the three Dockerfiles. `prerequisites.json` deliberately left alone — separate, already-stale generated file with no CI check verifying it; whether to keep it at all is an open question for later.
- **Excluded mobile from `react18-compat.yaml`**: that workflow's React 18 override doesn't reach `eds-mobile-components`'s own exact-pinned `react` devDependency, so mobile's suites were running under React 19 there for zero React 18 signal — pure cost, and a false-alarm risk for unrelated PRs. Confirmed non-blocking either way (that workflow isn't part of the required `"All checks passed"` status).
- **Removed an unused `react-dom` devDependency** from `eds-mobile-components` (zero imports in `src/`, already declared as a `peerDependency` separately).

## Verified

- `pnpm run test:mobile` — 13/13 suites, 109/109 tests pass, no regressions across every change above.
- `pnpm run build:mobile` / `check-types:mobile` still fail, but only on the already-tracked missing `packages/eds-mobile-components/tsconfig.json` base config (fixed on the separate step-3 PR, #5333) — confirmed not a regression by checking the failure is unrelated to anything in this diff.
- Confirmed no other hardcoded pnpm version pins exist in `.github/workflows/`, and the lockfile format (`lockfileVersion: '9.0'`) is unaffected by the pnpm minor bump.

## Not included

- `packages/eds-mobile-components/tsconfig.json`'s extends path and the missing `@expo/vector-icons` peerDependency — both fixed on the separate step-3 PR (#5333), stacked on top of this one.
- Mobile's test files aren't type-checked anywhere (pre-existing since step 1, not a regression here) — needs a dedicated `tsconfig.test.json` plus a `rootDir` fix, and can't be tested cleanly until #5333's tsconfig base lands. Tracked as its own follow-up.


Chibuzor-Nwemambu:
@claude review

millus:
engines.node: ">=22" is added to the root package.json, which means it applies to the entire monorepo — not just the mobile packages. I can see it aligns with .node-version (22.12.0), so Node 22 is arguably already the de facto requirement, but was it intentional to enforce this for all packages and not just packages/eds-mobile-components?

I'll wait a bit on the full review pass until the points raised by the claude review above are addressed — happy to take another look after that.

Chibuzor-Nwemambu:
Addressed finding #2 (`eslint-plugin-import` peered to an older resolver version) in 4b2bd0a9d.

Confirmed this was a real regression from this PR, not pre-existing — right after step 1 merged, root's `eslint-plugin-import` correctly peered to `eslint-import-resolver-typescript@4.4.5`. Something in this PR (most likely the pnpm `10.15.0` → `10.28.0` bump changing dedupe/tie-break behavior) flipped that to `3.10.1`, pulled in via `eslint-config-expo` (new from step 1's `apps/mobile-storybook`).

Fix: a `pnpm.overrides` entry forcing `eslint-import-resolver-typescript` to `4.4.5` everywhere.

Worth flagging explicitly: this has a wider reach than just mobile. The same override also forces `eslint-config-next`'s own internal resolver usage (used by `eds-demo` and `eds-color-palette-generator` — unrelated to this migration) from `3.10.1` to `4.4.5`. I tried scoping it to just root's own usage first (`"eslint-plugin-import>eslint-import-resolver-typescript"`), but confirmed empirically that selector syntax doesn't take effect with pnpm — only the plain top-level override does.

Smoke-tested both real consumers of the old v3:
- `apps/mobile-storybook`'s own lint (`eslint-config-expo`) — exit 0, clean.
- `eds-demo`'s lint (`eslint-config-next`) — fails with `Invalid project directory provided`, but I confirmed via `git stash` that this exact failure exists with or without this override — pre-existing, unrelated to this change.

`import/no-unresolved` (the only rule this resolver affects) is `warn`-level everywhere in root's config, so the worst case if v4 resolves anything differently for `eslint-config-next` is a changed set of warnings, not a build/lint failure.

millus (approved):
LGTM

Chibuzor-Nwemambu:
@millus Yes, adding `engines.node: ">=22"` at the root was intentional, and it's less of a scope expansion than it looks — `packages/eds-tokens-build` and `packages/eds-tokens-sync` already declare `"node": ">=22"` in their own `package.json`, and `.node-version` (`22.12.0`) already governs the whole repo today, not just mobile. This just formalizes what was already the de facto floor. It's non-blocking either way — `engine-strict` isn't set anywhere in this repo, so it produces a warning on an older Node, not a hard failure. One related thing I found while checking: `apps/design-system-docs` still declares its own `"node": ">=18.0"`, now stale relative to root's floor — unrelated to this PR, so I left it alone, but flagging it since it's now a real inconsistency.

Since your last look, I've also addressed the rest of the `@claude` review: fixed the stale `pnpm@10.15.0` references repo-wide (except `prerequisites.json`, deliberately left as a separate follow-up), excluded mobile from the React 18 compat workflow (it wasn't getting real React 18 signal there), fixed a real regression this PR introduced in `eslint-plugin-import`'s resolver peering (confirmed via `git show` against the pre-PR lockfile, not just asserted), corrected a misleading CLAUDE.md claim about `lint:all` covering mobile, scoped the `react-test-renderer` override to the v19 line, removed an unused `react-dom` devDependency, and renamed the `mobile:*` scripts to `<verb>:mobile` for consistency with the rest of the root scripts (added `start:mobile` as the cross-platform alternative to the macOS-only `dev:mobile`). The one item I deliberately did not fold in is mobile's test files not being type-checked anywhere — that needs a dedicated `tsconfig.test.json` and can't be tested cleanly until step 3's tsconfig fix lands on its own branch, so it's tracked as a separate follow-up rather than growing this PR further.

PR #5334, opened 2026-08-14
chore: switch mobile docs to workspace link, fix stale sourceUrls

Chibuzor-Nwemambu:
Part of #5138 (step 6: mobile docs via workspace link).

## Summary

Switches `eds-core-react`'s dependency on `@equinor/eds-mobile-components` from a real npm version to `workspace:^`, and fixes the hardcoded `sourceUrl` links that the dependency swap alone doesn't touch.

## Changes

- **Dependency swap:** `@equinor/eds-mobile-components` in `eds-core-react`'s `package.json` changed from `^0.3.0` to `workspace:^`. Verified via the lockfile and a direct symlink check that `docs/*.mdx` content now resolves from the local `packages/eds-mobile-components`, not a downloaded npm copy.
- **Scoped properly:** searched the whole repo for every consumer of the package rather than guessing filenames. Found 10 relevant files:
  - **7 files** import `MobileDocs`/`MobileIntro` *and* hardcode a `sourceUrl` pointing at the old external `design-system-mobile` repo — `Button`, `Checkbox`, `Input`, `Radio`, `Switch` (`packages/eds-core-react/src/components/next/`), `EdsProvider` (legacy tree, not under `next/` — it's version-agnostic, sitting above the EDS 1.0/2.0 split entirely), and `Typography.new.docs.mdx` (**not** `Foundation/Typography.docs.mdx`, a different, unrelated foundation/type-scale doc with no mobile content). Verified each target file actually exists at its new path before writing it — most just needed the repo/base-path swapped, but `Button`'s hardcoded path was flat (`src/components/Button.tsx`) while the real structure is now a nested folder (`src/components/Button/Button.tsx`).
  - **1 file** (`stories/mobile/About.mdx`) imports `MobileIntro` with no `sourceUrl` — the dependency swap alone was sufficient.
  - **2 files** (`stories/docs/Intro.mdx`, `stories/mobile/Components.mdx`) only reference the package by name/npm-badge/link — confirmed nothing to fix.
- **`Typography.new.docs.mdx` needed a judgment call, not a mechanical fix:** the original `sourceUrl` never pointed at a real file, even in the source repo — there's no flat `Typography.tsx`, only `TypographyUI.tsx` and `TypographyHeader.tsx`, composed together in `index.ts`. Pointed `sourceUrl` at `index.ts`, since that's where the actual exported `Typography`/`Typography.Header` pair is composed, and the embedded mobile docs on this page describe both.
- **Fixed a stale example in `packages/eds-mobile-components/CLAUDE.md`** (caught by self-review): its own "consumer-side wrapping pattern" documentation still showed the old `design-system-mobile` URL as the canonical pattern for future MDX handoffs — would have taught contributors to reintroduce this exact bug.

## Verified

- Dependency resolves via a real symlink (`node_modules/@equinor/eds-mobile-components -> ../../../eds-mobile-components`), confirmed in both the lockfile (`specifier: workspace:^`, `version: link:../eds-mobile-components`) and on disk.
- Every `sourceUrl` target file confirmed to exist at its new path — not just a syntax check.
- Self-review confirmed the `workspace:^` swap is isolated to `eds-core-react`'s own devDependencies, no other package affected, and the lockfile diff only reflects this one change.

## Not included

- Whether `TypographyNext`'s docs page should exist at all (reportedly the team has agreed EDS 2.0 won't have a `Typography` component) and what a fully standalone, web-less mobile doc page should look like if/when that page is removed — both flagged for team discussion, not actioned here.

### DIFF
--- a/packages/eds-core-react/package.json
+++ b/packages/eds-core-react/package.json
@@ -54,7 +54,7 @@
     "react"
   ],
   "devDependencies": {
-    "@equinor/eds-mobile-components": "^0.3.0",
+    "@equinor/eds-mobile-components": "workspace:^",
     "@figma/code-connect": "^1.4.9",
     "@playwright/test": "^1.62.0",
     "@rollup/plugin-babel": "^7.1.0",
--- a/packages/eds-core-react/src/components/EdsProvider/EdsProvider.docs.mdx
+++ b/packages/eds-core-react/src/components/EdsProvider/EdsProvider.docs.mdx
@@ -10,7 +10,7 @@ Provider used to set global values for EDS react components.
 
 <PlatformTabs mobile={<>
   <Links
-    sourceUrl="https://github.com/equinor/design-system-mobile/blob/main/packages/components/src/components/EDSProvider/EDSProvider.tsx"
+    sourceUrl="https://github.com/equinor/design-system/blob/main/packages/eds-mobile-components/src/components/EDSProvider/EDSProvider.tsx"
     npmUrl="https://www.npmjs.com/package/@equinor/eds-mobile-components"
   />
   <MobileDocs />
--- a/packages/eds-core-react/src/components/Typography/Typography.new.docs.mdx
+++ b/packages/eds-core-react/src/components/Typography/Typography.new.docs.mdx
@@ -9,7 +9,7 @@ import MobileDocs from '@equinor/eds-mobile-components/docs/Typography.mdx'
 
 <PlatformTabs mobile={<>
   <Links
-    sourceUrl="https://github.com/equinor/design-system-mobile/blob/main/packages/components/src/components/Typography/Typography.tsx"
+    sourceUrl="https://github.com/equinor/design-system/blob/main/packages/eds-mobile-components/src/components/Typography/index.ts"
     npmUrl="https://www.npmjs.com/package/@equinor/eds-mobile-components"
   />
   <MobileDocs />
--- a/packages/eds-core-react/src/components/next/Button/Button.docs.mdx
+++ b/packages/eds-core-react/src/components/next/Button/Button.docs.mdx
@@ -14,7 +14,7 @@ Button component for triggering actions.
 
 <PlatformTabs mobile={<>
   <Links
-    sourceUrl="https://github.com/equinor/design-system-mobile/blob/main/packages/components/src/components/Button.tsx"
+    sourceUrl="https://github.com/equinor/design-system/blob/main/packages/eds-mobile-components/src/components/Button/Button.tsx"
     npmUrl="https://www.npmjs.com/package/@equinor/eds-mobile-components"
   />
   <MobileDocs />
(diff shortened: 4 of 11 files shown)


Chibuzor-Nwemambu:
@claude review

Chibuzor-Nwemambu:
Heads up on how to review this one: the `workspace:^` swap and 6 of the 7 `sourceUrl` fixes are mechanical (verified each target file exists, most just needed the repo/base-path swapped) — quick sanity check is enough there.

The one thing worth real scrutiny: `Typography.new.docs.mdx`'s `sourceUrl` now points at `index.ts` instead of a specific component file, because there's no single flat `Typography.tsx` on the mobile side — `index.ts` composes `Typography`/`Typography.Header` from `TypographyUI.tsx` + `TypographyHeader.tsx`. That's a judgment call, not a verified fact like the other six, so a second opinion on whether `index.ts` is the right target (vs. `TypographyUI.tsx`, or something else) would help. Also worth knowing going in: this whole page may not be long for this world — there's a pending team discussion about removing `TypographyNext` entirely, so don't over-invest here.

Chibuzor-Nwemambu:
Fixed findings #1 and #2 in b8c0e0bc3.

- **`docs/Introduction.mdx`'s three hardcoded links** — confirmed this really does render live on the web Storybook's Mobile/About page, not just internal docs drift. Repository link now points at this repo; the two release-notes links now point at the package's own `CHANGELOG.md`, matching the existing convention already used in `eds-core-react`'s own README (checked before picking this rather than guessing).
- **The stale `CLAUDE.md` prose** above the snippet — corrected the "ships via npm" claim and the consumer path, including the `EdsProvider`/`Typography.new` outside-`next/` note.

Verified end to end rather than trusting the diff: ran `pnpm build:storybook` (exit 0), then grepped the actual compiled output — zero remaining `design-system-mobile` references, and confirmed the correct new URL is really present in `Button.docs*.js`. Good catch that no CI build had exercised this, since resolution alone doesn't prove the MDX compiles.

The 4 follow-ups (published `package.json` fields, a different `CLAUDE.md` section, `apps/mobile-storybook`'s own hardcoded GitHub URL) are logged on #5138 under this step rather than folded in here — agreed they're genuinely separate scope.

Kept `Typography` → `index.ts` as-is rather than switching to the folder link — all 6 other `sourceUrl`s point at a specific file, and `index.ts` actually shows the real `TypographyUI`/`TypographyHeader` composition, which felt more consistent and more informative than a bare directory listing. Still open to the alternative if you feel strongly.

millus (approved):
index.ts makes sense to me — the docs page covers both Typography and Typography.Header, so pointing at where they're composed together is more useful than either TypographyUI.tsx or TypographyHeader.tsx alone. Checked some of the sourceUrls in Storybook locally/manually — correct repo and file paths, 404s are expected until the full stack lands on main. Approving!

PR #5340, opened 2026-08-17
fix: Today button in Datepicker now selects and navigates correctly

### LINKED ISSUE #4933: 🐛 Datepicker Input | Today  button does not work correctly
**Describe the bug**

Datepicker button **Today** does not work as intended. I expect it to select today's date when pressed. But it does not behave as expected.

1. Pick any date
2. Click Today 👍🏻 
3. Click Today 👎🏻 
4. Click Today 👎🏻 
5. Switch month
6. Click Today 👎🏻 

**Expected behavior**

Today's date is selected ALWAYS when pressed. 

**Additional context**

Image

millus:
## Summary

Fixes #4933 — the **Today** button in `DatePicker` did not reliably select today's date.

### Root cause

react-stately's `selectDate` constrains the target date against `startDate` (the currently visible month). When the user navigated to a future month, `today < startDate` caused a silent early return — the date was never selected and `onChange` was never called. `setFocusedDate` (the previous implementation) only navigated the view; it never selected.

### Fix

Thread a dedicated `onSelectToday` callback from `DatePicker` down through `Calendar` → `CalendarHeader` → `TodayPicker`. This bypasses react-stately's visible-range constraint entirely. The callback:

- Guards against `isDisabled` / `isReadOnly` (the popover is reachable via keyboard on a `readOnly` field)
- Validates today against `minValue`, `maxValue`, and `isDateUnavailable` before calling `onChange` — today's button stays enabled and always navigates, but `onChange` is suppressed when today is not selectable
- Preserves the existing time component when `showTimeInput` is active
- Computes today using `today(timezone)` from `@internationalized/date` so the correct calendar date is used when the picker's timezone differs from the local runtime timezone
- Calls `onChange` with today's date
- Calls `setFocusedDate` to navigate the view back to today's month

The Today button disabled state is unchanged from `main`: `disabled={showYearPicker}` only. Navigation to today's month always works, even when today itself is not selectable.

`DateRangePicker` is not passed `onSelectToday`, so Today there remains navigate-only — identical to `main`.

### Design decision: picker stays open after clicking Today

The Today button lives in the **navigation header** alongside the prev/next month arrows, not in the date grid. Its original behaviour was to navigate to today's month (not select), and users have built muscle memory around this: click Today to jump to the current month, then click the cell to confirm. Closing the picker on Today would break that workflow. Keeping it open is therefore intentional — Today selects the date and navigates the view, but stays in the navigation mental model rather than acting like a date cell click.

## Test plan

- [ ] Open DatePicker set to a future date, click Today → today's date is selected, input updates, picker stays open showing today's month
- [ ] Navigate forward past today, click Today → same as above (the core bug scenario)
- [ ] Click Today multiple times → `onChange` fires each time, picker stays open
- [ ] Set `minValue` to tomorrow → Today navigates to today's month, `onChange` not called
- [ ] Set `isDateUnavailable` to block today → Today navigates to today's month, `onChange` not called
- [ ] Set `showTimeInput` with an existing time → clicking Today preserves the time
- [ ] Open a `readOnly` DatePicker via keyboard (Tab + Enter), click Today → `onChange` not called
- [ ] Open DateRangePicker with a past `maxValue` → Today button is enabled, navigates only
- [ ] All 41 unit tests pass

### DIFF
--- a/packages/eds-core-react/src/components/Datepicker/DatePicker.spec.tsx
+++ b/packages/eds-core-react/src/components/Datepicker/DatePicker.spec.tsx
@@ -313,6 +313,80 @@ describe('DatePicker', () => {
     expect(screen.getByRole('presentation')).toHaveTextContent('01/22/2025')
   })
 
+  describe('Bug #4933: Today button', () => {
+    it('selects today when clicked', async () => {
+      const onChange = jest.fn()
+      render(
+        <I18nProvider locale={'en-US'}>
+          <DatePicker
+            label={'Datepicker'}
+            value={new Date(2020, 0, 1)}
+            onChange={onChange}
+          />
+        </I18nProvider>,
+      )
+
+      await userEvent.click(screen.getByLabelText(/^Change date.*/))
+      await userEvent.click(screen.getByText('Today'))
+
+      expect(onChange).toHaveBeenCalledTimes(1)
+      const result = onChange.mock.calls[0][0] as Date
+      const today = new Date()
+      expect(result.getFullYear()).toBe(today.getFullYear())
+      expect(result.getMonth()).toBe(today.getMonth())
+      expect(result.getDate()).toBe(today.getDate())
+    })
+
+    it('selects today when the calendar is showing a future month', async () => {
+      const onChange = jest.fn()
+      // Use a future date so the calendar opens past today — this is the exact
+      // scenario from the bug: today < startDate causes selectDate to bail out
+      render(
+        <I18nProvider locale={'en-US'}>
+          <DatePicker
+            label={'Datepicker'}
+            value={new Date(2027, 0, 1)}
+            onChange={onChange}
+          />
+        </I18nProvider>,
+      )
+
+      await userEvent.click(screen.getByLabelText(/^Change date.*/))
+      await userEvent.click(screen.getByText('Today'))
+
+      expect(onChange).toHaveBeenCalledTimes(1)
+      const result = onChange.mock.calls[0][0] as Date
+      const today = new Date()
+      expect(result.getFullYear()).toBe(today.getFullYear())
+      expect(result.getMonth()).toBe(today.getMonth())
+      expect(result.getDate()).toBe(today.getDate())
+    })
+
+    it('selects today on repeated clicks', async () => {
+      const onChange = jest.fn()
+      render(
+        <I18nProvider locale={'en-US'}>
+          <DatePicker
+            label={'Datepicker'}
+            value={new Date(2020, 0, 1)}
+            onChange={onChange}
+          />
+        </I18nProvider>,
+      )
+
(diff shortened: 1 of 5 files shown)


millus:
@claude review

millus:
@claude please re-review

millus:
@claude review

millus:
**Addressing round 3 leftovers:**

- **Design call:** Reverted to `main`'s disabled behaviour — Today is disabled when today is outside min/max or unavailable. Always-navigate was scope beyond the reported bug, and navigating to a month where no dates are selectable is confusing UX. PR description updated.
- **Timezone test:** Added — `jest.useFakeTimers` + `Pacific/Kiritimati` at 20:00 UTC, asserts correct day is selected in that timezone.
- **Duplication comment:** Added above `_onSelectToday` explaining why it is separate from `_onChange` (picker must stay open) and why it re-validates instead of delegating to react-stately.
- **`Today` hardcoded English:** Pre-existing, not introduced by this PR. Follow-up issue opened: #5346.
- **Memo churn:** Acknowledged as harmless — memoising `useConvertedValidationFunctions` would fix it but is out of scope here.
- **`isDisabled`/`isReadOnly`:** Unreachable through the UI (`Toggle` hides the open button), left as is.

millus:
@claude please review and stay focused on the issue reported, we are not fixing everything with datepicker now, that is for later! But of course let me know if the fix in this pr breaks anything the users are using in datepicker and/or daterangepicker

millus:
@claude please review

millus:
@claude please review

Chibuzor-Nwemambu (approved):
LGTM 💪🏾 

---
Two small non blocking notes from review:

1. Memoization gap in _onSelectToday

In DatePicker.tsx, the useCallback wrapping _onSelectToday lists _isDateUnavailable as a dependency, but that function comes from useConvertedValidationFunctions and is a fresh reference on every render. So the useCallback never actually returns a cached function since one of its own dependencies always changes. Not a correctness problem since Calendar rerenders on most state changes anyway, but the memoization isn't doing anything right now. Might be worth memoizing _isDateUnavailable upstream at some point if callback identity stability matters here.

2. DateRangePicker still has the original bug

Clicking Today in DateRangePicker moves focus to today's date but does not select it, since RangeCalendar never passes onSelectToday into CalendarHeader. This is called out intentionally in the PR description as out of scope, since issue #4933 was filed against DatePicker only, and there is a test pinning the navigate only behavior there. Agreed this should stay out of scope for this PR since selecting today on a range is a separate design question (does it become the start or end of the range, does it collapse an existing range, etc). Worth opening a follow up issue so this does not get lost once this merges.

Nice work overall, the test coverage here (time zones, min and max, unavailable dates, readOnly via keyboard) is thorough.

PR #5369, opened 2026-08-26
fix: fold token math and discover export files when generating TS tokens

eddiman:
## Why

The Tokens Studio **v0.0.6** release ran `.github/workflows/tokens_studio_release.yaml` on 2026-08-25 and failed at the `generate:ts-tokens` step:

```
generate-ts-tokens: non-numeric dimension for --eds-density-corner-radius-rounded-outer:
calc(var(--eds-density-corner-radius-rounded) + var(--eds-primitives-spacing-6))
Exit status 1
```

That step sits before `generate:css-bundle` and `create-pull-request`, so the run never opened a PR and **the entire token release was dropped** — only the Slack failure notification remained.

## What was broken

Three independent problems, all in `packages/eds-tokens/scripts/generate-ts-tokens.mjs`. Only the first was visible in the CI log; the other two surfaced once the script stopped exiting on the first bad token.

### 1. Token math arrives as an unevaluated `calc()`

`density.corner-radius.rounded-outer` stopped being a plain alias (`{primitives.spacing.37}`) and became a math expression. The platform's formula engine folds colour formulas but **not** dimension arithmetic, so the CSS export emits the expression verbatim — while `dereference` only matched a value that was *entirely* one `var()` (`/^var\(--([\w-]+)\)$/`).

It now substitutes every reference in a value, including ones nested inside a larger expression and the fallback arm of `var(--a, 1px)`, then folds the arithmetic with a small recursive-descent evaluator using CSS unit algebra (matching units for `+`/`-`, at most one united factor for `*`, unitless divisor for `/`). Substituted values are parenthesised only when genuinely arithmetic, so `rgb(0 0 0 / 0.5)` keeps its exact form while `var(--a) * 2` over a sum cannot mis-associate.

### 2. `buildContext` listed its input files literally

v0.0.6 also added an `elevation/` export declaring the 20 `--eds-shadow-*` primitives, plus `elevation/{ambient,key}` colours in the color-scheme layer. Because the file was not in the hard-coded list, none of those variables reached any resolution context — 20 × `CSS variable not found`, plus `broken var() chain` for `--eds-elevation-{low,high}`, which compose them in `semantic/default.css`.

The dimension-independent files are now discovered rather than listed, so a new export folder is picked up automatically. One explicit exclusion: the committed `variables.css` bundle concatenates every dimension variant and predates the run (the workflow bundles *after* this script), so merging it would leak one density's values into another's.

### 3. No converter for `$type: shadow`

`semantic.elevation.{low,high}` are composite two-layer shadows. Every `$type` in the live export was enumerated to rule out a fourth surprise — `shadow` was the only unhandled one. They now emit the `box-shadow` string with lengths as authored and colours converted to hex:

```ts
elevation: {
  high: '0px 4px 12px 0px #00000033, 0px 12px 16px 6px #0000001f',
  low: '0px 1px 8px 0px #00000033, 0px 4px 8px 3px #0000001f',
},
```

The per-layer numbers are separately available in the new `elevation/default.ts`. A colour function the script cannot convert (`color-mix()`, relative colour syntax) fails loudly rather than passing through.

## Two supporting changes

- **Report every failure, not just the first.** Per-token problems are accumulated and printed together instead of `process.exit(1)` on the first. This is what surfaced breakages 2 and 3 — under the old behaviour they would have taken 22 sequential release retries to find one at a time.
- **Don't wipe committed output on failure.** `rm(OUT_DIR)` ran *before* resolution, so a failed run deleted `src/tokens/ts/` as a side effect. It now runs only once the whole export resolves.

No change to the workflow: a `generate:ts-tokens` failure still aborts the release. A PR pairing a fresh CSS/DTCG export with stale TS modules would silently disagree with itself, and Slack already alerts.

## Tests

The generator had **no test coverage** — `density-aliases.test.ts` and `typography-shape.test.ts` both target the legacy Style Dictionary pipeline. That is why a new value shape from the platform reached CI unguarded.

`src/__tests__/generate-ts-tokens.test.ts` adds 19 cases over temp-dir fixture exports: calc folding per density context, nested parentheses, `*`, `/`, negative factors, operator precedence when inlining a sum, `var()` fallback, discovery of a new export folder, exclusion of the `variables.css` bundle, shadow conversion, and each failure mode (mismatched units, unconvertible unit, divide-by-dimension, cycle, junk value, multi-token reporting, output preserved on failure).

## Verification

Ran the real pipeline against live platform state (all 20 exported file sizes match the failed CI log byte for byte, so `main` is still at v0.0.6 — this is the same input that broke):

| Step | Result |
| --- | --- |
| `studio tokens pull --dry-run` | 18 files, `1 unchanged` |
| `studio exports run` (EDS-CSS / EDS-DTCG) | 10 + 10 files |
| `generate:ts-tokens` | exit 0, 11 modules |
| `widen-semantic-scope` + `generate:css-bundle` | bundled 10 files incl. the new `elevation/` |

- Deterministic: two runs on identical input produce byte-identical trees.
- No `calc(`, `var(`, `oklch(` or `rgb(` left anywhere in the generated modules.
- `roundedOuter` folds to 3 / 5 / 7 for compact / comfortable / relaxed (`rounded + 1px`, per the new formula).
- **No regression:** the currently committed export still generates byte-identical output.
- 65/65 package tests pass; eslint clean.

## Note for whoever merges the release PR

Re-running the release workflow after this lands will produce a large token diff, unrelated to this fix: v0.0.6 restructures the concept layer (`bg-floating` → `background-floating`, `bg-backdrop` → `background-backdrop`, `border-focus` → `border-interactive-focus`, `text-link` → `text-interactive-link-{default,hover,pressed}`, all moved from `color-scheme/` into `semantic/`) and shifts dark-scheme palette values. Renames rather than deletions, but breaking for anyone on the beta line — worth a real review rather than a rubber stamp.


millus:
@claude review

eddiman:
Thanks — all five findings addressed in 103bc7a45. I reproduced finding 1 before acting on it (`contrast/{high,low}.css` → exit 0 with `border: 1`, the `high` variant silently discarded), and it is a real loudness regression from the discovery change.

**1. Unrecognised dimension folder** — fixed, but with a narrower guard than suggested. Failing on any conflicting duplicate across *all* export files would also catch the semantic ↔ color-scheme overlap that `generate-css-bundle.mjs` deliberately tolerates (#5221 — it orders color-scheme last precisely so it wins). Both work on today's data (I measured the live v0.0.6 export: zero duplicates, among base files and against the dimension files), but if that overlap returns upstream, a blanket guard blocks the release while the CSS bundle handles it fine.

So `buildContext` now has two guards instead:

- a discovered folder holding more than one file is a dimension the script does not know about → abort naming the folder and pointing at `DIMENSION_FOLDERS`
- two *base* files declaring one property with different values → abort, since precedence among them is alphabetical and that is no basis for picking a value

Together those cover both effects you identified, and the intentional dimension override survives. Three tests pin it, including one asserting a density file can still override a base value.

**2. Colour errors don't name the token** — agreed, and more so once failures are batched. `oklchToHex` / `rgbToHex` take `cssName` and interpolate it; all call sites updated.

**3. `rgbToHex` requires alpha** — fixed, alpha arm optional, 6-digit hex when absent.

**4. `BUNDLE_FILE`** — comment now points at the `OUT_FILE` default in `generate-css-bundle.mjs` it has to stay in step with. Not worth a cross-script import for one string.

**5. Tests** — `afterAll` removes the fixture trees (verified: no `eds-ts-tokens-*` left in `/tmp` after a run); both describe-body `run()` calls moved into `beforeAll`; added the two-layer comma shadow case and the file-precedence cases.

Re-verified after the changes: 71/71 tests pass, eslint clean, the committed export still generates byte-identical output, and a fresh pull of the live v0.0.6 export still produces 11 modules at exit 0 with `roundedOuter` 3 / 5 / 7 and the elevation strings intact.

No action on the `types` failure — agreed it is the pre-existing unbuilt `@equinor/eds-tokens-build` dependency.

eddiman:
@claude

Chibuzor-Nwemambu:
Ran the script against the committed export and got byte-identical output, tests pass. Evaluator reads well. A few things I ran into that might be worth a look before this merges.

**`convertShadow` doesn't fold `calc()`**

Shadow lengths come from the same `--eds-primitives-*` space that started emitting token math, so if an offset/blur/spread ever becomes an expression it hits the paren guard on 523:

```
--eds-shadow-y: calc(var(--eds-primitives-spacing-25) + var(--eds-primitives-spacing-6));

→ exit 1: unsupported shadow value for --eds-elevation-low:
          0px (calc(4px + 1px)) 8px 0px rgb(0 0 0 / 0.2)
```

Same failure mode as the one this fixes. The parens in the message come from `asOperand` rather than the platform, so it also reads like a platform bug. Might be as simple as passing each length through `evaluateExpression` first.

**Stale `src/tokens/css`**

`tokens_studio_release.yaml` exports over the previous release without clearing the directory. Harmless under the hard-coded list, but discovery picks the leftovers up: I saw a renamed folder abort on a conflict that isn't in the new export, and a removed one silently resolve to its old value. An `rm -rf` before the export step would sort it.

**Smaller**

`fail()` exits before flushing `problems`, so a structural error hides any token errors found first. And chain errors name the intermediate variable rather than the DTCG token, which collapses several tokens into one line.

**For mobile**

`$type: shadow` emits a web `box-shadow` string. Nothing's lost, but RN needs `shadowColor`/`shadowOffset`/`shadowRadius`/`shadowOpacity`, which the legacy Style Dictionary output already gives us. Wondering whether a structured object alongside the string is worth doing now rather than at migration time.

eddiman:
All four addressed in 2a80dcfa3 — thanks, the shadow one was going to bite us and the stale-export one was created by this branch.

**`convertShadow` and `calc()`** — reproduced exactly, including the misleading message: the parentheses in `0px (calc(4px + 1px)) 8px 0px …` come from `asOperand`, so the value shown exists in no export file. Rather than only folding the lengths, shadow layers are now parsed properly — split on top-level commas, then on top-level whitespace (so `rgb(0 0 0 / 0.2)` stays intact), each length through the same `evaluateExpression` as any other dimension, colour converted once. A layer with no colour, two colours, or fewer than 2 / more than 4 lengths fails with a specific message instead of the generic paren guard.

**React Native properties** — added, mirroring the legacy `build/ts/elevation/elevation.ts` shape rather than inventing one. Alpha is taken from the source colour before hex conversion, so it is exact rather than round-tripped through `#…33`:

```ts
elevation: {
  high: {
    boxShadow: '0px 4px 12px 0px #00000033, 0px 12px 16px 6px #0000001f',
    layers: [
      { shadowColor: '#000000', shadowOffset: { width: 0, height: 4 },
        shadowOpacity: 0.2, shadowRadius: 12, spread: 0 },
      { shadowColor: '#000000', shadowOffset: { width: 0, height: 12 },
        shadowOpacity: 0.12, shadowRadius: 16, spread: 6 },
    ],
  },
}
```

Layers are an ordered array rather than named `key`/`ambient`, because the composite CSS value carries no layer names — those exist in `elevation/default.ts`, which the new `elevation/` export already produces. `spread` is kept for parity even though React Native has no equivalent. This needed an array branch in `toLiteral`.

**Stale `src/tokens/css`** — the workflow now clears `src/tokens/{css,dtcg}` before the export runs, with a comment explaining that discovery is what makes pruning matter. Left `raw/` alone: `studio tokens pull` has its own manifest and `--partial` handling, so clearing it seemed likelier to break something than to fix it. Say if you would rather it went too.

**The two smaller ones** — `fail()` now flushes accumulated token problems before exiting, so a structural error no longer discards everything found before it (verified: a bad colour plus a missing density CSS file reports both). And a broken chain renders the full path from the token's own property — `--eds-space-inline → --eds-mid → --eds-never-exported` — so two tokens sharing one bad intermediate produce two distinct, actionable lines instead of deduplicating into one anonymous one.

74/74 tests pass (six new cases: calc in a shadow length, the shadow shape, the chain naming, and the flush-on-structural-error path), eslint clean, and the committed export still generates byte-identical output. My Studio session expired since yesterday, so the live end-to-end re-run needs a fresh login — the two-layer fixture uses the real v0.0.6 numbers in the meantime.

Chibuzor-Nwemambu (approved):
Very Nice @eddiman 🎉 

All four look really good. Ran the tests and regenerated against the committed export, still byte-identical.

Only thing I found: `#RGBA` shorthand falls through `colourAlpha` to a default of 1, so the alpha stays in the colour instead of being split out. Unlikely given the platform emits oklch/rgb, and everything else failed loudly like it should. Not a blocker, just something to note.

Nice work 💪🏾

PR #5371, opened 2026-08-26
feat: also trigger iOS build on eds-mobile-components releases

Chibuzor-Nwemambu:
Part of #5138 (resolves finding #8 from the earlier iOS pipeline review).

## Summary

Previously, the iOS TestFlight build only fired when `apps/mobile-storybook` itself got a release-please release. A component-only change (bug fix, refactor - no new demo screen) would publish to npm without triggering a fresh TestFlight build, meaning nobody would see it running on a real device until some unrelated app-side change eventually pulled it along.

## Decision

Considered and rejected release-please's `linked-versions` plugin (checked its actual source: it forces every package in a group to share one version number, propagating the highest bump to all members - even ones with zero real changes, via a synthetic commit). That's bidirectional: it would also force an app-only change to trigger a meaningless `eds-mobile-components` npm republish under a new version number - real semver pollution for a published package, with no corresponding benefit.

Instead: widened `trigger-mobile-ios`'s condition to fire on *either* package's release-please release, with no other change.

- No version bump or npm publish happens on `apps/mobile-storybook`'s side - it isn't published to npm at all.
- No versioning/semver concept is touched. This is purely a CI trigger change.
- `build_release_ios.yaml`'s build number is already independent of any committed app version (it's `date + CI run number`, computed fresh per run), so there's nothing for this to conflict with.
- Uploading a build to TestFlight is always separate from submitting a build for public App Store review, which stays a deliberate, manual step in App Store Connect - this can't accidentally ship anything to the public.

## Changes

- `trigger_publish.yml`: `trigger-mobile-ios`'s `if:` condition now checks `mobile-storybook == 'true' || mobile-components == 'true'`
- Updated the job's own comment and the neighbouring `trigger-mobile-components` job's comment to reflect that a components release now triggers two things (npm publish + iOS build), not one
- Updated `publish-summary`'s reporting logic to match - it now checks both outputs before saying "iOS build triggered", so a components-only release doesn't get wrongly reported as "skipped" for the app

## Verified

- YAML parses correctly, all 12 jobs present with no gaps/duplicates
- Embedded bash in the summary job syntax-checks clean
- Traced through `check-releases`' actual output names to confirm `mobile-components` exists and is spelled correctly

## Not testable yet

`trigger_publish.yml` has no manual trigger at all - it only fires on `pull_request: types: [closed]` for a real merged release-please PR against `main`. This can't be exercised live until the whole migration is merged to `main` and a subsequent real `eds-mobile-components` release happens. Static/logical verification is all that's possible before then.

### DIFF
--- a/.github/workflows/trigger_publish.yml
+++ b/.github/workflows/trigger_publish.yml
@@ -344,10 +344,11 @@ jobs:
           echo "::error title=Icons Publish Failed::Failed to trigger publish workflow for @equinor/eds-icons"
           echo "❌ Failed to trigger icons publish workflow" >> $GITHUB_STEP_SUMMARY
 
-  # Job: Trigger mobile-components publish workflow
+  # Job 8: Trigger mobile-components publish workflow
   # Only runs if the Release Please manifest bumped eds-mobile-components.
-  # apps/mobile-storybook's own release (which gates the iOS build/release
-  # pipeline) is handled separately, not here - see the tracking issue.
+  # The npm publish itself is handled here; the iOS build/release pipeline
+  # this release also gates is a separate job (trigger-mobile-ios) below,
+  # since a components release triggers both.
   trigger-mobile-components:
     needs: check-releases
     if: needs.check-releases.outputs.mobile-components == 'true'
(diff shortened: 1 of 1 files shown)


Chibuzor-Nwemambu:
@claude  review

millus (approved):
LGTM ✨ I think it could be worth fixing finding 1 and 2 from the bot review before merging, so there is less confusion later.

millus:
Nice work fixing findings 1 and 2, this one is ready to merge now ✅ 👍

Issue #5124, opened 2026-06-25
Revisit the dynamic-vs-static user-test recordings (spike)


eddiman:
## Description
The original user testing reportedly came out ~40% dynamic / 60% static, but the decision at the time was to **keep both** rather than pick one — part of how the current complexity arose. Revisit the recordings to validate the decision to consolidate on static.

## Scope
- Locate the original dynamic-vs-static user-test recordings (noted to exist).
- Re-review against the decision to drop dynamic.
- Capture anything that should adjust the approach.

## Definition of Done
- [x] Recordings located and reviewed
- [x] Findings summarised against the static-only decision
- [x] Any adjustments flagged

eddiman:
## Spike outcome — static-only decision validated

Revisited the Spring-2025 dynamic-vs-static user test against the decision to consolidate on static. Recordings and artifacts located and reviewed; findings summarised and adjustments flagged below. Full working notes live in the team-assistant token docs.

### The decision this validates

**Dynamic colour is out.** No data-attribute appearance switching and no appearance modes at the semantic-colour level — 100% removed for now. One static semantic colour system, full stop. (Dark/light and density remain as modes; those are a separate axis and are unaffected.)

### Recordings located

All Spring-2025 artifacts are linked from [equinor/design-system-internal#189](https://github.com/equinor/design-system-internal/discussions/189) (Resources): the user-test recordings (SharePoint: *Designsystem → General → Colours - User test - Spring 2025 → User Test Recordings*), the user-test notes (Figma), and the Usability-testing A/B protocol files. Review was done against the transcribed record in #189 (per-participant feedback for 10 designers + 8 developers, insight/trend rollups, and Victor's per-concept breakdown), not a fresh pass over the raw video.

### Verdict: removing dynamic is the right call

- **Static led both cohorts in 2025.** Developers 6/10 for static, all six confident; designers split 5/5, but the static-preferring half were markedly more confident (4/5 vs 2/5). Start-order had no effect on final preference.
- **"Keep both" did not come from the data.** Static won the head-to-head; the dual-library outcome came from *reframing* the library-level verdict into per-concept decisions, which is what kept dynamic alive as a promoted parallel library. That reframing is the documented root of the current dual-source complexity.
- **Every reason static won is durable, not an onboarding artifact:** familiarity, visual feedback while picking, separation of concerns, and clean mapping to utility/Tailwind classes.
[...]

Issue #4963, opened 2026-06-02
Define the Token Variable Architecture (Figma + code ADRs)


emirgens:
## Description
Define a clear, opinionated architecture for how token variables are structured in the redefined token system, based on the 23 Jun 2026 token decision ("redefine, don't rebuild"). This is the **foundational** sub-task — the rest of the colour & token work references it.

The architecture is delivered as **two ADRs**, tracked as the sub-tasks below:

- **Figma variable structure** (typography, spacing, corner-radius, density) — #5139
- **Code / token output architecture** (Tokens Studio export, build pipeline, CSS + TS shape) — #5140

Colour is **out of scope here** — the colour collection's structure and state-naming are handled in a separate track (#4742 / #4744) and will get their own ADR.

## Decided direction (applies across both ADRs)
- **Keep the foundation:** type scale + contrast algorithm, colour generator + values, the primitive (raw value) layer.
- **Layering:** primitive → semantic → text-style/components. References flow **one way only** — no sibling / cross-collection / semantic→semantic connections.
- **Decoupled groups:** typography, spacing, corner-radius — each built directly on primitives, independent of one another; density set once via one collection.
- **Modes scoped** to density (dark/light lives in the colour track) — not appearance or per-typography.
- **Clear design vs. code ownership.**

## Definition of Done
- [x] Figma variable structure ADR written, reviewed by design + engineering (#5139)
- [x] Code / output architecture ADR written, reviewed by design + engineering (#5140)
- [x] Both agreed as the shared mental model for the redefined tokens

eddiman:
**Token architecture work — workshop scheduled**

The token-architecture alignment this issue calls for is being kicked off through a two-part workshop:

- **17 Jun 2026 — diagnosis workshop** (whole team): build a shared understanding of the current token system and produce two artefacts — a prioritised weakness list and a set of decision criteria (model-first / tooling-second). The `apps/token-factory/` app is the teaching companion used to trace the alias chain live.
- **23 Jun 2026 — decision meeting**: turn the diagnosis into commitments — model direction and tooling direction — which is what produces the ADR this issue asks for.

Detailed agenda, pre-reading and roles are tracked internally in equinor/design-system-internal#264. The diagnosis output will be captured in a "Design tokens" discussion; the ADR lands after the 23 Jun decision.

eddiman:
Closing as completed. All three ADRs this issue called for are done:

- #5139 Token variable architecture, Figma structure (typography and spacing)
[...]

Issue #4745, opened 2026-03-30
📚 Update token documentation


emirgens:
## Description
Update and extend the token documentation to reflect all the redefined-token changes. The designer interviews made **"documentation of intent"** the #1 ask: abstract t-shirt sizes are only acceptable alongside a style guide with contextual examples showing *when* to use each — ideally in Storybook (the source of truth).

## Scope
- Disabled-state approach + rationale — **the colour scale is opaque; disabled uses opaque steps** (decided 2026-08-26, ~~alpha-channel~~ dropped). Rationale to carry over from #4741, including the scale-scoped rule: alpha only where the backdrop cannot be known (`--eds-elevation-*`, `overlay.scrim`, dataviz).
- Updated colour values; the "static only, don't mix" story and its rationale.
- Density rename (compact / comfortable) rationale.
- Naming updates; required-vs-nice-to-have clarity.
- Text-trim removal rationale.
- Extended typography usage docs with contextual examples in Storybook.
- Must be usable by AI agents (and tested by one).

## Definition of Done
- [ ] Documentation updated for the disabled-state approach (opaque, not alpha — see #4741)
- [ ] Updated colour token values reflected
- [ ] Density rename documented
- [ ] Naming updates included
- [ ] Typography usage docs extended with contextual examples (Storybook)
- [ ] Published and shared with the team
- [ ] Tested by an AI agent

eddiman:
## Rationale: why the trim-based baseline grid & text-trim were dropped

Capturing this for the **"text-trim removal rationale"** DoD item — and the closely-related **baseline-grid removal**, since they're the same mechanism. The reasoning currently lives mostly in the 23 Jun token decision notes and designer interviews; consolidating here so the docs can cite it. (`design-system-internal/docs/decisions.md` has been updated to stop describing the removed approach as current.)

### What existed, and why
A 4px **baseline grid** was a foundational EDS 2.0 principle — mandated in ADR 0004 (spacing) and ADR 0005 (typography) — to give type and spacing one shared vertical rhythm (print-typesetting heritage). It was implemented via a Figma **typography component using cap-size / vertical trim + padding** to snap each text block to a 4px multiple, and in code via `text-box: trim-both ex alphabetic` behind an `@supports` progressive enhancement (ADR 0005). **`text-trim` was the mechanism that made the baseline grid possible** — so dropping one drops the other.

### Why it was dropped (23 Jun 2026 token decision)
[...]

PR #5374, opened 2026-08-27
docs: add ADR-0016 colour approach for EDS 2.0

### LINKED ISSUE #5375: ADR: colour approach for EDS 2.0 (ADR-0016)
Tracks the colour ADR deliverable of #4963.

The decision record for the EDS 2.0 colour system — static vs. dynamic, opaque-only, the generation chain, step roles and state ladders, scheme handling, OKLCH vs. Figma's clamp, and the build-time checks the decision depends on.

- PR: #5374
- Done when the PR is merged.

eddiman:
Part of #4963 (Define the Token Variable Architecture, the parent ADR track) under epic #4740.

## Why

The colour system is the largest part of the EDS token redefinition (#4740), and none of it was written down as a decision. Settled questions (opaque-only, why the lightness inputs are hand-typed, step roles) kept getting re-litigated from issue comments and sync notes. Spacing and typography have ADRs; colour had nothing.

## What

Adds ADR-0016, recording the colour system as defined in Tokens Studio (the canonical source, state verified 2026-08-27):

- **Three contested axes** with sourced reasoning: dynamic vs. static colour (settled by the 23 Jun "redefine, don't rebuild" decision and the #5124 user-test review), opaque scale vs. alpha variants (#4741), step-numbered scale vs. semantic-only names
- **D1–D10**: static-only, the generation chain, hand-set lightness inputs (deliberately non-monotonic in dark), opaque-only with the sanctioned alpha carve-outs, step roles and state ladders, semantic layer shape, scheme handling, one-owner-per-name, OKLCH canonical vs. Figma's sRGB clamp, side-by-side migration
- **Confirmation**: the build-time checks the decision depends on (ladder progression, duplicate steps, alpha, duplicate emitted names, APCA targets against `background.surface`), none of which exist yet
- **Unresolved**: surface pairing, dataviz, orphaned steps 6 and 14, the dark step-4 border contrast observation

The ADR deliberately references no concrete token counts, since the set grows as roles are added. Consumer-facing token documentation (#4745) remains separate work; this is the decision record.

## Verification

Claims were verified against the live Tokens Studio state rather than the repo's lagging `src/tokens/raw/` snapshot. Concretely, that means the `tokens-studio-backup` branch: despite the name it is not a stale copy but the automated hourly mirror of the Studio project's `main` branch (`tokens_studio_backup.yaml` runs `studio tokens pull` every hour and commits the result), so it is at most ~1 hour behind Studio. The snapshot used was committed 2026-08-27 10:40 UTC. APCA figures were computed from the generated OKLCH values.


pomfrida:
@claude review

documentation/adr/0016-colour-approach-for-eds-2.md
```diff
@@ -0,0 +1,275 @@
...
+The values in Tokens Studio as of 2026-08-27:
+
+| Scheme | 1    | 2    | 3   | 4   | 5   | 6    | 7   | 8   | 9   | 10  | 11  | 12  | 13  |
+| ------ | ---- | ---- | --- | --- | --- | ---- | --- | --- | --- | --- | --- | --- | --- |
+| light  | .98  | .94  | .91 | .87 | .82 | .77  | .72 | .62 | .52 | .47 | .42 | .37 | .32 |
+| dark   | .215 | .226 | .47 | .30 | .45 | .595 | .61 | .76 | .82 | .88 | .93 | .96 | .99 |
+
+The ladder is **deliberately not monotonic**: the dark values dip at steps 4–5 because a strictly increasing ramp cannot hold the required contrast for every role the steps serve. The invariants are narrower, and they are what [#5282](https://github.com/equinor/design-system/issues/5282) actually established: no two steps within a scale may resolve to the same value, and every state ladder built on the scale must progress visibly in one direction. Those two properties are what _Confirmation_ asserts; global ordering is not a rule.
+
+Steps **14 and 15 are inverse-polarity**: text and icon values intended to sit _on_ an emphasis fill (light .90 / 1.00, dark .33 / .10). They are outside the ladder by design and must be excluded from any ladder assertion.
```
pomfrida: D3 says steps 14 and 15 are foreground only, but background.surface resolves to neutral.15 in semantic.json. Worth mentioning here so D3 and D5 tell the same story.

eddiman:
Pushed `7e7fc03c8`. Here's what changed, in plain language.

## The short version

Almost everything in the ADR was already right. I re-checked every factual claim against the actual token data and found **five things that were wrong or missing**, plus one that nobody had spotted. All of it is fixed.

To check the claims I used the `tokens-studio-backup` branch at `9b4cef6` — the automated hourly mirror of Tokens Studio, snapshotted 2026-08-27 10:40 UTC. Same source the PR body cites. I wrote a script that re-derives every number in the ADR from that data; it now passes 81 out of 81 checks.

## What was actually wrong

**1. The chain diagram had a layer that doesn't do anything.** The ADR said colour flows through six layers starting at `primitives`. But `primitives` contains no colours at all — it's font families, line heights and spacing. It's just listed first in the config file. So the colour chain is really five layers starting at `input/*`. Fixed, and the two places the doc disagreed with itself about where shadows live are now reconciled.

**2. "Only two of the eight old aliases survived."** Not true — all eight survived. When the flat names like `bg-input` were deleted, every one of them got a proper semantic replacement. Only the *names* went, no roles were dropped. As written, a reader would think six things got deleted.

**3. The step-roles table was missing a whole group of tokens.** The ADR called that table authoritative, so gaps there matter more than elsewhere. Missing: the entire `background.non-interactive.*` group (18 tokens), a second selection ladder, `border.interactive.link`, and five one-off roles. All added.

**4. `background.interactive.neutral.selected` is not neutral.** It resolves to `accent` colours, and on different steps (3/4/5) than the accent selection ladder (5/7/8). That's the kind of surprise an ADR exists to explain, so it's now called out explicitly rather than left for someone to discover.

**5. One check would have broken the build the day it landed.** The ADR said "measure foreground roles against the surface colour" without saying whether borders count as foreground. If they do, the check fails immediately on the dark step-4 borders the ADR itself flags as an open question. That's not an implementation detail — it decides whether the build goes red. Now settled: foreground means text and icons only; borders are explicitly out until the surface-pairing list exists.

## The one nobody caught

**There is no `icon.interactive.neutral` token.** The interactive icon ladder exists for accent, info, success, warning and danger — but not neutral. Every other per-tone group in the system covers all six. So if you're building a component that needs a neutral interactive icon, there's nothing to bind to. Now documented as an open question rather than an invisible hole.

## Where the review was wrong

The automated review said the `Lc ≈ 0` figures weren't reproducible and that it got "low single digits" by hand. I resolved the colours properly and ran the real APCA algorithm — it's **exactly 0**, not approximately. Both pairs fall below APCA's low-contrast cutoff, which floors the result to zero by design.

So I kept the numbers and added the recipe instead of softening them. And I went further: I'd only tested `neutral`. It turns out this holds for **all six tones** — step 4 is too low in chroma in every hue for the curve to separate it from the canvas. I also added the light-mode numbers (Lc 17–23), which show dark is the outlier rather than the whole system being broken.

Two token paths in the review were also slightly off (`background.read-only` and `border.interactive.selected`); I used the real names.

## Two things that came out of @pomfrida's comment

You were right that D3 and D5 were telling different stories about step 15. D3 said 14 and 15 are "text and icon values"; D5 said `background.surface` is step 15. Both true — step 15 does double duty. It's now explained as the *point* of the inverse polarity: it's what makes a panel lighter than the canvas in light mode and darker in dark mode.

Chasing that also turned up something bigger. The ADR claimed the reason the lightness values are hand-typed "was never written up". **It was** — it's in the team decisions log, and the reason is specific: when the scale was expanded for hover/pressed/selected, generating the base values failed twice over. Inserting a step cascade-renames the whole scale and breaks every consumer, and the generated values at those positions didn't produce usable state ladders. That's a much better answer than the hedge I'd written.

But the log and the ADR now **disagree on what to check for**. The log says "enforce strict monotonicity, except steps 13–15". I tested that against live values: it fails at dark 3 → 4 (`.47 → .30`, the deliberate dip), and step 13 isn't an exception at all — it continues the ramp. It's 14 and 15 that invert. The ADR's ladder-progression check is the correct version, and it now says so explicitly so the two docs don't quietly contradict each other.

Worth knowing: I found the log at `docs/architecture-decisions/decisions.md`. The path everyone cites, `docs/decisions.md`, doesn't exist and never has — it was wrong in the ADR, in `CLAUDE.md`, and in three other places in the internal repo. Fixed there too.

## Also fixed

Data visualisation turned out to be more interesting than "hand-picked hex". **56 of the 66 values differ between light and dark, and each categorical ramp is exactly inverted** — series 1 is the lightest colour in light mode and the darkest in dark mode. Nothing says whether that's intended or how to order a series across schemes. Now recorded.

On constraining the hand-set inputs: the log lists two proposals, constrain them or hide them. **Hiding is already done** — all 15 steps and all 7 anchors are marked `hiddenFromPublishing`, so they never reach Figma. Only the constraint question is still open.

Smaller: `#5227` and `#5280` are pull requests, linked as issues (`#5280` was also described as a "semantic layer restructure" when it's the token-sync PR that landed it). Dropped the ADR count that contradicted itself. Added a sentence separating **EDS 2.0** (the design system generation) from **2.x / 3.0** (the `eds-tokens` package version), since using both without explanation made the docs app look like it contradicted the title. Fixed `scaleexpansion`.

## Not changed

Status stays **Proposed** — the review's read is right that D5 and the checks are still moving.

documentation/adr/0016-colour-approach-for-eds-2.md
```diff
@@ -0,0 +1,297 @@
+# Colour approach for EDS 2.0
+
+- **Status:** Proposed
```
pomfrida:
```suggestion
- **Status:** Approved
```

pomfrida (dismissed):

PR #5392, opened 2026-09-02
fix: reflect inherited disabled state in Checkbox, Radio and Switch

### LINKED ISSUE #5349: Checkbox/Radio/Switch don't reflect an inherited :disabled (e.g. from a disabled ancestor <fieldset>)
## Description

`Checkbox`, `Radio` and `Switch` derive their visual disabled state (glyph fill, cursor, hover ring) from the React `disabled` **prop**, not the `:disabled` pseudo-class. When the underlying native input is disabled by an ancestor `<fieldset disabled>` (standard HTML form behaviour), the control becomes functionally non-interactive but still **renders as enabled** — coloured glyph, `cursor: pointer`, and the hover highlight still fires.

Native inputs, `Button`, and `TextField`/`Input` all reflect the inherited disabled state automatically. So the common pattern of wrapping a section in a disabled `<fieldset>` to disable it in one place leaves only these custom-glyph controls looking active.

## Reproduction

```tsx
<fieldset disabled>
  <Checkbox label="A" />   {/* looks enabled ✗ */}
  <Radio label="B" />      {/* looks enabled ✗ */}
  <TextField label="C" />  {/* looks disabled ✓ */}
</fieldset>
```

## Expected

Checkbox/Radio/Switch look disabled whenever their input is `:disabled`, including via an inherited disabled `<fieldset>`, matching native controls.

## Actual

Only the explicit `disabled` prop drives the disabled styling; an inherited `:disabled` is ignored visually.

## Root cause (v2.3.4)

In `Checkbox/Input.tsx` (and the Radio equivalent), `fill = disabled ? states.disabled.background : background`, and the `Input` `cursor` + the `InputWrapper` hover `::before` are all computed from the `disabled` prop. Keying these off the CSS `:disabled` state (e.g. `&:disabled ~ svg { fill: … }`, `&:disabled { cursor: not-allowed }`, and gating the hover `::before` on `:not(:disabled)`) would honour inherited disabling.

## Workaround

Scoped CSS overriding the svg fill / cursor / hover for `fieldset:disabled input:disabled`.

## Version

`@equinor/eds-core-react` 2.3.4

pomfrida:
## Description

Checkbox, Radio and Switch derived their disabled visuals (glyph fill, cursor, hover highlight, track/handle colours) from the React `disabled` prop only. When the underlying input was disabled by an ancestor `<fieldset disabled>`, the control was functionally non-interactive but still rendered as enabled.

This moves all disabled styling to the CSS `:disabled` state, which matches regardless of where the disabling comes from:

- **Checkbox/Radio**: cursor and SVG glyph fill are keyed off `&:disabled` / `&:disabled ~ svg`. The hover ring and label/wrapper cursor live on parent elements and use `:has(input:disabled)`.
- **Switch**: track/handle/hover colours are expressed as `&:disabled + span > span` sibling selectors from the input, removing the `$isDisabled` props on Track/Handle entirely. The small variant keeps its handle colour when disabled, as before.
- The prop-based styling on the label/wrapper elements is removed rather than kept as a fallback — `:has(input:disabled)` covers every case the prop covered. This also stops styled-components forwarding an invalid `disabled=""` attribute onto the wrapper `<span>`s in Checkbox and Radio (pre-existing).
- Tests per component rendering inside `<fieldset disabled>`, asserting both the `:disabled` rules and the `:has(input:disabled)` rules, plus a `size="small"` Switch case and regression guards against the wrapper attribute leak.

**Browser support trade-off:** `:has()` is supported by everything in the repo's browserslist target (last 2 versions; Chrome/Edge 105+, Safari 15.4+, Firefox 121+). In older browsers outside the target, the core disabled visuals (glyph, track, input cursor) still work via plain `:disabled`, but the two parent-level details (hover ring suppression, label cursor) are lost — including for the explicit `disabled` prop, which previously styled these via props. This is a deliberate simplification.

Verified in Storybook: all three controls show disabled styling (grey glyph/track, not-allowed cursor) inside a disabled fieldset, and default/explicitly-disabled states are unchanged.

Note: the `/next` (EDS 2.0) versions of these components have the same defect class (disabled visuals keyed off the prop via `data-*` attributes) — tracked in #5393.

Fixes #5349


pomfrida:
@claude review

pomfrida:
@claude review

millus (approved):
LGTM ✨ all selection controls in a fieldset get disabled styling when fieldset is disabled now 👍 (for legacy)

PR #5405, opened 2026-09-07
feat: add plug_connect icon

### LINKED ISSUE #5101: Adding icons for Azure Bastion
We are in need of an icon that matches being redirected to the Bastion blade in Azure. 

We would highly appreciate if the computer arrow up or connect icons from google material icons could be added to EDS. Or if it would be possible to add the icon for Azure Bastion.

Image

Image

Image

eddiman:
Adds the `plug_connect` icon, requested in #5101.

Pulled from the [Assets Figma file](https://www.figma.com/design/BQjYMxdSdgRkdhKTDDU7L4KU/Assets) with the icon broker (`pnpm icons --force --only plug_connect`), so all three generated outputs are in sync:

| File | Change |
| --- | --- |
| `assets/icons/system-icons/hardware/plug_connect.svg` | New SVG, Hardware group |
| `packages/eds-icons/src/data.ts` | `plug_connect` export appended |
| `packages/eds-core-react/stories/assets/icons/system-icons.json` | Storybook metadata entry |

`packages/eds-icons/src/index.ts` is `export * from './data'`, so no index change is needed — verified `plug_connect` lands in `dist/icons.cjs` after a package build.

Per the discussion in #5101, this is the Material-style connect icon rather than the Azure Bastion product icon, since EDS doesn't take product/vendor icons.

### Second commit: figma-broker README

Pulling this icon surfaced some stale setup docs, fixed in a separate `docs:` commit:

- The README asked for a token with vague "read access" and **edit** access to the Assets file. Now names the actual scope, `file_content:read`, and corrects file access to **view** — the broker only calls `GET /v1/files/{id}` and `GET /v1/images/{id}`.
- Adds a troubleshooting entry for an **expired token**. This blocked the icon pull: an expired token returns a bare `403` on the file fetch, which reads like a permissions problem. Documents checking `/v1/me` (returns `401 Token has expired`) to tell the two apart, and notes that `--force` doesn't help.
- Replaces the token check that printed the secret with one that prints only a match count, matching the secrets rule in `AGENTS.md`. The repo's own pre-tool hook blocks the old command.

## Verification

- [x] `pnpm --filter @equinor/eds-icons run build` passes; `plug_connect` present in build output
- [x] Prettier clean on all changed files
- [x] Dry run confirmed a single match (`1/719 icons`) before writing


pomfrida (approved):
LGTM 👍

PR #5406, opened 2026-09-07
fix: reflect inherited disabled state in Checkbox, Radio and Switch in /next

### LINKED ISSUE #5393: Checkbox/Radio/Switch (next) don't reflect an inherited :disabled (e.g. from a disabled ancestor <fieldset>)
## Description

The `/next` (EDS 2.0) versions of `Checkbox`, `Radio` and `Switch` derive their disabled visuals from the React `disabled` **prop**, via `data-disabled` and `data-color-appearance` attributes set at render time. When the underlying native input is disabled by an ancestor `<fieldset disabled>` (standard HTML form behaviour), the control becomes functionally non-interactive but still renders as enabled.

This is the `/next` equivalent of #5349, which was filed against (and fixed for) the legacy components in #5392.

## Affected code

- `next/Checkbox/Checkbox.tsx` — `data-color-appearance={disabled ? 'neutral' : 'accent'}`, `data-disabled={disabled || undefined}` (lines ~73, 94–95)
- `next/Radio/Radio.tsx` — same pattern (lines ~50, 64–65)
- `next/Switch/Switch.tsx` — same pattern (lines ~46, 50)

The component CSS (`checkbox.css`, `radio.css`, `switch.css`) then selects on these attributes rather than on `:disabled`, so an inherited `:disabled` is ignored visually.

The same prop-keyed pattern also exists in `next/Slider`, `next/Input` and `next/Label` and should be checked while fixing this.

## Reproduction

```tsx
<fieldset disabled>
  <Checkbox label="A" /> {/* from @equinor/eds-core-react/next — looks enabled ✗ */}
</fieldset>
```

## Expected

The controls look disabled whenever their input matches `:disabled`, including via an inherited disabled `<fieldset>`, matching native controls.

## Suggested approach

Key the disabled styling off the CSS `:disabled` state (and `:has(input:disabled)` for parent-level styles) instead of render-time attributes, as done for the legacy components in #5392. Since `/next` uses vanilla CSS, this can likely live entirely in the component CSS files.

Note the interaction with `data-color-appearance`: the disabled appearance switch (accent → neutral) is currently also prop-keyed, so the fix needs to cover the token switch, not just the local styles.

pomfrida:
Fixes #5393

The /next `Checkbox`, `Radio` and `Switch` derived their disabled visuals from the React `disabled` prop via `data-disabled` and a `data-color-appearance` flip to `neutral`. An input disabled by an ancestor `<fieldset disabled>` was functionally inert but still rendered as enabled. This is the /next counterpart to #5392 (legacy components, #5349).

## Changes

- **CSS keys disabled styling off the input's `:disabled` state** — all `[data-disabled='true']` selectors in `checkbox.css`, `radio.css` and `switch.css` are replaced with `:has(.input:disabled)`, which matches both the prop and inherited disabling. `/next` already depends on `:has()` for the checked state, so this adds no new browser requirement.
- **Appearance-independent disabled grey.** `--eds-color-text-disabled` is switched by `data-color-appearance` and resolves to a tinted accent shade (not grey) under `accent` — and the token CSS is unlayered, so it can't be re-pinned from `@layer eds-components`. Each component now defines a pseudo-private `--_<component>-disabled-color: var(--eds-color-neutral-7)` and uses it in the disabled rules. Switch track/handle use the static `--eds-color-bg-disabled` (identical value to the previous neutral muted fill).
- **The `data-color-appearance={disabled ? 'neutral' : 'accent'}` flip is removed** (now redundant); the wrapper stays `accent`. The standalone `<span>`s no longer set `data-disabled`. `Field` keeps setting `data-disabled` — that's a documented contract with other consumers.
- **`field.css`**: the helper-message colour rule now also matches `:has(:disabled)`, so inherited disabling greys the helper for all Field-based components.
- Tests: fieldset-inheritance behaviour tests, regression guards for the removed standalone `data-disabled`, and the new `data-color-appearance` invariant.

## Verified

- 80/80 tests green, no snapshot churn.
- Measured computed styles in Storybook: prop-disabled, inherited-disabled and inherited+checked are pixel-identical (icon/label `#aeaeae`, switch track/handle `#e1e1e1`, `not-allowed` cursor, hover suppressed) for all three components; enabled states unchanged vs production Storybook.

## One deliberate visual change

Disabled Checkbox helper text is now disabled-grey (`#aeaeae`). In production it stayed `#585858` because the checkbox-level helper-colour override accidentally shadowed Field's disabled rule at equal specificity; this restores Field's documented behaviour ("only helper message changes color when disabled").

## Out of scope / follow-ups

- The issue mentions `next/Slider` and `next/Label` — neither exists; `Label` is `Field.Label` (no disabled handling). `next/Input` (and `Search`'s icon tone) have the same prop-keyed pattern — follow-up: #5412. Until that lands, a TextField/TextArea/Search inside `<fieldset disabled>` shows a greyed helper under an enabled-looking control (transitional half-state, detailed in #5412).
- Token pipeline finding: the static concept token `--eds-color-text-disabled` (and `border-disabled`) is shadowed by the dynamic appearance set redefining the same name — the static one is effectively dead code, and "disabled" text renders tinted under non-neutral appearances. Worth its own issue.

## Review follow-up (second commit)

- **Scoped the `field.css` disabled rule to `input/select/textarea:disabled`** — a bare `:has(:disabled)` also matched the hidden `<option disabled>` a Select renders for its placeholder, greying helper text on fully enabled Selects.
- **Merged the consumer `className` with the `input` class on Checkbox/Radio** — previously a consumer className replaced it, silently detaching every `:has(.input:...)` state selector (pre-existing for `:checked`, widened by this PR for `:disabled`).
- **Disabled states now override pseudo-private variables instead of properties** (per AGENTS.md), incl. new `--_track-bg`/`--_handle-bg` on the Switch (declared on the elements so the appearance-switched tokens resolve under `.control`).
- **`field.css` pins the disabled helper grey appearance-independently** (`--eds-color-neutral-7`, same rationale/TODO as the components) instead of the appearance-switched `--eds-color-text-disabled`.
- **Updated `BUILDING_EDS_2_COMPONENTS.md`** so newly scaffolded components use `:disabled`/`:has(...:disabled)` instead of the prop-keyed `data-disabled` + appearance-flip pattern this PR removes.

All states re-verified in Storybook after the changes (same measured values as above; enabled Select helper stays subtle-grey).

## Review follow-up, round 2

- **`Inherited disabled` stories** for all three components (real `<fieldset disabled>` around unchecked/checked/indeterminate) + matching `<Canvas>` sections in the Storybook `.docs.mdx` pages, so the state is reviewable and anchors a future visual spec.
- **Fixed a dead selector the round-1 doc update introduced**: the scaffold template in `BUILDING_EDS_2_COMPONENTS.md` used `.eds-componentname:disabled` on a `<div>`-rooted component — now `:has(:disabled)`, with a note that bare `:disabled` applies only when the root *is* the form control, and a caveat that prop-set `data-disabled` remains correct for components with no native control (Field's pattern).
- **Hygiene from the round-2 review**: Switch hover suppression now uses variable overrides like Checkbox/Radio (the `:not(:has(...))` guards are gone — verified by real hover in Storybook: no ring, track/handle stay grey when disabled; enabled hover unchanged); adjacent `:has(.input:disabled)` blocks merged; `className` placement documented in all three `*.types.ts` (Checkbox/Radio → input element, Switch → Field root); `field.css` notes that the `--_eds-field-*` colour variables are intended override points; test query/organisation tidied.
- Not taken, deliberately: collapsing the disabled grey into one Field-level variable (the standalone Checkbox/Radio wrappers have no `.eds-field` root, so it doesn't truly collapse — deferred to the semantic-token follow-up), and restoring `data-disabled` on the standalone wrappers (open to it if reviewers prefer symmetric DOM).


pomfrida:
@claude review

pomfrida:
@claude review

pomfrida:
Status on the two bot reviews above, for human reviewers:

**Round 1:** #1 (over-broad `field.css` `:has()`) and #4 (consumer `className` clobbering the `input` class) — fixed in 775da4f5c with test guards. The checkbox helper-colour duplication from #3 was removed in the same commit.

**Round 2:** #2 (dead `:disabled` selector in the scaffold doc) — fixed in 8d0361fa0, incl. the Field caveat. The story part of #1 landed in c3c1edd2c (`Inherited disabled` stories + Storybook docs sections). The hygiene points #6–#10 (Switch hover via variable overrides, merged `:has()` blocks, `className` JSDoc, `--_eds-field-*` override-point comment, test tidying) — 7da8a2d16.

**Deliberately not taken:** #3/#5 (single Field-level disabled variable — the standalone Checkbox/Radio wrappers have no `.eds-field` root, so it doesn't truly collapse; deferred to the semantic-token follow-up) and restoring `data-disabled` on the standalone wrappers (open to it if preferred).

**Tracked elsewhere:** the Input-family half-state (round 2 #3) → #5412, executed via the colour-migration tasks #5260/#5267/#5264/#5251, DoD point proposed on #5119. A Playwright visual spec remains a candidate follow-up; the CSS is verified by measured computed styles in Storybook (values in the PR description).

millus (approved):
LGTM ✨

PR #5409, opened 2026-09-07
fix!: route className and style to root element in all next input components

### LINKED ISSUE #5322: Select `className` does not size the outer field
### Describe the bug

Passing `className` to `Select` reaches an inner wrapper rather than the outer labeled field, so consumers can't size the whole control through the documented prop.

### Steps to reproduce

1. Render `<Select className="w-64" label="..." items={[...]} />`
2. Inspect the DOM — the class lands on an inner element, not the outermost field wrapper.
3. The overall control doesn't resize as expected.

### Expected behavior

`className` sizes/styles the full labeled field.

### Specifications

- Version: `@equinor/eds-core-react@3.0.0-beta.1`

### Additional context

Workaround: size an external wrapper element around `Select` instead of relying on `className` for outer sizing.

millus:
## Summary

Fixes a bug reported in #5322 where `className` on `Select` landed on the inner wrapper div instead of the outer labeled field — then discovered the same issue existed across all `/next` field-input components.

All five components now forward `className` and `style` to their outermost root element, consistent with the pass-through prop convention established in #5456 and how every major design system works (MUI, Radix, Mantine, designsystemet.no).

### What changed per component

| Component | `className`/`style` lands on | Notes |
|---|---|---|
| `Input` | Outer `div.eds-input-container` (root) | `containerClassName` removed (non-standard) |
| `TextField` | Outer `<Field>` wrapper (root) | |
| `Select` | Outer `<Field>` wrapper (root) | Fixes #5322 |
| `Search` | Outer `<search>` element (root) | |
| `Autocomplete` | Outer `div.eds-autocomplete` (root) | |

### Breaking changes

- `className` and `style` now land on the root element of each component instead of passing through to the inner `<input>` element
- `Input.containerClassName` has been removed — use `className` instead (it now targets the visual container, which is the root)
- Consumers using `className` to style the native input directly should switch to a descendant selector, e.g. `className="[&_input]:rounded-full"`

### Also updated

- Storybook argTypes: `style` added, `containerClassName` removed, descriptions updated to clarify which element is targeted
- Tests: styling assertions updated to verify `className`/`style` land on `container.firstChild` (root)

Closes #5322. Contributes to #5456.

### DIFF
--- a/packages/eds-core-react/src/components/next/Select/Select.test.tsx
+++ b/packages/eds-core-react/src/components/next/Select/Select.test.tsx
@@ -43,6 +43,14 @@ describe('Select (next)', () => {
       expect(screen.getByLabelText('Element')).toBeInTheDocument()
     })
 
+    it('applies className to the outer field element, not the inner select wrapper', () => {
+      const { container } = render(
+        <Select label="Element" options={elements} className="custom-class" />,
+      )
+      expect(container.firstChild).toHaveClass('custom-class')
+      expect(container.firstChild).not.toHaveClass('eds-select')
+    })
+
     it('renders helper message', () => {
       render(
         <Select label="Element" options={elements} helperMessage="Pick one" />,
--- a/packages/eds-core-react/src/components/next/Select/Select.tsx
+++ b/packages/eds-core-react/src/components/next/Select/Select.tsx
@@ -65,7 +65,7 @@ function SelectInner<T = string>(
   const displayErrorIcon = invalid && !disabled && !readOnly
 
   return (
-    <Field disabled={disabled}>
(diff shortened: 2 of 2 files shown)


millus:
@claude please review

pomfrida (approved):
Very nice cleanup 🙌 

Could we add a migration section to the changelog covering:
- containerClassName is removed from Input → use className
- className on Input, TextField, Select, TextArea, Autocomplete and Search now targets the root element, not the native element
- How to reach the inner element, e.g. .my-field .eds-input { … }

The commit msg says "all next input components", Checkbox/Radio/Switch is not in this PR so could we change to something like `fix!: route className and style to root element in next text input components`

Issue #152, opened 2019-12-04
Create [name] component


vnys:
**Description**

What are the implementation details the developer needs to know about?

**Documentation**

_Edit the following links to point to relevant information_

- [The storefront](https://eds.equinor.com/components/navigation-tabs/)
- [Figma](https://www.figma.com/file/0bGXR2sCwMVSDNyyzu5BXrO5/UI%E2%80%94User-Interface?node-id=256%3A4)
- [WAI-ARIA Authoring Practices 1.1](https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel)

**Sub tasks**

– [ ] Write tests
- [ ] Create set of tokens (from base tokens)
- [ ] Create react component

**General guidelines**

We support the latest version of evergreen browsers, so use CSS grids where appropriate. 

_Optional footer which might have a pointer to another issue  (see #000)_

PR #5414, opened 2026-09-07
fix: keep component styles when a style prop is passed

### LINKED ISSUE #5411: Passing style to button component seems to break existing styles
**Specifications**

Version:  0.3.1
Component:  Button

**Describe the bug**

Adding any sort of style to the button component seems to cause a break in the button UI. For instance, adding minWidth using the style prop on a secondary variant button seems to take the border away.

1. Implement a secondary variant button.
2. Add the "style" prop to it with a value like 'minWidth' or 'height'
3. The border which should exist on a secondary variant button goes away

Chibuzor-Nwemambu:
Reported in #5411: adding `style` to a secondary `Button` removed its border.

`style` was never destructured, so it arrived inside the remaining props and was spread *over* the explicit `style={styles.container}`, replacing it instead of merging into it. On a secondary button that dropped `borderWidth` and `borderColor`, which is the reported symptom, plus `borderRadius` and `overflow` on every variant.

This is the same mistake the existing test at `Button.test.tsx` already documents for `accessibilityState`, one prop over.

## Scope

Since the same shape could exist elsewhere, I audited all 19 migrated components. "Migrated" here means the component's directory has left `tsconfig.json`'s `exclude` list, per the package CLAUDE.md.

| Outcome | Components |
|---|---|
| Fixed | `Button`, `Button.Icon` (`Button/`), `Badge`, `Label` |
| Reordered to match, already safe, behaviour unchanged | `Divider`, `TextArea` |
| Already merged correctly | `Paper`, `SelectionControls/Switch`, `Input`, `Typography`, `_internal/PopInContainer` |
| No `style` in the public prop type, nothing to fix | `SelectionControls/Checkbox`, `SelectionControls/Radio`, `Link` |
| No own container style to lose | `Icon`, `PressableHighlight`, `Portal`, `Scrim`, `EDSProvider`, `ErrorBoundary` |

That accounts for all 19 directories under `src/components/`.

One fix lands outside that set: `Menu` is unmigrated, but the lint rule below caught a real instance there, so it is fixed too. It read `style={[styles.innerContainer, rest.style]} {...rest}`, which looks like it merges, but the trailing spread re-applies `rest.style` and replaces the merged array outright.

## Preventing recurrence

A `no-restricted-syntax` rule now rejects a props spread placed after an explicit `style` or `accessibilityState` attribute. It reports zero violations on this branch, and reverting `Button.tsx` to its original attribute order fails lint with an actionable message.

`accessibilityState` is included because this codebase has already been bitten by the identical shape on that prop once.

## Implementation notes

`mergePressableStyle` lives in `src/styling/utility.ts` rather than beside `Button`, because several of the 14 unmigrated components are Pressable-based and will need it as they migrate. It is internal and is not re-exported from `src/index.ts`.

It only produces the function form of `style` when the caller used the function form, so the common case stays a plain array. Elsewhere the existing `[ownStyle, rest.style]` idiom is used, since `View` and `Text` cannot take a function.

## Behaviour note

Callers who passed `style` and relied on it *replacing* the container style will now get a merge. That is the intent, and the reason this is a `fix` rather than a `feat`.

## Testing

9 new tests, 117 passing. Each new test was checked against the unfixed source to confirm it actually fails, which caught two initially weak assertions, one in the `Label` test and one in `Badge`. `Menu` is unmigrated and excluded from Jest, so the lint rule is what guards it.

## Follow-ups, not in this PR

Four findings for the migration tracker (#152, still in `design-system-mobile` pending transfer). All are pre-existing and none are regressions from this PR.

- `TextField` and `Search` accept `style` and forward it to the inner `TextInput` rather than the outer container, so `<TextField style={{ marginTop: 20 }} />` moves the text box and leaves the label behind.
- `Label` has the same inner-versus-outer shape: `style` now merges onto its two inner `Typography` elements rather than the container. It also spreads the remaining props onto both of them, so `<Label testID="x" />` produces two nodes with the same `testID`.
- `Switch` reads `rest.style` but never spreads `rest`, so `testID`, `nativeID` and the rest of its `ViewProps` are silently dropped.
- `Button`'s fill is painted by a child view covering the pressable, so `style` cannot change its background, and padding set through `style` leaves an uncoloured band around the fill.

Separately, `Button` exposes only `small` and `default` on both web and mobile. `inset.lg` already exists unused in the token scale, so a third size would be a small code change, but whether the design intends one is a question for design rather than a defect.

Closes #5411


Chibuzor-Nwemambu:
@claude review

Chibuzor-Nwemambu:
Thanks, useful review. Actioned and pushed in 895cfad4: the `Badge` assertion is now a baseline comparison, since `not.toBe("transparent")` passed on the unfixed source where `borderColor` is `undefined`. Good catch.

One correction. `Dialog` isn't an instance of the mirror-image bug. It reads `rest.style` at `Dialog.tsx:19` and merges it through the stylesheet factory at `:33` (`StyleSheet.flatten([defaults, props.style])`), so `<Dialog style={{ maxWidth: 400 }} />` works. Unusual routing, but not a dropped style. The gap in the rule is real, but with no instance in the codebase I've left the rule comment alone rather than documenting a hypothetical.

millus (approved):
LGTM ✨

PR #5417, opened 2026-09-08
feat(eds-icons): add classification icon

asbjornhaland:


eddiman:
Thanks for this, we want the icon, so we are taking it in, just through a different route.

Icons in EDS are not hand-added: they are generated from the [EDS Assets Figma file](https://www.figma.com/design/BQjYMxdSdgRkdhKTDDU7L4KU/Assets) by the icon broker (`scripts/figma-broker`), which writes the SVG asset, the Storybook metadata and the `@equinor/eds-icons` export in one pass. Anything edited by hand in those files drifts from Figma and gets overwritten the next time the broker runs.

Since we use icons from material icons, we tend to keep the name from that for simplicity, so the icon is now called "shield".

So we have added the icons in Figma and pulled it in properly, as two icons rather than one:

- `shield` — filled
- `shield_outline` — outlined

Tracking issue: #5420
Replacement PR: #5421

Closing this one in favour of #5421. Thanks for raising it, it's what prompted adding the icons.

Issue #5394, opened 2026-09-03
chore: verify mobile migration post-merge and close out design-system-mobile


pomfrida:
Follow-up to #5138 / #5384. The migration PR closes #5138 on merge, but two things can only happen after it lands: verifying that the release machinery actually works end-to-end from this repo, and closing out the old repo. Tracked here so they don't disappear into a closed issue.

## Post-merge verification

- [x] The two baseline tags (`eds-mobile-components@v0.3.1`, `mobile-storybook@v0.3.0`) are real ancestors of `main` after the merge (`git merge-base --is-ancestor <tag> main`)
- [x] First release-please PR for the mobile packages has the correct version bump and a changelog covering only post-migration commits
- [x] npm publish of `@equinor/eds-mobile-components` goes through trusted publishing end-to-end (trusted publisher entry: `publish_mobile_components.yaml`, no environment set)
- [x] The iOS TestFlight build triggers on the release (components-only release should also trigger it, per trigger_publish.yml)
- [x] eds-core-react's mobile docs resolve via the `workspace:^` link to `packages/eds-mobile-components/docs`, not the npm package

## Close out design-system-mobile

- [x] Triage the ~23 open issues before transferring - separate active work (component tickets, the 3 Slice epics, #152, #100, #88) from stale ones; #16 is a candidate for closing with an explanation instead (see the discussion on #5138)
- [x] Transfer the issues worth keeping to this repo
- [x] Archive equinor/design-system-mobile with a redirect README pointing here

pomfrida:
Ran the first verification steps after the merge - two findings, one action needed before #5395 goes anywhere.

**1. The baseline tags don't point into main's history.** `eds-mobile-components@v0.3.1` and `mobile-storybook@v0.3.0` point at the *original* design-system-mobile commits (3a90a68, d45b522), not the subtree-split commits that were actually merged (1355c21, 683d02b). The originals aren't ancestors of main, so the first checkbox above currently fails, and the compare link in any next release spans two unrelated histories.

**2. #5395 shouldn't be merged as-is.** The 0.3.2 entry is triggered solely by 3f7ad9e, a migration-branch commit whose only package file is CLAUDE.md. That path *is* in exclude-paths, but release-please only skips a commit when *all* its files are excluded, and this one also touched root files (.prettierignore, AGENTS.md, eslint.config.mjs). Migration housekeeping shouldn't cut a release.

**Suggested fix for both at once** - re-point the two tags at the migration merge commit:

```
[...]

PR #5447, opened 2026-09-08
build: declare babel-preset-expo in mobile-storybook

Chibuzor-Nwemambu:
The iOS TestFlight build has never succeeded from this repository. Today's run, the first one ever, failed. This is the fix, and it's part of the post-merge verification in #5394.

## What failed

The native compile and link both succeeded. It failed in the JS bundling phase of the Xcode archive:

```
error: ../../node_modules/.pnpm/expo-router@55.0.17_.../node_modules/expo-router/entry.js:
       Cannot find module 'babel-preset-expo'
** ARCHIVE FAILED **
Process completed with exit code 65
```

## Why

`apps/mobile-storybook/babel.config.js` names `babel-preset-expo` as its preset:

```js
module.exports = function (api) {
  api.cache(true);
  return { presets: ["babel-preset-expo"] };
};
```

The app has therefore always depended on it, but never declared it. Under pnpm's strict layout a package can only resolve what it declares, so it was unresolvable. `require.resolve` fails locally too, not just in CI.

Checking the full history, it was never declared in `apps/mobile-storybook/package.json`, including in the pre-migration commits from `design-system-mobile`. It presumably resolved there by accident of that repo's directory layout. So this is not a regression from the migration; the workflow had simply never run far enough to reach bundling before.

## Why this fix rather than the alternatives

| Option | |
|---|---|
| Declare it in the app (this PR) | What Expo documents for any project with a custom `babel.config.js` |
| Hoist via `.npmrc` | Fixes the symptom by weakening pnpm's strictness across the whole monorepo, and would hide the next undeclared dependency |
| Resolve an absolute path in `babel.config.js` | Works, but replaces a standard generated config with a bespoke one |
| Delete `babel.config.js` | Not viable for a prebuilt Expo project |

## Why the lockfile shrinks by 116 lines

Nothing new is downloaded; `babel-preset-expo@55.0.24` was already in the store as a transitive dependency of `expo`. But declaring it re-keys `expo-router`'s peer-dependency hash:

```
- expo-router@55.0.17(db924759debce275f250b522099810c9)
+ expo-router@55.0.17(09ccd46779381dc0a2eec6c050f029a5)
```

`expo-router` previously resolved a private duplicate instance of `babel-preset-expo`; now it shares the app-level one. The deleted lines are that duplicate collapsing, including a `jest` subtree and the 86 `libc` fields belonging to its platform-specific binaries. The dependency graph got simpler.

## Verification

Reproduced the failing operation locally, transforming the exact `expo-router/entry.js` that CI failed on, with the app as Babel's root. It fails without this change and succeeds with it.

`build:` is a hidden type, so this deliberately does not bump `apps/mobile-storybook` or cut a release. The archive can be verified by dispatching `build_release_ios.yaml` directly.

### DIFF
--- a/apps/mobile-storybook/package.json
+++ b/apps/mobile-storybook/package.json
@@ -51,6 +51,7 @@
   "devDependencies": {
     "@babel/core": "^7.29.6",
     "@types/react": "~19.2.17",
+    "babel-preset-expo": "~55.0.24",
     "eslint": "^9.25.0",
     "eslint-config-expo": "~55.0.1",
     "eslint-import-resolver-typescript": "^4.4.4",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -392,7 +392,7 @@ importers:
         version: 55.0.16(expo@55.0.23)(react-native@0.83.6(@babel/core@7.29.7)(@types/react@19.2.18)(react@19.2.8))(react@19.2.8)
       expo-router:
         specifier: ~55.0.16
-        version: 55.0.17(db924759debce275f250b522099810c9)
+        version: 55.0.17(09ccd46779381dc0a2eec6c050f029a5)
       expo-splash-screen:
         specifier: ~55.0.21
         version: 55.0.23(expo@55.0.23)(typescript@5.9.3)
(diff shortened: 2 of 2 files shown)


Chibuzor-Nwemambu:
@claude review

pomfrida (approved):
Lgtm

Issue #5360, opened 2026-08-24
Validate refined Colour & Token System with designers


lavinia-chrystal:
**Purpose**
 Conduct a validation checkpoint with designers who previously provided input into the token work to ensure the redefined token structure is intuitive, usable, and supports real-world workflows.

**Description**
 Following the token redefinition work, run targeted feedback sessions with a representative group of designers  to validate:

1. 

The objective is not to redesign the solution, but to identify usability issues, ambiguities, or adoption risks before final release.

**Research Activities**

- Re-engage selected participants from previous token workshops/interviews
- Include additional developers who consume EDS tokens directly
- Walk through the proposed token structure and documentation
- Gather feedback through structured discussions and practical exercises
- Document findings and recommendations

**Acceptance Criteria**

- Validation sessions completed with designers
- Feedback synthesized and documented
- Key usability concerns identified and assessed

Required adjustments captured as follow-up issues
Team agreement on readiness of the token structure for release

Definition of Done

Research findings shared with EDS team
Any critical adoption or usability issues addressed or explicitly accepted
Evidence available that target users can successfully understand and use the revised token model

---

## Checklist — running the whole test

**Before (setup)**
- [X] Draft this blueprint (protocol, tasks, matrices) 
- [x] Prepare the reference card as a flat image / locked frame
- [x] Documentation

**Recruit — need 10 testers**
- [x] Contact users
- [x] Send calendar invitations (30–35 min each)
- [x] Send a short reminder the day before

**After all sessions**
- [ ] Combine matrices across all 10 testers
- [ ] Fill the Findings roll-up (every 0 and 1), ranked
- [ ] Work out the overall readiness score + verdict
- [ ] Note any dev-side items for Edward's handoff
- [ ] Share the summary with the team

alexlvrs:
# First draft for a blueprint on user testing
---

## Method

The tester works through a few tasks in Figma while you watch and score. No wrong answers — you're testing the system, not the person. Ask them to **think out loud**; their first, unfiltered reactions are what you're after.

There are two kinds of score, so keep them apart:
- **What you score while watching — 0 to 3** (Matrices A and B). This is your judgement of how the task went.
- **What the tester rates at the end — 1 to 5** (Matrix D). This is their own opinion, in their words.

**Your 0–3 scale:**

| Score | Meaning |
|---|---|
| **0** | Fail — couldn't do it, or got stuck |
| **1** | Weak — did it, but wrong or confused |
[...]

PR #5453, opened 2026-09-09
docs: 📝 document the 3.0.0-beta colour foundation, and version the docs

### LINKED ISSUE #4745: 📚 Update token documentation

### LINKED ISSUE #5399: 📚 Write colour migration guides (EDS 1.0 and 2.0.0-beta → redefined tokens)
## Description

Consumers on two different generations of EDS colour need a path to the redefined tokens, and neither can get there by mechanical find-and-replace. This covers **two migration guides** (or two clearly separated sections of one guide):

1. **EDS 1.0 → redefined tokens** — from the hand-picked hex palette (`eds-colors`, `theme.colors.*`) to the algorithmic OKLCH ladder.
2. **EDS 2.0.0-beta semantic/dynamic colours → redefined tokens** — from the beta's semantic + dynamic colour layers to the redefined semantic layer, which drops dynamic colour entirely.

Sub-task of #4745.

### Shared
- Explain *what changed and why* before the mapping tables — the ladder is contrast-driven, not lightness-driven, so "closest hex" is the wrong mental model.
- Old name → new name mapping tables, with an explicit **no direct equivalent** column rather than a forced nearest match.
- Call out where a 1:1 mapping does not exist and a decision is required (pick by role, not by value).
- Note the visual delta consumers should expect, and that APCA contrast is the acceptance criterion (measured against `background.surface`).
- Machine-readable mapping (JSON/CSV) alongside the prose, so agents and codemods can consume it.
- Usable by an AI agent, and tested by one — same bar as #4745.

### EDS 1.0 → redefined
- `eds-colors` / `theme.colors.*` / raw hex usage → semantic tokens.
- Interactive, text, border, and background groups mapped separately; the 1.0 palette has no equivalent of the step roles (1–2 are planes, fills start at 3).
- Dark mode: 1.0 had no real dark scheme, so this is a new-work section, not a mapping.

### 2.0.0-beta → redefined
- Semantic layer renames and regrouped tokens.
- **Dynamic colour is removed** — what replaces each dynamic token, and what to do where nothing does.
- Density rename (compact / comfortable) where colour tokens were density-coupled.
- Alpha → opaque: the scale is opaque, disabled uses opaque steps. Alpha survives only where the backdrop cannot be known (`--eds-elevation-*`, `overlay.scrim`, dataviz) — say so explicitly so consumers do not strip it there.
- Border tone ladder: subtle / medium / strong = steps 4 / 7 / 9 for all six tones (neutral's old 4/4/6 was a rename artifact).

## Definition of Done
- [ ] EDS 1.0 → redefined colour mapping written and reviewed
- [ ] 2.0.0-beta semantic/dynamic → redefined colour mapping written and reviewed
- [ ] Machine-readable mapping published next to the prose
- [ ] Tokens with no direct equivalent listed, with guidance on how to choose
- [ ] Published in the token docs and shared with the team
- [ ] Tested by an AI agent doing a real migration on a sample codebase

eddiman:
## What

Rewrites the colour foundation docs for 3.0.0-beta, adds a migration guide, and cuts a docs version
so the 2.0.0-beta pages stop being overwritten.

Resolves #4745 for colour.

Spacing and typography are deliberately **not** in scope; they are still drafted outside this repo
and will be ported separately.

## Why

The reasoning behind the 3.0.0-beta colour foundation was written down almost nowhere, so settled
questions kept being re-litigated and neither a new team member nor an agent had anything to check a
change against.

Two rounds of review shaped this beyond the first draft:

- **Alex's design-side test** ([results](https://github.com/equinor/design-system/issues/5360#issuecomment-5571395657)),
  7 testers. Colour scored best of any category, so nothing structural changed. What testers wanted
  was the *why and when*, plus swatches and a way to find a colour by sight.
- **A markup review of the three published pages**, 26 annotations. The dominant theme was
  "visualise it", six times, always where a page described a relationship between colours in prose.

## Versioning

The colour foundation was being rewritten in place, so 2.0.0-beta was being overwritten rather than
superseded. Three versions now:

| URL | Version | |
|---|---|---|
| `/docs/` | 1.1.0 | last stable, unchanged |
| `/docs/2.0.0-beta/` | 2.0.0-beta | frozen |
| `/docs/Next/` | 3.0.0-beta | current |

The snapshot was cut from `origin/main` rather than the working tree, so it holds the 2.0.0-beta
pages and not the rewrite.

Two judgement calls, both worth a look in review:

- **`lastVersion` is pinned to 1.1.0.** Without it Docusaurus promotes 2.0.0-beta to the root and
  pushes 1.1.0 to `/docs/1.1.0/`, moving every existing stable URL. It also keeps a beta off the
  default landing page.
- **2.0.0-beta carries no banner.** Docusaurus aims its "unmaintained" banner at `lastVersion`, so
  it told a 2.0.0-beta reader to go to 1.1.0, which is backwards. The useful destination is
  3.0.0-beta, which the banner cannot express, so the version dropdown carries that instead.

## Content

- **Introduction** and **Getting Started** rewritten: the level model is split into interactive and
  non-interactive, `link` is documented as a role, and the foreground pairing rules are rendered as
  painted specimens rather than a table of names.
- **The palette** is new. It shows all 15 steps for all 6 tones, so the jump between border steps
  and fill steps is visible rather than asserted, and the seven hue anchors are drawn rather than
  described.
- **Migration** is new. It maps both routes, 1.x and 2.0.0-beta, group by group, with the two colours
  painted side by side. Each table is headed by the version it maps from rather than by "old" and
  "new", and the 2.0.0-beta tokens are written dotted so both sides of a row share one notation.
- **The sidebar** gains the three new Colour pages, ordered so the four guidance pages (Introduction,
  Getting Started, The palette, Migration) read in sequence and `reference` sits last as lookup.
- **`src/components/` is reorganised** into `site/`, `tokens/` and `colour/`. `HomepageFeatures`,
  `Prerequisites` and `TypeSpecimen` move without changing behaviour; the import updates in
  `src/pages/index.tsx` and the two `getting_started_development.md` / `typography.md` touches are
  the knock-on. `tokens/TokenAnatomy` is the new diagram component; Getting Started is what
  exercises it, in colour mode. Its `DensitySwitch` and `spacing` mode ship unexercised on purpose:
  they are part of the component, and the spacing and typography pages that use them are the next
  port rather than dead code.

One finding shapes the migration page: **not one value carries over.** None of the 115 semantic
2.0.0-beta colours resolves to the same hex as any 3.0.0-beta token. Mapping by "closest hex"
produces a palette that fails contrast in one scheme or the other, so every pair is mapped by role,
and where there is no equivalent the guide says what to decide instead of forcing a match.

## Generation

The scripts that produce and verify these pages now live in `apps/design-system-docs/scripts` and
read only `packages/eds-tokens/src/tokens`. They previously lived outside the repo, which made the
claim that anyone could regenerate the numbers untrue.

```bash
pnpm generate:colour-reference
pnpm check:colour-docs
```

`check:colour-docs` is new. Docusaurus catches none of what it checks: a mistyped custom property
renders as an unstyled element and a stale count renders as a smaller grid, both without error. It
verifies token names in prose, `--eds-*` properties in prose and components, migration mapping
targets, and the structural counts the components hold as literals. Every guard was confirmed by
planting a fault.

Only `reference.mdx` between its `GENERATED` markers is generated. Everything else is hand-written.

## Note for reviewers

`src/css/custom.css` now imports the 3.x token bundle **by relative path**. The site previously
loaded only the 2.x build, so every `var(--eds-background-*)` these docs document was unresolvable on
our own site. The package exposes that output as `./next/css/*` on the beta tag only, injected at
publish time per ADR-0009, so a workspace app cannot reach it by specifier. It collapses into a
single import when the Tokens Studio output moves onto the final specifiers. Reasoning is in the docs
app README.

## Verification

- `pnpm typecheck` clean
- `pnpm build` succeeds; the one pre-existing broken anchor on the photography page
  (`resources/#external-resources`) is now reported three times, once per version
- `pnpm check:colour-docs` clean across 7 pages and 13 components: 319 semantic tokens accepted, 263
  of them colour, and 317 custom properties
- Rendered and checked in both colour schemes

## Still open

- **`background.backdrop` and `overlay.scrim` describe the same job.** Whether both tokens should
  exist is a token decision, not a docs one, so the guidance pages do not explain either of them:
  their entry on the Introduction page is replaced by a source-only stub recording why. They remain
  visible where leaving them out would be wrong: **Reference** lists both, because it is a complete
  index of what the source ships, and **Migration** still maps `ui.background__scrim`,
  `ui.background__overlay` and `bg.backdrop` onto `overlay.scrim`, with a note that the pair is under
  review. When the question is settled, the stub is the place to write the guidance back in, with
  fresh wording rather than the old.
- Data-visualisation ordering and accessibility guidance is not written. The palette is shown, the
  guidance needs whoever owns that track.

Closes #5399 — the colour migration guides for both routes, 1.x and 2.0.0-beta.


eddiman:
@claude

alexlvrs:
Screenshot 2026-09-15 at 14 56 20

Hello, it need read-only too here maybe?

eddiman:
> Screenshot 2026-09-15 at 14 56 20
> Hello, it need read-only too here maybe? http://localhost:3001/docs/Next/foundation/colour/getting_started#foreground-pairing

But read-only is a semantic state we define ourselves by using other colors, not defined directly by our colors. We don't have a `background/read-only` or `text/read-only`

alexlvrs:
Hey Edvard, my main things is that the interactive states aren't really discoverable. The pages only ever show the emphasis fill, so selected and the muted hover/pressed states are invisible unless you already know they exist. Same for interactive borders.

I'd like to propose a small reorg + a states section:

Structure — split Getting Started into two, keeping "The palette" as the underlying stuff (your suggestion):

- Getting Started → setup only: reading a token name, scheme switching, TypeScript, Figma/designers
- The Palette → all colours + tokens (unchanged)
- Usage (new) → the how-to: levels & states, foreground pairing, worked examples, do's/don'ts, mistakes

Why: it gives the states story one obvious home, and keeps Getting Started short. The split is written so Usage is self-contained — if it ever feels wrong it's a cut-and-move, not a rewrite.

What I'd add to Usage:

- A "pick a level → the states come attached" explainer, so it's clear there's more than one hover (muted.hover ≠ emphasis.hover ≠ selected.hover).
- A painted level × state matrix (mirrors the foundation Figma frame 2466-2), for both background and border.
- The border asymmetry made explicit: border ships muted + emphasis only, no selected — selection shows through the fill.
- A note that the same tokens colour a button and a table row (neutral.muted.hover = hovered row, neutral.selected.* = selected row), plus a selected-item example.
- On foreground: spell out that text.on-emphasis.accent and text.inverted are interchangeable on an emphasis fill — both pass contrast — using the accent button as the case, with the one trade-off noted (on-emphasis is per-tone and follows the fill; inverted is one fixed value).

I've already prototyped it locally to make it easy to visualize. I can try my little PR skills if that something that would help (if it works, it works, if not you have the main idea :) )

Issue #5442, opened 2026-04-20
Tracking: Findings from component migration


Chibuzor-Nwemambu:
As we redesign components to match EDS Figma, we're surfacing gaps, inconsistencies, and open questions that don't always fit the scope of the current PR. This issue is a running log so the team can see them in one place and decide what to action.

Each item is a short description + rationale. Items that become real work should be filed as their own issues and linked here.

## Findings

- [ ] **Radio has no `RadioGroup` wrapper** — mirrors `@equinor/eds-core-react`, but web radios get group semantics for free via the native `name` attribute. On mobile, that means no element reports `accessibilityRole="radiogroup"` and no arrow-key / swipe navigation between options. Material UI, Radix, Gluestack, and Tamagui all ship a group wrapper for this reason. Surfaced while writing the mobile Radio MDX doc.
- [ ] **Input error state is visual only** — when `invalid` is true, the component shows an error icon but does not expose the error semantically to assistive tech (no `accessibilityHint`, no `accessibilityLiveRegion`, no `aria-invalid` equivalent). Consumers must wire a11y themselves via `accessibilityHint` or an adjacent live region. Likely worth standardising at the component level. Surfaced while writing the mobile Input MDX doc.
- [ ] **Labelless Checkbox variant may overlap with Cell selection** — Checkbox renders as a circular hit target when `label` is omitted. The standard iOS multi-select pattern is for the row/cell to own the hit target and render a static checkmark indicator, not a tappable labelless Checkbox. If `Cell`/`ListItem` gains a selection pattern, shipping both is an API inconsistency worth resolving. Decision parked for pre-1.0 cleanup (we're on `beta`). Surfaced while reviewing the mobile Checkbox MDX doc.
- [ ] **Selection controls don't honor arbitrary native props** — Radio and Checkbox define no base prop type and don't spread a `...rest` onto the internal `Pressable`, so consumers cannot pass `accessibilityHint`, `testID`, `onLongPress`, `hitSlop`, custom `style`, etc. Switch extends `ViewProps` in its type but only forwards `rest.style` to the `Pressable` — other `ViewProps` are typed but silently ignored (type lie). Meanwhile Input extends `TextInputProps` and Button extends `PressableProps`, both forwarding correctly. This limits selection controls for real-world needs (e2e test IDs, custom a11y hints, gesture overrides). Surfaced while reviewing the mobile Radio MDX doc.
[...]

PR #5454, opened 2026-09-09
chore: stop shipping docs in the eds-mobile-components package

### LINKED ISSUE #5452: chore: stop shipping docs/ in the eds-mobile-components npm package
`packages/eds-mobile-components/package.json` lists `docs` in `files`, so all 14 MDX files ship in the npm tarball. That was how `eds-core-react` obtained them when the two lived in separate repositories.

Since the migration, `eds-core-react` reads them through the `workspace:^` devDependency instead, so the published copy is unused.

**The change:** remove `"docs"` from the `files` array.

**Why it's safe:** all 8 importers live in `packages/eds-core-react` and resolve through the workspace link. `files` only affects the published tarball, so in-repo resolution is unaffected.

**Why it's worth doing:** the npm copy only updates on release while the workspace copy is always current, so the two drift between releases. Two sources of the same doc is a trap for whoever finds them later. The size saving is incidental, roughly 40KB of a 1.1MB package.

**One caveat:** we can't tell whether any external consumer reads the docs out of `node_modules`. The package is pre-1.0 and mobile hasn't been announced widely, so the risk is low, but worth a thought before merging.

Surfaced while verifying the workspace-link checkbox in #5394.

Chibuzor-Nwemambu:
Closes #5452

## What

Removes `"docs"` from the `files` array in `packages/eds-mobile-components/package.json`, so the 14 MDX files stop going into the npm tarball. Adds a two-sentence note to the package's `CLAUDE.md` so the entry does not get added back.

## Why

`docs/` shipped in the tarball because that was how `eds-core-react` obtained the files while the two projects lived in separate repositories. Since the monorepo migration it reads them through the `workspace:^` devDependency, so the published copy is never read and drifts out of date between releases. Two sources of the same document is a trap for whoever finds them later. The 40KB saving is incidental.

## Why it is safe

| Check | Result |
| --- | --- |
| Importers | All 8 live in `eds-core-react`, 5 of them under `/next`. Every one resolves through the workspace symlink, which `files` does not affect |
| Published surface of the consumer | `eds-core-react` publishes only `dist/*` and `build/*` behind an `exports` map with a single `.` entry, so its `.docs.mdx` files never leave the repo and no external consumer ever sees the `@equinor/eds-mobile-components/docs/*.mdx` specifier |
| Other routes into the tarball | None. No `.npmignore`, no `exports` map, no `publishConfig`, and the build config never copies `docs` into `dist` |
| CI, turbo, docs site | No workflow or task reads the docs from an installed copy |
| Lockfile | Unchanged. `files` is publish metadata, not a dependency field |

## Verification

`npm pack --dry-run` against the published 0.3.2 listing:

| | Published 0.3.2 | This branch |
| --- | --- | --- |
| Entries | 394 | 380 |
| `.mdx` files | 14 | 0 |

The only difference between the two file lists is the 14 `docs/*.mdx` removals. `README.md` and `package.json` remain, as npm always includes those regardless of `files`.

Resolution was checked with the change applied. `packages/eds-core-react/node_modules/@equinor/eds-mobile-components` is a symlink to the workspace directory and all 14 MDX files stay reachable through it.

## Notes

Consistent with the rest of the repo. No other EDS package publishes its documentation, and `/next` in particular builds to `dist/esm-next` plus `build/` CSS with no MDX. The one apparent exception, `eds-tokens` shipping `instructions/*`, is the opposite case, since those exist for consumers' AI tooling to read out of `node_modules`.

`chore` is a hidden type, so this triggers no release. The slimmer tarball first reaches npm with the next `feat` or `fix` in mobile.

One caveat carried over from the issue: we cannot tell whether an external consumer reads the MDX out of `node_modules`. The package is pre-1.0 and mobile has not been announced widely, so the risk is low.

A follow-up worth deciding separately: neither the `/next` release unit nor `eds-mobile-components` excludes its co-located docs from `exclude-paths` in `release-please-config.json`, so a doc-only edit still bumps a version and republishes an identical `dist`. Stable `eds-core-react` does exclude its `stories` directory. Logging this on #5442 rather than fixing it here, so it gets decided once for both.

### DIFF
--- a/packages/eds-mobile-components/CLAUDE.md
+++ b/packages/eds-mobile-components/CLAUDE.md
@@ -213,6 +213,8 @@ If the finding needs real work, open a dedicated issue and link it next to the c
 />
 ```
 
+**`docs/` is not published.** It is left out of `files` in `package.json` on purpose, because every consumer reads it through the workspace symlink. Do not add it back (equinor/design-system#5452).
+
 ### Component Development Checklist
 
 - [ ] Uses `EDSStyleSheet.create` for all theming
--- a/packages/eds-mobile-components/package.json
+++ b/packages/eds-mobile-components/package.json
@@ -33,8 +33,7 @@
     "files": [
         "LICENSE",
         "CHANGELOG.md",
-        "dist",
-        "docs"
+        "dist"
     ],
     "homepage": "https://eds.equinor.com",
     "keywords": [


Chibuzor-Nwemambu:
@claude review

pomfrida (approved):

Issue #5387, opened 2026-09-02
docs: living "Breaking changes in EDS 2.0" overview page


pomfrida:
Part of #5383.

Create a consumer-facing page listing the **intentional** breaking changes per component between EDS 1.x and EDS 2.0, kept up to date as changes land during the beta. This lets beta testers tell intentional changes from accidental ones without waiting for the full migration guide, which ships at the stable 3.0.0 release. At graduation this page grows into (or feeds) that guide.

**Single source of truth (decided):** this page is the only living document. One section per component (explicit, stable anchor ids — not auto-generated from headings), and each component's Docusaurus page carries a static deep link to its section (e.g. `.../breaking-changes#button`) — no per-component prose to keep in sync. The standard link line goes into the component-doc template in `documentation/agent-instructions/COMPONENT_DOC_STYLE.md`. Where the 1.x → 2.0 mapping isn't 1:1 (TextField → Field composition, Autocomplete multi-select → Combobox), index by the **EDS 1.x component name** too, so consumers searching for the old name find the entry.

**Scope — what counts as a breaking change here:**

- ✅ API: renamed/removed props, changed accepted values, changed composition
- ✅ Behaviour and DOM/test contract: changed roles, markup structure, or test-ids that break consumer tests (cf. the 1.x popover change that required jest setup changes downstream)
- ❌ Fine-grained visual changes (spacing/colour tweaks) — not tracked per entry during beta; covered by the general beta disclaimer

Seed it with the known changes from #4484:

- **Button**: `variant` rename (`contained` → `primary`, `outlined` → `secondary`), `color` → `tone`
- **Input**: `leftAdornments`/`rightAdornments` → `startAdornment`/`endAdornment`
- **TextField**: `helperText` → `helperMessage`, new `Field` composition
- **Icon**: `size` from numeric to semantic values (`'sm'`, `'md'`, `'lg'`, …)

Plus, once their issues resolve:

- **Autocomplete**: single-select only; multi-select moves to Combobox (#5385)
- **Tooltip**: reduced `placement` union (#5386)

- [ ] Decide location (Docusaurus about-section page and/or the Storybook "EDS 2.0 (beta)/About" page)
- [ ] Write the seeded first version, one anchored section per component
[...]

Issue #5383, opened 2026-09-01
Document intentionally breaking changes in EDS 2.0, and how to migrate from 1.0


mhwaage:
In the present version of EDS 2.0, several breaking changes are introduced. On a high level:

 - Styling is changed, which can lead to layouts breaking in consuming applications
 - The documented use case of some components (notably: Autocomplete) is changed
 - Practically all interfaces are broken by removing/adding required props, changing accepted literal values, or a combination of the two
 - Some components are missing, although this appears to be due to EDS still rolling out new versions of components

While some of these are obviously intentional, others appear to be oversights, c.f. t[he recent slack discussion about tooltip positioning options](https://equinor.slack.com/archives/CJT20H1B9/p1787140458340689?thread_ts=1787042118.097619&cid=CJT20H1B9).

Given the amount of breaking changes currently being rolled out, it is not feasible for consumers of the library to infer whether a change is intentional or accidental. Moreover, in the case of intentional changes, the lack of a migration guide makes it hard to test the components, since the intended semantic replacement of current use cases is unknown.

Please document intentionally breaking changes, and recommendations on how to migrate from 1.0 to 2.0.

pomfrida:
Thanks for raising this, and you're right that it's currently hard to tell intentional changes from accidental ones.

Where this connects to existing work:

- #4484 tracks codemods for the 1.0 → 2.0 component API changes 
- #5121 and #5203 cover the consumer migration path for the token/CSS side (2.x → 3.0 tokens, colour-scheme changes)

We would also need a consumer-facing overview of intentional breaking changes per component — so we'll keep this issue open for that.

The full 1.0 → 2.0 migration guide is planned for when the APIs stabilise at the stable release; writing it now would mean another document to keep updated before everything is completely settled.  What we should do in the meantime is mark breaking changes explicitly in the beta release notes and keep a living "breaking changes so far" list per component, so you can at least tell intentional from accidental during testing.

On Autocomplete specifically: the narrowing to single-select is intentional (noted in the [new docs](https://eds.equinor.com/docs/Next/components/inputs/autocomplete#when-to-use)), and Combobox component is on the plan to replace today's Autocomplete functionality. 
But you're right that neither the changelog nor the docs say what replaces the multi-select use case so we'll make sure the breaking-changes list covers that.

mhwaage:
I don't understand how your can be ready for beta-testing if the API is so unstable that a provisional migration guideline would be counterproductive.

[...]

Issue #5323, opened 2026-08-11
Checkbox `className` replaces its internal input class


HaakonSvane:
### Describe the bug

Passing `className` to `Checkbox` replaces the class that visually hides the native checkbox input, instead of merging with it. Both the native checkbox and the EDS-styled checkbox render simultaneously.

### Steps to reproduce

1. Render `<Checkbox className="some-custom-class" />`
2. Observe the native browser checkbox renders alongside the styled EDS checkbox instead of being hidden.

### Expected behavior

`className` merges with the component's internal classes rather than overwriting the one that hides the native input.

### Specifications

- Version: `@equinor/eds-core-react@3.0.0-beta.1`

### Additional context

Workaround: avoid passing `className` directly to `Checkbox`; apply layout styling to an external wrapper element instead.

pomfrida:
Radio has the same bug: `className="input"` is set before `{...rest}` is spread on the `<input>` (`next/Radio/Radio.tsx:23,26`), so a consumer `className` replaces the class that hides the native radio. Suggest fixing both in the same PR. Found while writing the Breaking changes page (#5455); the general question of which element `className` should land on is #5456.

PR #5455, opened 2026-09-11
docs: document breaking changes in EDS 2.0 per component in Storybook

pomfrida:
Part of #5383, implements #5387.

Consumers testing EDS 2.0 cannot tell an intentional breaking change from a bug, and the full migration guide only ships when EDS 2.0 becomes stable. This adds the living "Breaking changes in EDS 2.0" page in Storybook, next to the About page, listing the intentional changes per component compared to EDS 1.0.

## Changes

- New Storybook page `EDS 2.0 (beta) / Breaking changes` (`stories/docs/BreakingChanges.mdx`):
  - what counts as breaking on the page (props, composition, markup and behaviour) and what is not tracked (visual changes)
  - overview table with one row per shipped `/next` component, indexed by the EDS 1.0 name as well
  - shared changes across the form components (`variant` → `invalid`, `helperText` → `helperMessage`, derived ids, `role="alert"` on helper messages)
  - one section per component with props, composition and markup changes. Hidden anchors for old names (`#nativeselect`, `#inputwrapper`, `#label`, `#helpertext`) so the old names can be deep-linked
  - an "Under review" section for changes the team has not yet confirmed as final
- Sorted right after About in the sidebar (`.storybook/preview.mjs`)
- Link from the About page under Learn More
- Convention added to `BETA_RELEASE_GUIDE.md` (Breaking Changes Policy) and the anti-patterns checklist in `BUILDING_EDS_2_COMPONENTS.md`: a PR that changes a published `/next` API or markup contract updates the page

## How the content was produced and verified

Each shipped `/next` component was diffed against its EDS 1.0 counterpart in this repo (types, component source, stories, `next/CHANGELOG.md`). Every claim on the page was then checked again, claim by claim, against source with file and line evidence; about 20 claims were wrong or imprecise and have been corrected (see the second commit). Only changes that can be pointed to in source are listed.

The page now separates three states, explained at the top of the page:

- **Unmarked**: intentional change.
- **Known issues**: behaviour we consider a bug. Listed so consumers are not surprised, each linked to its bug report. Filed from this work: #5457 (Switch native `size`), #5458 (Chip `aria-pressed`), #5459 (Tooltip `aria-describedby` and open delay), #5460 (Select `multiple`), #5461 (Dialog dismissable, scroll lock, `aria-describedby`), #5462 (Icon small dataset), #5463 (Accordion `disabled`). Already existing: #5323 (Checkbox/Radio `className`), #5322 (Select `className`), #5103 (Menu), #5456 (where `className`, `style` and `id` land).
- **Under review**: implemented as described, but design has not confirmed it is final: Banner default role `status`, Tooltip default placement `top`, no `warning`/`success` on the Input family, Divider variants.

## Also in this PR

- EDS 2.0 components on the page itself: a Banner pointing to the issue tracker, a Chip row for jumping to a component, and Badge markers for New / Known issues / Under review.
- `Checkbox.types.ts`: the `indeterminate` JSDoc said the consumer must set the native property via ref; the component does it itself. Comment-only fix.

## Not in this PR

The "Migrating from EDS 1.x?" link line for the component-doc template (`COMPONENT_DOC_STYLE.md`) and the Docusaurus component pages waits for #4989, like the rest of the docs site.

## Verification

- The MDX compiles with `@mdx-js/mdx`
- Prettier passes on the new and edited files. The pre-existing prettier warning in `BETA_RELEASE_GUIDE.md` is not from this change
- Anchor ids and the sidebar position are not verified in a running Storybook; please check the deploy preview

### DIFF
--- a/packages/eds-core-react/stories/docs/BreakingChanges.mdx
+++ b/packages/eds-core-react/stories/docs/BreakingChanges.mdx
@@ -1,0 +1,310 @@
+import { Meta } from '@storybook/addon-docs/blocks'
+
+<Meta title="EDS 2.0 (beta)/Breaking changes" />
+
+# Breaking changes in EDS 2.0
+
+This page lists the intentional breaking changes between EDS 1.0 components and their EDS 2.0 successors under `/next`. It is kept up to date while EDS 2.0 is in beta, so that you can tell an intended change from a bug. The full migration guide, together with codemods and AI-assisted rewrites, comes when EDS 2.0 becomes stable.
+
+**What counts as a breaking change on this page**
+
+- **Props**: removed or renamed props, changed accepted values, changed callback signatures.
+- **Composition**: added or removed sub-components, changed children API.
+- **Markup and behaviour**: changed elements, roles, ARIA attributes, class names or test ids that can break consumer tests or CSS.
+
+**Not tracked**: visual changes such as spacing, colour and typography. Every EDS 2.0 component is restyled, and styled-components hash class names are replaced by stable `eds-*` class names. Treat any CSS that targets EDS internals as broken until you have verified it.
+
+**Found a change that is not listed here?** It may be unintentional. Please [open an issue](https://github.com/equinor/design-system/issues) so we can either fix it or add it to this page.
+
+## Overview
+
+| EDS 1.0                                                   | EDS 2.0                                     | Most important changes                                                                                  |
+| --------------------------------------------------------- | ------------------------------------------- | ------------------------------------------------------------------------------------------------------- |
+| `Accordion`                                               | [Accordion](#accordion)                     | Native `<details>`; `isExpanded` → `open`; `HeaderTitle` and `HeaderActions` removed                    |
+| `Autocomplete`                                            | [Autocomplete](#autocomplete)               | Single-select only; `selectedOptions`/`onOptionsChange` → `value`/`onValueChange`                       |
+| none                                                      | [Badge](#badge)                             | New component for non-interactive labels                                                                |
+| `Banner`                                                  | [Banner](#banner)                           | `tone` moves to the root; default role `alert` → `status`; `elevation` removed                          |
+| `Button`                                                  | [Button](#button)                           | `contained`/`outlined` → `primary`/`secondary`; `color` → `tone`; `href` → `asChild`                    |
+| `Checkbox`                                                | [Checkbox](#checkbox)                       | Root `<label>` → `<div>`; `className` lands on the input                                                |
+| `Chip`                                                    | [Chip](#chip)                               | Real `<button>`; `variant="active"` → `selected`; delete fires on whole-chip click                       |
+| `Dialog`                                                  | [Dialog](#dialog)                           | `onClose` → required `onOpenChange(open)`; always dismissable; ref points at `<dialog>`                 |
+| `Divider`                                                 | [Divider](#divider)                         | `color`, `variant` and `size` removed; no built-in margins                                              |
+| `InputWrapper`, `Label`, `HelperText`                     | [Field](#field)                             | Compose `Field.Label`, `Field.Description` and `Field.HelperMessage` yourself                           |
+| `Icon`                                                    | [Icon](#icon)                               | Numeric `size` → semantic `size`; `name` and `Icon.add()` removed                                       |
+| `Input`                                                   | [Input](#input)                             | `leftAdornments`/`rightAdornments` → `startAdornment`/`endAdornment`; `variant` → `invalid`             |
+| none (`Typography link`)                                  | [Link](#link)                               | New component                                                                                           |
+| `Menu`                                                    | [Menu](#menu)                               | Not a replacement yet; keep using the EDS 1.0 `Menu`                                                    |
+| `Radio`                                                   | [Radio](#radio)                             | Root `<label>` → `<div>`; SVG `path[name]` hooks removed                                                |
+| `Search`                                                  | [Search](#search)                           | Clearing calls `onClear`, not `onChange`; native `<search>` element                                     |
+| `NativeSelect`                                            | [Select](#select)                           | `<option>` children → `options` prop; ref points at `<select>`                                          |
+| `Switch`                                                  | [Switch](#switch)                           | `label` required; `size` removed; `role="switch"`                                                       |
+| `Textarea`                                                | [TextArea](#textarea)                       | Renamed export; `rowsMax` → `maxRows`; `helperText` → `helperMessage`; auto-grows by default            |
+| `TextField`                                               | [TextField](#textfield)                     | `helperText` → `helperMessage`; `meta` → `indicator`; `variant` → `invalid`; `id` gets `-input` suffix  |
+| `Tooltip`                                                 | [Tooltip](#tooltip)                         | `placement` reduced to four values, default `top`; rendered inline without an arrow                     |
+
+Shared changes across the form components (Checkbox, Input, Radio, Search, Select, Switch, TextArea, TextField):
+
+- The `variant` prop with `error` / `warning` / `success` is replaced by a boolean `invalid`. There is no `warning` or `success` state.
+- `helperText` becomes `helperMessage` and accepts a ReactNode. `meta` becomes `indicator` and is meant for `(Required)` / `(Optional)`.
+- Components built on [Field](#field) derive their ids: `id="name"` puts `name-input` on the input and `name-label`, `name-description`, `name-helper-message` on the other parts.
+- The helper message gets `role="alert"` when the field is invalid.
+- The container is no longer forced to `width: 100%`.
+
+## Accordion
+
+EDS 1.0: `Accordion` with `Accordion.Item`, `Accordion.Header`, `Accordion.HeaderTitle`, `Accordion.HeaderActions` and `Accordion.Panel`.
+
+**Props**
+
+- Removed on `Accordion`: `headerLevel`, `chevronPosition`.
+- Removed on `Accordion.Item`: `disabled`.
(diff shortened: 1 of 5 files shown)


pomfrida:
@claude review

pomfrida:
Pushed two commits after a full claim-by-claim verification of the page against source (both the EDS 1.0 and `/next` side, with file and line evidence per claim). About 20 claims were wrong or imprecise and are corrected; the review findings above are all addressed:

- Shared-changes section is scoped to the components it actually applies to. Checkbox, Radio and Switch keep the `id` you pass; Checkbox's helper id is `${id}-helper`.
- Checkbox and Radio root markup is described for both the labelled and the standalone path.
- `getByRole('search')` does **not** find the `<search>` root with the aria-query version in this repo (verified in jsdom), so the page now says so and points to `getByRole('searchbox')`.
- Select: `multiple` is not blocked in the types. Filed as #5460 rather than papered over in prose.
- Dialog: required `open` and the blocked native `onClose` are called out.
- Accordion `onToggle`, the full Autocomplete blocked-prop list, Badge values, and the stale `indeterminate` JSDoc (fixed in `Checkbox.types.ts`) are in.
- Freshness line at the top (`3.0.0-beta.1`, September 2026) and hidden anchors for `#scrim`, `#useinputfield`, `#buttongroup`, `#buttontoggle`.

On the question raised outside GitHub, whether everything listed is really intentional: no. The page was produced by diffing code, so it lists implemented differences. It now separates three states, explained at the top:

- **Unmarked**: intentional.
- **Known issues**: behaviour we consider a bug, kept on the page so consumers are not surprised, each linked to its report. New from this work: #5457 Switch, #5458 Chip, #5459 Tooltip, #5460 Select, #5461 Dialog, #5462 Icon, #5463 Accordion. Existing: #5323, #5322, #5103, #5456.
- **Under review**: implemented as described but not confirmed by design. Banner default role, Tooltip default placement, no `warning`/`success` on the Input family, Divider variants.

Menu: agreed that "keep using the EDS 1.0 Menu" is contradicted by the public export. Noted on #5103 with a suggestion to mark it `@internal` until #4417 lands.

Note for reviewers: `*.mdx` is in `.prettierignore`, so the only automated check on this file is that it compiles as MDX, which it does. Anchor ids and the sidebar position still need the deploy preview.

pomfrida:
Heads-up on ordering: #5410 moves `className` and `style` to the root wrapper on Checkbox, Radio and Switch. The Checkbox, Radio and Switch sections here still describe the current behaviour (props landing on the `<input>`) and list #5323 under Known issues. If #5410 merges first, I will update those three sections and drop the #5323 entries before this PR is merged. If this PR merges first, the update comes as a small follow-up when #5410 lands.

pomfrida:
Some design questions will be talked about 17/09 - so holding this in draft until then so we can list the correct intended changes

Chibuzor-Nwemambu:
@pomfrida I just realised this is still a draft pr. Is that intentional?

Chibuzor-Nwemambu (changes requested):
Very Nice work 💪🏾 

Reviewed the whole page. The content checks out: I verified around 35 of the API claims against the source and all the in-page anchors. 
Three things:

### 1. The "Jump to a component" pills should be links, not buttons

They are `Chip`s, which render a `<button>` and only scroll the page. So:

- The address bar never updates, so you cannot link anyone to a section. The same component clicked in the Overview table does update it.
image

- Keyboard focus stays behind on the pill, so the next Tab carries on from the top of the page. It's difficult to show this with screenshot, but if you: 
-> navigate to the pill-shaped list, 
-> click on one component to get it active, 
-> click on Tab; that should focus on the next component, 
-> click enter, it navigates you to the component description. 
-> If you click on Tab again, you're directed back to the pill-shaped list. 
For comparison, the table behaves the right way.

- `Chip` adds `aria-pressed="false"` to all 22, so screen readers announce them as switches that are turned off. That is the `Chip` bug this page documents as #5458.

Plain anchors fix all three, matching the Overview table:

```jsx
<a key={name} href={`#${name.toLowerCase()}`} className="jump-link">{name}</a>
```

jumpTo can then go. Chip has no asChild, so keeping the pill look needs a few lines of CSS.

2. The "Form components" summary bullet should name the components

Every other bullet names specific components. This one does not, so it reads as though all form components are affected and have known issues, when only some are, and the reader cannot tell which.

3. The Checkbox JSDoc promises a bit more than the component delivers

Fair catch that the old note was stale, but the replacement is a touch absolute. "The component sets the native element's `indeterminate` property for you" holds for object refs but not callback refs, where `inputRef.current` is `undefined` and the effect never runs. (Pre-existing, not something this PR caused.) Only the DOM property is affected, but the deleted note was the only place that case was mentioned. Worth adding "when you pass an object ref".

2 and 3 are non-blocking. 1 is worth fixing, since this is the page teaching people about accessibility-affecting changes.

Also, I see this PR is still a draft. Is that intentional?

pomfrida:
@claude review

eddiman:
**Answers from the dev/design sync 17 September 2026.** All six items you were holding this draft for now have a decision, so the "Under review" entries can move.

| Item | Answer | Where it goes on the page |
| --- | --- | --- |
| Tooltip default placement (#5386) | Code is correct and matches EDS 1.0. Figma is the side that is wrong: it names `placement` after the arrow position, code names it after the tooltip position relative to the trigger, so the two read as reversed. Figma gets corrected in #5499. | Not a breaking change. Worth a line saying the default is unchanged from 1.0, since the Figma mismatch made it look otherwise. |
| Tooltip open delay (#5459) | 150 ms, matching the value already used for transitions across the components. The value is undocumented and becomes a token in #5498. | Behaviour change from 1.0's 100 ms. |
| Dialog blocking mode (#5461) | Supported. `isDismissable` comes back. One constraint: when a dialog has no action buttons, the dismiss control must be present, because backdrop click alone is not an acceptable way to close. Figma simply never showed a non-dismissable example, which is what made it look unsupported. | Not removed. |
| Icon small variants (#5462) | Retired. The dataset comes out of `@equinor/eds-icons`, the Figma broker and Figma, after a usage audit. Tracked in #5500. | Breaking, with the note that `xs` and `sm` now scale the 24 px geometry. |
| Accordion disabled items (#5463) | No disabled state. Two live usages across the organisation, and disabling an accordion signals the wrong component rather than a missing state. Full numbers on that issue. | Intentionally removed. |
| Button.Group and Button.Toggle | Toggle button is wanted. Lavinia confirmed the need, it is in the component table, and it is tracked as #5342, #5343, #5344 and #5345; it had fallen far down the slice ordering. `Button.Group` still has no answer and is now tracked in #5503. | Toggle button: coming, not removed. `Button.Group`: leave under review, pointing at #5503. |

So five of the six can move out of "Under review" and this can come out of draft. `Button.Group` is the one exception.

Chibuzor-Nwemambu (changes requested):
Really good work @pomfrida ! 👍🏾 

Found two real bugs while testing. Both are on the "Jump to a component" buttons at the top of the page (Accordion, Autocomplete, Badge, etc.).

1. Clicking any of those buttons kicks you out of Storybook entirely. You lose the sidebar and end up on a bare, plain version of the page. The table links seem to work perfectly fine, but not the button. The difference is a missing setting (target="_self") that tells the link "just update this page, don't jump away." The table's links have it automatically; these buttons don't. Adding it fixes all 22 of them at once.
2. 

Image

3. In Windows High Contrast mode, you can't tell which button is focused when tabbing through with a keyboard. These buttons were custom-styled to look like Chips, and copied Chip's focus style along with it, a style that High Contrast mode simply doesn't render. The table's links weren't given this custom styling, so they still show a normal, visible focus box.
   I think this is worth fixing, because High Contrast mode is a real accessibility setting people rely on daily, and someone using it would have no way to tell which section they're about to jump to while tabbing through, unlike a mouse user, who'd never notice anything's wrong. 
Bonus find: the same box-shadow-only focus style already exists on the real Chip component, so this might be worth a quick look there too, separately from this PR.

One more thing worth asking, not a blocker: the pill-shaped buttons and the table both link to the exact same sections. Is there a reason to keep both, or would it be simpler to just drop the buttons and rely on the table?

packages/eds-core-react/stories/docs/BreakingChanges.mdx
```diff
@@ -0,0 +1,470 @@
...
+    </Banner>
+  </div>
+</Unstyled>
+
+**Jump to a component**
+
+<Unstyled>
+  <nav aria-label="Jump to a component" data-color-scheme="light" className="eds-breaking-changes-jump-links">
+    {components.map((name) => (
+      <a key={name} href={`#${name.toLowerCase()}`}>
```
Chibuzor-Nwemambu: Needs a 
`<a key={name} href={`#${name.toLowerCase()}`} target="_self">`

packages/eds-core-react/stories/docs/breaking-changes.css
```diff
@@ -0,0 +1,40 @@
...
+.eds-breaking-changes-jump-links a:active {
+  background-color: var(--eds-color-bg-fill-muted-active);
+}
+
+.eds-breaking-changes-jump-links a:focus-visible {
+  outline: none;
+  box-shadow:
+    0 0 0 var(--eds-sizing-stroke-thin) var(--eds-color-bg-canvas),
+    0 0 0 calc(var(--eds-sizing-stroke-thin) * 2) var(--eds-color-border-focus);
+}
```
Chibuzor-Nwemambu: Adding this block after line 40 should fix the high contrast issue:

```
@media (forced-colors: active) {
  .eds-breaking-changes-jump-links a:focus-visible {
    outline: 2px solid;
    outline-offset: 2px;
  }
}
```

Chibuzor-Nwemambu (changes requested):
Sorry forgot the inline comments.

PR #5472, opened 2026-09-15
chore: bump pnpm overrides to resolve transitive dependabot alerts

pomfrida:
Bumps the `pnpm.overrides` in the root `package.json` so the remaining transitive Dependabot alerts resolve to patched versions. No direct dependencies change.

| Package | Before | After | Alert |
| --- | --- | --- | --- |
| `@xmldom/xmldom` | 0.8.13 / 0.9.10 | 0.8.15 / 0.9.12 | High (via Expo / plist) |
| `fast-uri` | 3.1.5 | 3.1.8 | High (via ajv) |
| `browserslist` | 4.28.6 + 4.28.9 | 4.28.9 | High (via autoprefixer) |
| `js-yaml` | 3.15.1 / 4.3.1 | 3.15.2 / 4.3.2 | High (via gray-matter) |
| `qs` | 6.15.3 | 6.16.0 | Medium (via express) |
| `joi` | 18.2.3 | 18.2.9 | Medium / Low (via Docusaurus) |
| `@humanfs/node` | 0.16.7 | 0.16.8 | Medium (via eslint) |

Deliberately left out:

- `decode-uri-component` 0.2.2 → 0.5.0: 0.5.0 is ESM-only and `query-string@7` (Expo) is CommonJS.
- `esbuild` 0.27 → 0.28 (low severity): pulled in by `tsup`, not worth the risk for a low.
- `image-size`: no patched version available.

The lockfile diff only touches the packages above and their direct resolutions. `pnpm install --frozen-lockfile` passes locally.


pomfrida:
@claude review

Chibuzor-Nwemambu (dismissed):
LGTM! 🔥 

Noticed the `vitest` and `@vitest/mocker` weren't updated as well. But it seems dependabot has a pr for that already so this looks good to merge.

pomfrida:
Follow-up on the review above, pushed in 52e2f78:

- Added upper bounds on `joi`, `qs`, `browserslist`, `@humanfs/node` and `@xmldom/xmldom@^0.9`. Floors verified against the first patched version in the GitHub advisories.
- Dropped the redundant `gray-matter>js-yaml` entry.
- Bumped `browserslist` to `^4.28.9` in eds-core-react and scoped the override to `autoprefixer>browserslist`, which was the only requester still on 4.28.6.

On the `libc:` fields: that one is not fixable in this PR. The pinned `pnpm@10.28.0` does not write `libc` to the lockfile at all. I regenerated twice with `corepack pnpm` (from main's lockfile and from this branch's) and got 0 both times. The 86 fields on main come from Dependabot, which runs a newer pnpm. Main history shows the same pattern: human commits (e.g. f2daa9690) have 0, Dependabot commits (9c28e8b42, ddf42eb8c, 474dbfbdc) have 86. They will come back with the next Dependabot PR either way. The real fix is to bump pnpm in `package.json` and the Dockerfiles so humans and Dependabot write the same lockfile. Tracked in #5483.

PR #5368, opened 2026-08-26
fix: override nanoid to >=3.3.18 to resolve high severity CVE

millus:
## Summary

Resolves Dependabot alert #548 — nanoid < 3.3.18 has a high severity vulnerability where custom generators can loop indefinitely when size is zero.

nanoid is a transitive dependency pulled in by `postcss` (via `stylelint`) so Dependabot cannot auto-create a PR. Fixed by adding a `pnpm.overrides` entry to force all instances to `>=3.3.18`.

## Changes

- Added `"nanoid": ">=3.3.18"` to `pnpm.overrides` in root `package.json`
- Updated `pnpm-lock.yaml`

## Test plan

- [ ] CI passes
- [ ] Verify nanoid resolves to 3.3.18: `pnpm why nanoid`

PR #5177, opened 2026-07-14
chore: add pnpm overrides for dependabot security alerts

pomfrida:
## What

Resolves all 14 open Dependabot alerts by forcing patched versions of transitive dependencies via `pnpm.overrides` in the root `package.json`, following the existing pattern. All affected packages are dev/build tooling (webpack dev-server chain, Docusaurus, express) — nothing ships in the published packages. Every override stays within the package's current major, except the ones where the only patched version is a new major (`serialize-javascript` 6→7, `uuid` 7/8→11).

| Override | Resolved version | Alerts |
| --- | --- | --- |
| `webpack-dev-server` | 5.2.6 | #488, #469 |
| `ws@^7` / `ws@^8` | 7.5.11 / 8.21.0 | #482, #481 (high), #470 |
| `serialize-javascript` | 7.0.7 | #360 (high, RCE), #472 |
| `http-proxy-middleware@^2` | 2.0.10 | #498 |
| `launch-editor` | 2.14.1 | #485 |
| `js-yaml@^4` (replaces old `js-yaml >=4.1.1`) | 4.3.0 | #499 |
| `joi` | 17.13.4 | #477 |
| `qs` | 6.15.3 | #473 |
| `postcss` (next pinned 8.4.31) | 8.5.19 | #425 |
| `uuid` | 11.1.1 | #471 |

Note on `uuid`: the advisory only affects `v3/v5/v6()` calls with a `buf` argument; our dependents (`sockjs`, `xcode`) only use `v4()`, so the jump to 11.1.1 is a formality to clear the alert.

## Verification

- `pnpm audit`: no known vulnerabilities
- Full monorepo build: green (Node 22)
- `docu:build` (Docusaurus/webpack chain where these packages are actually exercised): green
- `test:utils`: 29/29 passing

Issue #5249, opened 2026-07-29
docs: add Security and quality checks to the dependabot rotation


pomfrida:
## Background

The weekly dependabot rotation currently only covers reviewing and merging dependabot PRs. The person on duty should also check the repo's **Security and quality** alerts (GitHub Security tab: Dependabot alerts, code scanning) as part of the same weekly sweep, so vulnerabilities don't sit unnoticed between ad-hoc cleanups.

## Task

Add "Security and quality" to the dependabot rotation duty and update the documentation and reminder accordingly:

- [ ] `documentation/how-to/DEPENDABOT_GUIDE.md` — add a section describing the weekly security/quality check: where to find the alerts (Security tab), what to look at (Dependabot alerts, code scanning alerts), and what to do about them (dismiss with reason / fix / open follow-up issue)
- [ ] `.github/workflows/dependabot-rotation.yml` — extend the Monday Slack reminder with a link to the Security overview so the duty is visible in the weekly ping
- [ ] Check for other places that describe the rotation duty and update them if needed

## Definition of done

The Monday rotation reminder and the how-to guide both make it clear that checking Security and quality alerts is part of the weekly duty.

PR #5474, opened 2026-09-15
docs: add dependabot-duty playbook and harness entry points covering PRs and security alerts

pomfrida:
## Why

The Dependabot rotation has two halves: the open Dependabot PRs and the alerts on the Security tab that never get a PR. In practice only the first half gets done. The runbook did mention the alerts page, but as an afterthought below "When CI fails", and it said "log it" without saying where. It also had nothing on the `pnpm.overrides` recipe we actually use for transitive alerts (#5177, #5368, #5472).

This week's sweep also surfaced two things worth writing down: Dependabot opens a `npm_and_yarn` security PR and a regular version PR for the same bump, and the security one can fail on `ERR_PNPM_OUTDATED_LOCKFILE` because it lacks the lockfile change (#5449 vs #5451). And the root `pnpm run test` script does not run `eds-tokens` / `eds-tokens-build`, so a vitest major (#5470) went through CI without those suites running.

## What

Following the canonical-docs pattern used by `/audit-harnesses` and `/tokens-studio`:

- **New** `documentation/agent-instructions/DEPENDABOT_DUTY.md`: the long-form playbook. Prerequisites (`security_events` scope), boundaries (report first, never approve/merge/close without a go-ahead), PR triage with the duplicate-PR pattern and a decision table, alerts triage via `gh api`, the override recipe step by step, report format, and where to log alerts that cannot be fixed.
- **New** thin entry points: `.claude/commands/dependabot-duty.md`, `.github/prompts/dependabot-duty.prompt.md`, `.opencode/agent/dependabot-duty.md`. The OpenCode agent asks before any PR- or issue-mutating `gh` command and before git writes.
- **Updated** `documentation/how-to/DEPENDABOT_GUIDE.md`: "3-step" becomes "4-step" with the Security tab as step 4, the duplicate-PR and test-coverage checks added to step 2, a short "Fixing a transitive alert" recipe, a concrete "log it" destination, and a pointer to the agent entry points.
- **Registered** the new entry point in `AGENTS.md`, `.claude/CLAUDE.md` and `.opencode/README.md`.

No changes to workflows or config. Nothing here changes what Dependabot does, only what the person on duty (or their agent) does.


pomfrida:
@claude review

Chibuzor-Nwemambu (changes requested):
Very well done 🙌🏾 

Solid, well-cross-referenced docs.

However, there is one thing I think is worth addressing before merging:

- Documentation/agent-instructions/HARNESS_AUDIT.md : capability-parity matrix (Step 4) isn't updated. This PR adds the "Dependabot duty" workflow to all three harnesses, but the parity table (~lines 80-88) doesn't get a new row for it — every prior addition (e.g.  tokens-studio ) added one. Running  /audit-harnesses  today would flag this PR's own change as a gap.

pomfrida:
@claude review

pomfrida:
@Chibuzor-Nwemambu your point is done — `HARNESS_AUDIT.md` has the `Dependabot duty` parity row now, and the stale "three OpenCode agents" count is gone.

Fair warning that this grew a fair bit after your review. Code scanning is in scope now, since it was the other half of #5249. And there's a decision in here I'd like your view on: what to do with alerts we're not going to fix. The runbook has said "log it and move on" for a while without naming anywhere to log to, so nothing was ever recorded — this fills that gap rather than changing a practice. They now get dismissed with a reason and a comment, which keeps the reasoning on the alert. The trade-off is that an open high alert nags and a dismissed one doesn't, so the weekly sweep re-reads the dismissed list to compensate. Details in the updated description.

Chibuzor-Nwemambu:
Nice @pomfrida . I'll hve another look at the pr.

Chibuzor-Nwemambu (dismissed):
Well done! 🎉 
I checked the fixes against the live API and they all hold up (including the alert counts and the severity sorting).

One small thing I noticed though: the new "re-check what's dismissed" logic now wants to show two different things: alerts that recently became fixable, and alerts that are still stuck waiting (no fix, or "not enough time this week"). But the report template in Step 5 only has a table for the first one. So as written, the "still stuck" list has nowhere to go in the actual report, even though the whole point of this round was to keep that pile visible.

Suggestion: add a small "Still parked" table to the Step 5 report (and the same for code scanning's dismissed list), so those don't just get quietly dropped when someone follows the template.

Not a blocker, just don't want the visibility work in this PR to lose its ending.

pomfrida:
@Chibuzor-Nwemambu Good catch, thanks! Added a "Still parked" table (with a "Parked since" date instead of a sweep count, since the agent has no memory between sweeps) and a "Dismissed code scanning alerts" table to the Step 5 report, and pointed Step 2 and Step 4 at them.

PR #5479, opened 2026-09-16
fix: make Tooltip (next) fall back to a free side instead of shifting

### LINKED ISSUE #5473: Tooltip (next) should fall back to a free side instead of shifting; shifted arrow points at nothing
### Describe the bug

When a centred tooltip does not fit horizontally, for example on a button at the edge of the viewport, the browser shifts the tooltip body sideways to keep it inside the viewport (default `anchor-center` alignment from `position-area`). The shift itself is fine, but the arrow is a `::before` pseudo-element positioned relative to the tooltip body (`next/Tooltip/tooltip.css:29-60`, `margin: inherit`), so it stays centred on the body and ends up pointing at empty space.

Measured in Chromium on `3.0.0-beta.1`: viewport 1511 px, trigger 1411 to 1507 px, tooltip 1091 to 1507 px, arrow at about 1299 px, roughly 100 px left of the trigger.

EDS 1.0 handled the same situation with floating-ui's `arrow` middleware, which moved the arrow along when `shift` kicked in (`components/Tooltip/Tooltip.tsx:65-81,134-136`).

### Steps to reproduce

1. Render a `Tooltip` with a title longer than its trigger on a button placed at the right edge of the viewport (for example `position: fixed; right: 4px`).
2. Hover the button.
3. The tooltip is right-aligned with the trigger, but the arrow sits at the horizontal centre of the tooltip.

### Expected behavior (agreed with design, 2026-09-15)

The tooltip does not shift at all. When the centred placement does not fit, the component falls back to a free side, so a tooltip on a button in the top-left corner renders to the right of the button (`placement="right"` behaviour) with the arrow centred and pointing at the trigger. See the Figma reference from design attached below.

Consumers can already get this manually by choosing `placement="right"` or `"left"`, but the default has to handle it because triggers end up at edges through narrow viewports and dynamic layouts that the consumer cannot predict.

Implementation notes:

- `position-try-fallbacks` currently only lists the opposite side (`flip-block` / `flip-inline`). Add the perpendicular sides as fallbacks so `top` tries `bottom`, then `right`, then `left` (and equivalently for the others).
- The fallbacks only trigger if the box actually overflows. With `anchor-center` the browser shifts first and nothing overflows, so the alignment probably has to be plain centring (`justify-self: center`) for the fallback chain to kick in. Needs verification in Chromium and Safari.
- Removing the arrow (GitHub's approach) was considered and rejected by design.

### Docs follow-up

Add one line to the Tooltip docs: for triggers in corners or at edges, choose a side placement (`left` / `right`) if you want to control which side is used; otherwise the component picks a free side.

### Specifications

- Version: `@equinor/eds-core-react@3.0.0-beta.1`
- Browser: Chromium (CSS anchor positioning)

### Additional context

Found while checking whether the reduced `placement` set (#5386) loses the viewport-edge case. It does not: the browser keeps the tooltip inside the viewport on its own, and the corner case is solved by switching side, so aligned `-start`/`-end` placements are not needed for edges. Affects every tooltip on a control near a viewport edge, which is common for icon buttons at the end of toolbars and table rows.

pomfrida:
Closes #5473

## What

When a centred tooltip did not fit horizontally, the browser shifted the body sideways (`anchor-center`) while the arrow stayed centred on the body and pointed at empty space.

- The tooltip is now centred on the anchor with a symmetric inset (`--_center`) instead of `anchor-center`, so the browser cannot shift it. When the preferred side does not fit, `position-try-fallbacks` moves to the opposite side and then to the two perpendicular sides (e.g. top → bottom → right → left), as agreed with design.
- The arrow no longer relies on `margin: inherit`, which Firefox resolves against the base placement instead of the chosen `@position-try` option (pre-existing bug, made more visible by the new side fallbacks). It is now a fixed-position pseudo-element whose edges are computed with `anchor()` against both the trigger and the tooltip body (new `--tooltip-body-name` anchor set by the component).
- The tooltip margins are removed. They were only needed by the old arrow, and they made the side fallbacks overflow for triggers flush with the viewport edge in Firefox (where the tooltip is taller because `text-box-trim` is unsupported). The 4 px gap comes from the inset.
- `@position-try` options use longhands on purpose: Chromium ignores the `place-self` shorthand inside `@position-try`, and stylelint's `--fix` would otherwise merge them (`stylelint-disable` around the block).
- New `ViewportEdges` story, updated Storybook/JSDoc descriptions, and the docs line requested in the issue.

## Tests

- `tests/visual/Tooltip.next.spec.ts`: Playwright geometry tests (bounding boxes, not screenshots) for the four preferred placements and the four viewport-corner fallbacks. Each asserts side, 4 px gap, centring, inside-viewport, and that the arrow is extended towards the trigger. Runs in a new `firefox` project as well as `chromium`. 16/16 green locally.
- `playwright.config.ts`: `STORYBOOK_PORT` env var so the tests can run next to a Storybook from another checkout.
- Jest: one new test checks that both anchor names are exposed as CSS variables (jsdom cannot verify positioning).

Note: the visual tests are not wired into CI yet; they run with `pnpm test:visual` locally.

## Verified manually

- Chrome 153 (Storybook) and Firefox 156 / 153 (Playwright): corners, edges and centre for all four placements give the expected free side with a 4 px gap and the arrow centred on the trigger.
- Narrow viewport where no side fits: the browser keeps the tooltip inside the viewport (same as before).
- Safari not tested (not supported in Equinor).


pomfrida:
@claude review

pomfrida:
Follow-up on the review above, in the latest commit:

- **2 (spec robustness)**: all four taken. The trigger element is passed into `evaluate`, the arrow insets are resolved against `documentElement.clientWidth/Height`, the viewport is pinned to 1280×720, and the spec waits for `document.fonts.ready`. The arrow size is now read from the computed `--_arrow` token (it resolves in rem, so it is converted with the root font size). The `ViewportEdges` story notes that the spec depends on it.
- **3 (arrow degradation)**: documented in `tooltip.css` as an accepted trade-off for the supported browsers. No `@supports` guard, since a plain `inset` fallback would not help against invalid-at-computed-value-time.
- **4 (wording)**: docs rewritten so `placement` is described as a preference, `left`/`right` as "most likely to be used as given" near the top/bottom edge, and corners "pick a free side". The "always points at the trigger" claim is gone. README heading, CI paragraph and config JSDoc corrected; the README now says the suite is not in CI yet (#5480) and needs both browsers installed.
- **1 (no side fits)**: the wording is softened as above. `position-visibility: no-overflow` is a behaviour change (hide instead of clip), so I have left it out of this PR pending a decision with design. No test pins the degenerate case yet for the same reason.
- **5**: `STORYBOOK_PORT` stays in the config rather than the package script, because `${VAR:-9000}` in an npm script breaks on Windows shells. The `--tooltip-body-name` rename is skipped as cosmetic.

Playwright: 16/16 in Chromium and Firefox locally.

pomfrida:
@claude review

PR #5485, opened 2026-09-16
chore: bump pnpm to 11 and pin @types/react

### LINKED ISSUE #5483: chore: bump pinned pnpm so local and Dependabot lockfiles match
## Problem

The repo pins `pnpm@10.28.0` (`package.json` `packageManager`, `DockerfileEdsDocs`, `DockerfileEdsDemo`). Dependabot regenerates `pnpm-lock.yaml` with a newer pnpm that writes `libc:` platform metadata on Linux native optional dependencies (`sharp`, `lightningcss`, `@rollup/rollup-linux-*`, `@tailwindcss/oxide`, `@next/swc-linux-*`, etc.). pnpm 10.28.0 does not write that field.

Result: every human-authored lockfile change silently drops all 86 `libc:` lines, and the next Dependabot PR adds them back. Main history shows the flip-flop:

| Commit | Author | `libc:` lines |
| --- | --- | --- |
| f2daa9690 | human | 0 |
| 9c28e8b42 | dependabot | 86 |
| 57e8c6a3b | dependabot | 0 |
| ddf42eb8c, 474dbfbdc | dependabot | 86 |

The noise makes lockfile diffs hard to review (see the review thread on #5472) and means the Alpine Docker builds run with or without platform metadata depending on who touched the lockfile last.

## Task

Bump pnpm to a single version that matches what Dependabot uses, in all three places:

- `package.json` `packageManager`
- `DockerfileEdsDocs` and `DockerfileEdsDemo` (`npm install -g pnpm@...`)

CI picks the version up from `packageManager` via `pnpm/action-setup@v6`, so no workflow change is needed. Verify by regenerating the lockfile with the new pin and confirming `grep -c '^    libc:' pnpm-lock.yaml` returns 86 and the diff against main is empty apart from the header.

Note: pnpm 11+ requires Node 22+, which the repo already has (`engines.node >=22`). pnpm 11.6+ also infers platform fields from package names, so the `libc:` field becomes less important once bumped.

### LINKED ISSUE #5484: Build fails: duplicate @types/react (19.2.18 + 19.3.0) breaks eds-core-react tsc
**The `Build` job fails on `main`, and therefore on every open PR.**

`@equinor/eds-core-react`'s build (`rollup -c && tsc -p tsconfig.build.json`) throws 14 type errors, all variants of:

> `Type 'RefCallback<HTMLButtonElement>' is not assignable to type 'Ref<HTMLButtonElement>'` … **`Two different types with this name exist, but they are unrelated.`**

Example run: https://github.com/equinor/design-system/actions/runs/34976234265/job/104405110408 (surfaced on #5453, but that PR isn't the cause — it never touches the lockfile.)

## Cause

Two copies of `@types/react` in the pnpm tree. The override in root `package.json` is a range, not a pin:

```json
"pnpm": { "overrides": { "@types/react": ">=19.2.17" } }
```

So when `@types/react@19.3.0` was published, `apps/design-system-docs` resolved to **19.3.0** while every other importer stayed on **19.2.18**. Both copies reach `eds-core-react`'s `tsc` — the `TS2742` errors name the path outright: `.pnpm/@types+react@19.3.0/node_modules/@types/react`.

The 14 failing files (Datepicker, Dialog, Menu, Popover, Scrim, NativeSelect, SideBarAccordion, Table/Caption, Tooltip, Switch, Typography stories) are symptoms. None need changing.

## Repro

```bash
# on clean main
pnpm install && pnpm run build          # exits 1
grep -oE "@types/react@[0-9.]+" pnpm-lock.yaml | sort -u   # two versions
```

## Fix

```diff
-      "@types/react": ">=19.2.17",
+      "@types/react": "19.2.18",
```

then `pnpm install --lockfile-only && pnpm install`.

Verified locally: single `@types/react@19.2.18` in the lockfile, `tsc -p tsconfig.build.json` clean, full `pnpm run build` exits 0.

## Follow-ups (separate issues)

- Pin all `@types/*` overrides exactly — every `>=` entry has this same failure mode.
- Align the declared specs: `@types/react` is currently declared four different ways across the workspace (`^19.2.18`, `~19.2.17`, `19.2.14`), with only the override holding it together.
- Upgrading *to* 19.3.0 is a bigger job — its stricter `onSubmit` typing (`SubmitEventHandler<T>`) conflicts with styled-components 6.4.x polymorphic props. Don't bundle it with this unblock.

pomfrida:
Closes #5483, closes #5484.

Bumps the pinned pnpm from 10.28.0 to 11.27.0 and pins `@types/react` to a single version. Rebased on main after #5472 merged.

## Why pnpm 11

Dependabot regenerates `pnpm-lock.yaml` with pnpm 11 (11.25.0 per dependabot-core's Dockerfile), which writes `libc:` platform metadata on native optional dependencies. pnpm 10.x never writes that field, so every human-authored lockfile change dropped all 86 `libc:` lines and the next Dependabot PR put them back. pnpm 12 writes a different lockfile shape (`packageManagerDependencies`, 158 extra lines), so 11 is the version that matches Dependabot.

Verified: with this branch's lockfile, `pnpm install --lockfile-only` is a no-op under both 11.27.0 and 11.25.0.

## Breaking changes in pnpm 11 handled here

- **`pnpm` field in `package.json` is no longer read.** All overrides moved to `pnpm-workspace.yaml`. Without this the overrides were silently ignored and e.g. `fast-uri` fell back to a vulnerable version.
- **`.npmrc` is auth/registry only.** `sharedWorkspaceLockfile`, `linkWorkspacePackages`, `autoInstallPeers`, `strictPeerDependencies` moved to `pnpm-workspace.yaml`. `package-lock` dropped (npm-only setting).
- **`onlyBuiltDependencies` removed, replaced by `allowBuilds`.** Without this `pnpm install` fails with `ERR_PNPM_IGNORED_BUILDS` and `@tokens-studio/studio-cli` / `lightningcss-cli` never run their postinstall.

## `@types/react` (#5484)

The override was `>=19.2.17`. When 19.3.0 was published, Dependabot's next re-resolution bound the docs chain to 19.3.0 while everything else stayed on 19.2.18. Main still builds because 19.3.0 only reaches docs and react-native there, but #5453 adds `@types/react-dom` to the docs app, which pulls the 19.3.0 peer variant into eds-core-react's `tsc` and produces the TS2742 errors. Pinned to `19.2.18`; the lockfile now has a single `@types/react`.

## Lockfile diff

Large because pnpm 11 re-resolves the full graph when overrides change and records optional peers (e.g. `supports-color`) in the peer suffixes. Actual resolution changes: five packages removed (all dedupes: `@types/react@19.3.0`, `@napi-rs/wasm-runtime@1.1.6`, `@ungap/structured-clone@1.3.1`, `acorn@8.16.0`, `yargs@17.7.2`), nothing added.

## Other files

- Three Dockerfiles, README and AGENTS.md: pnpm version string.
- `TOKENS_STUDIO.md`: `onlyBuiltDependencies` → `allowBuilds`.
- `prerequisites.json`: regenerated by the `prebuild` step. It had been stale since April, so node and react also move (24.16.0 from `.nvmrc`, 18 from the peer range). Those are the script's own derivations.

## Verified locally

- `pnpm install` clean (no warnings), `--frozen-lockfile` passes, 86 `libc:` lines present
- full `pnpm run build` passes
- `test:utils` and `test:core-react` pass

## Risk to watch

pnpm 11 no longer delegates `pnpm publish` to the npm CLI. The publish workflows use `pnpm publish --provenance` with OIDC trusted publishing; pnpm 11 supports both natively (see 11.0.7, 11.1.3 release notes), but the first release after this merges is the real test.


pomfrida:
@claude review

pomfrida:
Thanks, all four checked. Pushed in 365db77a0:

- **1.** Added `@types/react` to the Dependabot ignore list and a lockstep comment above the override. We chose ignore over a comment alone so the divergence cannot happen silently.
- **3.** `extract-prerequisites.js` now says "pinned via the `packageManager` field" instead of "or higher", stale fallback default bumped; README regenerated.
- **4.** Devcontainer pinned to `pnpm@11.27.0`.
- **2.** No change. React 18 is exercised: `react18-compat.yaml` builds and tests all packages against React 18 on every PR touching `packages/**`, and passed on #5472 today. "React 18+" is the correct reading of the peer range; main showed 19 only because `prerequisites.json` had been stale since April.

PR #5497, opened 2026-09-18
docs: give every ADR a unique number and add a generated index

Chibuzor-Nwemambu:
Every ADR now has a number of its own.

Three numbers were in use twice, which made a reference like "see ADR-0005" ambiguous. The file that claimed each number first keeps it, and the other moves to the next free number:

| Was | Now |
| --- | --- |
| 0004 Spacing approach | 0017 |
| 0005 Typography approach | 0018 |

All ten links that pointed at the old filenames are updated, so nothing is left broken.

There is also a new `documentation/adr/README.md`. It is a table of all 18 records with title, status and date, so you can find the decision you need without opening every file. It is generated from the ADR files themselves on every build, which means it cannot drift out of date, and it fails CI if two ADRs ever share a number again.

One thing worth a reviewer's opinion: ADR 0005 and ADR 0016 say "Approved" while every other record says "Accepted". The index lists both at the bottom until we settle which word we use.


Chibuzor-Nwemambu:
@claude review

Chibuzor-Nwemambu:
Good catch on the first one, that was a real gap. Fixed in b8f94b1dd:

- The freshness check now also runs on docs-only PRs. It was sitting in a job that gets skipped when nothing outside `documentation/` changes, so a stale index could have merged green.
- ADR 0010's row now shows the date it was actually accepted, not just the date it was proposed.
- Long notes get shortened instead of dropped.
- Backslashes in a title are escaped now, which clears the CodeQL alert. Checked against GitHub's own renderer: the row still comes out with the right number of columns.
- The duplicate-number error no longer claims to know which file took the number first, because it can't.

Deliberately not in this PR: writing down that ADR numbers are just labels rather than a timeline. That belongs in a record of its own, not a rename cleanup, and it's on the list.

On Approved vs Accepted, agreed that Accepted is the one with a basis. Worth its own small change so this PR stays just the renaming.

One caveat: this PR touches `scripts/`, so CI takes the other path. The docs-only step gets its first real run on the next ADR-only PR.

eddiman (approved):
nize

